PR #58, opened 2020-07-30
feat: add icons

scurker:
This became larger than I expected:

* Refactored `packages/react` to build with rollup
* Demo site now uses `packages/react/lib` instead of building from `src`
* SVG Icons
* Removed `typechecks` as it's no longer needed since we export types by default

## Reviewer checks


docs/Demo/index.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import jsxStringify from 'react-element-to-jsx-string';
-import { Code } from '../../packages/react/src';
+import { Code } from '@deque/cauldron-react';
```
stephenmathieson, line 4: 👍

packages/react/src/components/Icon/index.tsx
```diff
@@ -8,24 +8,45 @@ export interface IconProps extends React.HTMLAttributes<HTMLDivElement> {
   type: string;
```
stephenmathieson, line 8: I know it's a pain, but I _strongly_ suggest making `type` stricter here.

Something like:

```ts
type: 'Thing' | 'OtherThing' | 'etc...'
```

packages/react/src/components/Icon/index.tsx
```diff
@@ -8,24 +8,45 @@ export interface IconProps extends React.HTMLAttributes<HTMLDivElement> {
...
-const Icon = ({ label, type, className, ...other }: IconProps) => {
+function Icon({ label, className, type, ...other }: IconProps) {
+  const [, name, direction] = type.match(/(.*)-(right|left|up|down)$/) || [
+    '',
+    type
+  ];
+  const [IconSVG, setIcon] = useState<React.ComponentType<any> | null>(null);
+
+  useEffect(() => {
+    import(`./icons/${name}.svg`)
```
stephenmathieson, line 19: What's wrong with `import`ing all of the icons and tree-shaking what we don't use out?

scurker: Dynamically importing means adding new icons is mostly drag and drop without having to modify the main `<Icon>` component. What's your concern here?

packages/react/src/components/Icon/index.tsx
scurker: I don't disagree, but we would need to make sure the type stay in sync with icons.

packages/react/src/components/Icon/index.tsx
scurker: Manually importing new icons and wrapping them with `<Icon>` just to import seems like unnecessary overhead when we can let rollup do the heavy lifting.

packages/react/src/components/Icon/index.tsx
scurker: Actually, typescript seems to be unhappy with dynamic strings when we define types.

```sh
src/index.ts → lib...
[!] (plugin typescript) Error: @rollup/plugin-typescript TS2322: Type 'string' is not assignable to type '"menu" | "code" | "arrow-up" | "arrow-down" | "arrow-left" | "arrow-right" | "bolt" | "checkbox-checked" | "checkbox-unchecked" | "chevron-up" | "chevron-down" | "chevron-left" | ... 21 more ... | "trash"'.
src/components/ExpandCollapsePanel/PanelTrigger.tsx (29:13)
29       <Icon type={`chevron-${open ? 'down' : 'right'}`} />
```

That's solvable by using `${open ? 'chevron-down' : 'chevron-right'}`, but I think for now I'd like to leave this as a non-breaking change.

stephenmathieson: Ah bummer, OK. Thanks for trying this out.

packages/react/src/components/Icon/index.tsx
stephenmathieson: I understand that manually `import`ing all of them will be a pain, but I'm wondering how frequently we'll add/remove icon SVGs.

My main concern is dependent packages - are the consumers of `cauldron` setup to handle these `import`s?

packages/react/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "outDir": "lib"
+    "typeRoots": ["node_modules/@types"]
```
stephenmathieson, line 4: Isn't this the default? Why do we need to specify this?

packages/react/src/components/Icon/index.tsx
scurker: Yup! Rollup and dynamic-import-variables handles this for us, it outputs a map like this inside of `<Icon>`:

```js
function __variableDynamicImportRuntime0__(path) {
   switch (path) {
     case './icons/arrow.svg': return Promise.resolve().then(function () { return { 'default': require('./icons/arrow.svg.js') }; });
     case './icons/bolt.svg': return Promise.resolve().then(function () { return { 'default': require('./icons/bolt.svg.js') }; });
     case './icons/checkbox-checked.svg': return Promise.resolve().then(function () { return { 'default': require('./icons/checkbox-checked.svg.js') }; });
     /* ... and the rest of the icons */
     default: return Promise.reject(new Error("Unknown variable dynamic import: " + path));
   }
}

function Icon(_a) {
    React.useEffect(function () {
        __variableDynamicImportRuntime0__("./icons/" + name + ".svg")
            .then(function (icon) {
            setIcon(function () { return icon["default"]; });
        })["catch"](function (ex) {
            console.error("Could not find icon type \"" + type + "\".");
            setIcon(null);
        });
    }, [type])
  ...
}
```

As far as dependent packages are concerned, they don't even know there's a dynamic import. Assuming a bundler is used downstream it becomes an inline require.

packages/react/tsconfig.json
scurker: Good call, looks like it's not needed and we can safely remove it.

packages/react/src/components/Icon/index.tsx
stephenmathieson: Awesome!!

PR #66, opened 2020-08-04
fix(styles): set correct height on thin buttons

schne324:
## Reviewer checks


schne324:
looks like the uxpin had a slightly larger height than the thin buttons currently in the extension (which is where I stole the styles from). This patch increases the height from 19 to 23px

Screen Shot 2020-08-04 at 10 02 16 AM

PR #75, opened 2020-08-25
feat(code): Add syntax highlighting as css (instead of inline styles in js) and explicitly import specific languages

### LINKED ISSUE #37: Code highlighter includes unused languages
I was investigating the axe extension bundle and noticed that `highlight.js` was significantly contributing to the size of one of the bundles:

image

It's contributing to roughly 28% (532 KB) of the bundle size and seems completely unnecessary. This is coming from `react-syntax-highlighter` for the `<Code>` component. It does provide a light bundle which you then have to import languages manually:

```js
import { Light as SyntaxHighlighter } from 'react-syntax-highlighter';
import js from 'react-syntax-highlighter/dist/esm/languages/hljs/javascript';
SyntaxHighlighter.registerLanguage('javascript', js);
```

I don't know if there's any way for us to leverage this given this component doesn't know what languages will be used ahead of time.

### LINKED ISSUE #72: Code component uses inline styles making it impossible to override anything

schne324:
## Reviewer checks

---

resolves #72 

resolves #37


packages/react/rollup.config.js
```diff
@@ -13,7 +13,6 @@ export default {
   output: {
     dir: 'lib',
     format: 'cjs',
-    preserveModules: true,
```
stephenmathieson, line 16: @scurker this flag causes the build to break. For some strange reason, it emits `lib/src` and `lib/node_modules/highlight.js`. I don't know why we added it in the first place. Is it safe to remove?

scurker: The old webpack code preserved the directory structure so I was trying to keep the same structure when migrating from webpack, so this option kept that output the same.

The only downside to not having it is debugging issues with cauldron, you'll see something like `cauldron/lib/index.js:23484` instead of `cauldron/lib/components/foo.js:234`. I'm not sure why it's just now causing issues but we can probably remove it.

stephenmathieson: It's causing issues now because of how we're "registering" languages in the Code component. I don't fully understand it either 🤷

netlify.toml
```diff
@@ -5,4 +5,4 @@
 [[redirects]]
   from = "/*"
   to = "/index.html"
-
+  status = 200
```
stephenmathieson, line 8: This fixes the "deployed" site, enabling it to act as a SPA. It should have been done in #76.

I know it's an unrelated change, but figured we didn't need a PR just for this.

packages/react/package.json
```diff
@@ -23,7 +23,7 @@
     "keyname": "^0.1.0",
     "prop-types": "^15.6.0",
     "rc-tooltip": "^3.7.2",
-    "react-syntax-highlighter": "^6.1.1",
+    "react-syntax-highlighter": "^13.5.1",
```
stephenmathieson, line 26: Updating this is technically a breaking change, [as `react-syntax-highlighter` does not work with Parcel (v1)](https://github.com/parcel-bundler/parcel/issues/3176).

This will prevent us from updating Cauldron in the ["axe Linter" frontend](https://github.com/dequelabs/axe-linter/tree/develop/src/frontend). I don't know if we need to worry about that right now, but it's something we should consider before merging this.

stephenmathieson: /cc @WilcoFiers @michael-siek heads up

packages/styles/code.css
```diff
@@ -0,0 +1,139 @@
+.hljs {
```
stephenmathieson, line 1: To avoid conflicts with existing "highlight.js" stylesheets, I would like do something like this:

```ts
// packages/react/src/components/Code/index.tsx
<SyntaxHighlighter {...props} useInlineStyles={false} className="Code">
  // ...
</SyntaxHighlighter>
```

And prefix these classes with:

```css
.Code .hljs {
 /* .. */
}
```

packages/react/package.json
schne324: > Updating this is technically a breaking change

Yes it is, but I figured I didn't have to worry about it since we are still pre v1. If we want to go ahead and tag v1 now (from the dev branch) before we merge this, I'm fine with that. If we do, I will add `BREAKING CHANGE` to the commit message body.

stephenmathieson: No need. The axe Linter folks just said this is fine (on Slack).

packages/react/src/components/Code/index.tsx
```diff
@@ -1,27 +1,44 @@
...
-  <SyntaxHighlighter {...props} style={theme}>
+const Code: React.ComponentType<Props> = ({
+  children,
+  className,
+  ...props
+}) => (
+  <SyntaxHighlighter
+    {...props}
+    useInlineStyles={false}
+    className={classNames('Code', className)}
```
stephenmathieson, line 30: 👍

stephenmathieson:
# REVIEWED FOR SECURITY 🚓 🎨

PR #79, opened 2020-09-11
fix(react): Adds aria-haspopup to OptionsMenuTrigger

schne324:
## Reviewer checks

- [ ] Follows the commit message policy, appropriate for next version
- [ ] Tested for accessibility
- [ ] Code is reviewed for security


packages/react/src/components/OptionsMenu/OptionsMenuTrigger.tsx
```diff
@@ -19,6 +19,7 @@ function OptionsMenuTriggerComponent({
   return (
     <button
       type="button"
+      aria-haspopup="true"
```
schne324, line 22: Let me do some testing here...I'm thinking we might want to `aria-haspopup="menu"` here

PR #80, opened 2020-09-11
chore: Prevent test output from being polluted with Icon exceptions

schne324:
## Reviewer checks

### DIFF
--- a/packages/react/src/components/Icon/index.tsx
+++ b/packages/react/src/components/Icon/index.tsx
@@ -16,6 +16,9 @@ function Icon({ label, className, type, ...other }: IconProps) {
   const [IconSVG, setIcon] = useState<React.ComponentType<any> | null>(null);
 
   useEffect(() => {
+    if (process.env.JEST_WORKER_ID !== undefined) {
+      return;
+    }
     import(`./icons/${name}.svg`)
       .then(icon => {
         setIcon(() => icon.default);
--- a/packages/react/src/components/Icon/index.tsx
+++ b/packages/react/src/components/Icon/index.tsx
@@ -16,9 +16,12 @@ function Icon({ label, className, type, ...other }: IconProps) {
   const [IconSVG, setIcon] = useState<React.ComponentType<any> | null>(null);
 
   useEffect(() => {
-    if (process.env.JEST_WORKER_ID !== undefined) {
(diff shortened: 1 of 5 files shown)


packages/react/src/components/Icon/index.tsx
```diff
@@ -16,6 +16,9 @@ function Icon({ label, className, type, ...other }: IconProps) {
   const [IconSVG, setIcon] = useState<React.ComponentType<any> | null>(null);
 
   useEffect(() => {
+    if (process.env.JEST_WORKER_ID !== undefined) {
```
stephenmathieson, line 19: What if we checked `NODE_ENV` instead?

```suggestion
    if (process.env.NODE_ENV === "test") {
```

tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "strictNullChecks": true,
     "strictFunctionTypes": true,
     "typeRoots": ["node_modules/@types"],
-    "jsx": "react"
+    "jsx": "react",
+    "types": ["node"]
```
stephenmathieson, line 13: Don't need this. Should remove it.

```suggestion
```

schne324: I'm pretty sure without it the [build script broke](https://app.circleci.com/pipelines/github/dequelabs/cauldron/180/workflows/7fcfcafd-4583-48af-8d3c-8b52021253cb/jobs/736) but I'll remove it and give it a try 🤷‍♂️ 

```
[!] (plugin typescript) Error: @rollup/plugin-typescript TS2580: Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i @types/node`.
src/components/Icon/index.tsx (21:9)
```

schne324: Yep. Removing `"types": ["node"]` broke the build: https://app.circleci.com/pipelines/github/dequelabs/cauldron/183/workflows/933ecaeb-99a4-47cc-941a-730939dcf718/jobs/747

stephenmathieson: Strange.. might have something to do with setting `typeRoots` (which is unnecessary most of the time), or using Babel to compile TypeScript. It's probably worth looking into this in the future. Not going to block this PR for that tho.

stephenmathieson (approved):

stephenmathieson (approved):

PR #88, opened 2020-09-25
feat: implement new versions of modals and alerts

scurker:
## Reviewer checks


schne324:
The focus indication on the header appears to be off: 
Screen Shot 2020-10-05 at 11 04 04 AM

awpearlm:
The way it's coded now with focus going all the way to the edge for the close X is fine

schne324 (approved):
LGTM -- looks like our alerts/modals don't technically "remove" focusable elements from the tab order outside of the modal but that issue pre-existed this patch. I will create a ticket to address this (maybe using `inert`)

PR #104, opened 2020-10-28
feat: support adding className to dialog header

schne324:
## Reviewer checks

- [ ] Follows the commit message policy, appropriate for next version
- [ ] Tested for accessibility
- [ ] Code is reviewed for security

### DIFF
--- a/packages/react/__tests__/src/components/Alert/index.js
+++ b/packages/react/__tests__/src/components/Alert/index.js
@@ -7,16 +7,23 @@ const defaults = { show: false, heading: { text: 'hi' } };
 
 test('returns null if passed a falsey "show" prop', () => {
   const alert = mount(<Alert {...defaults}>{'hello'}</Alert>);
-  console.log(alert.debug())
-  expect(alert.html()).toBe("");
+  expect(alert.html()).toBe('');
 });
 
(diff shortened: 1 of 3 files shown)


packages/react/__tests__/src/components/Alert/index.js
```diff
@@ -7,16 +7,23 @@ const defaults = { show: false, heading: { text: 'hi' } };
 
 test('returns null if passed a falsey "show" prop', () => {
   const alert = mount(<Alert {...defaults}>{'hello'}</Alert>);
-  console.log(alert.debug())
```
schne324, line 10: noticed this lingering log statement

packages/react/src/components/Dialog/index.tsx
```diff
@@ -42,6 +43,7 @@ export default class Dialog extends React.Component<DialogProps, DialogState> {
 
   static propTypes = {
     className: PropTypes.string,
+    headerClassName: PropTypes.string,
```
scurker, line 46: Why can't we use the className from the dialog?

schne324: looks like css module's `:global` will solve my specific problem so I am declining this PR

PR #125, opened 2020-11-20
feat: support any node type in toast children

schne324:
BREAKING CHANGE: children are no longer rendered within a direct-child span element

### DIFF
--- a/packages/styles/toast.css
+++ b/packages/styles/toast.css
@@ -49,7 +49,7 @@
   align-items: center;
 }
 
-.Toast__message > span {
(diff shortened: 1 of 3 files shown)


stephenmathieson (approved):
Thanks dude 👌

packages/styles/toast.css
```diff
@@ -49,7 +49,7 @@
   align-items: center;
 }
 
-.Toast__message > span {
```
schne324, line 52: very happy to see these weird span-specific styles go away 😄

PR #126, opened 2020-11-20
feat(styles): Adds full color palette

schne324:
Leaving this in draft for now...we should discuss how we have quite a few colors that are extremely similar and maybe we can consolidate. There is still plenty of work to do here once we finalize the palette (referencing the palette colors in other vars / component styles)


schne324:
closing in favor of #265

PR #129, opened 2020-12-01
 fix(react): allow tooltip to be hidden on escape keypress

scurker:
This addresses the dismissable portion of [WCAG 2.1 - 1.4.13: Content on Hover or Focus](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html).

### DIFF
--- a/packages/react/src/components/Tooltip/index.tsx
+++ b/packages/react/src/components/Tooltip/index.tsx
@@ -82,7 +82,11 @@ export default function Tooltip({
 
   useEffect(() => {
     const handleEscape = (event: KeyboardEvent) => {
-      if (event.key === 'Escape') {
+      if (
+        event.key === 'Escape' ||
+        event.key === 'Esc' ||
+        event.keyCode === 27
+      ) {
         setShowTooltip(false);
       }
     };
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Tooltip/index.tsx
```diff
@@ -80,6 +80,25 @@ export default function Tooltip({
     }
   }, [popperPlacement]);
 
+  useEffect(() => {
+    const handleEscape = (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
```
stephenmathieson, line 85: In Edge, this is `Esc`. If we support Edge, we should check for both:

```suggestion
      if (event.key === 'Escape' || event.key === 'Esc') {
```

stephenmathieson (changes requested):
LGTM with the exception of Edge support.

If we don't care about Edge, dismiss this review and I'll approve.

packages/react/src/components/Tooltip/index.tsx
scurker: Good call. We should, I'll implement a graceful fallback.

stephenmathieson (approved):

Issue #87, opened 2020-09-24
DEMO: Text fields take up 100% width


schne324:
the [`TextField`](https://cauldron.netlify.app/components/TextField) redesign added [`width: 100%;`](https://github.com/dequelabs/cauldron/pull/85/files#diff-952e7233b4dc492bd8645985deb94341R10) 

the uxpin has the width of text fields at `300px`

schne324:
After speaking with @awpearlm, this is expected. So the solution to this ticket is to add a container with a width of `300px` to the fields like so:

<img width="357" alt="Screen Shot 2020-09-24 at 8 22 52 AM" src="https://user-images.githubusercontent.com/3180185/94165536-2f689800-fe3f-11ea-992c-eb431b5504d6.png">

PR #130, opened 2020-12-02
fix(docs): set 300px width wrapper around textfields

noahcoop:
Issue resolution for https://github.com/dequelabs/cauldron/issues/87

### DIFF
--- a/packages/styles/forms.css
+++ b/packages/styles/forms.css
@@ -80,10 +80,6 @@ textarea.Field--has-error:focus,
   padding-left: var(--space-half);
 }
 
-.TextFieldContainer {
-  width: 300px;
-}
-
 .Field {
   display: block;
   position: relative;
--- a/packages/styles/forms.css
+++ b/packages/styles/forms.css
@@ -80,6 +80,10 @@ textarea.Field--has-error:focus,
   padding-left: var(--space-half);
 }
 
+.TextFieldContainer {
+  width: 300px;
+}
+
 .Field {
   display: block;
   position: relative;
--- a/docs/patterns/components/TextField/index.css
+++ b/docs/patterns/components/TextField/index.css
@@ -1,3 +1,7 @@
 .TextField button {
   margin: 0;
 }
+
+.TextField .TextFieldContainer {
+  width: 300px;
+}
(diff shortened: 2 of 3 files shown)


allisonkench:
Hello Cauldron team! Myself and @noahcoop are contributing to this project as an assignment for UM. Looking forward to hearing your feedback!

packages/styles/forms.css
```diff
@@ -80,6 +80,10 @@ textarea.Field--has-error:focus,
   padding-left: var(--space-half);
 }
 
+.TextFieldContainer {
```
schne324, line 83: `TextFieldContainer` is something specific to the demo/docs app and shouldn't be part of the actual `@deque/cauldron-styles` package.  All you need to do is move this css block to 
https://github.com/dequelabs/cauldron/blob/develop/docs/patterns/components/TextField/index.css

schne324 (changes requested):
:+1: there's just one minor thing needing to be addressed

packages/styles/forms.css
noahcoop: Awesome, thanks for the feedback. Moved the css to the correct file in the latest commit

schne324:
👍 thanks for the contribution! The change looks good I just need to figure out why circleci isn't running in this branch. I will get this merged once I sort that out

schne324 (approved):

schne324 (changes requested):
Sorry for the back-and-forth. I completely overlooked the fact that you updated the actual react component rather than just the demo app. You should update [the TextField documentation/demo component](https://github.com/dequelabs/cauldron/blob/develop/docs/patterns/components/TextField/index.js) instead of  `packages/react/src/components/TextField/index.tsx` because we intentionally leave our inputs at 100% without a max width container to make them work in various contexts.

schne324:
> You should update the TextField documentation/demo component instead of `packages/react/src/components/TextField/index.tsx`

I think all you'll need to do is back out the changes to `packages/react/src/components/TextField/index.tsx` and update the css to:

```css
.TextField form {
  max-width: 300px;
}
```

noahcoop:
Awesome, no worries re the back-and-forth. Just reverted the changes and updated the docs css

schne324 (approved):

schne324:
# Reviewed for security (gh action still isn't working)

Issue #103, opened 2020-10-27
Button Link variant should have pointer cursor


scurker:
One thing I noticed in the Link button variant is that it's the normal cursor instead of the pointer which is a minor difference from normal anchor links.

allisonkench:
@noahcoop and I fixed this for Link Buttons, but we also noticed that the rest of the buttons have normal cursors instead of the pointer. Should we also implement this for the other types of buttons? Or is there a reason that only the Link Buttons should have this?

scurker:
This was most recently changed with #42 with the reason being that it closely mirrors what browsers do natively. The one difference for links is that natively it appears with a pointer cursor, and the button link variant should as well.

I kind of agree that buttons should have visual pointer change on hover, but it's ultimately @awpearlm's call.

awpearlm:
I'm fine with it being a pointer. Also I'm not 100% on board with the the change from #42 tbh. While that is the native cursor that [buttons] use the arrow cursor, there is tons of practical application showing that the pointer is used for interactive elements universally. IMO we should be using the pointer icon for both buttons and links. Look at amazon, youtube, even deque.com all buttons and links use the pointer style.

allisonkench:
To clarify, does this mean we should change the PR to include making all buttons have the hand cursor on hover? We can make this change. Thanks!

schne324:
I don't have strong opinions on cursors tbh but I've definitely seen conversations on web-a11y slack where people have asserted that changing default cursors introduces understandability issues for users with cognitive issues. @allisonkench let's just change the link variant for now. Sounds like we need to discuss making the change for **all** buttons/links.

schne324:
this has been completed

PR #131, opened 2020-12-02
fix(styles): Adds pointer cursor to Button with variant link

noahcoop:
Resolution to issue https://github.com/dequelabs/cauldron/issues/103 (UM assignment with @allisonkench)

### DIFF
--- a/packages/styles/button.css
+++ b/packages/styles/button.css
@@ -15,7 +15,7 @@
   --button-hover-outline-color: var(--button-background-color-primary);
 }
 
-.Link {
+button.Link {
   cursor: pointer;
 }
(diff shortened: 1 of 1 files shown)


packages/styles/button.css
```diff
@@ -15,6 +15,10 @@
   --button-hover-outline-color: var(--button-background-color-primary);
 }
 
+.Link {
```
schne324, line 18:
```suggestion
button.Link {
```

schne324 (approved):

schne324:
# Reviewed for security (gh action is not working at the moment)

Issue #102, opened 2020-10-27
font-size of TextField should be var(--text-size-small)


schne324:
https://github.com/dequelabs/cauldron/blob/develop/packages/styles/variables.css#L98

PR #133, opened 2020-12-08
fix(styles): Updates font-size of text field to `--text-size-small`

noahcoop:
Resolve issue https://github.com/dequelabs/cauldron/issues/102

### DIFF
--- a/docs/patterns/components/TextField/index.css
+++ b/docs/patterns/components/TextField/index.css
@@ -1,7 +1,3 @@
 .TextField button {
   margin: 0;
 }
-
-.TextField .Field__text-input {
-  font-size: var(--text-size-small);
-}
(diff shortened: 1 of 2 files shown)


schne324 (changes requested):
Thank you for the contribution! Unlike your [other pr](https://github.com/dequelabs/cauldron/pull/130), we actually want this change to be a part of the core library. Basically, all we need is to update [this loc](https://github.com/dequelabs/cauldron/blob/752be9519630f6174ed5652fe60862ddef92346e/packages/styles/forms.css#L151)

change it form `font-size: var(--text-size-smaller);` to `font-size: var(--text-size-small);`

noahcoop:
Sounds good, just made a new commit with the font size change in the correct file.

schne324:
# Reviewed for security (gh action isn't working at the moment)

schne324 (approved):

PR #134, opened 2020-12-11
feat: Adds big Tooltip component

schne324:
see https://github.com/dequelabs/walnut/issues/1300

### DIFF
--- a/packages/styles/line.css
+++ b/packages/styles/line.css
@@ -1,0 +1,21 @@
+.Line {
+  border: none;
+  border-top: 1px solid #ccc;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  overflow: visible;
+  margin: 26px auto;
+  max-width: 300px;
+}
+
+.Line::after {
+  content: '';
+  display: inline-block;
+  margin-top: -4px;
+  width: 7px;
+  height: 7px;
+  border-radius: 50%;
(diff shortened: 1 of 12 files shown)


schne324:
Naming is hard...so I went with `variant='big'`. I'm open to something better if anyone can think of anything

dylanb (approved):

packages/styles/line.css
```diff
@@ -0,0 +1,21 @@
+.Line {
+  border: none;
+  border-top: 1px solid #ccc;
```
stephenmathieson, line 3: Should have used variables here.

packages/styles/line.css
```diff
@@ -0,0 +1,21 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  overflow: visible;
+  margin: 26px auto;
```
stephenmathieson, line 8: And here

packages/styles/line.css
```diff
@@ -0,0 +1,21 @@
...
+
+.Line::after {
+  content: '';
+  display: inline-block;
+  margin-top: -4px;
+  width: 7px;
+  height: 7px;
+  border-radius: 50%;
+  background: #fdfdfe;
+  border: 1px solid #ccc;
```
stephenmathieson, line 20: and here

Issue #52, opened 2020-04-20
link: support React Router's <Link/>


stephenmathieson:
I'm building an app that uses React Router. In order to get cauldron styles on my links, I need to use its `<Link/>` component. However, this breaks React Router as it expects me to use its `<Link/>` component.

It'd be nice if we could pass a `component={RouterLink}` prop [like Material UI does](https://material-ui.com/guides/composition/#link) to support this.

PR #136, opened 2020-12-13
React Router for Link

noahcoop:
Resolution to https://github.com/dequelabs/cauldron/issues/52

### DIFF
--- a/packages/react/src/components/Link/index.tsx
+++ b/packages/react/src/components/Link/index.tsx
@@ -2,10 +2,15 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
+// @ts-ignore
+import { BrowserRouter, Link as RouterLink } from 'react-router-dom';
+
 export interface LinkProps
   extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
   linkRef?: React.Ref<HTMLAnchorElement>;
+  reactRouter?: false;
   variant?: 'button';
+  useRouter?: false;
 }
 
 const Link = ({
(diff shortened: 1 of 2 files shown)


noahcoop:
Added prop to Link component called `useRouter`. This prop will return a react-router Link component instead of an a tag so that the user can use react router.

packages/react/src/components/Link/index.tsx
```diff
@@ -2,35 +2,56 @@ import React from 'react';
...
-      'Button--primary': variant === 'button'
-    })}
-    {...other}
-  >
-    {children}
-  </a>
-);
+}: LinkProps) =>
+  useRouter ? (
+    <BrowserRouter>
```
schne324, line 25: We don't want to include `BrowserRouter` for every single link...The idea of this feature would be to be able to:

```js
import { Link as CauldronLink } from '@deque/cauldron-react'
import { Link } from 'react-router-dom' 
 
const Foo = () => {
  return <CauldronLink component={Link} href="/about" />
}
```

schne324 (changes requested):
I think we want to go with a simpler approach (see comment)

packages/react/src/components/Link/index.tsx
stephenmathieson: Yeah, we can't include multiple routers here. This would break our apps.

We also don't want to _only_ support React Router. We might decide to build an app with [`Wouter`](https://github.com/molefrog/wouter) and need its [`<Link/>`](https://github.com/molefrog/wouter/#link-hrefpath), or maybe [Next.js's `<Link/>`](https://nextjs.org/docs/api-reference/next/link) instead.

Harris' approach of passing a `component={SomeLinkComponent}` prop seems like the best solution to me.

Issue #141, opened 2020-12-16
When an Icon is added as a child to a button element, it doesn't align with sibling elements.


okry:
It seems like the svg is not vertically centered within the Icon div. Possibly related to #137 

<img width="554" alt="Screen Shot 2020-12-16 at 9 12 14 AM" src="https://user-images.githubusercontent.com/25124347/102359447-cc3e5b80-3f7e-11eb-9e72-9193d62a07f6.png">

PR #142, opened 2020-12-17
feat(button): handle style icons as children inside of buttons

akornmeier:
- **add**: startIcon and endIcon props that will place an `Icon` in `Button` either before or after button text
- **add**: tests to verify proper elements are added when an Icon is passed to either prop
- **refactor**: update `Button` code to be a Stateless Functional Component

Closes: #141

### DIFF
--- a/packages/react/src/components/Button/index.tsx
+++ b/packages/react/src/components/Button/index.tsx
@@ -1,42 +1,16 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { ButtonHTMLAttributes, forwardRef } from 'react';
 import classNames from 'classnames';
 
-export interface ButtonProps
-  extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
   variant?: 'primary' | 'secondary' | 'error' | 'link';
-  buttonRef?: React.Ref<HTMLButtonElement>;
-  disabled?: boolean; // todo, investigate why disabled is needed
   thin?: boolean;
 }
 
-export default class Button extends React.Component<ButtonProps> {
-  static defaultProps = {
-    variant: 'primary',
-    buttonRef: () => {}
-  };
-
-  static propTypes = {
-    variant: PropTypes.oneOf(['primary', 'secondary', 'error', 'link']),
-    children: PropTypes.node,
-    className: PropTypes.string,
-    buttonRef: PropTypes.oneOfType([
-      PropTypes.func,
-      PropTypes.shape({ current: PropTypes.any })
-    ])
-  };
-
-  static displayName = 'Button';
-
-  render() {
-    const {
-      variant,
-      children,
-      className,
-      buttonRef,
-      thin,
-      ...other
-    } = this.props;
+const Button = forwardRef<HTMLButtonElement, ButtonProps>(
+  (
+    { variant = 'primary', thin, children, className, ...other }: ButtonProps,
+    buttonRef
+  ) => {
     return (
       <button
         type={'button'}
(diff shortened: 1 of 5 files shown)


packages/react/src/components/Button/index.tsx
```diff
@@ -1,42 +1,40 @@
...
+const Button = forwardRef<HTMLButtonElement, ButtonProps>(
+  (
+    {
+      variant = 'primary',
+      thin,
       children,
       className,
-      buttonRef,
-      thin,
+      startIcon,
```
scurker, line 18: I think we should favor composition over `propTypes` here. The feels way too prescriptive. The icon should naturally be placed in the button where it's needed:

```jsx
<Button>
  <Icon type="icon-type">
  Submit
</Button>
```

or

```jsx
<Button>
  Submit
  <Icon type="icon-type">
</Button>
```

It would be better to have inheritance styles [here](https://github.com/dequelabs/cauldron/blob/develop/packages/styles/button.css) so the icon would naturally inherit styles from the button. You're already on the right track with setting `flex` on the button, just need the appropriate styles for the icon.

```css
.Button--primary .Icon,
.Button--secondary .Icon,
.Button--clear .Icon,
.Button--error .Icon {
  /* icon styles */
}
```

packages/styles/button.css
```diff
@@ -80,6 +83,24 @@ button.Link {
...
+
+.Button--end-icon {
+  display: inherit;
+  margin-right: -4px;
+  margin-left: 4px;
+}
+
+.Button--thin .Button--start-icon svg,
+.Button--thin .Button--end-icon svg {
+  width: 15px;
```
scurker, line 100: This should be calculated from the button height. Something like:

```css
width: calc(var(--button-thin-height) - 8px);
height: calc(var(--button-thin-height) - 8px);
```

packages/styles/button.css
```diff
@@ -80,6 +83,24 @@ button.Link {
   background-color: var(--button-background-color-error-active);
 }
 
+.Button--start-icon {
+  display: inherit;
+  margin-right: 4px;
```
scurker, line 88: What are the margins used for? Is there any way we can avoid this?

packages/react/src/components/Button/index.tsx
```diff
@@ -1,42 +1,40 @@
...
-      PropTypes.shape({ current: PropTypes.any })
-    ])
-  };
-
-  static displayName = 'Button';
-
-  render() {
-    const {
-      variant,
+const Button = forwardRef<HTMLButtonElement, ButtonProps>(
```
scurker, line 11: I'm in favor of using `forwardRef` here, but this is a breaking change. I think we should eventually change this, but I don't think it should be a part of this PR.

scurker (changes requested):

akornmeier:
Update: 
- Removed Icon related props in favor of composition
- Added support for both buttonRef and forwardRef in component (buttonRef will trigger a console.warn deprecated message)
- Updated css to handle the positioning/alignment of the <Icon> element passed in as children
- updated tests

packages/react/src/components/Button/index.tsx
```diff
@@ -1,42 +1,31 @@
...
-    } = this.props;
+    }: ButtonProps,
+    ref
+  ) => {
+    if (buttonRef) {
+      console.warn(
+        "%c Warning: 'buttonRef' prop is deprecated, please use 'ref'. ",
+        'background: #222; color: #bada44'
+      );
+    }
```
scurker, line 27: Let's leave out the `console.warn` here.

scurker (changes requested):
One small nitpick

packages/react/src/components/Button/index.tsx
akornmeier: removed in the latest commit

scurker (approved):

PR #143, opened 2020-12-21
feat(react): allow toast auto-focus to be disabled

### LINKED ISSUE #32: Support not automatically focusing toast
Add a prop `focus` that defaults to true. When falsy, a role of alert should be added (but still allowed to be overridden)

scurker:
closes #32


schne324 (approved):
LGTM

PR #150, opened 2021-01-05
Refactor: functional checkbox

okry:
- Refactors Checkbox to use functional React
- Converts component to use `forwardRef`
- Deprecates `checkboxRef`
- Accepts `onChange` prop to give user access to `onChange` functionality like a semantic element


schne324:
as @scurker said #142  (https://github.com/dequelabs/cauldron/pull/142#discussion_r545437593) I am in favor of these forward ref changes but it should probably be an all-at-once thing rather than piecemeal

schne324:
I'm also not a fan of polluting everyone's console with a warning 🤷

okry:
> as @scurker said #142 ([#142 (comment)](https://github.com/dequelabs/cauldron/pull/142#discussion_r545437593)) I am in favor of these forward ref changes but it should probably be an all-at-once thing rather than piecemeal

That would be rather hard though since implementing forward refs would require to change all of the class based React to functional React. While it's not ideal, I feel like it's a lot more feasible converting them on a per component basis.

schne324 (approved):

PR #153, opened 2021-01-07
fix: move icon.css to styles package

okry:
The `index.css` file for the Icon component in the `React` package was previously not being exported by Cauldron. I have moved it into the `Styles` package so that it will be exported in `@deque/cauldron-styles`.


okry:
Needs further discussion.

PR #154, opened 2021-01-07
fix: make icons inside IconButtons flex to fit button container

okry:
- Icon SVGs now inherit height and width of wrapping div element.
- IconButtons make SVG div wrappers expand/contract to fit the button container wrapping them
- Had to remove padding from IconButtons since it was causing the interior icons to be offset by a significant amount

New hover indicator:
Screen Shot 2021-01-11 at 10 48 13 AM

New focus indicator:
Screen Shot 2021-01-11 at 10 47 45 AM

These indicators were taken from what we were using for the vanilla buttons and were tested to work with stretched buttons.

### DIFF
--- a/packages/styles/icon-button.css
+++ b/packages/styles/icon-button.css
@@ -10,7 +10,7 @@
   color: var(--icon-button-icon-color);
   margin: 2px;
   padding: 0;
-  --button-hover-outline-color: var(--header-text-color-dark);
+  --button-hover-outline-color: var(--top-bar-text-color);
 }
 
 .IconButton div {
--- a/packages/styles/icon-button.css
+++ b/packages/styles/icon-button.css
@@ -9,11 +9,12 @@
   background-color: var(--icon-button-background-color);
   color: var(--icon-button-icon-color);
   margin: 2px;
+  padding: 0;
 }
 
-.IconButton svg {
-  height: calc(var(--button-thin-height) - 4px);
-  width: calc(var(--button-thin-height) - 4px);
+.IconButton div {
+  width: 100%;
+  height: 100%;
 }
 
 .IconButton:before {
--- a/packages/styles/icon-button.css
+++ b/packages/styles/icon-button.css
@@ -13,7 +13,7 @@
   padding: 0;
 }
 
-.IconButton div {
+.IconButton .Icon {
   width: 100%;
   height: 100%;
 }
--- a/packages/styles/icon-button.css
+++ b/packages/styles/icon-button.css
@@ -11,7 +11,6 @@
   color: var(--icon-button-icon-color);
   margin: 2px;
   padding: 0;
-  --button-hover-outline-color: var(--top-bar-text-color);
 }
 
 .IconButton div {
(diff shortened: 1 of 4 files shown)


scurker (changes requested):
The outlines for the focus/hover states are now offset. 

react cauldron icon in a hover state with an offset hover border

react cauldron icon in a focus state with an offset focus border

I think one of the things that might make this difficult, is the specificity of the hover states depend on a known height/width of the button.

schne324:
it looks like that focus ring issue already exists: https://cauldron.netlify.app/components/IconButton but I agree we should fix it

scurker:
> it looks like that focus ring issue already exists: https://cauldron.netlify.app/components/IconButton but I agree we should fix it

???

This is what I see on https://cauldron.netlify.app/components/IconButton:
image

There's a 1px spacing around the button icon, which doesn't exist in this PR. 

For comparison:

image

okry:
I've updated the indicators to reflect Aaron's designs by taking from what we were using for the other buttons. Updated icons are in the main description.

scurker (changes requested):
One more small minor change: Add `pointer-events: none` to the `.IconButton:before` pseudo element. We don't want the pseudo element to be triggering click events or activating when it's not supposed to. `pointer-events: none` should prevent both of these behaviors.

packages/styles/icon-button.css
```diff
@@ -1,42 +1,42 @@
...
   height: var(--button-thin-height);
   width: var(--button-thin-height);
-  border: 1px solid transparent;
   border-radius: 3px;
+  border: none;
   background-color: var(--icon-button-background-color);
   color: var(--icon-button-icon-color);
   margin: 2px;
+  padding: 0;
+  --button-hover-outline-color: var(--top-bar-text-color);
```
stephenmathieson, line 14: This looks like a typo. I don't think `--button-hover-outline-color` is a valid CSS property.

stephenmathieson (changes requested):

packages/styles/icon-button.css
okry: Wasn't a typo, it's a custom key that we were using for the hover outline. I essentially stole this methodology from the vanilla button css. Since it isn't repeated, I'll remove it for now.

stephenmathieson: CSS is crazy 🤯

packages/styles/icon-button.css
```diff
@@ -1,42 +1,42 @@
...
   background-color: var(--icon-button-background-color);
   color: var(--icon-button-icon-color);
   margin: 2px;
+  padding: 0;
 }
 
-.IconButton svg {
-  height: calc(var(--button-thin-height) - 4px);
-  width: calc(var(--button-thin-height) - 4px);
+.IconButton div {
```
scurker, line 16: One more small thing I noticed, do you think this should be more specific? i.e. `.IconButton .Icon`? 

`div` seems a bit too generic.

okry: Fixed

scurker (approved):

PR #161, opened 2021-01-11
feat(react): fire custom show/hide event when tooltip is shown/hidden

schne324:


scurker (approved):

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -17,6 +17,23 @@ export interface TooltipProps extends React.HTMLAttributes<HTMLDivElement> {
...
 }
 
+// fires a custom "cauldron:tooltip:show" / "cauldron:tooltip:hide" event
+// to allow projects using cauldron to hook into when a tooltip is shown/hidden
+const fireCustomEvent = (show: boolean, button?: HTMLElement | null) => {
+  if (!button) {
+    return;
+  }
+
+  const event = new Event(
```
scurker, line 27: LGTM! I wonder if this should (eventually) be an event factory.

Something like `new CauldronEvents('tooltip:show', ...details)`. We can wait to refactor when implementing #162

Issue #162, opened 2021-01-11
Cauldron component events


scurker:
Cauldron components should fire custom events on specific interactions to allow for implementers to handle global side effects:

* Alert/Modal/Dialog (do these need to be specific events?)
  * `cauldron:dialog:show`
  * `cauldron:dialogt:hide`
* ExpandCollapsePanel
  * `cauldron:expandcollapsepanel:expand`
  * `cauldron:expandcollapsepanel:collapse`  
* Tooltip (implemented with #161)
  * `cauldron:tooltip:show`
  * `cauldron:tooltip:hide`

PR #163, opened 2021-01-12
feat(react): fire custom events for components

schne324:
see #162


packages/react/src/utils/cauldron-event.ts
```diff
@@ -0,0 +1,8 @@
+const cauldronEvent = (type: string, details?: EventInit) => {
```
schne324, line 1: @scurker, this is a bit different than [what you suggested](https://github.com/dequelabs/cauldron/pull/161/files#r555412328)...what do you think about this util?

scurker: Yeah, this is fine. I was just shooting off some psuedo code

scurker (approved):

PR #167, opened 2021-01-12
fix(react): preventDefault on tooltip close

okry:
Prevents tooltip close event from propagating.
This is important for scenarios where the tooltip in inside an `escape` sensitive component like a modal.


scurker (changes requested):
I'm a little torn on this, generally I think you should avoid `event.stopPropagation()` as it can create side effects that are extremely hard to debug and prevents ancestor elements from handling that specific event. I understand we're trying to prevent a specific side effect here with modal, but I think this is the wrong approach.

There's a couple of different ways we could handle this:

* Detect this specific event in modal
  * Using a custom event
  * Using the event target (to detect that it was the tooltip that triggered this event) 
* Use `event.preventDefault()`

Of these options, I feel like `event.preventDefault()` is the most optimal. Event cascading still works as expected, AND we can detect the event inside of the modal.

Example:

```js
// tooltip event
if (showTooltip) {
  setShowTooltip(false)
  event.preventDefault()
}
```

```js
// modal event handler
if (event.defaultPrevented) {
  // don't close the modal
}
```

This more naturally protects event propagation and uses existing event behavior to fill our needs. What do you think?

okry:
The only problem with that is what if you're using a component that has an `Escape` listener that isn't exposed to users?

okry:
One alternative is that we could expose the `onKeyDown` event so that users can modify what happens with this component when certain keys are pressed.

scurker:
> The only problem with that is what if you're using a component that has an `Escape` listener that isn't exposed to users?

I think we should be checking for `event.defaultPrevented` anywhere we're using an event that we would want to potentially be cancelled.

scurker (commented):
Changes LGTM. One question though:

Don't we want to check for this in `Dialog`?

```tsx
// dialog.tsx
close(event) {
  if (event.isDefaultPrevented) return;
  ...
}
```

okry:
Fixing that might be a bit difficult with how it interacts with focusTrap due to how the `escapeDeactivates` prop works ```<FocusTrap
        focusTrapOptions={{
          onDeactivate: this.close,
          escapeDeactivates: !forceAction,
          fallbackFocus: '.Dialog__heading'
        }}
      >```

It might just have to be something we would handle when we plan on bringing in Radix and using their dialog.

scurker (commented):
Change LGTM. Please update the title to be descriptive since it will go into our changelog.

scurker (changes requested):
Last change (I promise) 😅 

We probably want to verify in the test that we're calling `preventDefault`

scurker:
Closing as I'm not sure we need this at the moment. We can raise a new PR when we encounter an issue.

PR #171, opened 2021-01-14
fix(react): Remove onClose prop from Alert

### LINKED ISSUE #105: onClose should not be a required (or supported) prop in the `<Alert />` component

stephenmathieson:
The inherited `onClose` prop doesn't make sense in the `Alert` component. The patch from #168 made it optional, enabling this patch to remove it entirely.

Closes #105

### DIFF
--- a/packages/react/src/components/Alert/index.tsx
+++ b/packages/react/src/components/Alert/index.tsx
@@ -3,7 +3,7 @@ import classnames from 'classnames';
 import Icon from '../Icon';
 import { Dialog, DialogContent, DialogFooter, DialogProps } from '../Dialog';
 
-interface AlertProps extends Omit<DialogProps, 'forceAction'> {
+interface AlertProps extends Omit<Omit<DialogProps, 'forceAction'>, 'onClose'> {
   variant?: 'default' | 'warning';
 }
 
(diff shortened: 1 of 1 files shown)


scurker (dismissed):

packages/react/src/components/Alert/index.tsx
```diff
@@ -3,7 +3,7 @@ import classnames from 'classnames';
 import Icon from '../Icon';
 import { Dialog, DialogContent, DialogFooter, DialogProps } from '../Dialog';
 
-interface AlertProps extends Omit<DialogProps, 'forceAction'> {
+interface AlertProps extends Omit<Omit<DialogProps, 'forceAction'>, 'onClose'> {
```
scurker, line 6: Not a big deal, but I believe you can also do `Omit<DialogProps, 'forceAction' | 'onClose'>`

stephenmathieson: Oh neat! I'll give it a try.

stephenmathieson:
I noticed that the [demo site makes use of the `onClose` prop](https://github.com/dequelabs/cauldron/blob/35dd790f6482c48ef9a59a310fad97509c80cbe6/docs/patterns/components/Alert/index.js#L30), so I'm not sure we do want to remove it. However, I think the linked issue (#105) suggests otherwise.

@schne324 did you want the `onClose` prop removed ("not supported"), or was making it optional sufficient?

schne324:
> @schne324 did you want the onClose prop removed ("not supported"), or was making it optional sufficient?

making it optional is technically sufficient but I agree that we can just remove this entirely (for alerts you have to manage the show state yourself because it isn't generically dismissible with an `x` button)

schne324 (approved):

schne324:
closing in favor of #180

PR #174, opened 2021-01-14
fix: use Pattern Library lock icon

okry:
We were using an version of the lock icon that was in a different format from our other icons. Updated lock.svg to use FontAwesome like everything else.
Screen Shot 2021-01-14 at 3 48 33 PM


scurker:
The icons should come from our [pattern library](https://preview.uxpin.com/5ca5c040146a4ebb4536c10998dd84d1199e5d79#/pages/134488442/specification/sitemap) not font awesome. There might be some outdated ones, but I don't think the lock icon is one of them.

okry:
@scurker It seems like the Icon was outdated and missing attributes. I have updated it to use the pattern library svg now which was different to what we had before this PR.

scurker:
It's not a 1:1 mapping I usually have to do a little bit of cleanup when importing an svg, React can't do `xmlns:` attributes so the `xmlns` isn't needed.

okry:
That's weird, we were having issues scaling the SVG when the `xmlns` attribute was missing.

scurker:
My understanding of the HTML spec is the only time you need the `xmlns` on an svg is if it's the root element.

SVGs in cauldron embed the svg inline so I don't think the `xmlns` is needed.

okry:
I see, so should we remove it from the PR? Most of the other icons have the `xmlns` attribute in them.

scurker:
IMO, I don't think we need the xmlns here but I'm not going to hold up the PR. It looks like we have a mixed bag of SVGs, so we should probably go in and clean them up at some point.

scurker (approved):

PR #180, opened 2021-01-29
fix(react): remove unsupported props for alert component

### LINKED ISSUE #105: onClose should not be a required (or supported) prop in the `<Alert />` component

scurker:
closes #105


stephenmathieson (changes requested):
The demo site currently uses these props and suggests to users that they should/can:

- https://github.com/dequelabs/cauldron/blob/develop/docs/patterns/components/Alert/index.js#L30
- https://github.com/dequelabs/cauldron/blob/develop/docs/patterns/components/Alert/index.js#L44
- https://github.com/dequelabs/cauldron/blob/develop/docs/patterns/components/Alert/index.js#L79
- https://github.com/dequelabs/cauldron/blob/develop/docs/patterns/components/Alert/index.js#L93

We should either remove them from the documentation site or leave them.

Also this is a duplicate of https://github.com/dequelabs/cauldron/pull/171 (I forgot to pick this back up/finish it).

scurker:
> The demo site currently uses these props and suggests to users that they should/can:
> 
> ...
> 
> We should either remove them from the documentation site or leave them.

Ah frick, forgot to check demo. This could've been caught if the demo site was in typescript. 🤦

schne324 (approved):

schne324:
🎉 this means we can finally tag v1 🎊

PR #181, opened 2021-01-29
feat(react): implement icon button variants

scurker:
closes #177 

This should bring `<IconButton>` in pretty close alignment with `<Button>`.

PR #184, opened 2021-02-01
feat: add disabled styles to IconButtons

okry:
Utilizes the disabled styles that `SecondaryButton` uses. We will have to add more states when we institute more `IconButton` variants.

Disabled Buttons:
Primary- 
Screenshot 2021-04-06 at 1 43 39 PM
Secondary- 
Screenshot 2021-04-06 at 1 43 58 PM
Error- 
Screenshot 2021-04-06 at 1 44 21 PM
Light- 
Screenshot 2021-04-06 at 1 44 37 PM
Dark- 
Screenshot 2021-04-06 at 1 44 52 PM


scurker:
I started to work on the variants in #181, perhaps this change should be included there?

okry:
Yeah, that would probably be a good idea since this doesn't hold up well when a user puts in a variant icon button.

schne324 (approved):

PR #185, opened 2021-02-03
feat: allow pointouts to specify their positioning in alignment with their target element

scurker:
This is needed for the extension to address some positioning issues with pointouts, and to allow for some more flexibility in where the pointout is.


schne324 (approved):
changes LGTM. You don't have to necessarily do this here in this PR but it'd be nice to get some documentation on `arrowPosition` and `position`

schne324:
foo testin123

PR #193, opened 2021-02-12
feat: Adds tag component

### LINKED ISSUE #176: Implement tags
[UXPin (Pattern Library)](https://preview.uxpin.com/5ca5c040146a4ebb4536c10998dd84d1199e5d79#/pages/134488442/specification/sitemap)

cauldron tags pattern

schne324:
resolves #176

---

tag component unit test coverage 100%

### DIFF
--- a/docs/patterns/components/Tag/index.js
+++ b/docs/patterns/components/Tag/index.js
@@ -1,0 +1,36 @@
+import React, { Component } from 'react';
+import Demo from '../../../Demo';
+import { Tag, TagLabel } from '@deque/cauldron-react/';
+import { children, className } from '../../../props';
+
+export default class LinkDemo extends Component {
+  render() {
+    return (
+      <div>
+        <Demo
+          component={Tag}
+          states={[
+            {
+              children: 'Value'
+            },
+            {
+              children: (
+                <>
+                  <TagLabel>Label: </TagLabel>value
+                </>
+              )
+            }
+          ]}
+          propDocs={{
+            children: {
+              ...children,
+              required: true
+            },
+            className
+          }}
+        />
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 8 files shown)


packages/styles/variables.css
```diff
@@ -152,4 +152,9 @@
 
   /* list */
   --list-separator: rgba(153, 153, 153, 0.15);
+
+  /* tags */
+  --tag-text-color: #333333;
```
schne324, line 157: we still haven't implemented our full color palette so for now, I'm adding new vars for tags...once we implement our palette I can replace references to these with the palette var names

stephenmathieson: Is there an issue for adding the remaining variables?

docs/patterns/components/Tag/index.js
```diff
@@ -0,0 +1,36 @@
...
+      <div>
+        <Demo
+          component={Tag}
+          states={[
+            {
+              children: 'Value'
+            },
+            {
+              children: (
+                <>
```
stephenmathieson, line 18: What's the Fragment for?

stephenmathieson (approved):

docs/patterns/components/Tag/index.js
schne324: there are 2 sibling reactNode's here: 
- `<TagLabel />`
- "value" (string)

packages/styles/variables.css
schne324: yes I believe so...there's a draft PR too

docs/patterns/components/Tag/index.js
stephenmathieson: Ah yep, I thought it was a child of the `<TagLabel>`. My bad 👍

PR #199, opened 2021-03-01
feat(icon-button): exposes ref prop

akornmeier:
*update*: icon button ref was just internal for the tooltip, adds ability to use ref freely

### DIFF
--- a/packages/react/src/components/IconButton/index.tsx
+++ b/packages/react/src/components/IconButton/index.tsx
@@ -1,52 +1,50 @@
-import React, { useRef } from 'react';
-import PropTypes from 'prop-types';
+import React, { useRef, forwardRef } from 'react';
 import classnames from 'classnames';
 import { Placement } from '@popperjs/core';
 import Icon from '../Icon';
 import Tooltip from '../Tooltip';
-
 export interface IconButtonProps
   extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   icon: string;
   label: string;
   tooltipPlacement?: Placement;
 }
-
-export default function IconButton({
-  icon,
-  label,
-  tooltipPlacement = 'auto',
-  className,
-  ...other
-}: IconButtonProps) {
-  const buttonRef = useRef<HTMLButtonElement>(null);
-  return (
-    <React.Fragment>
-      <button
-        type={'button'}
-        className={classnames('IconButton', className)}
-        ref={buttonRef}
-        {...other}
-      >
-        <Icon type={icon} />
-      </button>
-      <Tooltip
-        target={buttonRef}
-        placement={tooltipPlacement}
-        association="aria-labelledby"
-        hideElementOnHidden
-      >
-        {label}
-      </Tooltip>
-    </React.Fragment>
-  );
-}
-
+const IconButton = forwardRef<HTMLButtonElement, IconButtonProps>(
+  (
+    {
+      icon,
+      label,
+      tooltipPlacement = 'auto',
+      className,
+      ...other
+    }: IconButtonProps,
+    ref
+  ): JSX.Element => {
+    const buttonRef = useRef<HTMLButtonElement>(null);
+    if (typeof ref === 'function') {
+      ref(buttonRef.current);
+    }
+    return (
+      <React.Fragment>
+        <button
+          type={'button'}
+          className={classnames('IconButton', className)}
+          ref={typeof ref === 'function' || !ref ? buttonRef : ref}
+          {...other}
+        >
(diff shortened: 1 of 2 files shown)


packages/react/src/components/IconButton/index.tsx
```diff
@@ -1,52 +1,50 @@
...
+      tooltipPlacement = 'auto',
+      className,
+      ...other
+    }: IconButtonProps,
+    ref
+  ): JSX.Element => {
+    const buttonRef = useRef<HTMLButtonElement>(null);
+    if (typeof ref === 'function') {
+      ref(buttonRef.current);
+    }
```
scurker, line 26: I think we may want to use [`useImperitiveHandle`](https://reactjs.org/docs/hooks-reference.html#useimperativehandle) here:

```jsx
useImperitiveHandle(ref, () => buttonRef.current)
```

That way the `ref` passed as props below can continue to use the combined ref value, whether it was passed from the parent as a ref or functional ref, or not.

scurker (changes requested):

packages/react/src/components/IconButton/index.tsx
akornmeier: What is the difference between the imperativeHandle and what is implemented in the PR? The TS compiler really REALLY does not like: ```useImperativeHandle(ref, () => buttonRef.current)```

akornmeier: From what i see `useImperativeHandle` is more suited to provide a subset of (in this case ) <HTMLButton> properties to the forwardRef like "focus". I would think we want to expose the full ref, which is what is implemented.

akornmeier: ah, nevermind... i got this to compile, changes inbound and down.

packages/react/src/components/IconButton/index.tsx
```diff
@@ -1,52 +1,55 @@
...
+  }
+);
 IconButton.displayName = 'IconButton';
-
-IconButton.propTypes = {
-  icon: PropTypes.string,
-  label: PropTypes.string,
-  tooltipPlayemnt: PropTypes.string,
-  buttonRef: PropTypes.any
-};
```
scurker, line 52: I don't care about `propTypes` here as I find them pretty useless, but others have complained when they were removed. cc @schne324

packages/react/src/components/IconButton/index.tsx
scurker: FWIW, I was able to get it working with this as well:

```tsx
function IconButton({
  icon,
  label,
  tooltipPlacement = 'auto',
  className,
  ...other
}: IconButtonProps, ref: React.Ref<HTMLButtonElement>) {
  const buttonRef = useRef<HTMLButtonElement>(null);
  useImperativeHandle(ref, () => buttonRef.current as HTMLButtonElement)
  return (...)
}
```

What you have is probably fine too.

scurker (approved):
Approving, but I'd like @schne324's to weigh in on removing the `propTypes` as well.

packages/react/src/components/IconButton/index.tsx
akornmeier: prop types were being problematic with forwardRef + Typescript, and they become extraneous

akornmeier: also: `tooltipPlayemnt`

akornmeier: I can put them back, however they were not even properly enforcing required props.

schne324:
> to weigh in on removing the propTypes as well.

@stephenmathieson iirc you had some concerns with removing them. Thoughts here?

packages/react/src/components/IconButton/index.tsx
schne324: Put them back, the rest of the library has them so for now (even though I think they're pointless) let's keep them

akornmeier: word. i'll fix them up to enforce the required props n stuff

akornmeier: Added the PropTypes object back.

schne324 (approved):

stephenmathieson:
> @stephenmathieson iirc you had some concerns with removing them. Thoughts here?

Only reason I suggest we keep them around is for non-TypeScript users. I agree that they're pointless if you're using TS, but for JS apps, they're pretty handy.

Not a big deal to me either way tho 🤷

akornmeier:
valid point.

PR #209, opened 2021-03-24
docs: Adds contributing guidelines

schne324:


akornmeier (commented):
This is a good start. I would like us to address some of the following in the contributing guide: 

- ref vs forwardRef preference in components moving forward
- including prop types for non-typescript users

schne324:
> ref vs forwardRef preference in components moving forward

👍 I will document that it is preferred that we use `forwardRef`

> including prop types for non-typescript users

@stephenmathieson, you were an advocate of leaving propTypes for non-typescript users...is that something you still stand by?

stephenmathieson:
> @stephenmathieson, you were an advocate of leaving propTypes for non-typescript users...is that something you still stand by?

Yes

akornmeier (approved):
Looks good, we can add more context to some of these topics in the future if we think it is needed.

erik-deque:
Review for security, button is currently broken.

PR #232, opened 2021-04-30
feat(react): allows RadioGroup component to be 'controlled'

### LINKED ISSUE #231: Allow radio buttons to be controlled
Currently, `<RadioGroup />` does not support programmatically controlling the selected radios. Similar to our `<TextField />` component, we should allow for both controlled and uncontrolled radio groups.

schne324:
Closes #231


packages/react/__tests__/src/components/Alert/index.js
```diff
@@ -7,16 +7,23 @@ const defaults = { show: false, heading: { text: 'hi' } };
 
 test('returns null if passed a falsey "show" prop', () => {
   const alert = mount(<Alert {...defaults}>{'hello'}</Alert>);
-  console.log(alert.debug())
```
schne324, line 10: a log was accidentally checked in. The rest of the diff in this file is a result of prettier not being run on this file 😦

PR #234, opened 2021-05-03
feat(react): allows select component to be controlled

### LINKED ISSUE #233: Allow `<Select />` to be "controlled" (via `value` prop)

schne324:
Closes #233

### DIFF
--- a/packages/react/src/components/Select/index.tsx
+++ b/packages/react/src/components/Select/index.tsx
@@ -79,6 +79,8 @@ const Select = React.forwardRef(
       dynamicProps.defaultValue = defaultValue;
     }
 
+    // In order to support controlled selects, we
+    // have to attach an `onChange` to the select.
     /* eslint-disable jsx-a11y/no-onchange */
     return (
       <div className="Field__select">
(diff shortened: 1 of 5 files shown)


packages/react/src/components/Select/index.tsx
```diff
@@ -14,19 +14,74 @@ export interface SelectProps
   label: string;
   options?: SelectOption[];
   children?: React.ReactElement<HTMLOptionElement | HTMLOptGroupElement>[];
+  value?: any;
```
schne324, line 17: I'm copying [material-ui](https://material-ui.com/api/select/#props) in allowing `any` on `value` and `defaultValue`

packages/react/src/components/Select/index.tsx
```diff
@@ -14,19 +14,74 @@ export interface SelectProps
...
+    }: SelectProps,
     ref: Ref<HTMLSelectElement>
   ): React.ReactElement<HTMLSelectElement> => {
     if (options && children) {
       console.error(
         'The Select component only takes the options props or child option elements, not both.'
       );
     }
+
+    const isControlled = typeof value !== 'undefined';
```
scurker, line 45: This shouldn't hold up the PR in any way as I'm just thinking out loud: Do we want to account for `null` values?

scurker (approved):

packages/react/src/components/Select/index.tsx
schne324: > Do we want to account for `null` values?

Good question. When in doubt I check and see what other react libraries do. I can follow up with a new ticket to address this once I do some research.

PR #254, opened 2021-05-24
chore: adds shell script to create files for new component

schne324:


schne324:
If this is something people find useful - awesome! If not, I can just close this PR 🤷

schne324:
sounds like no one cares about this  - closing

PR #255, opened 2021-05-25
feat: Adds simple table

schne324:
This is a minimal implementation of the cauldron table (#202)

what this does **NOT** support:
- sortability
- responsiveness (the "mobile view" table repeater thing from the uxpin)

### DIFF
--- a/docs/index.js
+++ b/docs/index.js
@@ -19,6 +19,7 @@ import {
 } from '@deque/cauldron-react';
 import logo from './assets/img/logo.svg';
 import 'fontsource-roboto';
+import 'fontsource-lato';
 
 // styles
 import '../packages/styles';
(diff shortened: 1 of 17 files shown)


docs/index.js
```diff
@@ -19,6 +19,7 @@ import {
 } from '@deque/cauldron-react';
 import logo from './assets/img/logo.svg';
 import 'fontsource-roboto';
+import 'fontsource-lato';
```
scurker, line 22: Interesting, tables are using a different font?

scurker (approved):

docs/index.js
schne324: yea, the `<TableHeader />` s are `Lato` according to the uxpin

PR #258, opened 2021-05-27
feat: Adds DescriptionList components

schne324:
I think it'd be cool if we added some "recipes" to the docs site. It'd be a good place to show the `<Table />` into `<DescriptionList />` into `<DescriptionList collapsed />` (going from wide to narrow viewport). For now, I've implemented this in the [demo component's prop docs](https://github.com/dequelabs/cauldron/pull/258/files#diff-1fbca86d25fcd8f078b5075cf4a63c1ebbed3362233a0a73034120936132481d).

- [x] design implementation reviewed/approved by @bobbyomari


schne324:
https://user-images.githubusercontent.com/3180185/119770462-14887080-be71-11eb-9623-ef24ff507931.mov

packages/styles/utilities.css
```diff
@@ -70,3 +70,9 @@ body .is--hidden {
   display: flex;
   align-items: center;
 }
+
+ul.semantic-only {
```
stephenmathieson, line 74: This class doesn't seem to follow the existing patterns. Is this something we care about?

stephenmathieson (approved):
Nice 👍

packages/styles/utilities.css
schne324: naming is hard. I'm open to suggestions 😄

stephenmathieson: I donno, maybe `SemanticList`? Could even be turned into a component:

```
<SemanticList ordered={false} items={[ 'foo', 'bar' }) />
```

Not sure how useful it'd actually be tho 🤷‍♂️

No need to change this if you don't think it matters. I certainly don't think it does, just something I figured I'd ask about so others could give their opinions.

PR #265, opened 2021-06-02
feat(styles): Adds full color palette

schne324:
see [colors documented in uxpin](https://preview.uxpin.com/5ca5c040146a4ebb4536c10998dd84d1199e5d79\#/pages/133873611/simulate/no-panels)

implements all the colors aside from the 'Corrected derivative shades/tints' colors :shrug: afaict we don't use those anywhere


packages/styles/variables.css
```diff
@@ -5,57 +5,87 @@
...
 :root {
+  /* color palette */
+  --white: #ffffff;
+  --gray-10: #fdfdfe;
+  --gray-20: #f2f2f2;
+  --gray-30: #e0e0e0;
+  --gray-40: #cccccc;
+  --gray-60: #666666;
+  --gray-80: #4d4d4d;
+  --gray-90: #333333;
```
scurker, line 16: We should be sure to cross check the variables in the axe extension. I think we define them after cauldron so may not be an issue, but will likely want to clean them up in our extension styles once this lands.

scurker (dismissed):

packages/styles/variables.css
schne324: 👍  here are the conflicts in extension:

```
--gray-20: #cccccc; /* in updated palette this is #f2f2f2 (--gray-40 is ccc) */
--gray-80: #333333; /* in updated palette this is #4d4d4d (--gray-90 is 333) */
--gray-90: #02101a; /* in updated palette this is #333333 (--accent-dark is #02101a) */
```

scurker (approved):

PR #292, opened 2021-07-22
feat(react): Adds theme provider

### LINKED ISSUE #283: Allow for simple switching between light and dark modes
completely open to feedback but here are our initial thoughts...

Create a generic `<CauldronProvider />` which accepts a `theme` prop ("light" or "dark"). Under-the-hood this will actually instantiate a `<ThemeProvider />` and pass the theme prop through. This leaves us open to future enhancements that could benefit from a provider-like state.  Consumers can still directly use `<ThemeProvider />` if they want.

## Prior art

- https://github.com/chakra-ui/chakra-ui/blob/cb027c13b52059779931cf1874c496ae8253ef11/packages/react/src/chakra-provider.tsx

schne324:
resolves #283

### DIFF
--- a/docs/index.js
+++ b/docs/index.js
@@ -15,7 +15,8 @@ import {
   SkipLink,
   OptionsMenuList,
   TopBarMenu,
-  Icon
+  Icon,
+  ThemeProvider
 } from '@deque/cauldron-react';
 import logo from './assets/img/logo.svg';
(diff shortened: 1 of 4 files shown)


scurker (changes requested):
Missing tests for different `context` values or that the class gets correctly set.

scurker (approved):

PR #298, opened 2021-07-26
feat(react): add "info" variant for modal that uses an unstyled heading

vicdq:
Add support for plain modal.

Supports modal without header and allows to override default styles.


schne324:
This stripped down modal will force use to constantly write 1-off css. The "plain" modal should look like the modal you are implementing in walnut.

Screen Shot 2021-07-26 at 9 09 25 AM


I'd prefer to have something like this:

```js
<Modal variant="simple">
  <ModalHeader>
    <h2>axe DevTools Pro</h2>
    <p>Maximize your testing coverage blah blah</p>
  </ModalHeader>
  <ModalContent>
     <FeaturesListThing />
  </ModalContent>
</Modal>
```

Screen Shot 2021-07-26 at 9 11 26 AM

**NOTE**: I think the background image bubbles can be a 1-off style in walnut rather than shipping the cauldron modal with it

scurker:
I agree with @schne324. 

I think we really only need 2 things:

* `heading` becomes optional
* _Either_ a variant to indicate that the heading "chrome" should not be displayed like @schne324 has indicated (`variant="info"`) or alternatively only if the `heading` prop exists. I don't have super strong opinions here.

I don't think we should be shipping something in cauldron that doesn't have a default design requiring an implementer to "finish" it.

scurker (dismissed):
☝️  See requested changes

schne324:
afaict "axe DevTools Pro" is the heading here...if we don't make `heading` a required thing, then we need to enforce that the dialog element has an accessible name....

```js
// pseudo-logic

if (!props.heading && !props['aria-label'] && !props['aria-labelledby']) {
  throw new Error('THOU SHALT ADD AN ACCESSIBLE NAME TO THY MODAL!')
}

// (there's probably a way to enforce this via `propTypes` or something better than the above 1-off error 🤷 )
```

schne324 (changes requested):
👏 I like the change! One more thing though: 
The the code block on the [modal demo page](https://deploy-preview-298--cauldron.netlify.app/components/Modal) needs to be updated to reflect the `variant=info` change

packages/styles/dialog.css
```diff
@@ -110,10 +110,13 @@
   justify-content: center;
   border: none;
   background-color: white;
+  height: fit-content;
```
scurker, line 113: I don't think has wide support yet. https://caniuse.com/?search=fit-content

What does this look like when `fit-content` is not supported?

dequejosie: I changed it into "auto". So far the behavior is the same, and when heading is too long it will wrap. Let me know if there is other issues.

scurker (dismissed):

packages/styles/dialog.css
```diff
@@ -100,3 +100,41 @@
   border-top: 1px solid var(--dialog-border-color);
   padding: calc(var(--space-small) - 4px) var(--dialog-padding);
 }
+
+/**
+* style for plain dialog
+*/
+
+.Plain__Dialog .Dialog__header {
```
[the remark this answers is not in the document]
dequejosie: Got it!

packages/styles/dialog.css
```diff
@@ -100,3 +100,41 @@
...
+
+/**
+* style for plain dialog
+*/
+
+.Modal--info .Dialog__header {
+  display: flex;
+  justify-content: center;
+  border: none;
+  background-color: white;
```
scurker, line 112:
```suggestion
  background-color: transparent;
```

The current value of `white` would be layered on top of any custom background colors.

packages/styles/dialog.css
```diff
@@ -100,3 +100,41 @@
...
   padding: calc(var(--space-small) - 4px) var(--dialog-padding);
 }
+
+/**
+* style for plain dialog
+*/
+
+.Modal--info .Dialog__header {
+  display: flex;
+  justify-content: center;
```
scurker, line 110:
```suggestion
```

This is not needed (see below suggestion on `.Dialog__heading`

packages/styles/dialog.css
```diff
@@ -100,3 +100,41 @@
...
+  background-color: white;
+  height: auto;
+}
+
+.Modal--info .Dialog__heading {
+  padding-top: 24px;
+  height: auto;
+  margin-right: 16px;
+  text-align: center;
+}
```
scurker, line 121:
```suggestion
.Modal--info .Dialog__heading {
  flex: 1;
  justify-content: center;
  padding-top: var(--space-large);
  height: auto;
  margin-right: var(--space-small);
}
```

This is based off changes to `.Dialog__header`:

* Uses appropriate css variables
* Center aligns the text using flex values

scurker (dismissed):

packages/styles/dialog.css
dequejosie: I am keeping text-align, otherwise if I pass in a <h1>, it will be left-aligned. Or is there a better way?

dequejosie: <h1> This is fun!

dequejosie: hmm, now I think about it. I guess we can leave that out of Cauldron

packages/styles/dialog.css
```diff
@@ -100,3 +100,41 @@
...
+
+.Modal--info .Dialog__content {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+.Modal--info .Dialog__close {
+  padding: 0;
+  width: 30px;
```
scurker, line 132: We shouldn't reduce the target size of the close button, see [WCAG 2.1 2.5.5](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html).

While the current size does not meet AAA standards, we shouldn't make the alternate variation harder to click than the default modal by reducing the target area.

scurker (changes requested):

scurker (dismissed):
Everything looks good but I think we need to at least add a test for this.

It should be as simple as validating that the `Modal--info` class exists when variant is set.

dequejosie:
I actually also thought of that!!

packages/react/__tests__/src/components/Modal/index.js
```diff
@@ -7,15 +7,32 @@ const defaults = { show: false, heading: { text: 'hi' } };
...
+  expect(modal.find('.Modal--info').exists()).toBe(true);
 });
 
 test('should return no axe violations', async () => {
-  const modal = mount(<Modal show={true} heading={'title'}>Hello!</Modal>);
+  const modal = mount(
+    <Modal show={true} heading={'title'}>
+      Hello!
+    </Modal>
+  );
```
scurker, line 36: We should test `variant="info"` here as well

scurker (changes requested):
One last thing

scurker:
> Like this? No need to write a separate test?

We need to test _both_ the variant and the non-variant pass axe, but it can be in a single test.

dequejosie:
Got it!

scurker (approved):

PR #301, opened 2021-07-28
feat: Adds stepper (progress) component

schne324:
This patch implements the "light theme" stepper. Support for the dark theme will be added in a separate PR (which we can then use in the extensions IGT progress thing)

### DIFF
--- a/packages/react/src/components/Stepper/index.tsx
+++ b/packages/react/src/components/Stepper/index.tsx
@@ -1,0 +1,54 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+interface StepperProps {
+  children: React.ReactNode;
+  status: 'current' | 'complete' | 'future';
+  className?: string;
+}
+
+export const Step = ({
+  children,
+  status,
+  className,
+  ...other
+}: StepperProps) => (
+  <li
+    className={classNames(
+      'Stepper__step',
+      `Stepper__step--${status}`,
+      className
+    )}
+    {...other}
+  >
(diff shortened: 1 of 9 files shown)


packages/react/src/components/Stepper/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+import TooltipTabstop from '../TooltipTabstop';
+
+interface BaseStepProps {
+  status: 'current' | 'complete' | 'future';
+  className?: string;
+}
+
+interface StepWithChildren extends BaseStepProps {
+  children: React.ReactNode;
+  tooltip?: never;
```
schne324, line 13: @scurker I forgot to circle back and try out your suggestion here for replacing `never` here with something like: 

```
function propsWithTooltip(props: StepWithChildren | StepWithTooltip): props is StepWithTooltip {
    return !!(props as StepWithTooltip).tooltip
}
```

packages/styles/stepper.css
```diff
@@ -0,0 +1,87 @@
+.Stepper {
+  --step-indicator-size: 25px;
+  --step-line-height: 8px;
+  --step-line-offset: 18px;
```
scurker, line 4: No vars? Wouldn't we at least want to make these relative to the font size?

scurker (commented):
Just one question, code LGTM otherwise

packages/styles/stepper.css
schne324: Made the change to `em` -- good call 👍

scurker (approved):

PR #304, opened 2021-07-29
fix: make Info modal heading symmetrical

dequejosie:
Moved variable `target-size` to the root and it is not undefined in `.Dialogue__heading`


packages/react/src/components/Dialog/index.tsx
```diff
@@ -51,8 +50,7 @@ export default class Dialog extends React.Component<DialogProps, DialogState> {
     forceAction: PropTypes.bool,
     heading: PropTypes.oneOfType([PropTypes.object, PropTypes.node]).isRequired,
     closeButtonText: PropTypes.string,
-    portal: PropTypes.any,
-    variant: PropTypes.string
```
schne324, line 55: Won't this make it so non-typescript consumers of cauldron will get a prop type error in their console if they try to use `variant="info"`? I think we want to leave it in `propTypes`...

schne324 (changes requested):
question about prop types

packages/react/src/components/Dialog/index.tsx
dequejosie: Thanks Harris! I thought we only needed `variant` in Modal. It adds a className to Modal and then we change the modal style accordingly. Do we need it in Dialog as well? I removed `variant `from both typescript interface and propTypes of Dialog

schne324 (approved):

PR #310, opened 2021-08-06
fix(IconButton): sets aria-label when disabled

### LINKED ISSUE #309: IconButton with label has no accessible name if the tooltip has not been created
Solution: Set the `aria-label` on the button _only_ when the button is currently disabled.

schne324:
Closes #309

### DIFF
--- a/docs/patterns/components/IconButton/index.js
+++ b/docs/patterns/components/IconButton/index.js
@@ -13,12 +13,15 @@ const IconButtonDemo = () => (
         { icon: 'pencil', label: 'Edit', variant: 'error' },
         { icon: 'pencil', label: 'Edit', variant: 'light' },
         { icon: 'pencil', label: 'Edit', variant: 'dark' },
+        { icon: 'pencil', label: 'Edit', disabled: true },
         { icon: 'trash', label: 'Delete', tooltipPlacement: 'bottom' }
       ]}
       propDocs={{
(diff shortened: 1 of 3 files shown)


scurker (commented):
I think we're missing one thing here.

The tooltip adds a dynamic attribute according to its `association` prop. We should probably make it so that when it becomes unmounted or the association changes it removes the dynamic prop from the Button.

https://github.com/dequelabs/cauldron/blob/3ad6ac953a2f500dfe63d9592d962af80f5438f5/packages/react/src/components/Tooltip/index.tsx#L174-L182

schne324:
I don't think so...we literally don't render the tooltip when disabled: https://github.com/dequelabs/cauldron/blob/eb0d7b059034998e8081e7053a5f38e1bb87ba4e/packages/react/src/components/IconButton/index.tsx#L54-L63

scurker:
Steps where this could happen:

1. IconButton gets mounted with no `disabled`
1. IconButton gets focused/hovered and Tooltip is created that adds `association` attribute value to the target button element
1. IconButton then gets disabled, removing the tooltip
1. The IconButton has both `aria-label` and the `association` that was dynamically added via the Tooltip

schne324:
> Steps where this could happen

Yea good catch! This one might be tricky... 🤔

Issue #308, opened 2021-08-06
IconButton should extend Tooltip props


scurker:
Currently can't pass the target element into `IconButton` button, but it only looks like `IconButton` only allows a subset of tooltip props

PR #311, opened 2021-08-06
fix(react): include additional tooltip props for IconButton

scurker:
see #308

### DIFF
--- a/packages/react/src/components/IconButton/index.tsx
+++ b/packages/react/src/components/IconButton/index.tsx
@@ -6,7 +6,6 @@ import React, {
 } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import { Placement } from '@popperjs/core';
 import Icon from '../Icon';
 import Tooltip, { TooltipProps } from '../Tooltip';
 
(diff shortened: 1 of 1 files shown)


packages/react/src/components/IconButton/index.tsx
```diff
@@ -71,6 +77,9 @@ IconButton.propTypes = {
   // @ts-ignore
   tooltipPlacement: PropTypes.string,
   // @ts-ignore
+  tooltipVariant: PropTypes.string,
```
stephenmathieson, line 80: Just curious - why do we have to `ts-ignore` this (but only sometimes)?

stephenmathieson (approved):

schne324 (approved):

packages/react/src/components/IconButton/index.tsx
scurker: At one point, typescript did not validate `prop-types` but we upgraded and it validated the typescript interface _and_ the prop-type values so some of that came during the change where we are trying not to break any of the existing `prop-type` values.

IMO, I don't think we need to have both `prop-types` and typescript.

stephenmathieson: > IMO, I don't think we need to have both prop-types and typescript.

It'd be a breaking change to remove them, and would make using the components extremely difficult for non-TS users.

This said, we could just add https://github.com/milesj/babel-plugin-typescript-to-proptypes and forget about them entirely.

PR #317, opened 2021-09-07
fix: skip link text missing space

vicdq:
fix: skip link text missing space
axe-extension#1761


scurker (approved):
This is fine for a short term fix, but after discussing with @schne324 we need to make the skip link text a single tag/prop to allow for proper translations which would completely negate the need for this.
https://github.com/dequelabs/cauldron/issues/318

PR #320, opened 2021-09-08
fix: optional icon button props

vicdq:
fix: optional icon button props


vicdq:
Don't think we need a test for this. Let me know how to test this if you think otherwise. Thanks.

scurker (approved):

PR #321, opened 2021-09-14
feat: support for dark theme buttons and links

vicdq:
feat: support for dark theme buttons and links

Dark theme support for buttons and links that use cauldron's ThemeProvider and '.theme-dark' in any context.

#322


vicdq:
Screen Shot 2021-09-14 at 8 53 35 AMScreen Shot 2021-09-14 at 8 53 52 AM

vicdq:
Please notice the following:
- single file for the dark-related CSS. If this looks good enough, we can keep adding the necessary updates for checkboxes, lists, stepper, etc in this file.
- Support for both cauldron--theme-dark (cauldron's theme-provider) and theme-dark in general (current axe-extension theme provider).

packages/styles/dark-theme.css
```diff
@@ -0,0 +1,77 @@
+.cauldron--theme-dark,
+.theme-dark {
```
scurker, line 2: Sorry, I may not have made myself clear when we spoke in slack.

Components in cauldron should only support the `.cauldron--theme-dark` scoped class. `.theme-dark` is something local to the extension, so in the extension the theme context will need to set both classes for backwards compatibility support, but cauldron itself should be agnostic to that.

packages/styles/dark-theme.css
```diff
@@ -0,0 +1,77 @@
...
+  --header-text-color-dark: var(--white);
+
+  --button-background-color-primary-active: #a7e9f7;
+  --button-background-color-secondary-active: var(--accent-light);
+  --button-background-color-error-active: #fea7a6;
+
+  --header-text-color-dark: var(--white);
+}
+
+.cauldron--theme-dark .Layout,
```
scurker, line 17: These styles should be co-located in their own individual css files.

For example, `button` styles should be in `button.css`, layout styles should be in `layout.css` and so on. We've been trying to move each styles to their own component CSS as it's easier to maintain individually as opposed to lumped together in a single file.

docs/Demo/ThemeSelector.js
```diff
@@ -0,0 +1,18 @@
...
+    <Checkbox
+      label="Dark Theme"
+      id="toggle-dark-light-theme"
+      checked={theme === 'dark'}
+      onChange={toggleTheme}
+    />
+  );
+};
+
+export default ThemeSelector;
```
scurker, line 18: We should put this in the header (see the thin/not-thin topbar toggle).

scurker (changes requested):

packages/styles/button.css
```diff
@@ -175,3 +175,50 @@ button.Link {
 [class*='Button--'] + [class*='Button--'] {
   margin-left: var(--space-smallest);
 }
+
+/* Dark Theme */
+
+.cauldron--theme-dark .Button--primary,
+.theme-dark .Button--primary {
```
scurker, line 182: No `.theme-dark`

packages/styles/variables.css
```diff
@@ -109,3 +109,21 @@
   /* ensure this is 1 higher than the top bar z-index */
   --z-index-skip-container: calc(var(--z-index-top-bar) + 1);
 }
+
+.cauldron--theme-dark {
```
scurker, line 113: This can go in the `button.css` file

scurker (changes requested):
Please remove the `.theme-dark` classes and put the button variables in the appropriate file

vicdq:
@scurker , should be good now. The updates for topbar and icon buttons are in this other pr: https://github.com/dequelabs/cauldron/pull/326

scurker (dismissed):
It looks like none of the hover states have been implemented correctly for the dark mode.

For example, here's how the primary button should appear when hovered:

image

There should be a 1px outline of `#6CDAF2` when hovered with a 1px gap between the outline and the button border.

Also please resolve previous comments if they have been addressed, it makes it difficult to do follow up reviews.

packages/styles/variables.css
```diff
@@ -7,6 +7,7 @@
 :root {
   /* color palette */
   --white: #ffffff;
+  --gray: #53636e;
```
scurker, line 10: This isn't gray, if you look at the rgb values `83, 99, 110`, it's primarily blue with a bit of red and green. I don't think we should assign a generic variable name to something that isn't purely gray. We should name this something that will be unlikely to conflict with anything else.

Something like `--dark-accent`, we can always update it later.

scurker (changes requested):
One minor request

scurker (approved):

Issue #323, opened 2021-09-14
Support dark/light theme for IconButton & TopBar


vicdq:
Remove light/dark variant from these components and support cauldron's theme-provider/context.

somaalapati:
Caldron stuff. Not possible to verify it on Axe Pro.

PR #326, opened 2021-09-15
feat: supports dark/light theme for topbar & icon buttons

vicdq:
feat: supports dark/light theme for topbar & icon buttons

Replaces light/dark variants with the appropriate styles for the theme provider.

#323

https://preview.uxpin.com/5a35f510bfb55e3e77f9dbd54a49203736113855#/pages/139000272/specification/sitemap

https://preview.uxpin.com/5a35f510bfb55e3e77f9dbd54a49203736113855#/pages/139030171/specification/sitemap

BREAKING CHANGE: This pr removes the light/dark variant options for IconButton and TopBar.


vicdq:
Screen Shot 2021-09-15 at 9 22 21 AMScreen Shot 2021-09-15 at 9 22 38 AM

vicdq:
This needs to be reviewed/merged after this:
https://github.com/dequelabs/cauldron/pull/321

packages/styles/variables.css
```diff
@@ -11,7 +11,9 @@
   --gray-20: #f2f2f2;
   --gray-30: #e0e0e0;
   --gray-40: #cccccc;
+  --gray-50: #b3bfc6;
```
scurker, line 14: This doesn't look like a grayscale color, should this just be local to the sidebar?

packages/styles/variables.css
```diff
@@ -11,7 +11,9 @@
...
   --gray-60: #666666;
+  --gray-70: #5d676f;
```
scurker, line 16: This will conflict with the extension and is not a grayscale color, perhaps it should just be local to where its needed?

scurker (commented):
I'm not quite following where the icon buttons are coming from in the ux pin for the dark mode. I see a `#6CDAF2` variation and a `#A5DB75` variation, but I'm not seeing either one of those in the preview. Am I missing something?

vicdq:
> I'm not quite following where the icon buttons are coming from in the ux pin for the dark mode. I see a `#6CDAF2` variation and a `#A5DB75` variation, but I'm not seeing either one of those in the preview. Am I missing something?

I added  the uxpins for the different icon buttons that we need to support in the new igt-results. The light and dark versions will use the secondary icon button. We don't currently use the error-option or any of the alternative colors. The idea would be to add support for those when we need them and probably create a version in both light/dark even if the color/settings already exist; the alternatives for the dark version might just be the same primary for the light version.

packages/react/src/components/IconButton/index.tsx
```diff
@@ -16,7 +16,7 @@ interface IconButtonProps
   tooltipPlacement?: TooltipProps['placement'];
   tooltipVariant?: TooltipProps['variant'];
   tooltipPortal?: TooltipProps['portal'];
-  variant?: 'light' | 'dark' | 'primary' | 'secondary' | 'error';
```
scurker, line 19: This is a breaking change, so ensure that the body of the pr includes `BREAKING CHANGE:` with a description of what is breaking.

PR #328, opened 2021-09-17
feat: Dark theme support for checkbuttons/radios/stepper

vicdq:
feat: Dark theme support for checkbuttons/radios/stepper

#324


vicdq:
Screen Shot 2021-09-17 at 6 42 37 AM

Screen Shot 2021-09-17 at 6 42 15 AM


Screen Shot 2021-09-17 at 6 41 41 AM

packages/react/src/components/IconButton/index.tsx
```diff
@@ -16,7 +16,7 @@ interface IconButtonProps
   tooltipPlacement?: TooltipProps['placement'];
   tooltipVariant?: TooltipProps['variant'];
   tooltipPortal?: TooltipProps['portal'];
-  variant?: 'light' | 'dark' | 'primary' | 'secondary' | 'error';
```
scurker, line 19: This is a breaking change, so we need to ensure the body of the merge request includes `BREAKING CHANGE:` with a description.

scurker (changes requested):
Radio Group dark mode does not match uxpin:

radio group buttons not matching ux pin

The stepper label has a color contrast issue in dark mode:

stepper with text description color contrast issue

scurker (approved):
Looks like I was looking at an outdated deploy preview, approved

Issue #316, opened 2021-09-07
Add new light and dark issue element


erik-deque:
UXPin: https://preview.uxpin.com/5a35f510bfb55e3e77f9dbd54a49203736113855#/pages/139000272/simulate/sitemap

![image](https://user-images.githubusercontent.com/57361142/132360360-9ebdfabc-d578-4285-aa3d-e8c0ece07376.png)

PR #344, opened 2021-09-21
feat: new panel component, dark/light support for expansible-panel/code

vicdq:
feat: new panel component, support for dark/light expansible panel.

Support for new panel component (igt results) and dark/light theme for expansible-panel component.

#316


packages/styles/panel.css
```diff
@@ -0,0 +1,58 @@
...
+.Panel {
+  border: 1px solid var(--panel-border-color);
+}
+
+.Panel__Header {
+  background-color: var(--panel-header-background-color);
+  color: var(--panel-header-color);
+  font-size: var(--text-size-normal);
+  display: flex;
+  height: 35px;
```
scurker, line 18:
```suggestion
```

If we size the appropriate things below, we won't need an implicit height here.

packages/styles/panel.css
```diff
@@ -0,0 +1,58 @@
...
+  border-bottom: 1px solid var(--panel-border-color);
+}
+
+.Panel__Header-title {
+  flex-grow: 1;
+  display: flex;
+  line-height: var(--space-gutter);
+  margin-left: var(--space-small);
+}
```
scurker, line 27:
```suggestion
.Panel__Header-title {
  flex-grow: 1;
  display: flex;
  padding: var(--space-half) 0;
}
```

The margin isn't needed because we're using flex/padding to space things appropriately. Without the line height/height changes from the container the `.Panel__Header` can "grow" appropriately if users make font size adjustments.

We can also add a min-height `min-height: var(--text-size-large-medium);` to enforce the minimum size of the header if we want to address that.

packages/styles/panel.css
```diff
@@ -0,0 +1,58 @@
...
+
+.Panel__Header-actions {
+  flex-grow: 1;
+  text-align: right;
+  line-height: var(--space-gutter);
```
scurker, line 32:
```suggestion
```

This is not needed with the above changes.

packages/styles/panel.css
```diff
@@ -0,0 +1,58 @@
...
+  margin-right: var(--space-smallest);
+}
+
+.Panel__Header-actions button {
+  margin-bottom: var(--space-smallest);
+}
+
```
scurker, line 39:
```suggestion
```

If we set vertical alignment with `align-items` above, this shouldn't be needed anymore.

packages/react/src/components/ExpandCollapsePanel/index.css
```diff
@@ -1,5 +1,35 @@
+:root {
+  --expandcollapse-trigger-background-color: var(--gray-10);
+  --expandcollapse-trigger-color: var(--gray-90);
+  --expandcollapse-panel-background-color: var(--gray-10);
+  --expandcollapse-panel-color: var(--gray-90);
```
scurker, line 5:
```suggestion
  --expandcollapse-trigger-background-color: transparent;
  --expandcollapse-trigger-color: var(--gray-90);
  --expandcollapse-panel-background-color: transparent;
  --expandcollapse-panel-color: var(--gray-90);
```

This component isn't intended to be styled by default. The only thing we really need to set is the appropriate text color for light/dark backgrounds.

scurker (dismissed):

packages/react/src/components/ExpandCollapsePanel/index.css
```diff
@@ -1,5 +1,35 @@
...
+  width: 100%;
+}
+
+.ExpandCollapse__trigger-title {
+  flex-grow: 1;
+  text-align: left;
+}
+
+.ExpandCollapse__panel {
+  padding: var(--space-smallest) 0;
```
scurker, line 26: Noticed this one thing, the padding mucks around with smooth transitions creating some jank so let's remove this for now.

```suggestion
```

scurker (changes requested):
Sorry, noticed one minor thing and we should be good to go.

scurker (approved):

PR #350, opened 2021-09-22
feat: Add new Tabs component

dequejosie:
Tabs are needed for our new record and setting screens: https://github.com/dequelabs/axe-extension/issues/1809; https://github.com/dequelabs/axe-extension/issues/1763


dequejosie:
Everything else is ready except tests

packages/react/src/components/Tabs/TabPanel.tsx
```diff
@@ -0,0 +1,38 @@
...
+  children?: React.ReactNode;
+  className?: string;
+}
+
+const TabPanel = (props: TabPanelProps) => {
+  const { children, index, value, className, ...other } = props;
+
+  return (
+    <div
+      tabIndex={0}
```
schne324, line 17: The panel shouldn't be a tab stop...

```suggestion
```

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,104 @@
...
+        'Tabs--thin': thin
+      })}
+    >
+      <ul
+        role="tablist"
+        className="Tablist"
+        aria-label="Tablist"
+        onClick={handleClick}
+        onKeyDown={handleKeyDown}
+        tabIndex={0}
```
schne324, line 87: The tablist shouldn't be a tab stop...

```suggestion
```

packages/react/src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,45 @@
...
+  index: number;
+  label?: string;
+  children?: React.ReactNode;
+}
+
+const Tab = ({ value, label, index, children }: TabProps) => {
+  const tabRef = useRef<HTMLLIElement>(null);
+  const tabIndex = index === value ? 0 : -1;
+  const selected = tabIndex === 0;
+  if (selected) {
```
schne324, line 16: this effect should be inside of a `useEffect`

maybe something like this: 

```js
useEffect(() => {
  if (index !== value) {
    return
  }
  
  tabRef?.current?.focus();
}, [value])
```

packages/react/src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,45 @@
...
+      className={classNames('Tab', {
+        'Tab--active': selected
+      })}
+      id={`tab-${index}`}
+      aria-controls={`tabpanel-${index}`}
+      tabIndex={tabIndex}
+      aria-selected={selected}
+      ref={tabRef}
+    >
+      {label || children}
```
schne324, line 32: What's the point of even having a special `label` prop here?

Is `<Tab label="Foo" />` better/easier/etc than `<Tab>Foo</Tab>`?

packages/react/src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,49 @@
...
+  index: number;
+  children?: React.ReactNode;
+}
+
+const Tab = ({ value, index, children }: TabProps) => {
+  const tabRef = useRef<HTMLLIElement>(null);
+  const tabIndex = index === value ? 0 : -1;
+  const selected = tabIndex === 0;
+
+  useEffect(() => {
```
schne324, line 16: sorry, I should've been more specific when I made this suggestion but this useEffect runs on mount. This means that anytime anyone uses tabs focus will be stolen and set on the currently active tab. We only want this effect to run _after_ mounting. In a different project, I've setup a custom hook like this:

```ts
import React from 'react';

/**
 * Hook to be used similarly to the React.Component#componentDidMount.
 * Executes the provided `effect` when `dependencies` change but does not
 * execute the effect initially (on mount) - only on update.
 *
 * @param effect {Function} function to be executed when dependencies update
 * @param dependencies {Any} any valid dependency argument to React.useEffect
 */
const useDidUpdate = (effect: () => void, dependencies: unknown[]): void => {
  const mounted = React.useRef(false);

  React.useEffect(() => {
    if (!mounted.current) {
      mounted.current = true;
      return;
    }

    effect();
  }, dependencies);
};

export default useDidUpdate;

```

dequejosie: What do you mean by `anytime anyone uses tabs focus will be stolen and set on the currently active tab`?

dequejosie: I see. The 'Tab 1' is automatically focused on page loads.

packages/react/src/components/Tabs/TabPanel.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+interface TabPanelProps {
```
scurker, line 5:
```suggestion
interface TabPanelProps extends React.HTMLAttributes<HTMLDivElement> {
```

This will allow any consumer to add additional props to the panel.

packages/react/src/components/Tabs/TabPanel.tsx
```diff
@@ -0,0 +1,37 @@
...
+  className?: string;
+}
+
+const TabPanel = (props: TabPanelProps) => {
+  const { children, index, value, className, ...other } = props;
+
+  return (
+    <div
+      hidden={value !== index}
+      id={`tabpanel-${index}`}
```
scurker, line 18: This is a bit problematic, [`aria-controls` is specified by the tab](https://github.com/dequelabs/cauldron/blob/e65928306ad11abec9b2cb319b8cfd0ba5cdb4aa/packages/react/src/components/Tabs/Tab.tsx#L32) but is potentially pointing to an id that is not unique because you could have multiple tab panels on the same page which would lead to duplicate ids.

Ideally we need some way to uniquely identify a set of tabs and their panels while avoiding duplicate ids.

`aria-labelledby` has the same issue directly below this.

There's a couple of different ways we could solve this, we could uniquely generate an id for each `<Tabs>` component that gets created and pass that identifier down to each tab/panel to control it.

```jsx
<Tabs>
  <Tab>1</Tab>
  <Tab>2</Tab>
  <Tab>3</Tab>
  <TabPanel>Panel 1</TabPanel>
  <TabPanel>Panel 1</TabPanel>
  <TabPanel>Panel 1</TabPanel>
</Tabs>
```

This allows for an accessible by default behavior. If the consumer cares about the specific ids, they should still be allowed to override the autogenerated default.

```jsx
<Tabs>
  <Tab id="tab-1" aria-controls="panel-1">1</Tab>
  <Tab id="tab-2" aria-controls="panel-1">2</Tab>
  <Tab id="tab-3" aria-controls="panel-1">3</Tab>
  <TabPanel id="panel-1" aria-labelledby="tab-1">Panel 1</TabPanel>
  <TabPanel id="panel-2" aria-labelledby="tab-1">Panel 1</TabPanel>
  <TabPanel id="panel-3" aria-labelledby="tab-1">Panel 1</TabPanel>
</Tabs>
```

packages/react/src/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useRef } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { useDidUpdate } from '../../index';
+
+interface TabProps {
```
scurker, line 6:
```suggestion
interface TabProps extends React.HTMLAttributes<HTMLLIElement> {
```

Allow specific attributes to be added to the element

packages/react/src/components/Tabs/TabPanel.tsx
```diff
@@ -0,0 +1,37 @@
...
+  children?: React.ReactNode;
+  className?: string;
+}
+
+const TabPanel = (props: TabPanelProps) => {
+  const { children, index, value, className, ...other } = props;
+
+  return (
+    <div
+      hidden={value !== index}
```
scurker, line 17: `hidden` isn't supported universally, so may be safer to conditionally display only the currently visible panel or just ensure that's visibly hidden via css.

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,108 @@
...
+  return (
+    <div
+      className={classNames('Tabs', className, {
+        'Tabs--thin': thin
+      })}
+    >
+      <ul
+        role="tablist"
+        className="Tablist"
+        aria-label="Tablist"
```
scurker, line 89: We should allow the user to override this

scurker (changes requested):
There's a potentially huge a11y issue and some other comments

docs/patterns/components/Tabs/index.js
```diff
@@ -0,0 +1,86 @@
...
+  const [value, setValue] = useState(0);
+  const handleChange = newValue => setValue(newValue);
+
+  return (
+    <div className="TabsDemo">
+      <h1>Tabs</h1>
+      <h2>Demo</h2>
+      <h3>Baisc Tabs</h3>
+      <Tabs value={value} handleChange={handleChange}>
+        <Tab index={0} value={value}>
```
scurker, line 15: The `index` seems a bit kludgy, perhaps using a `ref` would be better here?

```jsx
<Tabs>
  <Tab target={panel1}>1</Tab>
  <Tab target={panel2}>2</Tab>
  <TabPanel ref={panel1}>1</TabPanel>
  <TabPanel ref={panel2}>2</TabPanel>
</Tabs>
```

We could also default to the element order, so that the tabs/tab panels will match their index unless otherwise specified:

```jsx
<Tabs>
  <Tab>1</Tab>
  <Tab>2</Tab>
  <TabPanel>1</TabPanel>
  <TabPanel>2</TabPanel>
</Tabs>
```

scurker: For the second example, we could still use `refs` behind the scenes.

packages/react/src/components/Tabs/Tabs.tsx
dequejosie: You mean aria-label?

scurker: Yes, otherwise you could have multiple panels that are named "Tablist" plus we would want to be able to translate this.

packages/styles/tabs.css
```diff
@@ -0,0 +1,66 @@
...
+  color: var(--tabs-active-text-color);
+}
+
+.Tab--active {
+  color: var(--tabs-active-text-color);
+  background-color: var(--tab-active-background-color);
+  font-weight: bold;
+  text-decoration: none;
+  box-shadow: inset 0px 4px 0px var(--tab-shadow-color);
+  outline-offset: -1px;
```
scurker, line 53: Why `outline-offset: -1px`?

dequejosie: The focus outline is cut off on the right side for two of the tabs

docs/patterns/components/Tabs/index.js
```diff
@@ -0,0 +1,74 @@
...
+    <div className="TabsDemo">
+      <h1>Tabs</h1>
+      <h2>Demo</h2>
+      <h3>Baisc Tabs</h3>
+      <Tabs value={value} handleChange={handleChange}>
+        <Tab>Tab 1</Tab>
+        <Tab>Tab 2</Tab>
+        <Tab>Tab 3</Tab>
+        <TabPanel>
+          <p>Insert content for tab panle 1 here...</p>
```
scurker:
```suggestion
          <p>Insert content for tab panel 1 here...</p>
```

Typo here and below

docs/patterns/components/Tabs/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+const Demo = () => {
+  const [value, setValue] = useState(0);
+  const handleChange = newValue => setValue(newValue);
+
+  return (
+    <div className="TabsDemo">
+      <h1>Tabs</h1>
+      <h2>Demo</h2>
+      <h3>Baisc Tabs</h3>
```
schne324, line 13:
```suggestion
      <h3>Basic Tabs</h3>
```

docs/patterns/components/Tabs/index.js
```diff
@@ -0,0 +1,74 @@
...
+      <Tabs value={value} handleChange={handleChange}>
+        <Tab>Tab 1</Tab>
+        <Tab>Tab 2</Tab>
+        <Tab>Tab 3</Tab>
+        <TabPanel>
+          <p>Insert content for tab panle 1 here...</p>
+        </TabPanel>
+        <TabPanel>
+          <p>Insert content for tab panle 2 here...</p>
```
schne324, line 22:
```suggestion
          <p>Insert content for tab panel 2 here...</p>
```

docs/patterns/components/Tabs/index.js
```diff
@@ -0,0 +1,74 @@
...
+      <Tab>Tab 3</Tab>
+      <TabPanel>
+        <p>Insert content for tab panle 1 here...</p>
+      </TabPanel>
+      <TabPanel>
+        <p>Insert content for tab panle 2 here...</p>
+      </TabPanel>
+      <TabPanel>
+        <p>Insert content for tab panle 3 here...</p>
+      </TabPanel>
```
schne324, line 63:
```suggestion
      <TabPanel>
        <p>Insert content for tab panel 1 here...</p>
      </TabPanel>
      <TabPanel>
        <p>Insert content for tab panel 2 here...</p>
      </TabPanel>
      <TabPanel>
        <p>Insert content for tab panel 3 here...</p>
      </TabPanel>
```

schne324: sorry, forgot about multiline suggestions until this last one 😄

docs/patterns/components/Tabs/index.js
schne324: 👮 Beat ya to it! 😆

packages/styles/tabs.css
```diff
@@ -0,0 +1,71 @@
...
+  border-left: 1px solid var(--tabs-border-color);
+}
+
+.Tab {
+  height: 2.875rem;
+  list-style-type: none;
+  border-top: 1px solid var(--tabs-border-color);
+  border-right: 1px solid var(--tabs-border-color);
+  text-align: center;
+  padding-top: 14px;
```
scurker, line 35: I'm assuming this is for centering the text?

If so, this can be more equality achieved with flex:

```suggestion
  display: flex;
  justify-content: center;
  align-items: center;
```

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,165 @@
...
+  return (
+    <div
+      className={classNames('Tabs', className, {
+        'Tabs--thin': thin
+      })}
+    >
+      <ul
+        role="tablist"
+        className="Tablist"
+        aria-label={ariaLabelForTablist || 'Tablist'}
```
scurker, line 143: We should not specify a default here and instead require the user to provide it.
```suggestion
        aria-label={ariaLabelForTablist}
```

scurker (changes requested):

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,165 @@
...
+import Tab from './Tab';
+
+const [left, right, home, end] = [37, 39, 36, 35];
+
+interface TargetElement extends EventTarget {
+  id?: string;
+}
+
+interface TabsProps {
+  value: number;
```
scurker, line 16: `value` implies it could be anything, so having a more specific name would better indicate its usage.

Something like `activeTabIndex` would make it more clear. We could also make it optional, so that by default it defaults to the first tab.

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,165 @@
...
+        setActiveIndex(newIndex);
+        handleChange(newIndex);
+        break;
+      }
+    }
+  };
+
+  const tabComponents = tabs.map((child, index) => {
+    const { ...other } = (child as React.ReactElement<any>).props;
+    const selected = index === value;
```
scurker, line 100: Shouldn't this be `activeIndex`?
```suggestion
    const selected = index === activeIndex;
```

scurker (dismissed):
A few more things I noticed, the `initialActiveIndex` is just optional if you want the others I do think we should resolve.

docs/patterns/components/Tabs/index.js
```diff
@@ -0,0 +1,60 @@
...
+    <Tab>Tab 3</Tab>
+    <TabPanel>
+      <p>Insert content for tab panle 1 here...</p>
+    </TabPanel>
+    <TabPanel>
+      <p>Insert content for tab panle 2 here...</p>
+    </TabPanel>
+    <TabPanel>
+      <p>Insert content for tab panle 3 here...</p>
+    </TabPanel>
```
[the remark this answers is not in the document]
dequejosie: I thought I changed it, but realized I just did it for the real code, not the demo code...

scurker (approved):

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,157 @@
...
+      className={classNames('Tabs', className, {
+        'Tabs--thin': thin
+      })}
+    >
+      <ul
+        role="tablist"
+        className={classNames('Tablist', {
+          'Tablist--full-width': variant === 'full-width'
+        })}
+        aria-label={label}
```
schne324, line 136: I hate to keep going back and forth on this... @scurker and I talked about this yesterday but there may have been a misunderstanding...I don't think we should force an `aria-label` on the tablist because `aria-labelledby` is just as good, if not better in some situations. imo we should force the caller to either provide an `aria-label` prop or an `aria-labelledby` prop

schne324: To expand on my rationale a bit more - 

Often times tabs will appear directly beneath a heading and in those situations it is best to just make the tablist aria-labelledby the heading like so:


```js
<h3 id="foo">Stuff/things/etc</h3>
<div role="tablist" aria-labelledby="foo">...</div>
```

scurker: I think @schne324 is right here, and I overlooked that.

You can add an intersecting type allowing `aria-labelledby` or `aria-label`:

```ts
type props = {
  foo: 'string'
} & ({ labelledby: string } | { label: string })

type LabelProps = { 'aria-label': string } | { 'aria-labelledby': string }
type Props = {
  foo: 'bar'
} & LabelProps

// fails because it does not have aria-label or aria-labelledby
const props1: Props = { 
  foo: 'bar'
}

// These pass because they satisfy one of the union conditions
const props2: Props = {
  foo: 'bar',
  'aria-label': 'hello'
}
const props3: Props = {
  foo: 'bar',
  'aria-labelledby': 'hello'
}
```

docs/patterns/components/Tabs/index.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import { Tabs, Tab, TabPanel, Code } from '@deque/cauldron-react';
+import './index.css';
+
+const Demo = () => (
+  <div className="TabsDemo">
+    <h1>Tabs</h1>
+    <h2>Demo</h2>
+    <h3>Basic Tabs</h3>
+    <Tabs>
```
schne324, line 10: Shouldn't the demo not omit required props? I think we should add an `aria-label` or `aria-labelledby` here

schne324 (changes requested):

packages/styles/tabs.css
```diff
@@ -0,0 +1,89 @@
...
+  display: flex;
+  flex-direction: row;
+  border-left: 1px solid var(--tabs-border-color);
+}
+
+.Tab {
+  display: block;
+  text-align: center;
+  height: 2.875rem;
+  width: 9.375rem;
```
scurker, line 33: NOT NECESSARILY BLOCKING:

I think that we should change this. Anything beyond the width gets truncated, and I do think if we have space to display the text in the tabs we should display it instead of just ellipsizing the text. This leads to a loss of content that there is no alternative way to view.

scurker (approved):
Noticed a non-blocking change. We don't have to address it here since it's not critical to implementing tabs, so if you want to get this merged in to get you unblocked please create an issue so we can track it.

packages/styles/tabs.css
dequejosie: Yeah, we need to discuss which is the better approach.  I think in most cases we should be able to extend the width of the Tab and show the entire title, but we do want to prevent the Tab title from getting too long. @bobbyomari @awpearlm

somaalapati:
@dequejosie Is this testable by QA? If yes, please provide some details on how to verify this.

Issue #352, opened 2021-09-28
TextField: Cursor jumps to end of field when using value property


longmatthewh:
**Scenario**
Consider the use of `TextField` below, which uses the `value` prop: 
```
<TextField
     value={textValue}
     onChange={(changedValue)=>{
          setTextValue(changedValue);
     }}
/>
```
and
`const [textValue, setTextValue] = useState('');`

**Issue**
The cursor jumps to the end of the text in the input field produced by the `TextField` component as the user types.

**Steps to Reproduce**
- in the browser, focus on the input associated with the `TextField` component
- type some text like, "ADEFG"
- move the cursor somewhere in the middle of the text, like after the "A" in "ADEFG"
- type some additional text, like "BC"
- notice after typing the first additional character, the cursor jumps to the end of the `input`

PR #353, opened 2021-09-28
fix(react): cursor jump caused by updating state on componentDidUpdate

longmatthewh:
closes: #352 
When a `value` prop is passed in, set the input `value` directly instead of from state. There's also no longer any need to update state with value on `componentDidUpdate` because of this change. This eliminates the issue where the cursor would jump to the end upon updating the value described in #352.

- the input value prop was always being set from state
- the `onChange` function of `TextField` was not updating state if a `value` prop was present
- `setState` was being called with the `value` prop on `componentDidUpdate`, which was causing the cursor to jump

### DIFF
--- a/packages/react/__tests__/src/components/TextField/index.js
+++ b/packages/react/__tests__/src/components/TextField/index.js
@@ -1,9 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { mount } from 'enzyme';
 import TextField from 'src/components/TextField';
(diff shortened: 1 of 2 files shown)


longmatthewh:
I also added a couple more tests for required and error props.

longmatthewh:
The other option is to always update state in onChange, which won't force the implementer to manage state

schne324 (approved):

Issue #361, opened 2021-10-05
Add "simple" variant for `<Card>`


ryomtoob:
Add a smaller variant for the `<Card>` component that is meant to be used without `<CardFooter>` 
<img width="307" alt="Screen Shot 2021-10-05 at 10 53 04 AM" src="https://user-images.githubusercontent.com/31857969/136076844-5693e8ee-be40-4660-9198-f1e7e588dfb8.png">

PR #365, opened 2021-10-05
feat(Card): Add simple variant.

ryomtoob:
Closes: #361


schne324:
I know this is a bit of scope creep for you but do you mind addressing the couple of minor issues with the cards when in dark/light mode?

## light mode

the simple variant looks great! the header of the default variant is white text on white bg...could we change that text to be the same as in your simple variant?

Screen Shot 2021-10-11 at 12 19 23 PM




## dark mode

could you update the colors of the simple variant to match the default variant? 


Screen Shot 2021-10-11 at 12 19 01 PM

schne324 (approved):

PR #366, opened 2021-10-06
feat: Implement dark theme for Select and TextField

dequejosie:
Should solve this issue: https://github.com/dequelabs/axe-extension/pull/1879

### DIFF
--- a/packages/styles/forms.css
+++ b/packages/styles/forms.css
@@ -23,6 +23,7 @@
   --field-border-color-hover: var(--accent-info-light);
   --field-border-color-focus: var(--accent-info);
   --field-border-color-focus-hover: var(--accent-info-light);
+  --field-border-color-focus-glow: var(--accent-info);
   --field-border-color-error: var(--accent-danger);
   --field-border-color-error-hover: #fea7a6;
   --field-background-color-disabled: #5d676f;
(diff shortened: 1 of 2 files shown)


packages/styles/forms.css
```diff
@@ -29,6 +45,19 @@ textarea,
   border-bottom: 1px solid var(--field-border-color-underline);
   margin-bottom: var(--space-half);
   box-shadow: inset 0 1px 2px rgba(0, 0, 0, 0.05);
+  background-color: var(--field-background-color);
+}
+
+.cauldron--theme-dark input,
+select,
```
stephenmathieson, line 52: Are we intentionally applying styles to _all_ `<select>` elements here? IMO Cauldron should only style elements it renders.

schne324: @stephenmathieson is correct - cauldron should only style its components. Given that this was already there, I'm cool if we want to address this issue in a separate PR

schne324: Digging deeper, this file is full of cauldron styles that'd apply to non-cauldron components (see https://github.com/dequelabs/cauldron/pull/366/files#diff-d2f5588b0cde6a6cbe8f128481e2f436d72c46d49ecd4a507c36b28f001a9b6dR34-R42

dequejosie: Created an issue for this

PR #368, opened 2021-10-07
refactor: Move TabPanels outside Tabs

dequejosie:


packages/react/src/components/Tabs/Tab.tsx
```diff
@@ -2,19 +2,25 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 interface TabProps extends React.HTMLAttributes<HTMLLIElement> {
+  targetref: React.RefObject<HTMLDivElement>;
```
scurker, line 5: This might be considered nitpicky, but I do think it's important we try to maintain a consistent API across components.

```suggestion
  target: React.RefObject<HTMLDivElement> | HTMLElement;
```

This keeps it consistent with other component apis (such as [Tooltip](https://github.com/dequelabs/cauldron/blob/8f02a508ac963edc8ff33e1ae580d2f6d55c53e4/packages/react/src/components/Tooltip/index.tsx#L13)) that need references to other elements to maintain associations.

packages/react/src/components/Tabs/Tab.tsx
```diff
@@ -2,19 +2,25 @@ import React from 'react';
...
+      <li id={propId} ref={ref} role="tab" {...other}>
+        {children}
+      </li>
+    );
+  }
 );
 
 Tab.displayName = 'Tab';
 Tab.propTypes = {
+  targetref: PropTypes.any.isRequired,
```
scurker, line 22:
```suggestion
  target: PropTypes.any.isRequired,
```

packages/react/src/components/Tabs/TabPanel.tsx
```diff
@@ -1,19 +1,40 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import { useId } from 'react-id-generator';
 
 interface TabPanelProps extends React.HTMLAttributes<HTMLDivElement> {
+  panelref: React.RefObject<HTMLDivElement>;
```
scurker, line 7: Why not just `ref`? You would need to use `React.forwardRef` here. We've tried to move away from having named refs, except in cases where we wanted to maintain backwards compatibility (see [Button](https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/Button/index.tsx)).

```suggestion
  ref: React.RefObject<HTMLDivElement>;
```

packages/styles/tabs.css
```diff
@@ -26,14 +26,13 @@
   border-left: 1px solid var(--tabs-border-color);
 }
 
+/* Texts that are longer the Tab's width will be filtered by design group */
```
scurker, line 29: I don't quite understand this comment. Since this is a public facing component library, this may be used by more than just deque.

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -130,29 +117,25 @@ const Tabs = ({
     >
       <ul
         role="tablist"
-        className={classNames('Tablist', {
-          'Tablist--full-width': variant === 'full-width'
-        })}
```
scurker, line 135: Why is this being removed?

scurker, line 135: Why is this being removed?

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -85,17 +74,28 @@ const Tabs = ({
...
   const tabComponents = tabs.map((child, index) => {
-    const { ...other } = (child as React.ReactElement<any>).props;
+    const { targetref, id: propId, ...other } = (child as React.ReactElement<
+      any
+    >).props;
     const selected = index === activeIndex;
+    const [id] = propId ? [propId] : useId(1, 'tab');
+
+    useEffect(() => {
+      targetref.current?.setAttribute('aria-controlledby', id);
```
scurker, line 84: This isn't a valid aria property

scurker (dismissed):
We need to be sure this gets merged as a `BREAKING CHANGE` there's several things in here that are api changes that would be breaking.

packages/react/src/components/Tabs/Tabs.tsx
dequejosie: The styling is very minimum for the 'full-width' variant, and we need users to set their own width of Tablist and TabPanel anyway. So I decided to just leave it to the users.

packages/styles/tabs.css
dequejosie: Good point

packages/react/src/components/Tabs/Tabs.tsx
dequejosie: I am setting aria-controls here.
Screen Shot 2021-10-07 at 3 08 09 PM
›

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -85,17 +74,28 @@ const Tabs = ({
...
+    }, [target]);
+
+    useEffect(() => {
+      index === activeIndex
+        ? target.current?.classList.remove('TabPanel--hidden')
+        : target.current?.classList.add('TabPanel--hidden');
+    }, [activeIndex]);
 
     const config = {
-      id: `${id}-${index}`,
```
scurker, line 92: Still need to set id here. Right now, it's only set if the user has supplied it. It should be safe just to set id because it's set above:

```jsx
const [id] = propId ? [propId] : useId(1, 'tab')
```

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -130,29 +118,25 @@ const Tabs = ({
...
 };
 
 Tabs.displayName = 'Tabs';
 Tabs.propTypes = {
   children: PropTypes.node.isRequired,
-  label: PropTypes.string,
-  labelledby: PropTypes.string,
+  'aria-label': PropTypes.string,
+  'aria-labelledby': PropTypes.string,
   id: PropTypes.string,
```
scurker, line 136: Looks like you removed the typescript property, so should probably remove this one too
```suggestion
```

scurker (commented):
One small minor thing, LGTM!

PR #373, opened 2021-10-13
fix(Icon): Support intellisense for `type=` prop

### LINKED ISSUE #281: Icon's type= prop is deceiving
The `<Icon>` component supports a `type=` prop and according to its TypeScript definitions, `type` can be _anything_. This isn't true, nor is it helpful. Ideally we'd get autocomplete/intelisense when working with the component.

Can we update the build and have to dynamically create a list of valid Icon types?

stephenmathieson:
This patch adds a build step which dynamically creates our list of `<Icon>` types. This enables intellisense and prevents us from having to manually edit the demo page.

Closes #281


packages/react/scripts/buildIconTypes.js
```diff
@@ -0,0 +1,63 @@
...
+const ICONS_DIRECTORY = path.join(ICON_COMPONENT_DIRECTORY, 'icons');
+const ICON_TYPES_FILE = path.join(ICON_COMPONENT_DIRECTORY, 'types.ts');
+const ICONS_WITH_DIRECTIONS = [
+  'arrow',
+  'arrow-circle',
+  'chevron',
+  'chevron-double',
+  'triangle'
+];
+const DIRECTIONS = ['up', 'down', 'left', 'right'];
```
scurker, line 21: In theory, any icon could have a direction since it's just a modifier. I don't have strong opinions here since these are all the icons that make sense to have directions, we would just need to remember to add them when adding new icons. 🤷

Issue #382, opened 2021-10-25
Implement nav bar component


dylanb:
Steal as much of the implementation from Walnut as possible

Light: https://preview.uxpin.com/68df515d0023cc4b79f25d43f7663e24e363b1b9#/pages/141465415

somaalapati:
@dylanb Is this testable by QA?

Keerthi-Penukonda:
Working as expected.

**Screenshots:**
![image](https://user-images.githubusercontent.com/40262884/142982873-f9ba2b0f-0ac3-49a6-8a3a-3f05f2fed7dc.png)
![image](https://user-images.githubusercontent.com/40262884/142982950-b4c88a3c-1dde-4450-b481-c0e6b8793e69.png)

PR #387, opened 2021-11-02
feat: Add NavBar component

dequejosie:
For the ticket, please check: https://github.com/dequelabs/cauldron/issues/382


docs/index.css
```diff
@@ -46,3 +46,8 @@ code {
 .Demo ul {
   margin: var(--space-large);
 }
+
+/* add border between topbar and navbar just in the demo */
```
schne324, line 50: what's the backstory here? I think the demo/docs site should have as few 1-off styles as possible

packages/react/src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,196 @@
...
+      }
+
+      if (direction === 'next') {
+        // the maximum offset is navItemCount minus limit
+        const nextOffset = offset + limit;
+        setOffset(nextOffset > navItemCount ? offset : nextOffset);
+      }
+    };
+
+    return (
```
schne324, line 159: This control is inaccessible to keyboard and assistive technology users. It isn't keyboard-focusable and it doesn't have an accessible name.

I have some thoughts here that would be easier to discuss on a call. I think we could probably make it so keyboard users don't have to click the next/prev button and we could just scroll the overflowed items into view and focus them (so it'd kinda follow your focus).

schne324 (changes requested):
We'll have to make a decision here on these next/prev issues. I will bring this up as a standup topic tomorrow!

packages/react/src/components/NavBar/NavBar.tsx
schne324: This "pagination" stuff seems so weird to me...I don't think our navbar needs to support it. If callers of these components have tons of items, they should be able to tell the navbar component to turn into the hamburger-style menu

schne324:
I don't think the docs app should have every single component rendered as both a navbar item and sidebar item. We should leave them as sidebar items and have a "toggle navbar" button on the navbar documentation page where we render a few dummy nav items

Screen Shot 2021-11-04 at 2 51 21 PM

docs/index.css
dequejosie: If we don't demo NavBar in the documentation page, then we won't need this.

packages/react/src/components/NavBar/NavBar.tsx
dequejosie: Yeah, also noticed that the next/previous buttons are not focusable when reviewing with Aaron. Jason also raised the same concern. I will remove the pagination part for this pr. What we are trying out is to render all items and make it scrollable (without showing the scrollbar). We will still need next/previous buttons to scroll forwards/backwards. Anyway, this will be another pr, and yes, let's discuss it tomorrow.

schne324:
I think the way we have set up the hamburger menu is slightly off. The trigger will falsley add an extra item in the item count (for screen readers). A [really great article on this very component](https://inclusive-components.design/menus-menu-buttons/) suggests the ideal markup for this scenario is:

```html
<nav id="navigation">
  <button aria-expanded="false" aria-controls="menu-list">Menu</button>
  <ul id="menu-list" hidden>
    <li><a href="/">Home</a></li>
    <li><a href="/about">About</a></li>
    <li><a href="/shop">Shop</a></li>
    <li><a href="/contact">Contact</a></li>
  </ul>
</nav>
```

it also looks like we have aria-haspopup on both the parent `<li />` and the `<button />` when I'd expect to only see it on the `<button />`

mobile navbar markup

schne324 (changes requested):
I think the way we have set up the hamburger menu is slightly off. The trigger will falsley add an extra item in the item count (for screen readers). A [really great article on this very component](https://inclusive-components.design/menus-menu-buttons/) suggests the ideal markup for this scenario is:

```html
<nav>
  <button aria-expanded="false" aria-controls="menu-list" aria-haspopup="true">Menu</button>
  <ul id="menu-list">
    <li><a href="/">Home</a></li>
    <li><a href="/about">About</a></li>
    <li><a href="/shop">Shop</a></li>
    <li><a href="/contact">Contact</a></li>
  </ul>
</nav>
```

it also looks like we have aria-haspopup on both the parent `<li />` and the `<button />` when I'd expect to only see it on the `<button />`

mobile navbar markup

schne324:
Sorry, I hadn't popped the preview open and tested stuff with the keyboard but I still think there are a couple things missing here for the 🍔 menu:

- tabbing outside outside of the dropdown should close the dropdown (that means both <kbd>tab</kbd> and <kbd>shift + tab</kbd>)
  - I think we need something that auto-closes the dropdown menu if focus leaves the menu at all (like a user clicking outside of it). This is especially important since this version of the navbar comes with a `<Scrim />` when expanded
- a nice to have is  to allow <kbd>esc</kbd> to close the menu and shift focus to the trigger (🍔 )


and lastly, more of a question for @awpearlm... The icon + text have different heights making it funky to style in a way in which both are perfectly center aligned (vertically). Do we want to size the icon down (or text up) so this alignment is easier on the eyes?

Screen Shot 2021-11-09 at 10 12 49 AM
Screen Shot 2021-11-09 at 10 12 41 AM

awpearlm:
Let's try sizing the icon up a little to get the alignment centered.

schne324:
We talked about this on slack but for transparency / sharing this info with any other reviewers, I threw together a simplified approach to this component: https://codesandbox.io/s/winter-worker-bvr16?file=/src/App.js:1369-1692

akornmeier:
Was there a plan to incorporate sub-menu items in the navbar design as seen in this design? https://preview.uxpin.com/db46cd9bc4ccdc119aa1a6f8f57385bc2250a19f#/pages/141510737/simulate/no-panels

schne324:
> Was there a plan to incorporate sub-menu items in the navbar design as seen in this design? https://preview.uxpin.com/db46cd9bc4ccdc119aa1a6f8f57385bc2250a19f#/pages/141510737/simulate/no-panels

No plan in this patch to do so but I think you should take a look at what we end up with here and ensure there's nothing that'll make adding the subnav feature to the navbar difficult.

dequejosie:
@schne324 I made changes based on the simplified approach. One thing is that, non-interactive elements like `ul` cannot take onKeyDown `eventlistener` directly. I had to use react `useRef`, which works fine in demo. However, it is hard to test events through react `useRef`. Would appreciate some help with that!

schne324:
>  non-interactive elements like ul cannot take onKeyDown eventlistener directly

I don't think that is true. Are you referring to jsx-a11y-linter yelling about it? Delegating keydown listeners on a parent is perfectly acceptable, you'll just have to turn that lint rule off for this 1 loc

docs/patterns/components/NavBar/index.js
```diff
@@ -0,0 +1,65 @@
...
+        {componentsList.map((name, index) => {
+          return (
+            <NavItem key={`${name}-${index}`} active={index === 2}>
+              <a href="#">{`${name} ${index + 1}`}</a>
+            </NavItem>
+          );
+        })}
+      </NavBar>
+      <Button onClick={handleToggle} className="NavBarButton">
+        Toggle between mobil and non-mobile mode
```
schne324, line 26: not a huge deal but I think we mean "mobile" here (as in mobile device)

```suggestion
        Toggle between mobile and non-mobile mode
```

packages/react/src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  return (
+    <nav
+      className={classNames('NavBar', className, {
+        'NavBar--collapsed': collapsed
+      })}
+      ref={navRef}
+    >
+      <Scrim show={showDropdown} />
```
schne324, line 74: I don't see the scrim tied to this component anywhere in the uxpins....I think we should remove it because with a scrim comes a whole bunch of a11y considerations (like preventing screen readers from traversing to content behind scrim)


```suggestion
```

packages/react/src/components/NavBar/NavBar.tsx
```diff
@@ -0,0 +1,109 @@
+import React, {
+  useState,
+  useRef,
+  useEffect,
+  KeyboardEventHandler
+} from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import Scrim from '../Scrim';
```
schne324, line 9:
```suggestion
```

schne324 (commented):
One last question about the use of `<Scrim />`

schne324 (commented):
Another thing I just thought of -- If I'm using this in my SPA, I'd want the dropdown to collapse itself when an item is clicked (because clicking it will cause a new SPA page to be rendered, and I'd then want to move focus to the main content).

How do we want to support this use case?

BTW this is something that can be decided after this patch

packages/react/src/components/NavBar/NavBar.tsx
schne324: just confirmed with the designers and we should in fact remove the scrim here

schne324 (approved):

dequejosie:
> Another thing I just thought of -- If I'm using this in my SPA, I'd want the dropdown to collapse itself when an item is clicked (because clicking it will cause a new SPA page to be rendered, and I'd then want to move focus to the main content).
> 
> How do we want to support this use case?
> 
> BTW this is something that can be decided after this patch

Created a ticket for that!

Issue #389, opened 2021-11-04
Update <Select /> `onChange` prop to be optional


akornmeier:
**Current functionality** - The `onChange` prop for the `<Select />` component is required. It requires adding an empty void function to satisfy the compiler

**Desired functionality** - Make the `onChange` prop type optional, default the onChange method to an empty void method if the prop is nullish

schne324:
This was resolved by #390

PR #390, opened 2021-11-04
fix(select): updates onChange prop to be optional

akornmeier:
- **update**: `onChange` prop to be optional
- **update**: removes default value from function signature to be assigned later if prop value is nullish

Closes issue: #389

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3129,6 +3129,13 @@ create-hmac@^1.1.0, create-hmac@^1.1.2, create-hmac@^1.1.4:
     safe-buffer "^5.0.1"
     sha.js "^2.4.8"
 
+cross-env@^7.0.3:
+  version "7.0.3"
+  resolved "https://agora.dequecloud.com/artifactory/api/npm/dequelabs/cross-env/-/cross-env-7.0.3.tgz#865264b29677dc015ba8418918965dd232fc54cf"
+  integrity sha1-hlJkspZ33AFbqEGJGJZd0jL8VM8=
+  dependencies:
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -3129,6 +3129,13 @@ create-hmac@^1.1.0, create-hmac@^1.1.2, create-hmac@^1.1.4:
     safe-buffer "^5.0.1"
     sha.js "^2.4.8"
 
+cross-env@^7.0.3:
```
schne324, line 3132: all of this diff seems weird but I remember you told me yesterday that there were agora deps checked into source? I'm not seeing those removed here...I'm also not seeing any agora deps in the [`yarn.lock` in the develop branch](https://raw.githubusercontent.com/dequelabs/cauldron/develop/yarn.lock) either 🤔

akornmeier: They were from my initial commit. I removed them by uninstalling and reinstalling the cross-env package

akornmeier: i had my Deque npmrc pointed to agora when I checked in the first time.

schne324: so this patch doesn't update any deps but has yarn.lock diffs?

schne324 (approved):
Not gonna hold this up for some yarn.lock nonsense...ideally we back those changes out though 🤷

yarn.lock
akornmeier: @schne324 - I reverted the `yarn.lock` changes that were caused by my DQ .npmrc config. I will need a re-thumb.

schne324 (approved):

PR #394, opened 2021-11-08
fix(styles): Adds expected dark/light theme focus rings

schne324:
Ref: dequelabs/axe-extension/issues/1930

### DIFF
--- a/packages/styles/button.css
+++ b/packages/styles/button.css
@@ -16,7 +16,6 @@
 
   --button-text-color-dark: var(--gray-90);
   --button-text-color-light: var(--white);
-  --button-focus-ring-color: var(--focus-light);
   --button-thin-height: 23px;
 }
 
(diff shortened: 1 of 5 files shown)


scurker:
Just curious: Why remove `button-focus-ring-color`? I think there's plenty of spots where we have a global variable, then a local variable that just extends the global one. This feels like an unnecessary breaking change.

schne324:
> Just curious: Why remove button-focus-ring-color? I think there's plenty of spots where we have a global variable, then a local variable that just extends the global one. This feels like an unnecessary breaking change.

I tried and failed :(

**UPDATE**: @scurker and I looked at this yesterday and are a bit stumped by how variable overriding works. A great example of this problem can be found here in jason's codepen: https://codepen.io/scurker/pen/WNEKpLK

packages/styles/button.css
```diff
@@ -81,7 +80,7 @@ button.Link {
 .Button--primary:focus:before,
 .Button--secondary:focus:before,
 .Button--error:focus:before {
-  box-shadow: 0 0 1px 2px var(--button-focus-ring-color);
+  box-shadow: 0 0 1px 2px var(--focus);
```
scurker, line 83: I think what we were going for is to allow a consumer to locally override the focus ring just for button, I still think we can do this be providing a fallback:

```suggestion
  box-shadow: 0 0 1px 2px var(--button-focus-ring-color, --focus);
```

This should allow us to maintain backwards compat - and probably something we need to make a note of in the future.

scurker (approved):

PR #414, opened 2021-11-10
fix: type issues of OptionsMenu

dequejosie:
This is a continuation of https://github.com/dequelabs/cauldron/pull/393; `OnClick `and `ref` in `OptionsMenu` also have type discrepancies with types in `OptionMenuTriggerProps`.

Is there anything we can do to catch this earlier? Our demo is mostly javaScript, if we use typescript, will we be able to catch these issues before we merge?

### DIFF
--- a/packages/react/src/components/OptionsMenu/OptionsMenu.tsx
+++ b/packages/react/src/components/OptionsMenu/OptionsMenu.tsx
@@ -11,10 +11,10 @@ export interface OptionsMenuAlignmentProps {
 }
 
 export interface OptionsMenuRenderTriggerProps {
-  onClick: (event: Event) => void;
+  onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
   onKeyDown: (e: React.KeyboardEvent<HTMLButtonElement>) => void;
   'aria-expanded': boolean;
-  ref: React.RefObject<HTMLButtonElement>;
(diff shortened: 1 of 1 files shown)


packages/react/src/components/OptionsMenu/OptionsMenu.tsx
```diff
@@ -11,10 +11,10 @@ export interface OptionsMenuAlignmentProps {
 }
 
 export interface OptionsMenuRenderTriggerProps {
-  onClick: (event: Event) => void;
+  onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
   onKeyDown: (e: React.KeyboardEvent<HTMLButtonElement>) => void;
   'aria-expanded': boolean;
-  ref: React.RefObject<HTMLButtonElement>;
+  ref: React.Ref<HTMLButtonElement>;
```
schne324, line 17: Does this support "both" ways of reffing?

schne324, line 17: does this support both ways of reffing?

dequejosie:
@schne324  Not sure. I tried the other way by changing the `React.Ref` into `React.RefObject` in `OptionsMenuTrigger` and typescript was not happy with that. I just changed it to be safe.

PR #418, opened 2021-11-12
fix(styles): adjust styles of checkbox/radio to match set designs

### LINKED ISSUE #400: Radio groups and checkbox groups need more space between each control
8px of spacing is expected between each radio / checkbox

### LINKED ISSUE #401: remove border from "hidden" input behind checkboxes and radios
remove border from html native radio / checkbox elements (behind icon)...this creates a weird cursor flashing issue

### LINKED ISSUE #403: Apply radio/checkbox hover styles even when label is hovered
...this will likely require making use of a sibling selector

scurker:
fixes #400 
fixes #401
fixes #403

### DIFF
--- a/packages/react/src/components/Checkbox/index.tsx
+++ b/packages/react/src/components/Checkbox/index.tsx
@@ -75,6 +75,14 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
             }}
             {...other}
           />
+          <label
+            className={classNames('Field__label', {
+              'Field__label--disabled': disabled
+            })}
+            htmlFor={id}
+          >
+            {label}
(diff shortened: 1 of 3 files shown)


schne324 (commented):
curious here -- why did you move the labels before the icon and then use `row-reverse` to restore the order?

scurker:
> curious here -- why did you move the labels before the icon and then use `row-reverse` to restore the order?

So I could use the general sibling selector (`~`) to target the checkbox when the label is hovered. I don't think we want to style based on hovering the wrapping component as there are gaps in the component that do not currently trigger the checkbox/radio.

schne324 (approved):

Issue #205, opened 2021-03-12
<Pagination /> component


schne324:
![design for pagination](https://user-images.githubusercontent.com/1062039/110979375-2191cb00-832a-11eb-8b49-d95316ffdc88.png)

see uxpin for detailed design

# implementation details:

```ts
interface Props {
  totalItems: number;
  itemsPerPage?: number; // defaults to 10
  currentPage?: number; // defaults to 1
  statusLabel?: React.ReactElement; // defaults to "Showing x-y of {total}"
  firstPageLabel?: React.ReactElement; // defaults to "First page"
  previousPageLabel?: React.ReactElement; // defaults to "Previous page"
  nextPageLabel?: React.ReactElement; // defaults to "Next page"
  lastPageLabel?: React.ReactElement; // defaults to "Last page"
  onNextPageClick?: () => void;
  onPreviousPageClick?: () => void;
  onFirstPageClick?: () => void;
  onLastPageClick?: () => void;
}

<Pagination
  totalItems={15}
  itemsPerPage={5} // optional
  currentPage={1} // optional
  statusLabel={i18n`Showing 1-5 of 15`}
/>
```

- Use the `<IconButton />` with *label* (tooltip with *Label props)
- Component should manage (derive) disabled state of buttons. For example, on the first page, 'first page' and 'previous page' should be disabled. For each disabled control use `<TooltipTabstop />` so it can remain focusable/tooltip-able (obviously don't invoke the associated click listeners on disabled controls).
- wrap the `statusLabel` in a `role="log" aria-atomic="true"` (and test that this works as you navigate through the pages

schne324:
```ts
interface Props {
  totalItems: number;
  itemsPerPage?: number; // defaults to 10
  currentPage?: number; // defaults to 1
  statusLabel?: React.ReactElement; // defaults to "Showing x-y of {total}"
  firstPageLabel?: React.ReactElement; // defaults to "First page"
  previousPageLabel?: React.ReactElement; // defaults to "Previous page"
  nextPageLabel?: React.ReactElement; // defaults to "Next page"
  lastPageLabel?: React.ReactElement; // defaults to "Last page"
  onNextPageClick?: () => void;
  onPreviousPageClick?: () => void;
  onFirstPageClick?: () => void;
  onLastPageClick?: () => void;
}

<Pagination
  totalItems={15}
  itemsPerPage={5} // optional
  currentPage={1} // optional
  statusLabel={i18n`Showing 1-5 of 15`}
[...]

PR #425, opened 2021-11-16
feat/add pagination component

lpriyasiva:
Issue #205 

Adds the pagination component with a mobile view option.
Added the docs page and tests as well

### DIFF
--- a/packages/react/src/components/Pagination/index.tsx
+++ b/packages/react/src/components/Pagination/index.tsx
@@ -1,4 +1,5 @@
 import React, { ReactElement } from 'react';
+import PropTypes from 'prop-types';
 import IconButton from '../IconButton';
 import Select from '../Select';
 
@@ -10,13 +11,13 @@ export interface paginationProps {
   currentItemCount: number;
   hasPrevious: boolean;
   hasNext: boolean;
-  mobile_view?: boolean;
+  mobileView?: boolean;
   handlePageSizeChange: (pagesize: string) => void;
   handlePageChange: (pagenumber: number) => void;
 }
 
-const Pagination = (props: paginationProps): ReactElement => {
-  const mobileView = props.mobile_view ? props.mobile_view : false;
+const Pagination = (props: paginationProps) => {
+  const mobileView = props.mobileView ? props.mobileView : false;
   const previousPage = props.currentPage - 1;
   const startIndex =
     props.currentPage === 1 ? 1 : props.pageSize * previousPage + 1;
@@ -120,5 +121,20 @@ const Pagination = (props: paginationProps): ReactElement => {
     </div>
   );
 };
+
+Pagination.propTypes = {
+  pageSize: PropTypes.number,
+  totalPages: PropTypes.number,
+  currentPage: PropTypes.number,
+  totalItems: PropTypes.number,
+  currentItemCount: PropTypes.number,
+  hasPrevious: PropTypes.bool,
+  hasNext: PropTypes.bool,
+  mobileView: PropTypes.bool,
+  handlePageSizeChange: PropTypes.func,
+  handlePageChange: PropTypes.func
+};
+
+Pagination.displayName = 'Pagination';
 
 export default Pagination;
--- a/packages/react/src/components/Pagination/index.tsx
+++ b/packages/react/src/components/Pagination/index.tsx
@@ -18,18 +18,14 @@ export interface paginationProps {
 
 const Pagination = (props: paginationProps) => {
   const mobileView = props.mobileView ? props.mobileView : false;
+
   const previousPage = props.currentPage - 1;
   const startIndex =
     props.currentPage === 1 ? 1 : props.pageSize * previousPage + 1;
   const endIndex = startIndex + props.currentItemCount - 1;
   const current_page_index = (
     <p className={mobileView ? 'Page_Index_Sm' : ''}>
-      {'Showing '}{' '}
-      <span>
-        {' '}
-        {startIndex}-{endIndex}
-      </span>{' '}
-      {' of '} <span>{props.totalItems}</span>
+      Showing {startIndex}-{endIndex} of {props.totalItems}
     </p>
   );
   const changePage = (e: React.MouseEvent, pagenumber: number) => {
--- a/packages/react/src/components/Pagination/index.tsx
+++ b/packages/react/src/components/Pagination/index.tsx
@@ -1,0 +1,124 @@
+import React, { ReactElement } from 'react';
+import IconButton from '../IconButton';
+import Select from '../Select';
+
+export interface paginationProps {
+  pageSize: number;
+  totalPages: number;
+  currentPage: number;
+  totalItems: number;
+  currentItemCount: number;
+  hasPrevious: boolean;
+  hasNext: boolean;
+  mobile_view?: boolean;
+  handlePageSizeChange: (pagesize: string) => void;
+  handlePageChange: (pagenumber: number) => void;
+}
+
+const Pagination = (props: paginationProps): ReactElement => {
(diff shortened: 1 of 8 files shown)


packages/react/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+              key: '100',
+              value: '100',
+              label: '100 records'
+            }
+          ]}
+        />
+      </div>
+      <nav
+        id="page_index"
+        className={'Page_Index_Wrapper'}
```
scurker, line 71: I'm not sure this class name is needed, we should be able to target uniquely in css with the following selector: `.Pagination nav`.

packages/styles/pagination.css
```diff
@@ -0,0 +1,77 @@
...
+  display: flex;
+  flex-direction: row;
+  margin-top: 10px;
+}
+
+.Page_Size_Selector {
+  width: 150px;
+  height: 33px;
+  padding-top: 13px;
+  font-family: var(--monitor-font-family);
```
scurker, line 11: This looks like it shouldn't be here.

```suggestion
```

scurker (changes requested):
I haven't made it all the way through the PR, but I do have some general requested changes:

* All labels need to be able to be passed via props
* It looks like there's monitor specific details in the component

I will take a look tomorrow once I have more time.

packages/react/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+  pageSize: number;
+  totalPages: number;
+  currentPage: number;
+  totalItems: number;
+  currentItemCount: number;
+  hasPrevious: boolean;
+  hasNext: boolean;
+  mobileView?: boolean;
+  handlePageSizeChange: (pagesize: string) => void;
+  handlePageChange: (pagenumber: number) => void;
```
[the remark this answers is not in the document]
lpriyasiva: > I think we can simplify a lot of the props here, and have some of the state calculated internally. I think we may just be able to have the following props:
> 
> * `totalItems` - (required)
> * `itemsPerPage` - (optional, could default to something like `10`)
> * `currentPage` - (optional, would default to `1`)
> 
> I think everything else, we can just calculate inside of the component. Is there a particular reason we need to pass all those props in?

I originally did that. Then thought, better to reduce the logic in the component and have it be presentational.


> I'm not sure why `handlePageSizeChange` or `handlePageChange` is needed. Can we not have `on*` handlers for each button? `onNextClick`, `onPreviousClick`, `onFirstClick`, `onLastClick`?

Not sure what you mean by this tbh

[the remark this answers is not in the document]
scurker: We typically use `on*` props for event functions from components, see an example [with Dialog](https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/Dialog/index.tsx#L17).

Having a imperative prop event name makes the api much clearer without having to handle internal props, here's an example:

```jsx
function Component({ items }) {
  const [currentPage, setCurrentPage] = useState(1)
  const itemsPerPage = 10

  const lastPage = Math.ceil(itemsPerPage / items.length)
  const handleNextClick = () => setCurrentPage(Math.min(currentPage + 1, lastPage))
  const handlePreviousClick = () => setCurrentPage(Math.min(currentPage - 1, 0))
  const handleFirstClick = () => setCurrentPage(1)
  const handleLastClick = () => setCurrentPage(lastPage)

  return (
    <Pagination 
      currentPage={currentPage}
      itemsPerPage={itemsPerPage}
      onNextClick={handleNextClick}
      onPreviousClick={handlePreviousClick}
      onFirstClick={handleFirstClick}
      onLastClick={handleLastClick}
    />
  )
}
```

We have this pattern in the extension for our pagination and we have different events that happen depending on pagination. If there were only a single handler for all page events, we'd have to have a bunch of things nested in if statements to determine which pagination button was clicked. This is a bit cleaner to read IMHO.

packages/react/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+}
+
+const Pagination = (props: paginationProps) => {
```
scurker, line 19: We typically use destructuring for our props:

```tsx
const Pagination = (props: paginationProps) => {
```

...would become

```tsx
const Pagination = ({ currentPage, pageSize, ...otherProps }: paginationProps) => {
```

packages/react/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const mobileView = props.mobileView ? props.mobileView : false;
+
+  const previousPage = props.currentPage - 1;
+  const startIndex =
+    props.currentPage === 1 ? 1 : props.pageSize * previousPage + 1;
+  const endIndex = startIndex + props.currentItemCount - 1;
+  const current_page_index = (
+    <p className={mobileView ? 'Page_Index_Sm' : ''}>
```
scurker, line 27: If we're following the collapsed prop, this would become a modifier according to [BEM]:(http://getbem.com/naming/)

```suggestion
    <p className={collapsed ? 'Pagination--collapsed' : ''}>
```

Following this approach ensures that css variables are specific to their components and we're not leaking "global" styles.

packages/styles/pagination.css
```diff
@@ -0,0 +1,77 @@
...
+@media screen and (max-width: 576px) {
+  .Pagination {
+    flex-direction: column;
+    margin: 0 auto;
+  }
+
+  .Page_Index_Wrapper {
+    margin-top: 10px;
+  }
+}
```
scurker, line 77: We don't typically include `@media` styles in our components but rather let them be controlled with a prop.

packages/styles/pagination.css
```diff
@@ -0,0 +1,77 @@
...
+
+.Page_Size_Selector > div:first-of-type {
+  flex-direction: row;
+}
+
+.Page_Size_Selector label {
+  padding-right: 8px;
+  margin: 6px;
+  font-family: var(--monitor-font-family);
```
scurker, line 25:
```suggestion
```

packages/styles/pagination.css
```diff
@@ -0,0 +1,77 @@
...
+  font-weight: var(--font-weight-medium);
+}
+
+.Page_Size_Options {
+  width: 150px;
+}
+
+.Page_Index_Wrapper {
+  margin-left: auto;
+  font-family: var(--monitor-font-family);
```
scurker, line 36:
```suggestion
```

packages/styles/pagination.css
```diff
@@ -0,0 +1,77 @@
+.Pagination {
```
scurker, line 1: For all the `px` values in this file:

* For spacing/margins please check that it exists in our [variable vars](https://github.com/dequelabs/cauldron/blob/6d33223cbca82d028aa5fcefb71d29633e021cb2/packages/styles/variables.css#L60-L70). 
* For sizing/widths/heights we should be using relative units, preferably `rem`s.

scurker (changes requested):

packages/react/src/components/Pagination/index.tsx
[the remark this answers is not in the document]
scurker: > I originally did that. Then thought, better to reduce the logic in the component and have it be presentational.

Having logic in the component is okay and reduces error cases when we can derive calculated values from props.

It also lets us validate and test specific scenarios in unit tests so we can guarantee that the component will display correctly based on the current props values. We wouldn't want the the next button to be enabled if `hasNext` was passed when the combination of props wouldn't be a valid use case for doing so.

schne324:
pagination has been implemented in #485

PR #434, opened 2021-11-23
feat: add Address component

thuey:
Add a set of Address components for https://github.com/dequelabs/walnut/issues/2488

## Dark Theme
image

## Light Theme
image

### DIFF
--- a/packages/react/src/components/Address/index.tsx
+++ b/packages/react/src/components/Address/index.tsx
@@ -2,9 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-interface AddressProps {
+interface AddressProps extends React.HTMLAttributes<HTMLElement> {
   children: React.ReactNode;
-  className?: string;
 }
 
 export const Address = ({ children, className, ...other }: AddressProps) => (
@@ -19,10 +18,7 @@ Address.propTypes = {
   className: PropTypes.string
 };
 
-interface AddressLineProps {
-  children?: React.ReactNode;
-  className?: string;
-}
+type AddressLineProps = React.HTMLAttributes<HTMLElement>;
 
 export const AddressLine = ({
   children,
@@ -41,11 +37,10 @@ AddressLine.propTypes = {
   className: PropTypes.string
 };
 
-interface AddressCityStateZipProps {
+interface AddressCityStateZipProps extends React.HTMLAttributes<HTMLElement> {
   city: React.ReactNode;
   state: React.ReactNode;
   zip: React.ReactNode;
-  className?: string;
 }
 
 export const AddressCityStateZip = ({
--- a/packages/react/src/components/Address/index.tsx
+++ b/packages/react/src/components/Address/index.tsx
@@ -1,0 +1,75 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+interface AddressProps {
+  children: React.ReactNode;
+  className?: string;
+}
+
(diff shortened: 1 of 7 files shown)


packages/react/src/components/Address/index.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+interface AddressProps {
+  children: React.ReactNode;
+  className?: string;
+}
```
stephenmathieson, line 8: TypeScript users will not be able to pass "other" props unless we explicitly state that they're supported. We also get `className` for free (as it's a valid prop for any element), so this should be:

```suggestion
interface AddressProps extends React.HTMLAttributes<HTMLElement> {
  children: React.ReactNode;
}
```

packages/react/src/components/Address/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+Address.displayName = 'Address';
+Address.propTypes = {
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string
+};
+
+interface AddressLineProps {
+  children?: React.ReactNode;
+  className?: string;
+}
```
stephenmathieson, line 25: This could just be:

```suggestion
type AddressLineProps = React.HTMLAttributes<HTMLElement>
```

packages/react/src/components/Address/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  children: PropTypes.node,
+  className: PropTypes.string
+};
+
+interface AddressCityStateZipProps {
+  city: React.ReactNode;
+  state: React.ReactNode;
+  zip: React.ReactNode;
+  className?: string;
+}
```
stephenmathieson, line 49:
```suggestion
interface AddressCityStateZipProps extends React.HTMLAttributes<HTMLElement> {
  city: React.ReactNode;
  state: React.ReactNode;
  zip: React.ReactNode;
}
```

stephenmathieson: Also, should `country` be supported here?

packages/styles/address.css
```diff
@@ -0,0 +1,7 @@
+:root {
+  --address-line-height: 1.618;
```
stephenmathieson, line 2: I don't know what our standards are here, but this variable doesn't seem very useful as it's only used in a single place 🤷

stephenmathieson (changes requested):
Minor TypeScript nits. Otherwise LGTM!

packages/react/src/components/Address/index.tsx
thuey: I was patterning this off of [DescriptionList](https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/DescriptionList/index.tsx), but I believe you are correct. I suspect DescriptionList would have similar issues for someone using TypeScript.

packages/react/src/components/Address/index.tsx
thuey: Good question, but no. Country should have its own line: https://faq.usps.com/s/article/International-addressing-requirements

packages/styles/address.css
thuey: This is another area where I was trying to pattern after what we do for [DescriptionList](https://github.com/dequelabs/cauldron/blob/develop/packages/styles/description-list.css#L1-L8)

stephenmathieson (approved):

stephenmathieson:
It would be nice if we had some bottom padding on the demo pages, as the Netlify tool bar truncates the page.

demo page truncated

PR #441, opened 2021-12-03
fix(react): Only render form field error when field is erroneous

dequejosie:


schne324 (dismissed):

packages/styles/forms.css
```diff
@@ -163,6 +163,10 @@ textarea.Field--has-error:focus:hover,
   padding-left: var(--space-half);
 }
 
+.Error:empty {
```
thuey, line 166: A few points of feedback:

1. Although the PR description says "BREAKING CHANGE", I wouldn't consider this a breaking change per-say. Have we flagged similar changes to this as breaking?
2. In my experience, I typically use `:empty` as a last resort because it is a little finicky. For example, someone might do something like this:
```js
  <div className="Error">
    <span className="errorMessage">
      {error}
    </span>
  </div>
```
If `{error}` is null or empty here, it wouldn't matter because the `<span>` would make CSS think `div.Error` isn't empty. In my opinion, it would be better and more reliable if we just didn't render the div container at all when we shouldn't be:
```js
  {error && (
    <div className="Error">
      <span className="errorMessage">
        {error}
      </span>
    </div>
  )}
```

3. If we were to take this approach, though, we should probably modify `.Error` itself to be `.Error:not(:empty)`, rather than add set of overrides `.Error:empty` because we might forget some properties (for example, in order for this truly not to take up space, you probably need to add overrides for `margin-bottom` and `padding-left` as well.

thuey (changes requested):
We should consider changing our approach here, unless there's some context I am missing. (See comment)

schne324:
>  I wouldn't consider this a breaking change per-say. Have we flagged similar changes to this as breaking?

I can't think of a time in the past where we were in a similar situation so unfortunately we don't have anything historical to lean on here. I told Josie to mark it as a breaking change originally but don't feel too strongly about it. I am starting to agree with you that this isn't a breaking change. @dequejosie , you can remove that from the commit message body.

> In my experience, I typically use :empty as a last resort because it is a little finicky. For example, someone might do something like this

imo we shouldn't worry much about people using our css hooks for 1-offs like you've posted above. We should treat our react components as the top-priorities in our styles package.  


---

I think there are many ways to solve this problem...I usually prefer to do this type of thing in react to keep our styles as minimally opinionated as possible. This means if I were to implement this "fix", I'd probably do something like `error && <div className="Error">{error}</div>`

packages/styles/forms.css
dequejosie: 1. @schne324 probably would have better idea;
2. For the particular three components, `Select`, `TextField`, and `CheckBox`, they don't have children inside the `Error` div, but I could totally picture cases similar to your `span` example occur at some point.
3. I was taking a shortcut and I agree add conditional rendering is probably safer.

thuey (approved):

PR #453, opened 2021-12-10
feat(react): add onChange prop to tabs when the active tab changes

scurker:
This is needed for dequelabs/axe-extension#1809 to react to when the active tab changes.

### DIFF
--- a/packages/react/src/components/Tabs/Tabs.tsx
+++ b/packages/react/src/components/Tabs/Tabs.tsx
@@ -11,6 +11,7 @@ type TabsProps = {
   initialActiveIndex?: number;
   thin?: boolean;
   className?: string;
+  onChange?: (this: HTMLDivElement, event: Event) => void;
 } & Cauldron.LabelProps;
 
 const Tabs = ({
@@ -18,9 +19,11 @@ const Tabs = ({
   thin,
   initialActiveIndex = 0,
   className,
+  onChange,
   ...labelProp
 }: TabsProps): JSX.Element => {
   const [activeIndex, setActiveIndex] = useState(initialActiveIndex);
+  const tabsRef = useRef<HTMLDivElement>(null);
   const focusedTabRef = useRef<HTMLLIElement>(null);
 
   const tabs = React.Children.toArray(children).filter(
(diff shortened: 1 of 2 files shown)


dequejosie (dismissed):

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -107,13 +110,28 @@ const Tabs = ({
 
   useDidUpdate(() => {
     focusedTabRef.current?.focus();
+    const event = new Event('change', { bubbles: true, cancelable: false });
+    focusedTabRef.current?.dispatchEvent(event);
```
schne324, line 114: why use a custom event when we could just call their onChange handler manually here? Are you thinking it'd be nice for the reason of delegation or something?

To me, calling their function here is simpler/better

schne324: Also, `change` is not a standard event for non form fields afaik. This could cause problems for apps using cauldron tabs who also have a useCapture `change` event

```js

document.body.addEventListener('change', e => {
  console.log('I expect e.target to be a checkbox or radio!')
}, true)
```

schne324: One more thing - if we _do_ need to go the event route, would it not make sense to [fire a custom event instead](https://github.com/dequelabs/cauldron/issues/162)?

scurker: This was only to have an `event` with the correct `target` property to match how `onChange` is typically called with form change events. I'm not opposed to just calling `onChange` directly if you think that's cleaner. I was originally thinking we might extend this later to allow `preventDefault`, but based on our slack conversation I don't think that's a route we would go down.

We could do `onChange: function({ activeTabIndex: number, target: HTMLLiElement })`, thoughts?

schne324: > We could do onChange: function({ activeTabIndex: number, target: HTMLLiElement }), thoughts?

👍 I like that

schne324 (approved):

PR #469, opened 2021-12-15
feat: replaced the panel implementation with a new one to match our pattern library

thuey:
BREAKING CHANGE: The panel implementation has been replaced, including removing props that are not used in the new implementation

This will be used in https://github.com/dequelabs/walnut/issues/2558

image

The old Panel has been renamed to IssuePanel

### DIFF
--- a/packages/react/src/components/Panel/index.tsx
+++ b/packages/react/src/components/Panel/index.tsx
@@ -1,33 +1,41 @@
-import React from 'react';
+import React, { HTMLAttributes, ReactElement, ReactNode } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import IconButton from '../IconButton';
 
-const Panel = ({
-  className,
-  title = '',
-  actions,
-  children
-}: {
+interface PanelProps extends HTMLAttributes<HTMLElement> {
+  children: ReactNode;
+  heading:
+    | ReactElement<any>
+    | {
+        text: ReactElement<any>;
+        level: number | undefined;
+      };
   className?: string;
-  title?: string;
-  actions?: typeof IconButton[];
-  children: React.ReactNode;
-}) => {
+}
+
+const Panel = ({ children, className, heading, ...other }: PanelProps) => {
+  const Heading = `h${
+    typeof heading === 'object' && 'level' in heading && !!heading.level
+      ? heading.level
+      : 2
+  }` as 'h1';
+
   return (
-    <div className={classNames('Panel', className)}>
-      <div className="Panel__Header">
-        {title && <div className="Panel__Header-title">{title}</div>}
-        {actions && <div className="Panel__Header-actions">{actions}</div>}
-      </div>
-      <div className="Panel__Content">{children}</div>
-    </div>
+    <section className={classNames('Panel', className)} {...other}>
+      <Heading className="Panel__Heading">
+        {typeof heading === 'object' && 'text' in heading
+          ? heading.text
+          : heading}
+      </Heading>
+      {children}
+    </section>
   );
 };
 
 Panel.displayName = 'Panel';
 Panel.propTypes = {
   children: PropTypes.node.isRequired,
+  heading: PropTypes.oneOfType([PropTypes.object, PropTypes.node]).isRequired,
   className: PropTypes.string
 };
 
--- a/packages/react/src/components/Panel/index.tsx
+++ b/packages/react/src/components/Panel/index.tsx
@@ -1,12 +1,14 @@
 import React, { HTMLAttributes, ReactElement, ReactNode } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import rndid from '../../utils/rndid';
 
 interface PanelProps extends HTMLAttributes<HTMLElement> {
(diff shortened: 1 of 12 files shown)


packages/react/src/components/Panel/index.tsx
```diff
@@ -1,33 +1,41 @@
...
 
-const Panel = ({
-  className,
-  title = '',
-  actions,
-  children
-}: {
+interface PanelProps extends HTMLAttributes<HTMLElement> {
+  children: ReactNode;
+  heading:
```
thuey, line 7: I did this to match what we are doing in the Dialog component in order to encourage good semantics and allow customization: https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/Dialog/index.tsx#L19-L24

packages/react/src/components/Panel/index.tsx
```diff
@@ -1,33 +1,41 @@
...
+
   return (
-    <div className={classNames('Panel', className)}>
-      <div className="Panel__Header">
-        {title && <div className="Panel__Header-title">{title}</div>}
-        {actions && <div className="Panel__Header-actions">{actions}</div>}
-      </div>
-      <div className="Panel__Content">{children}</div>
-    </div>
+    <section className={classNames('Panel', className)} {...other}>
```
thuey, line 24: This seems like it should be a section to me. However, I'm open to being corrected.

packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -2,9 +2,9 @@ import { Cauldron } from '../../types';
 import React, { useState, useEffect, useRef } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { useDidUpdate } from '../../index';
 import Tab from './Tab';
 import { useId } from 'react-id-generator';
+import useDidUpdate from '../../utils/use-did-update';
```
thuey, line 7: We were getting a warning about a cyclic dependency. I made a quick tweak to address it.

scurker:
This does show up in a future design for IGT results:

image

Let's discuss this for a path forward next time we're all at standup.

thuey:
> This does show up in a future design for IGT results:

@scurker Yeah, I posted about that in our slack thread:

> I spoke with @aaronpe about this, and we agree that the usage of that parent component ("Your Products") in the pattern library is more like a container than a card. Also, that the current <Panel> component wasn't necessarily intended to be added to the pattern library, since its only known usage is for IGT results. Since the current <Panel> component is unused (and as Josie referenced in [#413](https://github.com/dequelabs/cauldron/issues/413) we were already thinking of getting rid of it), I'd like to totally replace the <Panel> component with one that fits the design patterns we have. I'll include a comment on https://github.com/dequelabs/axe-extension/issues/1771 noting where the source code for the old <Panel> can be found

I'm happy to discuss this in standup, though 👍

thuey:
In standup, we discussed leaving the old Panel's code in place, but renaming it to `IssuePanel`

dequejosie:
I know we don't need dark theme for the web app, but we might need to consider implementing dark theme as well, or at least make the content visible for dark theme. Do we have a design for dark theme panel? @awpearlm 

Screen Shot 2021-12-17 at 4 56 51 PM

packages/styles/panel.css
```diff
@@ -1,54 +1,22 @@
...
+  --panel-heading-color: var(--gray-90);
 }
 
 .Panel {
+  padding: var(--space-large);
   border: 1px solid var(--panel-border-color);
+  border-radius: 3px;
+  box-shadow: 0px 2px 3px rgba(0, 0, 0, 0.15);
+  background-color: var(--panel-background-color);
+  width: 100%;
```
dequejosie, line 13: We might be able to remove this line?

packages/react/src/components/Panel/index.tsx
dequejosie: I agree with you here

dylanb: you should `aria-labelledby` this `<section>` by pointing it at the heading (accessibility best practice)

dylanb (changes requested):
just the aria-labelledby on the section

packages/react/src/components/Panel/index.tsx
thuey: @dylanb Thanks. I looked elsewhere in our apps where we were using `<section>`, and it didn't seem like we were using `aria-labelledby`, which seemed odd to me. I will go ahead and add it here.

thuey:
> I know we don't need dark theme for the web app, but we might need to consider implementing dark theme as well, or at least make the content visible for dark theme. Do we have a design for dark theme panel? @awpearlm
> 
> Screen Shot 2021-12-17 at 4 56 51 PM

Good catch, @dequejosie! I got this from @awpearlm : https://preview.uxpin.com/4052c8b6f05ce03e28813f28deb3a682950c2ac6#/pages/144239874

packages/styles/panel.css
thuey: Yep, I think you're right.

dylanb (approved):

PR #470, opened 2021-12-15
feat: Add variant 'dismiss' to Tag

dequejosie:
UXPin: https://preview.uxpin.com/68df515d0023cc4b79f25d43f7663e24e363b1b9#/pages/141465421/specification/sitemap;
https://preview.uxpin.com/4052c8b6f05ce03e28813f28deb3a682950c2ac6#/pages/140808128/specification/sitemap

### DIFF
--- a/packages/react/src/components/Tag/index.tsx
+++ b/packages/react/src/components/Tag/index.tsx
@@ -1,10 +1,15 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import Icon from '../Icon';
+
+type Variant = 'dismiss';
 
 interface TagProps {
   children: React.ReactNode;
   className?: string;
+  variant?: Variant;
+  onDismiss?: (event: React.MouseEvent<HTMLButtonElement>) => void;
 }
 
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Tag/index.tsx
```diff
@@ -18,14 +23,45 @@ TagLabel.propTypes = {
...
+      className={classNames('Tag', className, {
+        'Tag--hidden': !show,
+        'Tag--dismiss': variant === 'dismiss'
+      })}
+      {...other}
+    >
+      {children}
+      {variant === 'dismiss' && <Icon type="close" />}
+    </Component>
+  );
```
schne324, line 58: I think this could be simplified to something like this:

```suggestion
  return (
    <Component
      className={classNames('Tag', className, {
        'Tag--hidden': !show,
        'Tag--dismiss': variant === 'dismiss'
      })}
      {...other}
      onClick={handleDismiss}
    >
      {children}
      {variant === 'dismiss' && <Icon type="close" />}
    </Component>
  );
```

dequejosie: Combined the two variant prs into one and used Union type for better typing!

PR #480, opened 2021-12-17
feat: Add 'dismiss' and 'toggle' variants to Tag

dequejosie:
UXPin: https://preview.uxpin.com/68df515d0023cc4b79f25d43f7663e24e363b1b9#/pages/141465421/specification/sitemap;
https://preview.uxpin.com/4052c8b6f05ce03e28813f28deb3a682950c2ac6#/pages/140808128/specification/sitemap

This pr took an approach that is more strict on types.


packages/react/src/components/Tag/index.tsx
```diff
@@ -18,14 +45,79 @@ TagLabel.propTypes = {
...
 
-const Tag = ({ children, className, ...other }: TagProps) => (
-  <div className={classNames('Tag', className)} {...other}>
-    {children}
-  </div>
-);
+const Tag = (props: TagProps) => {
+  // render dismiss Tag
+  if (props.variant === 'dismiss') {
+    const [show, setShow] = useState(true);
```
thuey, line 51: It seems like it would be good if `show` came in via a prop, similar to how the toggle variant gets `toggleState`. It would be more consistent and flexible.

packages/react/src/components/Tag/index.tsx
```diff
@@ -18,14 +45,79 @@ TagLabel.propTypes = {
...
+  // render toggle Tag
+  if (props.variant === 'toggle') {
+    const {
+      children,
+      className,
+      toggleState,
+      buttonLabel,
+      onToggle,
+      toggleOnText = 'ON',
+      toggleOffText = 'OFF',
```
thuey, line 83: Two things:

1. Will there always be toggle text? If so, I suppose this is okay. If not, though, we may want to leave it up to the children that are passed in.
2. It doesn't look like we have i18n in cauldron, so I am concerned about how this will get translated.

schne324: > It doesn't look like we have i18n in cauldron, so I am concerned about how this will get translated.

Isn't that precisely why we'd want to support a prop (or as you said children)?

thuey: @schne324 What I meant by that comment was if we did have specific toggle props, then we should not provide a default value for them of `'ON'` and `'OFF'` because they would not be translated. If we do have separate props, then they should not have a default value because. However, if we just rely on `children` alone, that point is moot.

thuey: To follow up a bit on this, @scurker pointed out that we do have default strings elsewhere (see: https://github.com/dequelabs/cauldron/pull/501#discussion_r783125464).

However, the point still stands here that I think this should just come in via children anyway, rather than standalone props.

schne324: >  I think this should just come in via children

sounds good to me

scurker:
Closing as this PR > 1 year old. Please re-open if you would like to address it.

PR #485, opened 2021-12-23
feat(react,styles): Adds pagination component

schne324:
Ref #205

### DIFF
--- a/docs/patterns/components/Pagination/index.js
+++ b/docs/patterns/components/Pagination/index.js
@@ -1,0 +1,59 @@
+import React, { useState } from 'react';
+import Demo from '../../../Demo';
+import { Pagination } from '@deque/cauldron-react/';
+import { children, className } from '../../../props';
+
+const PaginationDemo = () => {
+  const totalItems = 111;
+  const itemsPerPage = 25;
+  const [currentPage, setCurrentPage] = useState(3);
+  const itemStart = currentPage * itemsPerPage - itemsPerPage + 1;
+  const itemEnd = Math.min(itemStart + itemsPerPage - 1, totalItems);
+
+  const onNextPageClick = () => setCurrentPage(currentPage + 1);
+  const onFirstPageClick = () => setCurrentPage(1);
+  const onPreviousPageClick = () => setCurrentPage(currentPage - 1);
+  const onLastPageClick = () => setCurrentPage(5);
+
+  return (
+    <div>
+      <Demo
+        component={Pagination}
+        states={[
+          { totalItems: 15 },
+          // TODO: don't render this when in light mode?
+          { totalItems: 15, variant: 'dark' },
+          {
+            totalItems,
+            itemsPerPage,
+            currentPage,
+            statusLabel: (
+              <span>
+                <strong>{itemStart}</strong> - <strong>{itemEnd}</strong> of{' '}
+                <strong>{totalItems}</strong>
+              </span>
+            ),
+            firstPageLabel: 'FIRST PAGE!!',
+            nextPageLabel: 'NEXT PAGE!!',
+            previousPageLabel: 'PREV PAGE!!',
+            lastPageLabel: 'JIMMY PAGE!!',
+            onNextPageClick,
+            onFirstPageClick,
+            onPreviousPageClick,
+            onLastPageClick
+          }
+        ]}
+        propDocs={{
+          totalItems: {
+            type: 'number',
+            description:
+              'The total number of items being rendered (not the total number of pages)',
+            required: true
+          }
+        }}
+      />
+    </div>
+  );
+};
+
+export default PaginationDemo;
--- a/docs/patterns/components/Pagination/index.js
+++ b/docs/patterns/components/Pagination/index.js
@@ -21,8 +21,6 @@ const PaginationDemo = () => {
         component={Pagination}
         states={[
           { totalItems: 15 },
-          // TODO: don't render this when in light mode?
-          { totalItems: 15, variant: 'dark' },
           {
             totalItems,
             itemsPerPage,
--- a/docs/patterns/components/Pagination/index.js
+++ b/docs/patterns/components/Pagination/index.js
@@ -45,8 +45,89 @@ const PaginationDemo = () => {
           totalItems: {
             type: 'number',
             description:
-              'The total number of items being rendered (not the total number of pages)',
+              'The total number of items being rendered (not the total number of pages).',
             required: true
+          },
+          itemsPerPage: {
+            type: 'number',
+            description: 'The total number of items per page.',
+            required: false,
+            default: 10
+          },
+          currentPage: {
+            type: 'number',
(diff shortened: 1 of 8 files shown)


docs/patterns/components/Pagination/index.js
```diff
@@ -0,0 +1,59 @@
...
+  const onPreviousPageClick = () => setCurrentPage(currentPage - 1);
+  const onLastPageClick = () => setCurrentPage(5);
+
+  return (
+    <div>
+      <Demo
+        component={Pagination}
+        states={[
+          { totalItems: 15 },
+          // TODO: don't render this when in light mode?
```
schne324, line 24: I'll need to get clarification on this. The designs only have this additional variant in dark mode

anitradeque: Do not include the additional variant at all, per @awpearlm

packages/react/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <div
+        ref={ref}
+        className={classNames('Pagination', className, {
+          'Pagination--dark': variant === 'dark'
+        })}
+        {...other}
+      >
+        <ul>
+          <li>
+            {currentPage === 1 ? (
```
scurker, line 62: Why not `isFirstPage` as well?

```jsx
const isFirstPage = currentPage === 1
```

packages/react/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,157 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import IconButton from '../IconButton';
+import TooltipTabstop from '../TooltipTabstop';
+import Icon from '../Icon';
+
+type PaginationVariant = 'dark';
+
+interface Props {
```
scurker, line 10:
```suggestion
interface Props extends React.HTMLAttributes<HTMLDivElement> {
```

packages/react/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+      ...other
+    },
+    ref
+  ) => {
+    const itemStart = currentPage * itemsPerPage - itemsPerPage + 1;
+    const itemEnd = Math.min(itemStart + itemsPerPage - 1, totalItems);
+    const isLastPage = itemEnd === totalItems;
+
+    return (
+      <div
```
scurker, line 53: Should this be `nav` or `role="navigation"`?

packages/react/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+        ref={ref}
+        className={classNames('Pagination', className, {
+          'Pagination--dark': variant === 'dark'
+        })}
+        {...other}
+      >
+        <ul>
+          <li>
+            {currentPage === 1 ? (
+              <TooltipTabstop tooltip={firstPageLabel} placement="bottom">
```
scurker, line 63: Default placement seems odd, I guess we're always assuming the pagination is on the bottom of the thing its paginating?

packages/styles/pagination.css
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.Pagination > ul > li {
+  margin-right: 5px;
+}
+
+.Pagination button {
+  padding: var(--space-three-quarters);
+  height: 36px;
+  width: 36px;
```
scurker, line 14: Nitpick: Button height should scale with user set font-size, so rems would be preferable here.

scurker (dismissed):
Only props extending whatever the element should be is blocking, everything else is just some nitpicky things I noticed.

packages/react/src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+              <TooltipTabstop tooltip={previousPageLabel} placement="bottom">
+                <Icon type="chevron-left" />
+              </TooltipTabstop>
+            ) : (
+              <IconButton
+                icon="chevron-left"
+                tooltipPlacement="bottom"
+                label={previousPageLabel}
+                onClick={onPreviousPageClick}
+              />
```
scurker, line 87: Some thoughts here (not anything we need to change for this PR):

We may want to consider eventually adding a custom render function here because this is making an assumption that pagination is going to be working in state, but it's possible they could be navigation based pagination. It doesn't seem like it's anything we need _right now_, but something to keep in mind for the future.

packages/react/src/components/Pagination/index.tsx
schne324: 🤷 I originally coded it as a `<nav />` but felt kinda funky considering all the controls are buttons. I think having it as a list conveys everything we need but am more than happy to change this back to a `<nav />`.

scurker: I'm not picky here as we can slate this for a future improvement™ per comment https://github.com/dequelabs/cauldron/pull/485#discussion_r775977006

packages/react/src/components/Pagination/index.tsx
schne324: I agree. I made it bottom because I assume pagination will primarily be below a table. The default was unacceptable because it covered up other buttons in the pagination component.  I'm gathering that we should probably make this configurable via a prop (defaulting to `bottom` still makes sense to me 🤷 )

packages/styles/pagination.css
schne324: good catch

docs/patterns/components/Pagination/index.js
```diff
@@ -0,0 +1,57 @@
...
+            nextPageLabel: 'NEXT PAGE!!',
+            previousPageLabel: 'PREV PAGE!!',
+            lastPageLabel: 'JIMMY PAGE!!',
+            onNextPageClick,
+            onFirstPageClick,
+            onPreviousPageClick,
+            onLastPageClick
+          }
+        ]}
+        propDocs={{
```
schne324, line 44: just realized I haven't documented all the props...will update tomorrow

packages/styles/pagination.css
```diff
@@ -0,0 +1,63 @@
...
+  border-color: transparent;
+}
+
+.cauldron--theme-dark .Pagination button[aria-disabled='true'] {
+  color: #5d676f;
+}
+
+.cauldron--theme-dark .Pagination [role='log'] {
+  color: var(--accent-light);
+}
```
scurker, line 60: Might be good to add [`font-variant-numeric`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric):

```suggestion
.cauldron--theme-dark .Pagination [role='log'] {
  color: var(--accent-light);
  font-variant-numeric: tabular-nums;
}
```

I came across this yesterday, and this will keep more consistent numeric spacing when navigating through the pagination of items.

scurker (changes requested):
1 very small minor thing

scurker (approved):

PR #501, opened 2022-01-07
feat(react,styles): add two-column panel component

### LINKED ISSUE #488: Implement 2-column component pattern
A `<TwoColumnPanel>` component will need to support housing for vertical tabs or a navigation list and collapse down to a single column when necessary.

# Two-Column Panel

```ts
interface TwoColumnLayoutProps {
  initialCollapsed?: boolean // default = false
  showCollapsedPanelLabel?: string // default = "Show Panel"
  hideCollapsedPanelLabel?: string // default = "Hide Panel"
}
```

## Example with nav list

```jsx
<TwoColumnPanel>
  <ColumnLeft aria-labelledby="sidebar">
    <ColumnHeader id="sidebar">Items</ColumnHeader>
    <nav aria-label="Sidebar navigation">
      <ul>
        {items.map(item => (
          <li key={item.id}>
            <a
              href={item.url}
              onClick={selectItem(item.id)} aria-current={item.id === selectedItem.id}
              aria-current={item.id === selectedItem.id}
            >
              {item.name}
              <em>{item.description}</em>
            </a>
          </li>
        ))}
      </ul>
    </nav>
  </ColumnLeft>
  <ColumnRight aria-labelledby="item-header">
    <ColumnHeader id="item-header">{selectedItem?.name}</ColumnHeader>
    <div>
      {selectedItem?.contents}
    </div>
  </ColumnRight>
</TwoColumnPanel>
```

scurker:
closes #488

### DIFF
--- a/docs/patterns/components/TwoColumnPanel/index.js
+++ b/docs/patterns/components/TwoColumnPanel/index.js
@@ -5,8 +5,14 @@ import {
   ColumnRight,
   ColumnHeader,
   ColumnGroupHeader,
-  ColumnList
+  ColumnList,
+  Breadcrumb,
+  BreadcrumbLink,
+  BreadcrumbItem,
+  Code
 } from '@deque/cauldron-react/';
+import PropDocs from '../../../Demo/PropDocs';
+import { children, className } from '../../../props';
 
 function TwoColumnPanelDemo() {
   const [selected, setSelected] = useState(1);
@@ -38,13 +44,13 @@ function TwoColumnPanelDemo() {
       id: 1,
       name: 'One',
       description: 'Short description of one',
-      contents: 'Content for One'
+      contents: 'Content for grouped One'
     },
     {
       id: 2,
       name: 'Two',
       description: 'Short description of two',
-      contents: 'Content for Two'
+      contents: 'Content for grouped Two'
     }
   ];
   const groupTwo = [
@@ -52,13 +58,13 @@ function TwoColumnPanelDemo() {
       id: 3,
       name: 'Three',
       description: 'Short description of three',
-      contents: 'Content for Three'
+      contents: 'Content for grouped Three'
     },
     {
       id: 4,
       name: 'Four',
       description: 'Short description of four',
-      contents: 'Content for Four'
+      contents: 'Content for grouped Four'
     }
   ];
 
@@ -79,9 +85,16 @@ function TwoColumnPanelDemo() {
     <div className="twocolumnpanel-demo">
       <h1>Two Column Panel</h1>
 
+      <p>
+        Two column panel is a container component intended to provide a
+        navigable list of items that can be collapsed if not needed.
+      </p>
+
+      <h2>Basic</h2>
+
       <TwoColumnPanel>
         <ColumnLeft aria-labelledby="sidebar">
-          <ColumnHeader id="sidebar">Sidebar label</ColumnHeader>
+          <ColumnHeader id="sidebar">Items</ColumnHeader>
           <nav aria-label="Sidebar navigation">
             <ul>
               {items.map(item => (
@@ -119,11 +132,41 @@ function TwoColumnPanelDemo() {
         </ColumnRight>
       </TwoColumnPanel>
 
+      <h3>Example</h3>
+
+      <Code>{`<TwoColumnPanel>
+  <ColumnLeft aria-labelledby="sidebar">
+    <ColumnHeader id="sidebar">Items</ColumnHeader>
+    <nav aria-label="Sidebar navigation">
+      <ul>
+        {items.map(item => (
+          <li key={item.id}>
+            <a
+              href={item.url}
+              onClick={selectItem(item.id)} aria-current={item.id === selectedItem.id}
+              aria-current={item.id === selectedItem.id}
+            >
+              {item.name}
+              <em>{item.description}</em>
+            </a>
+          </li>
+        ))}
+      </ul>
+    </nav>
+  </ColumnLeft>
+  <ColumnRight aria-labelledby="item-header">
+    <ColumnHeader id="item-header">{selectedItem?.name}</ColumnHeader>
+    <div>
+      {selectedItem?.contents}
+    </div>
+  </ColumnRight>
+</TwoColumnPanel>`}</Code>
+
       <h2>With Optional Group Heading</h2>
 
       <TwoColumnPanel>
         <ColumnLeft aria-labelledby="group-heading">
-          <ColumnHeader>Items</ColumnHeader>
+          <ColumnHeader>Grouped Items</ColumnHeader>
           <ColumnList>
             <ColumnGroupHeader id="group-heading">
               <h3>Optional group heading</h3>
@@ -166,8 +209,14 @@ function TwoColumnPanelDemo() {
           </ColumnList>
         </ColumnLeft>
         <ColumnRight aria-labelledby="group-contents-heading">
-          <ColumnHeader id="group-contents-heading">
-            {selectedGroupItem?.name}
+          <ColumnHeader>
+            <Breadcrumb aria-label="content breadcrumbs">
+              <BreadcrumbLink href="#">Context A</BreadcrumbLink>
+              <BreadcrumbLink href="#">Context B</BreadcrumbLink>
+              <BreadcrumbItem id="group-contents-heading">
+                Grouped {selectedGroupItem?.name}
+              </BreadcrumbItem>
+            </Breadcrumb>
           </ColumnHeader>
           <div>
             {selectedGroupItem?.contents}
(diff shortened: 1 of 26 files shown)


docs/patterns/components/TwoColumnPanel/index.js
```diff
@@ -0,0 +1,361 @@
+import React, { useRef, useState } from 'react';
```
thuey, line 1: Not sure if it's just me, but when I toggle back and forth a few times, the page gets slower and slower until it crashes.

https://user-images.githubusercontent.com/6416791/149158736-2c458c83-2dc9-40dc-830f-3bfc9bdf0ed9.mov

packages/react/src/components/TwoColumnPanel/TwoColumnPanel.tsx
```diff
@@ -0,0 +1,244 @@
...
+}
+
+const TwoColumnPanel = forwardRef<HTMLDivElement, TwoColumnPanelProps>(
+  (
+    {
+      className,
+      children,
+      initialCollapsed = false,
+      showCollapsedPanelLabel = 'Show Panel',
+      hideCollapsedPanelLabel = 'Hide Panel',
```
thuey, line 31: I don't think we should provide default values for this, since they wouldn't be "internationalized". I think we should instead make them required so the consumer knows they are responsible for providing a translated string.

packages/react/src/components/TwoColumnPanel/TwoColumnPanel.tsx
```diff
@@ -0,0 +1,244 @@
...
+      ...props
+    },
+    ref
+  ) => {
+    const [isCollapsed, setCollapsed] = useState(initialCollapsed);
```
thuey, line 36: We don't have to do this now, but it might make for a more flexible component if the collapsed state could be managed by the parent component.

packages/react/src/components/TwoColumnPanel/TwoColumnPanel.tsx
```diff
@@ -0,0 +1,244 @@
...
+        columnLeftRef.current?.removeEventListener(
+          'transitionend',
+          handleTransitionEnd
+        );
+      };
+    }, [columnLeftRef.current, isCollapsed]);
+
+    // When the collapsable panel starts to overlay content, it needs to become a focus trap and collapsed by default
+    useLayoutEffect(() => {
+      const mediaQueryList = matchMedia('(max-width: 45rem)');
```
thuey, line 166: Two suggestions:

1. Should this media query be configurable?
2. Should we go ahead and create a `useMediaQuery` utility in cauldron to make this more reusable?

docs/patterns/components/TwoColumnPanel/index.js
```diff
@@ -0,0 +1,361 @@
...
+        Two column panel is a container component intended to provide a
+        navigable list of items that can be collapsed if not needed.
+      </p>
+
+      <h2>Basic</h2>
+
+      <TwoColumnPanel>
+        <ColumnLeft aria-labelledby="sidebar">
+          <ColumnHeader id="sidebar">Items</ColumnHeader>
+          <nav aria-label="Sidebar navigation">
```
thuey, line 98: For my own learning on accessibility, should these follow the keyboard navigation patterns for menus? https://www.w3.org/TR/wai-aria-practices/#menu

However, I noticed that the walnut app does not allow for keyboard navigation with arrow keys. Why is that?

docs/patterns/components/TwoColumnPanel/index.js
```diff
@@ -0,0 +1,361 @@
...
+      <h2>With Optional Group Heading</h2>
+
+      <TwoColumnPanel>
+        <ColumnLeft aria-labelledby="group-heading">
+          <ColumnHeader>Grouped Items</ColumnHeader>
+          <ColumnList>
+            <ColumnGroupHeader id="group-heading">
+              <h3>Optional group heading</h3>
+            </ColumnGroupHeader>
+            <nav aria-label="Sidebar group 1 navigation">
```
thuey, line 174: https://github.com/dequelabs/cauldron/issues/488 mentions supporting vertical tabs. It might be nice to include an example of how to do that in a semantic tab + tab panel way.

packages/styles/two-column-panel.css
```diff
@@ -0,0 +1,279 @@
...
+  .TwoColumnPanel__ButtonToggle button {
+    border-bottom: none;
+  }
+
+  .TwoColumnPanel--show .TwoColumnPanel__Left {
+    transform: translateX(100%);
+  }
+
+  .TwoColumnPanel--hide .TwoColumnPanel__Left {
+    transform: translateX(-10vw);
```
thuey, line 277: IMO, it is a little weird that the left panel flies in from off screen, as opposed to just from the left edge of the container. If there was anything to the left of the two column layout, it would appear as though it was sliding on top of the stuff to the left. That being said, I can't image the two column layout ever being used with content to the left of it, so it is probably okay.

thuey (changes requested):
This is looking really great!

docs/patterns/components/TwoColumnPanel/index.js
scurker: I need to investigate this, I think the tooltip (popper.js) has some performance issues in the past so ideally we need to resolve this.

packages/styles/two-column-panel.css
scurker: I agree, but I had to cheat a little bit here for a couple of reasons:

In order to properly animate the component to the left edge, we would need `overflow: hidden` on the wrapping container. This unfortunately would also clip things that overflow (like focus rings, tooltip) which is an undesirable behavior.

If we knew what the distance between the left edge of the container to the left edge of the viewport, we could calculate a fixed value to ensure the panel animates the correct amount but unfortunately we don't know that from css. The `10vw` is an approximation to try to get the panel off the screen. Alternatively, I suppose we could set the css transform via javascript and calculate the proper animation distance if we really wanted.

docs/patterns/components/TwoColumnPanel/index.js
scurker: @schne324 and I went back and forth on this a couple of times and decided not to support tabs because there were some significant accessibility concerns with supporting a tab panel:

* `TabPanel` has an aria relationship with its associated tabs meaning that hiding the tabs potentially has accessibility issues at worst, or confusing at the least when the tabs are hidden
* Depending on how the left panel was hidden we would have a couple of options of ensuring hidden content was not able to be navigated to: We could use `inert` with a polyfill, `tabindex="-1`" on other focusable elements, or implement a focus trap.

We decided to go with a focus trap, and given that it's not a common pattern to have a tab's controls hidden the focus trap + hiding the panel felt like the least worst option. If you think it's worth documenting some of these decisions in the component I'm happy to oblige.

packages/styles/two-column-panel.css
thuey: That makes sense. I think it's fine for now. We can revisit if we need to.

docs/patterns/components/TwoColumnPanel/index.js
thuey: Yeah, I think documenting the usage/limitations with reasonings of the component would be good for posterity.

packages/react/src/components/TwoColumnPanel/TwoColumnPanel.tsx
scurker: There's a couple of other places where we provide default text values:

* https://github.com/dequelabs/cauldron/blob/3127ddebe7d616863d47dcbe76cd4f30589db5fa/packages/react/src/components/Dialog/index.tsx#L39
* https://github.com/dequelabs/cauldron/blob/3127ddebe7d616863d47dcbe76cd4f30589db5fa/packages/react/src/components/Pagination/index.tsx#L33-L37
* https://github.com/dequelabs/cauldron/blob/3127ddebe7d616863d47dcbe76cd4f30589db5fa/packages/react/src/components/Pointout/index.tsx#L56-L57
* https://github.com/dequelabs/cauldron/blob/3127ddebe7d616863d47dcbe76cd4f30589db5fa/packages/react/src/components/SkipLink/index.tsx#L23-L24

This provides at a minimum accessible by default since the lack of a label would be an accessibility violation. I hope anyone using it provides a translated string but it would be hard to enforce that from cauldron, so at a minimum ensuring that some label is sets a baseline of accessibility.

scurker: > Interesting. Wouldn't making them required have the same effect of making sure we have a label?

Not necessarily.

If they don't use typescript, we can't enforce it with typescript. PropTypes isn't really an enforcement as it just prompts a console log in devtools when rendering about missing props.

docs/patterns/components/TwoColumnPanel/index.js
scurker: Found the issue, will be addressed with #505

docs/patterns/components/TwoColumnPanel/index.js
scurker: From a UX perspective, I wouldn't really consider this a "menu" since it's not a fixed set of items or actions but rather navigation links that change context on the page. There's also a [good article by Heydon Pickering](https://inclusive-components.design/menus-menu-buttons/) on this behavior. I think for similar reasons that are mentioned in the above article are likely why walnut does not use this pattern for its navigation.

thuey:
@scurker The hide/show tooltip appears to be a bit all over the place:
image

thuey:
@scurker  I believe https://github.com/dequelabs/cauldron/pull/547 will fix the tooltip issue

thuey (approved):

PR #502, opened 2022-01-07
feat(react,styles): add breadcrumb component

scurker:
this is needed for #488 

two column component with breadcrumb in the header

### DIFF
--- a/docs/index.js
+++ b/docs/index.js
@@ -71,7 +71,8 @@ const componentsList = [
   'Address',
   'Pagination',
   'IssuePanel',
-  'FieldWrap'
+  'FieldWrap',
+  'Breadcrumb'
 ].sort();
 
 const App = () => {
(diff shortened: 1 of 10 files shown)


schne324 (changes requested):
we should be setting `aria-current="page"` or `aria-current="location"`

some great examples here: https://scottaohara.github.io/a11y_breadcrumbs/

scurker:
> we should be setting `aria-current="page"` or `aria-current="location"`
> 
> some great examples here: https://scottaohara.github.io/a11y_breadcrumbs/

That wasn't my interpretation of the practices from [wai-aria](https://www.w3.org/TR/wai-aria-practices/#wai-aria-roles-states-and-properties-2):

> The link to the current page has aria-current set to page. If the element representing the current page is not a link, aria-current is optional.

I quickly threw this together because it's a part of Aaron's design but wasn't sure how it was exactly going to be used. I'm not sure if we can always assume if the breadcrumb will be a link, but we can bring it up for discussion next week.

schne324 (approved):

PR #504, opened 2022-01-11
fix: fix loader

thuey:
Fixing a couple of regressions with the loader:

- Allow `label` and `variant` props to be set for the loader overlay. (Since they are not included in the propTypes, they can't be set)
- Restore the default layout styles for the loader (previously, was `50px auto`)

### DIFF
--- a/packages/react/src/components/LoaderOverlay/index.tsx
+++ b/packages/react/src/components/LoaderOverlay/index.tsx
@@ -9,34 +9,35 @@ interface LoaderOverlayProps extends React.HTMLAttributes<HTMLDivElement> {
   label?: string;
 }
 
-const LoaderOverlay = React.forwardRef<
-  HTMLDivElement,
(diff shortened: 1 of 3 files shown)


packages/react/src/components/LoaderOverlay/index.tsx
```diff
@@ -9,34 +9,35 @@ interface LoaderOverlayProps extends React.HTMLAttributes<HTMLDivElement> {
...
+      </div>
+      {label ? <span className="Loader__overlay__label">{label}</span> : null}
+      {other.children}
     </div>
-    {label ? <span className="Loader__overlay__label">{label}</span> : null}
-    {other.children}
-  </div>
-));
+  )
+);
```
thuey, line 35: All that changed here was `React.HTMLAttributes<HTMLDivElement>` -> `LoaderOverlayProps`

scurker (approved):

Issue #489, opened 2022-01-04
Support sortability of `<Table />`


schne324:
[see uxpin's sortable table examples](https://preview.uxpin.com/4052c8b6f05ce03e28813f28deb3a682950c2ac6#/pages/140808126/simulate/sitemap)

As far as accessibility goes, feel free to ping @schne324 or check out [dqu's example](https://dequeuniversity.com/library/aria/table-sortable)

The callers of the component will handle the actual sorting of data and it is up to them to inform the `<Table />` component of what the sort state is.

dequejosie:
This is also needed by https://github.com/dequelabs/walnut/issues/2665

somaalapati:
Verified this ticket by testing the sorting feature on the users table in ticket https://github.com/dequelabs/walnut/issues/2663. Everything looks good, so closing the ticket.

PR #511, opened 2022-01-14
feat: Make Table support sorting

dequejosie:
Ref: https://github.com/dequelabs/cauldron/issues/489


docs/patterns/components/Table/index.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
```
thuey, line 1: Also, with the demo, after changing the sort direction with the keyboard, focus gets lost. I believe it should stay on the sort button

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -1,21 +1,53 @@
...
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import Icon from '../Icon';
+
+// these match aria-sort's values
+type SortDirection = 'ascending' | 'descending' | 'none';
 
 interface TableHeaderProps extends ThHTMLAttributes<HTMLTableCellElement> {
   children: React.ReactNode;
+  sortDir?: SortDirection;
```
thuey, line 11: I'm not sure what we have done elsewhere/if we have a standard, but my preference is to avoid abbreviations generally (`sortDir` -> `sortDirection`)

docs/patterns/components/Table/index.js
```diff
@@ -19,63 +19,110 @@ const tableComponentNames = [
...
+    );
+
+  const sortedData = sortBy ? comparator(sampleData) : sampleData;
+
+  const getSortDir = column => (column === sortBy ? sortDir : 'none');
+  /* asc => desc
+     none => asc
+     desc => asc */
+  const setSortDir = column =>
+    column === sortBy && sortDir === 'ascending' ? 'descending' : 'ascending';
```
thuey, line 63: This function name is a little misleading, since it doesn't actually set the sort direction. I think it would be less confusing (and more useful) if this function actually did set the sort direction. It could then be called with `setSortDir('last_name')` instead of doing `setSort(['last_name', setSortDir('last_name')])`

docs/patterns/components/Table/index.js
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
```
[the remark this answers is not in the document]
dequejosie: Yeah, I notice that too. Will look into what exactly is causing it.

docs/patterns/components/Table/index.js
dequejosie: I will go with your suggestion unless others have a different opinion. @schne324 @scurker

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -1,21 +1,64 @@
...
+      aria-sort={sortDirection}
+      className={classNames('TableHeader', className, {
+        'TableHeader--sorting': sortDirection && sortDirection !== 'none'
+      })}
+      {...other}
+    >
+      {onSort && sortDirection ? (
+        <button ref={sortButtonRef} onClick={onSort} className="sort-button">
+          {children}
+          <span>
```
scurker, line 41: Minor nitpick: We should be able to add `aria-hidden="true"` here
```suggestion
          <span aria-hidden="true">
```

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -1,21 +1,64 @@
...
+  return (
+    <th
+      aria-sort={sortDirection}
+      className={classNames('TableHeader', className, {
+        'TableHeader--sorting': sortDirection && sortDirection !== 'none'
+      })}
+      {...other}
+    >
+      {onSort && sortDirection ? (
+        <button ref={sortButtonRef} onClick={onSort} className="sort-button">
```
scurker, line 39: This is too generic of a class name, we should follow our [documented BEM syntax](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#styles).

An example of a name that would be appropriate would be `TableHeader__sort-button`

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -1,21 +1,64 @@
...
+    if (!sortDirection || sortDirection === 'none') {
+      return;
+    }
+    sortButtonRef.current?.focus();
+  }, [sortDirection]);
+  return (
+    <th
+      aria-sort={sortDirection}
+      className={classNames('TableHeader', className, {
+        'TableHeader--sorting': sortDirection && sortDirection !== 'none'
```
scurker, line 34: I don't think we would want different styles, but might be nice to have the full state here:

* ascending: `TableHeader--sort-ascending`
* descending: `TableHeader--sort-descending`

packages/styles/table.css
```diff
@@ -30,6 +36,44 @@
...
+.TableHeader .sort-button {
+  background-color: transparent;
+  width: 100%;
+  /* a fixed height prevent TableHeader from being jumpy */
+  height: 1.625rem;
+  display: flex;
+  align-items: center;
+  align-content: center;
+  padding-left: 0;
+  font-family: 'Roboto', Helvetica, Arial, sans-serif;
```
scurker, line 48: Why is `font-family` being set here, shouldn't that be inherited?

packages/react/src/components/Table/TableHeader.tsx
scurker, line 39:
```suggestion
        <button ref={sortButtonRef} onClick={onSort} className="sort-button" type="button">
```

[`button` defaults to  `submit`](https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element) in the html spec, so we should always specify the type.

packages/styles/table.css
```diff
@@ -30,6 +36,44 @@
   border-bottom: 2px solid var(--table-row-border-color);
 }
 
+.TableHeader .sort-button {
+  background-color: transparent;
+  width: 100%;
+  /* a fixed height prevent TableHeader from being jumpy */
+  height: 1.625rem;
```
scurker, line 43: I see you added a fixed height to fix the header from jumping, but it seems like something else may be going on behind the scenes that may need more investigating (such as the table header briefly rendering without a button causing the height to be different). It would be nice to try to determine the root cause to see if we're not rendering something correctly.

scurker (changes requested):
Not necessarily a blocker, but the documentation seems a bit lacking for this component now:

We should try to document the various states of the table (sorted, not-sorted) to showcase different examples of usage. Since we are requiring author implemented code to maintain sorting the documentation doesn't provide a minimal example of how to implement this.

packages/styles/table.css
dequejosie: I know, somehow the button does not inherit the font-family here. It would not allow me to change font-weight until font-family is set here.

packages/styles/table.css
dequejosie: Turns out that sometimes there is a brief moment when no Icon is rendered yet. I set the span's height to the height of an Icon to prevent the brief moment changes the height of the button.

thuey: It looks as though the entire table is re-mounting (the HTML elements are being totally replaced). That shouldn't be necessary for all of the elements, especially the non-sorting headers. React should be able to tell that some of this doesn't need to change. There also shouldn't be a lag in rendering the Icon. I agree with @scurker that I'm not sure CSS is the right fix here. I'd be happy to jump on a call and look at this together with you.

scurker: I think you hit on something @thuey, it's probably re-rendering the icon when the sort value changes. I wonder if we should handle this differently here since the browser has to perform a layout calculation each time the svg renders - there's probably an optimization we can make here to ensure that doesn't happen. Have you tried wrapping the icon with `React.memo`?

dequejosie: Let's jump on a call tomorrow and discuss it.

dequejosie: We realized that the Demo component causes extra renders (usually for complex component). Therefore in the following changes we manually set up the demo.

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -1,21 +1,72 @@
...
-  </th>
-);
+const TableHeader = ({
+  children,
+  sortDirection,
+  onSort,
+  className,
+  ...other
+}: TableHeaderProps) => {
+  const sortButtonRef = useRef<HTMLButtonElement>(null);
```
thuey, line 23: I think if all of the re-rendering bugs are fixed, then you shouldn't have to manage focus manually. The reason why the focus was being lost was because the button was being removed and added back to the DOM.

As an aside, right now, I can get into this state by clicking on a sortable header and then clicking on a non-sortable header:
image

As an aside aside, is that the highlight color Aaron requested? It seems odd that the highlight color would be the same color as the background.

packages/styles/table.css
```diff
@@ -30,6 +37,43 @@
   border-bottom: 2px solid var(--table-row-border-color);
 }
 
+.TableHeader[aria-sort] {
+  padding: 0;
+}
+
+.TableHeader .TableHeader__sort-button {
```
thuey, line 44: Not to be too OCD about it, but would it be feasible to make the border a just be around the button instead of overlapping on the sides?
image

packages/react/src/components/Table/TableHeader.tsx
dequejosie: Good call for the focus! And yes, I double checked. That's the color for the highlight.
Screen Shot 2022-01-19 at 2 59 18 PM

packages/styles/table.css
dequejosie: It is overlapping on the sides because for the active column, we add a 2px border at the bottom and our universal offset for focus outline is 2px. I could manually modify the offset here.

dequejosie: It is supposed to be like this.
Screen Shot 2022-01-19 at 3 16 47 PM

dequejosie: It is supposed to be like this.
Screen Shot 2022-01-19 at 3 37 20 PM

thuey (approved):
LGTM. It'd be good to get one last review from @scurker, though

scurker:
Looks good!

Issue #513, opened 2022-01-17
Add an optional `labelDescription` prop to items in Radiogroup component 


akornmeier:
To accommodate a design comp for exporting data from axe Reports, the Radiogroup component needs to accept secondary text that describes an item. 

<img src="https://user-images.githubusercontent.com/1696398/149811155-49a779b5-5057-4ffd-b996-09aaa4279cef.png" width=50% height=50%>

Feature Spec: 

- Add optional prop `labelDescription` to RadioItem interface that will be displayed below the radio label ONLY when the `labelDescription` prop is defined in the RadioItem array
- When rendered, link `labelDescription` to the Radio label via aria-describedBy attribute
- Add related CSS styles for `Field__label-description`

akornmeier:
Additional context shown here with error and disabled states:  
<img width="400" alt="UX pin showing checkbox and radio options w/ description text " src="https://user-images.githubusercontent.com/1696398/150377423-fb86fceb-dee1-4801-b4bb-922d7bc5f8e3.png">

PR #514, opened 2022-01-17
feat(radiogroup): add labelDescription prop to radio items

akornmeier:
closes: #513

### DIFF
--- a/packages/react/src/components/RadioGroup/index.tsx
+++ b/packages/react/src/components/RadioGroup/index.tsx
@@ -111,17 +111,14 @@ const RadioGroup = ({
           onClick={() => onRadioClick(index)}
         />
         {labelDescription && (
-          <>
-            <div className="Field__break"></div>
-            <span
-              id={`${id}Desc`}
-              className={classNames('Field__labelDescription', {
-                'Field__labelDescription--disabled': disabled
-              })}
-            >
-              {labelDescription}
-            </span>
-          </>
+          <span
+            id={`${id}Desc`}
+            className={classNames('Field__labelDescription', {
+              'Field__labelDescription--disabled': disabled
+            })}
+          >
+            {labelDescription}
+          </span>
         )}
       </div>
     );
--- a/packages/react/src/components/RadioGroup/index.tsx
+++ b/packages/react/src/components/RadioGroup/index.tsx
@@ -6,6 +6,7 @@ import Icon from '../Icon';
 export interface RadioItem extends React.InputHTMLAttributes<HTMLInputElement> {
   label: string;
   value?: string;
+  labelDescription?: string;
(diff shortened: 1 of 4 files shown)


packages/react/src/components/RadioGroup/index.tsx
```diff
@@ -107,6 +110,19 @@ const RadioGroup = ({
           aria-hidden="true"
           onClick={() => onRadioClick(index)}
         />
+        {labelDescription && (
+          <>
+            <div className="Field__break"></div>
```
scurker, line 115: Why the extra element? Why not just apply `flex-basis: 100%` to the description?

akornmeier: i could try that, was having a heck of a time getting the wrap to work properly.

akornmeier: that works, pushing that change now

okry (approved):

packages/styles/forms.css
```diff
@@ -267,6 +270,16 @@ textarea.Field--has-error:focus:hover,
   border: 1px solid var(--field-border-color);
 }
 
+.Field__labelDescription {
+  flex-basis: 100%;
+  color: var(--field-label-description-text-color);
+  font-size: var(--text-size-small);
+  font-weight: var(--font-weight-normal);
+  line-height: 1;
+  margin: 0 0 0 30px;
```
scurker, line 279: Nitpick: This should probably match the computed width of the icon, unfortunately we don't have icon size anywhere as a variable so that is something that we should probably refactor in cauldron somewhere else. At a minimum, we could use calc:

```css
margin-right: calc(24px + 2px + var(--space-half));
```

* `24px` - icon size
* `2px` - borders (left and right)
* `var(--space-half)` matching the natural spacing between the input and the text

I don't recall why we put an invisible border around the input though 🤔

akornmeier: fair enough, i was looking for an icon size var and didn't see one. I will update to use the proposed value.

scurker (approved):

scurker:
Might want to pass this by @awpearlm as well if he's not already aware

packages/styles/forms.css
akornmeier: > Might want to pass this by @awpearlm as well if he's not already aware

@awpearlm - Had a CSS change for the desc text that @bobbyomari will be working on. Once he has a design, he will link it to the original issue #513 and I will make the update. Aaron also was interested in adding this prop to the checkbox element as well. I will create a new issue for that in GH.

akornmeier: Reviewed the changes with @awpearlm, he approves.

PR #530, opened 2022-01-24
fix: Make Panel heading optional

dequejosie:
Needed in https://github.com/dequelabs/walnut/issues/2665

### DIFF
--- a/docs/patterns/components/Panel/index.js
+++ b/docs/patterns/components/Panel/index.js
@@ -27,7 +27,7 @@ const PanelDemo = () => {
             children: 'Collapsed panel content.'
           },
           {
-            children: 'Content of Panel with a custom heading'
+            children: 'Content of Panel with no heading'
           }
         ]}
         propDocs={{
--- a/docs/patterns/components/Panel/index.js
+++ b/docs/patterns/components/Panel/index.js
@@ -27,16 +27,14 @@ const PanelDemo = () => {
             children: 'Collapsed panel content.'
           },
           {
-            'aria-labelledby': 'custom-heading',
-            children:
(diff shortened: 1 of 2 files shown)


scurker (changes requested):
Heading can't be optional, Panel uses a section meaning it's a region and it [must have a label](https://www.w3.org/TR/wai-aria-practices/examples/landmarks/region.html). 

https://github.com/dequelabs/cauldron/blob/78d6a7430df952abf22abcdfdcc6afea5e0bf42a/packages/react/src/components/Panel/index.tsx#L35-L41

For context, why is a heading not needed here?

dequejosie:
@scurker It is for the User Management Table, which does not have a Panel level heading. Can we have a label without having a visible heading for the Panel?

Screen Shot 2022-01-25 at 12 21 09 PM

scurker:
> @scurker It is for the User Management Table, which does not have a Panel level heading. Can we have a label without a visible heading for the Panel? Like when heading is not there, we require a label.

Is _Manage Users_ not a heading in this instance? We can't make this change alone to make the heading optional because that would introduce an A11Y issue where the section has an invalid `aria-labelledby`.

_If_ we want panel's heading to be optional, there's a couple of options

* Make `heading` optional _if_ an `aria-labelledby` attribute is passed into the component

_OR_

* Make `heading` optional and only include auto `aria-labelledby` when the `heading` is present

scurker (commented):
Just a few small nitpicks, and we need some tests to check for the conditional rendering of the heading.

It would also be nice to update docs.

scurker (approved):

PR #535, opened 2022-01-28
fix(styles): fix unnecessary border from lists inside of cards

scurker:
I have a pagination inside of a card pattern, but the list items there are showing a border on the bottom: 

pagination items with thin 1px line under the components

### DIFF
--- a/packages/styles/card.css
+++ b/packages/styles/card.css
@@ -75,10 +75,6 @@
   box-sizing: border-box;
 }
 
-.Card__content ul li {
-  border-bottom: 1px solid var(--list-separator);
-}
-
 .Card__footer {
   padding: var(--space-small);
   text-align: center;


packages/styles/card.css
```diff
@@ -75,10 +75,6 @@
   box-sizing: border-box;
 }
 
-.Card__content ul li {
```
dequejosie, line 78: Do we have any list inside .Card__content that actually requires the border? I don't recall any, but maybe we should check with @awpearlm

scurker: I asked @schne324 about this and he doesn't know, but regardless this is way too generic of a selector. If we intended to have some sort of list components inside of a card we need something more specific.

schne324 (approved):

PR #540, opened 2022-02-07
feat: add support for button-secondary to link

thuey:
Needed for https://github.com/dequelabs/walnut/issues/2696


scurker (approved):
Looks like we need to fix line-heights for links that appear as buttons, not blocking your PR but just noticed this.

thuey:
> Looks like we need to fix line-heights for links that appear as buttons, not blocking your PR but just noticed this.

Yeah, there seem to be a few things that are off. I created a separate ticket: https://github.com/dequelabs/cauldron/issues/541

Issue #538, opened 2022-02-02
Missing bottom margin for select lists, text inputs and text areas


awpearlm:
There should be ~24px~ 16px from the bottom of the input to the element below it.

Example: 
![image](https://user-images.githubusercontent.com/5720243/152193400-753f2c51-2b4c-4c01-9605-421fc06f4c01.png)

When error is present (additionally push the below elements down the distance of the error footprint ~15px <-- confirm this):
![image](https://user-images.githubusercontent.com/5720243/152194236-bfc346c4-06dd-4100-8ea7-0aa2fc5604a3.png)

dequejosie:
If I recall correctly, we should fix Checkbox as well.

awpearlm:
Small note: I spoke with Bobby and Josie. The space value should be 16px, not 24px. So Josie has it changed in the PR to be 16px and that is correct.

somaalapati:
@dequejosie I struggling to understand this. Which component of axe DevTools Pro I can test this change? Or this issue only with Caldron components?

dequejosie:
@somaalapati this ticket itself is for Cauldron, but once we update Cauldron version in Webapp and axe extension, we will see all the areas with TextFields and Select change. Actually I will spend some time tomorrow to update and take a look at the changes.

dequejosie:
@somaalapati For the impact of this ticket on Walnut, please check dev once this pr is merged: https://github.com/dequelabs/walnut/pull/3036, which is a part of https://github.com/dequelabs/walnut/issues/2961.

dequejosie:
@somaalapati you should be able to see the enlarged margins after this pr https://github.com/dequelabs/axe-extension/pull/2348 gets merged.

dequejosie:
This ticket should be ready for test both in Walnut and extension repo. @somaalapati @Keerthi-Penukonda @karthikeyan2716

karthikeyan2716:
**Retesting details:**
- This is working as expected 

**Screenshot:**

![30-03-2022-04-58-02](https://user-images.githubusercontent.com/79319620/160824333-5f0d4f3f-0c5e-4bb2-9993-1c6ede427f77.png)

![30-03-2022-04-57-19](https://user-images.githubusercontent.com/79319620/160824222-f30bf3a9-dbac-4cae-8488-2274c7685570.png)

[...]

PR #543, opened 2022-02-11
fix: Add missing margin to TextField and Select

dequejosie:
Ref: https://github.com/dequelabs/cauldron/issues/538

### DIFF
--- a/packages/styles/forms.css
+++ b/packages/styles/forms.css
@@ -156,7 +156,6 @@ textarea.Field--has-error:focus:hover,
 }
 
 .Error {
-  min-height: var(--space-large);
   color: var(--error);
   text-align: left;
   font-size: var(--text-size-smallest);
@@ -169,10 +168,6 @@ textarea.Field--has-error:focus:hover,
   position: relative;
   box-sizing: border-box;
   margin-bottom: var(--space-small);
-}
-
-.Field .Error {
-  min-height: unset;
 }
 
 .Field.is--flex-row {
@@ -195,10 +190,6 @@ textarea.Field--has-error:focus:hover,
 
 .Checkbox {
   margin-bottom: var(--space-smallest);
-}
-
-.Checkbox .Error {
-  min-height: unset;
 }
 
 .Checkbox + .Checkbox,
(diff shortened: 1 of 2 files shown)


packages/styles/forms.css
```diff
@@ -168,7 +168,11 @@ textarea.Field--has-error:focus:hover,
   display: block;
   position: relative;
   box-sizing: border-box;
-  margin-bottom: var(--space-smallest);
+  margin-bottom: var(--space-small);
```
dequejosie, line 171: Having discussed with Aaron, we decided to go with 16px

packages/styles/forms.css
```diff
@@ -168,7 +168,11 @@ textarea.Field--has-error:focus:hover,
...
+}
+
+.Field .Error {
+  min-height: unset;
```
dequejosie, line 175: Error has its min-height for 24px, which makes the gap between erroring TextField or Select too large.

packages/styles/forms.css
```diff
@@ -210,7 +218,7 @@ textarea.Field--has-error:focus:hover,
   text-align: left;
   color: var(--field-label-text-color);
   font-size: var(--text-size-small);
-  font-weight: var(--font-weight-bold);
+  font-weight: var(--font-weight-medium);
```
dequejosie, line 221: Not directly related to this ticket, but Bobby noticed the font-weight for Field__label was off.

packages/styles/forms.css
```diff
@@ -193,6 +197,10 @@ textarea.Field--has-error:focus:hover,
   margin-bottom: var(--space-smallest);
 }
 
+.Checkbox .Error {
+  min-height: unset;
```
dequejosie, line 201: Not directly related to this ticket, but the `Error` margin issue also applies to `Checkbox`.

packages/styles/forms.css
thuey: @schne324 @scurker Do you think we could just get rid of the min-height from the `.Error` class? Looking at walnut and the extension, we only ever conditionally render the element that has an error on it (e.g. `{error && <div className="Error">Something went wrong</div>}`

thuey (approved):

Issue #539, opened 2022-02-03
The position of Tooltip does not match the position of Icon Button after sorting


dequejosie:
Once the default order of the items changes, the position of Tooltip stops matching the position of Icon Button. In the screenshot below, I was hovering the icon on the third row, which used to be the first row before sorting, and the tooltip of the icon on the first row pops up. It is probably related to how Popper works. Additional information is, after resizing the viewport, this bizarre behavior disappears.

To reproduce this issue, please go to User Management on WebApp dev with multiple items. Sort, and then the issue will occur.

Given that sortable Table plus Icon Buttons appear in multiple places in our WebApp, we probably need to fix this soon.

<img width="1247" alt="Screen Shot 2022-02-03 at 5 44 02 PM" src="https://user-images.githubusercontent.com/83524297/152445624-c94c372d-ae0d-4bd7-958f-7b95f3486612.png">

somaalapati:
@karthikeyan2716 Please test this ticket.

karthikeyan2716:
**Retesting details:**

- It is working as expected
- The position of Tooltip matches the position of Icon Button after sorting

**Recording:**

https://user-images.githubusercontent.com/79319620/155658035-a7476093-2643-480d-91f9-4044192d1d01.mp4




**Tested Environment:**
**Walnut Version:** 4.41.1 build
**Chrome Browser:** Version 98.0.4758.102 (Official Build) (64-bit)
**Server:** https://axe.dequelabs.com/
**OS:** MS Windows 10

Issue #546, opened 2022-02-15
Tooltip show prop does not work as expected


thuey:
There are a couple of issues with the Tooltip show prop:

1. According to the documentation, `show` is supposed to manually control the visibility of the tooltip. However, it doesn't work because `show` is only used to initialize the state of whether or not the tooltip appears.
2. `show` is defaulted to `false` (see: https://github.com/dequelabs/cauldron/blob/16adb90060f3962eb6a7df4bce421c1dd8f056b3/packages/react/src/components/Tooltip/index.tsx#L48), so if we started treating `show` as a manual override, the current automatic hide/show code would stop working (because it would always be overridden by `show`)

Fixing these issues will be a breaking change

scurker:
After further evaluation, this should _not_ be a breaking change. We have the prop documented as following: 

> Manually control the show state of the tooltip.

With the original intention being that this was to change the component from uncontrolled to controlled when the `show` prop was set. I could not find any internal usages of the component in this way so we can move forward.

If the "initial show" behavior is desired, a new prop `defaultShow` will be added to accommodate this.

PR #547, opened 2022-02-15
fix: fix issue with tooltip not updating placement

thuey:
Ref: https://github.com/dequelabs/cauldron/issues/539

The issue was that `usePopper`'s `update` can be null to start with (see: https://popper.js.org/react-popper/v2/hook/#update-forceupdate-and-state). However, the `show` method was only being bound on the first render when `update` was still `null` (I believe this was introduced by https://github.com/dequelabs/cauldron/pull/505). So the tooltip's placement was not getting updated on `show`, which is a problem with table sorting because React shuffles elements around without popper knowing about it.

While I was in here, I also addressed some other issues and tried to simplify the code some. I have called these out as comments in this PR.

### DIFF
--- a/docs/patterns/components/Table/index.js
+++ b/docs/patterns/components/Table/index.js
@@ -7,7 +7,8 @@ import {
   TableHead,
   TableHeader,
   TableRow,
-  Code
+  Code,
+  IconButton
 } from '@deque/cauldron-react/';
 import { children, className } from '../../../props';
 
@@ -162,14 +163,24 @@ const BasicTable = () => (
             >
               Email
             </TableHeader>
+            <TableHeader scope="col">Actions</TableHeader>
           </TableRow>
         </TableHead>
         <TableBody>
           {sortedData.map(contact => (
-            <TableRow>
+            <TableRow key={contact.email}>
               <TableCell>{contact.first_name}</TableCell>
               <TableCell>{contact.last_name}</TableCell>
               <TableCell>{contact.email}</TableCell>
+              <TableCell>
+                <IconButton
+                  icon="trash"
+                  label="Delete"
+                  onClick={() => {
+                    console.log(`Delete ${contact.email}`);
+                  }}
+                />
+              </TableCell>
             </TableRow>
           ))}
         </TableBody>
--- a/docs/patterns/components/Tooltip/index.js
+++ b/docs/patterns/components/Tooltip/index.js
@@ -178,7 +178,7 @@ const Demo = () => {
           },
           show: {
             type: 'boolean',
-            description: 'Manually control the show state of the tooltip'
+            description: 'Whether or not the tooltip is shown initially.'
           },
           placement: {
             type: 'string',
--- a/packages/react/src/components/Tooltip/index.tsx
+++ b/packages/react/src/components/Tooltip/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect } from 'react';
+import React, { useState, useRef, useEffect, useCallback } from 'react';
 import classnames from 'classnames';
 import { createPortal } from 'react-dom';
 import PropTypes from 'prop-types';
@@ -44,15 +44,14 @@ export default function Tooltip({
   target,
   association = 'aria-describedby',
   variant = 'text',
-  show: showProp = false,
+  show: initialShow = false,
   hideElementOnHidden = false,
   className,
   ...props
 }: TooltipProps) {
   const [id] = propId ? [propId] : useId(1, 'tooltip');
-  const hovering = useRef(false);
-  const [placement, setPlacement] = useState(initialPlacement);
-  const [showTooltip, setShowTooltip] = useState(!!showProp);
+  const hideTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const [showTooltip, setShowTooltip] = useState(!!initialShow);
   const [targetElement, setTargetElement] = useState<HTMLElement | null>(null);
   const [tooltipElement, setTooltipElement] = useState<HTMLElement | null>(
     null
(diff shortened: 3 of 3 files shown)


docs/patterns/components/Table/index.js
```diff
@@ -162,14 +163,24 @@ const BasicTable = () => (
...
               <TableCell>{contact.email}</TableCell>
+              <TableCell>
+                <IconButton
+                  icon="trash"
+                  label="Delete"
+                  onClick={() => {
+                    console.log(`Delete ${contact.email}`);
+                  }}
+                />
+              </TableCell>
```
thuey, line 183: Initially, I just did this for testing, but I thought it was nice to leave in there.

docs/patterns/components/Tooltip/index.js
```diff
@@ -178,7 +178,7 @@ const Demo = () => {
           },
           show: {
             type: 'boolean',
-            description: 'Manually control the show state of the tooltip'
+            description: 'Whether or not the tooltip is shown initially.'
```
thuey, line 181: See: https://github.com/dequelabs/cauldron/issues/546

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -44,15 +44,14 @@ export default function Tooltip({
   target,
   association = 'aria-describedby',
   variant = 'text',
-  show: showProp = false,
+  show: initialShow = false,
```
thuey, line 47: See: https://github.com/dequelabs/cauldron/issues/546

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -44,15 +44,14 @@ export default function Tooltip({
...
   hideElementOnHidden = false,
   className,
   ...props
 }: TooltipProps) {
   const [id] = propId ? [propId] : useId(1, 'tooltip');
-  const hovering = useRef(false);
-  const [placement, setPlacement] = useState(initialPlacement);
```
thuey, line 54: We don't need to store this as state. Doing so was basically a no-op: https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/Tooltip/index.tsx#L115-L123

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -44,15 +44,14 @@ export default function Tooltip({
...
   association = 'aria-describedby',
   variant = 'text',
-  show: showProp = false,
+  show: initialShow = false,
   hideElementOnHidden = false,
   className,
   ...props
 }: TooltipProps) {
   const [id] = propId ? [propId] : useId(1, 'tooltip');
-  const hovering = useRef(false);
```
thuey, line 53: Instead of trying to maintain the state of whether or not we are focussed, I changed the implementation to a "deadman switch"--if you perform a "blurring" action, we will hide the tooltip unless you perform a "hovering" action within the timeout.

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -73,55 +72,51 @@ export default function Tooltip({
...
-    hovering.current = false;
-    hide(e);
-  };
-  const handleTipMouseEnter = () => {
-    hovering.current = true;
-  };
-  const handleTipMouseLeave = (e: MouseEvent) => {
-    hovering.current = false;
-    hide(e);
-  };
```
thuey, line 107: Replaced by the deadman switch

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -143,35 +138,35 @@ export default function Tooltip({
     return () => {
       targetElement.removeEventListener('keyup', handleEscape);
     };
-  }, [showProp]);
+  }, [showTooltip]);
```
thuey, line 141: This was a bug--`showProp` doesn't currently drive the visibility of the tooltip. As you can see on line 132, this effect is checking `showTooltip`

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -143,35 +138,35 @@ export default function Tooltip({
...
-      targetElement?.addEventListener('focusout', hide);
-    }
+    targetElement?.addEventListener('mouseenter', show);
+    targetElement?.addEventListener('mouseleave', hide);
+    targetElement?.addEventListener('focusin', show);
+    targetElement?.addEventListener('focusout', hide);
+
     return () => {
       targetElement?.removeEventListener('mouseenter', show);
       targetElement?.removeEventListener('mouseleave', hide);
```
thuey, line 152: We actually had a mismatch here with "adding" event listeners for `handleTriggerMouseEnter` and `handleTriggerMouseLeave` but then removing `hide` and `show`. Hopefully, now that there is only `hide` and `show`, it will make it easier to do the right thing.

I considered trying to couple the add and remove event listeners more, but I struggled to come up with something that wouldn't just suffer from other short comings (plus, this PR has enough changes already :) )

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -143,35 +138,35 @@ export default function Tooltip({
...
       targetElement?.removeEventListener('focusin', show);
       targetElement?.removeEventListener('focusout', hide);
     };
-  }, [targetElement, showProp]);
+  }, [targetElement, show, hide]);
```
thuey, line 156: We need to update our listeners when there are changes to the `show` and `hide` callbacks.

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -143,35 +138,35 @@ export default function Tooltip({
     return () => {
       targetElement.removeEventListener('keyup', handleEscape);
     };
-  }, [showProp]);
+  }, [showTooltip]);
 
+  // Handle hover and focus events for the targetElement
   useEffect(() => {
-    if (typeof showProp !== undefined) {
```
thuey, line 149: `showProp` is always defined: https://github.com/dequelabs/cauldron/issues/546

docs/patterns/components/Tooltip/index.js
scurker: I'm debating on if this should be a breaking change or not - the description implies that it changes the functionality of the prop. If so, I _think_ this PR should have _BREAKING CHANGE_ because we're changing the original intended functionality. I don't think we use it like this anywhere internally, but it's possible someone else was.

scurker: Right, but if someone is using this component based on the current expected documented functionality, it still is a breaking change because we're changing the documented functionality.

scurker: So we should either:

* Fix the issue to match what is currently documented
* Or keep the code as is and mark this change as a breaking change

thuey: I reverted the description and added a note that there is a known issue.

scurker (approved):

Issue #551, opened 2022-02-28
Sortable Columns of Table are missing hover state


dequejosie:
Expected:

<img width="355" alt="Screen Shot 2022-02-28 at 3 52 48 PM" src="https://user-images.githubusercontent.com/83524297/156056857-1ccd8b32-65b6-48d8-8e2c-7f21ed6d4c70.png">

dequejosie:
Please test this with walnut User Management's table. Hover the sortable columns (with triangle icons) and see if the color changes.

somaalapati:
working as expected

https://user-images.githubusercontent.com/40261789/157207572-f1dbb42c-3f9f-4cbd-8154-7372a8ab7df2.mp4


"version":"4.42.1"

Issue #550, opened 2022-02-28
Confirmation toast has the wrong icon


dequejosie:
Expected:
<img width="586" alt="Screen Shot 2022-02-28 at 3 50 19 PM" src="https://user-images.githubusercontent.com/83524297/156056591-580da6f0-7199-4434-8d66-40823bcbf911.png">

Actual:
<img width="230" alt="Screen Shot 2022-02-28 at 3 49 26 PM" src="https://user-images.githubusercontent.com/83524297/156056444-12288c5f-5ecc-40fd-b2ff-cddd9d05f7d1.png">

dequejosie:
Please test this in Walnut user management. This confirmation toast should appear whenever an action (add user, edit user, resend invitation, and remove user) is done successfully.

PR #554, opened 2022-02-28
fix: Add hover state to Table and correct Toast icon

dequejosie:
Ref: https://github.com/dequelabs/cauldron/issues/550; https://github.com/dequelabs/cauldron/issues/551
Part of https://github.com/dequelabs/walnut/issues/2961

### DIFF
--- a/packages/react/src/components/DescriptionList/index.tsx
+++ b/packages/react/src/components/DescriptionList/index.tsx
@@ -67,7 +67,7 @@ export const DescriptionDetails = ({
   ...other
 }: DescriptionProps) => (
   <dd className={classNames('DescriptionList__details', className)} {...other}>
-    {children}
+    <div>{children}</div>
   </dd>
 );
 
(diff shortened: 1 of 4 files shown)


thuey (dismissed):
Adding that div shouldn't be necessary and might throw off some styling people have.

packages/react/src/components/DescriptionList/index.tsx
```diff
@@ -67,7 +67,7 @@ export const DescriptionDetails = ({
   ...other
 }: DescriptionProps) => (
   <dd className={classNames('DescriptionList__details', className)} {...other}>
-    {children}
+    <div>{children}</div>
```
schne324, line 70: why this div?

dequejosie:
Yes, it should not be there. It was for another issue and I didn't clean it up.

thuey (approved):

PR #562, opened 2022-03-21
Fix/move focus to loader

tbusillo:
ref: https://github.com/dequelabs/walnut/issues/3039

Additional comments at above ref as well. 

## Purpose

To address moving focus to the loader, which was less of an issue with the component itself and more of an issue with its implementation in and need for handling focus outside of the component (e.g., via forwardRef).  

This solution addresses one part of the problem, with the second part needing to be addressed where the loader is being used without implementing some kind of focus (this PR will make it easier to fix that, however). 

See comments in https://github.com/dequelabs/walnut/issues/3039 and the diffs for additional information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,12 @@ See [`CONTRIBUTING.md`](./CONTRIBUTING.md)
 $ yarn
 ```
 
+### build (if first time building dev environment, must be run before yarn dev)
+
+```sh
+$ yarn build
+```
+
 ### dev
 
 (watches/rebuilds react, styles, and docs)
@@ -30,12 +36,6 @@ $ yarn dev
 ```
 
 navigate browser to http://localhost:8000
-
-### build
-
-```sh
-$ yarn build
-```
 
 ### run tests
 
--- a/docs/patterns/components/LoaderOverlay/index.js
+++ b/docs/patterns/components/LoaderOverlay/index.js
@@ -29,9 +29,11 @@ const LoaderOverlayDemo = () => {
       return;
     }
 
+    /*
     if (loading) {
       return loaderRef.current.focus();
     }
+    */
 
     buttonRef.current.focus();
   }, [loading]);
@@ -62,6 +64,7 @@ const LoaderOverlayDemo = () => {
               ref={loaderRef}
               label="Loading..."
               variant="large"
+              focus
             >
               <p>
                 Explanatory secondary text goes here. Let them know what's
--- a/docs/patterns/components/LoaderOverlay/index.js
+++ b/docs/patterns/components/LoaderOverlay/index.js
@@ -50,8 +50,10 @@ const LoaderOverlayDemo = () => {
           type: 'boolean',
           description: 'whether or not to focus the loader on initial render'
         },
-        ref: {
-          type: 'ref'
+        loaderRef: {
+          type: 'function',
+          description:
+            'DEPRECATED - replace with "focus" prop - optional ref function to move focus to loader on mount'
         }
       }}
     >
--- a/docs/patterns/components/LoaderOverlay/index.js
+++ b/docs/patterns/components/LoaderOverlay/index.js
@@ -49,6 +49,9 @@ const LoaderOverlayDemo = () => {
         focus: {
           type: 'boolean',
           description: 'whether or not to focus the loader on initial render'
+        },
+        ref: {
+          type: 'ref'
         }
       }}
     >
--- a/docs/patterns/components/LoaderOverlay/index.js
+++ b/docs/patterns/components/LoaderOverlay/index.js
@@ -28,7 +28,6 @@ const LoaderOverlayDemo = () => {
       mounted.current = true;
       return;
     }
-
     buttonRef.current.focus();
   }, [loading]);
 
--- a/docs/patterns/components/LoaderOverlay/index.js
+++ b/docs/patterns/components/LoaderOverlay/index.js
@@ -14,6 +14,7 @@ const LOADING_DURATION = 5000;
 const LoaderOverlayDemo = () => {
   const mounted = useRef(false);
   const buttonRef = useRef();
+  const loaderRef = useRef();
   const [loading, setLoading] = useState(false);
   const onClick = () => {
     setLoading(true);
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -21,6 +21,12 @@ See [`CONTRIBUTING.md`](./CONTRIBUTING.md)
 $ yarn
 ```
 
+### build (if first time building dev environment, must be run before yarn dev)
+
+```sh
+$ yarn build
+```
+
```
tbusillo, line 29: This is a bit unrelated to the overlay issue, but added to the README that upon first run, must run `yarn build` prior to `yarn dev`.

docs/patterns/components/LoaderOverlay/index.js
```diff
@@ -50,6 +45,15 @@ const LoaderOverlayDemo = () => {
         variant: {
           type: 'string',
           description: 'Loader variant, can be "small" or "large".'
+        },
+        focus: {
+          type: 'boolean',
+          description: 'whether or not to focus the loader on initial render'
+        },
```
tbusillo, line 52: Updated the documentation/demo to reflect the new usage of the `focus` prop, which is similar to how the `Toast` component currently works.

packages/react/__tests__/src/components/LoaderOverlay/index.js
```diff
@@ -37,6 +37,44 @@ test('handles variants', () => {
...
+    <LoaderOverlay className="baz" role="alert" label="loading" ref={loaderRef}>
+      Some text
+    </LoaderOverlay>
+  );
+
+  setTimeout(() => {
+    expect(document.activeElement).toStrictEqual(loaderOverlay.getDOMNode());
+  });
+});
+
```
tbusillo, line 77: Test cases above this test for implementation of the new `focus` prop, but also added in a test to ensure backwards compatibility. As it stands, if we were to remove backwards compatibility for use of `forwardRef`, it would be a breaking change for current implementations where that prop is used.

packages/react/src/components/LoaderOverlay/index.tsx
```diff
@@ -7,37 +7,65 @@ import AxeLoader from './axe-loader';
 interface LoaderOverlayProps extends React.HTMLAttributes<HTMLDivElement> {
   variant?: 'large' | 'small';
   label?: string;
+  focus?: boolean;
+  children?: React.ReactNode;
```
tbusillo, line 11: Added props for focus and children. Typescript does not seem to like leaving `children` as usable under `...others` without it having a type definition.

packages/react/src/components/LoaderOverlay/index.tsx
```diff
@@ -7,37 +7,65 @@ import AxeLoader from './axe-loader';
...
+        className={classNames(
+          'Loader__overlay',
+          className,
+          variant === 'large'
+            ? 'Loader__overlay--large'
+            : variant === 'small'
+            ? 'Loader__overlay--small'
+            : ''
+        )}
+        ref={ref ? ref : overlayRef}
```
tbusillo, line 48: For backwards compatibility purposes, checks if a ref was passed via prop, and if so, uses that ref instead of the overlayRef.

docs/patterns/components/LoaderOverlay/index.js
```diff
@@ -50,6 +45,15 @@ const LoaderOverlayDemo = () => {
...
+        loaderRef: {
+          type: 'function',
+          description:
+            'DEPRECATED - replace with "focus" prop - optional ref function to move focus to loader on mount'
```
tbusillo, line 56: While backwards compatibility will be supported with this PR, still think it should be encouraged to make use of the `focus` prop instead.

schne324: A couple of thoughts here...
- needing an handle on the loader (`loaderRef`) goes beyond just `focus`
- There are scenarios, like loading state on page load, where one might not want to focus the loader

I am just bringing this up to ensure we are cautious with how we document this

tbusillo: > While backwards compatibility will be supported with this PR, still think it should be encouraged to make use of the `focus` prop instead.

That's a good point - think just leave loaderRef as a more ambiguous statement for now, then?

packages/react/src/components/LoaderOverlay/index.tsx
```diff
@@ -7,37 +7,65 @@ import AxeLoader from './axe-loader';
...
+      focus,
+      ...other
+    }: LoaderOverlayProps,
+    ref
+  ) => {
+    const overlayRef = createRef<HTMLDivElement>();
+
+    useEffect(() => {
+      if ((!!focus && overlayRef) || ref) {
+        setTimeout(() => {
```
tbusillo, line 30: Trying to move focus prior to the next tick was causing issues where the styling for the focus indicator would not be applied, even thought he element was the active element.

packages/react/src/components/LoaderOverlay/index.tsx
```diff
@@ -7,37 +7,65 @@ import AxeLoader from './axe-loader';
...
-      <div className="Loader__overlay__loader">
-        <Loader variant={variant} />
-        <AxeLoader />
+  (
+    {
+      className,
+      variant,
+      label,
+      children,
+      focus,
```
scurker, line 21: This prop is a bit ambiguous as it doesn't describe what the prop does. I would like to think we could give this a name that's a little more self documenting, like `focusOnInitialRender`.

packages/react/src/components/LoaderOverlay/index.tsx
```diff
@@ -7,37 +7,65 @@ import AxeLoader from './axe-loader';
...
+      ...other
+    }: LoaderOverlayProps,
+    ref
+  ) => {
+    const overlayRef = createRef<HTMLDivElement>();
+
+    useEffect(() => {
+      if ((!!focus && overlayRef) || ref) {
```
scurker, line 29: This looks like this will always focus if there's a user provided ref

scurker (changes requested):

scurker (changes requested):
Just a few minor nitpicks then I think we're good to go!

scurker:
The github status for security reviews isn't available so 

🚨  🚨  REVIEWED FOR SECURITY 🚨  🚨  

Small nitpick: You may want to update the PR title to accurately reflect the changes we made

tbusillo:
> The github status for security reviews isn't available so
> 
> 🚨 🚨 REVIEWED FOR SECURITY 🚨 🚨
> 
> Small nitpick: You may want to update the PR title to accurately reflect the changes we made

Will do and thank you again for the help!

PR #601, opened 2022-03-25
fix(styles): allow button text to wrap to prevent text from being cut off

### LINKED ISSUE #567: Docs site issue: `<Button />` text is not properly visible at 320px width
This should probably also be fixed in the component itself...Hard-coding the height is what is causing this issue:

image

<dl><dt>Description</dt><dd>Content or functionality is lost due to overlapping content when page width is adjusted to 320px.
'I am a link that looks like a button'  button text is not properly visible at 400% zoom setting on browser. </dd><dt>Impact</dt><dd>Moderate</dd><dt>Recommendation</dt><dd>RULE : 

HOW TO FIX: 

REFERENCE: 
WCAG Understanding document:

BACKGROUND: 
People with low vision often read content on a desktop computer by scaling content using the browser zoom - up to 400%. When zooming causes content to overlap other content, users with low vision may not be able to see or use that content. When content is created so that it reflows without losing information or functionality at screen widths down to 320px, users can view a web page in a 1280px width browser screen and zoom the content to 400% (320px) without having to scroll side to side to read and without losing access to content.</dd><dt>Source code</dt><dd><pre><code><a class="Button--primary" href="#">I am a link that looks like a button</a></code></pre></dd><dt>Checkpoint</dt><dd>Reflow (1.4.10.a)</dd><dt>URL</dt><dd>https://cauldron.dequelabs.com/components/Link</dd></dl>

scurker:
closes #567

### DIFF
--- a/packages/styles/button.css
+++ b/packages/styles/button.css
@@ -32,7 +32,7 @@
   position: relative;
   text-align: center;
   text-transform: uppercase;
-  height: 36px;
+  min-height: 36px;
   min-width: 100px;
   display: inline-grid;
   grid-auto-flow: column;


isner (approved):
👍

Issue #455, opened 2021-12-14
Migrate over the multi-link Skiplink pattern from old Cauldron over to new repo


awpearlm:
this one:
![image](https://user-images.githubusercontent.com/5720243/146052642-ae7aeb07-7e3c-4b69-b6f8-78d75fbef0ce.png)

awpearlm:
Related #332

PR #607, opened 2022-03-30
fix(skip-link): fix skip-link styles with new design

straker:
Closes issue: #455

The whitespace between the spans isn't needed as the `display: block` causes a whitespace.

VoiceOver output of link reading 'Skip to Main Content'

Chrome Accessibility Tree output of link reading 'Skip to Main Content'

### DIFF
--- a/packages/react/src/components/SkipLink/index.tsx
+++ b/packages/react/src/components/SkipLink/index.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Icon from '../Icon';
+import { isBrowser } from '../../utils/is-browser';
 
 export interface SkipLinkProps {
   target: string;
@@ -77,6 +78,9 @@ export default class SkipLink extends React.Component<
   }
 
   private onClick() {
+    if (!isBrowser()) {
+      return;
+    }
     const element = document.querySelector(this.props.target) as HTMLElement;
 
     if (element) {
--- a/packages/react/src/components/SkipLink/index.tsx
+++ b/packages/react/src/components/SkipLink/index.tsx
@@ -69,7 +69,7 @@ export default class SkipLink extends React.Component<
         >
           <span className="SkipLink__item--first">{skipText}</span>
           <span className="SkipLink__item--second">
-            <Icon type="triangle-right" aria-hidden="true" />
+            <Icon type="triangle-right" />
             {targetText}
           </span>
         </a>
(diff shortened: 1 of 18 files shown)


packages/react/src/components/SkipLink/index.tsx
```diff
@@ -66,8 +67,11 @@ export default class SkipLink extends React.Component<
           onFocus={this.onFocus}
           onBlur={this.onBlur}
         >
-          <span className="SkipLink__item--first">{skipText}</span>&nbsp;
-          <span className="SkipLink__item--second">{targetText}</span>
+          <span className="SkipLink__item--first">{skipText}</span>
+          <span className="SkipLink__item--second">
+            <Icon type="triangle-right" aria-hidden="true" />
```
scurker, line 72:
```suggestion
            <Icon type="triangle-right" />
```

`aria-hidden` gets added automatically because `<Icon />` does not have a label: https://github.com/dequelabs/cauldron/blob/cea3659486f910cae66d2ab75e2e90616541b14d/packages/react/src/components/Icon/index.tsx#L47

packages/styles/skip-link.css
```diff
@@ -1,19 +1,17 @@
...
 }
 
 .SkipLink {
-  min-width: 173px;
-  height: 58px;
-  background-color: var(--skip-link-background-color);
   position: absolute;
   box-sizing: border-box;
-  padding: var(--space-half);
+  padding: 8px;
```
scurker, line 10: We use [variables](https://github.com/dequelabs/cauldron/blob/develop/packages/styles/variables.css) for a majority of our padding in cauldron, this should probably be the following:

```suggestion
  padding: var(--space-smallest);
```

packages/styles/skip-link.css
```diff
@@ -27,10 +25,13 @@
 }
 
 .SkipLink a {
+  min-width: 208px;
```
scurker, line 28: Nitpick: If we want to enforce a min-width, we should probably use rems here.

`208px` === `13rem`

packages/styles/skip-link.css
```diff
@@ -1,19 +1,17 @@
 :root {
-  --skip-link-background-color: rgba(11, 14, 17, 0.9);
+  --skip-link-background-color: rgba(255, 255, 255, 0.95);
```
scurker, line 2: Did you check with @awpearlm on this? The UX isn't clear, the notes say it's 90% opacity but the actual spec shows 95%.

scurker (changes requested):
A few small things

packages/styles/skip-link.css
scurker: I don't know if anything needs to change where, just want to validate that 95% is correct.

straker: I was just going off the uxpin notes.

scurker (approved):

PR #618, opened 2022-04-04
fix(styles): allow buttons to wrap

thuey:
Adds an additional fix for: https://github.com/dequelabs/cauldron/issues/567

Allows text in buttons to wrap to prevent content from being cut off at 320px


scurker (approved):
From screenshot tests done elsewhere, I think this change doesn't appear to be breaking anything major.

PR #622, opened 2022-04-06
feat(react): announce changes to sort on sortable TableHeaders

scurker:
Changes to the `aria-sort` attribute on table column cells is not yet [widely supported](https://a11ysupport.io/tech/aria/aria-sort_attribute) and is only currently announced in Firefox/JAWS.

This does mean that anyone using Firefox/JAWS will get a _very verbose_ announcement for sorting changes - but every other SR combination will at least have a polite announcement of the sort change.

### DIFF
--- a/packages/react/src/components/Table/TableHeader.tsx
+++ b/packages/react/src/components/Table/TableHeader.tsx
@@ -13,7 +13,7 @@ interface TableHeaderProps extends ThHTMLAttributes<HTMLTableCellElement> {
   onSort?: () => void;
   className?: string;
   sortAscendingAnnouncement?: string;
-  sortDescendingAnnoucnement?: string;
+  sortDescendingAnnouncemen?: string;
 }
 
 const TableHeader = ({
@@ -22,7 +22,7 @@ const TableHeader = ({
   onSort,
   className,
   sortAscendingAnnouncement = 'sorted ascending',
-  sortDescendingAnnoucnement = 'sorted descending',
+  sortDescendingAnnouncemen = 'sorted descending',
   ...other
 }: TableHeaderProps) => {
   // When the sort direction changes, we want to announce the change in a live region
@@ -34,7 +34,7 @@ const TableHeader = ({
       setAnnouncement(
         sortDirection === 'ascending'
           ? sortAscendingAnnouncement
-          : sortDescendingAnnoucnement
+          : sortDescendingAnnouncemen
       );
     }
   }, [sortDirection]);
--- a/packages/react/src/components/Table/TableHeader.tsx
+++ b/packages/react/src/components/Table/TableHeader.tsx
@@ -1,6 +1,7 @@
-import React, { ThHTMLAttributes } from 'react';
+import React, { ThHTMLAttributes, useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import Offscreen from '../Offscreen';
 import Icon from '../Icon';
 
 // these match aria-sort's values
@@ -11,6 +12,8 @@ interface TableHeaderProps extends ThHTMLAttributes<HTMLTableCellElement> {
   sortDirection?: SortDirection;
   onSort?: () => void;
   className?: string;
+  sortAscendingAnnouncement?: string;
+  sortDescendingAnnoucnement?: string;
 }
 
 const TableHeader = ({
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -11,45 +12,69 @@ interface TableHeaderProps extends ThHTMLAttributes<HTMLTableCellElement> {
...
+            {sortDirection === 'none' ? (
+              <Icon type="sort-triangle" />
+            ) : sortDirection === 'ascending' ? (
+              <Icon type="triangle-up" />
+            ) : (
+              <Icon type="triangle-down" />
+            )}
+          </span>
+          <Offscreen>
+            <span role="alert" aria-live="polite">
```
schne324, line 67: given our convo about the NVDA quirks that this patch addresses, do you think `role=status` would be better than `role=alert`?

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -11,45 +12,69 @@ interface TableHeaderProps extends ThHTMLAttributes<HTMLTableCellElement> {
...
-      </button>
-    ) : (
-      children
-    )}
-  </th>
-);
+}: TableHeaderProps) => {
+  // When the sort direction changes, we want to announce the change in a live region
+  // because changes to the sort value is not widely supported yet
+  // see: https://a11ysupport.io/tech/aria/aria-sort_attribute
```
schne324, line 30: good call on adding this comment...our future selves thank you!

packages/react/src/components/Table/TableHeader.tsx
schne324: both inherit `aria-atomic=true` which is good

schne324 (dismissed):

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -11,45 +12,69 @@ interface TableHeaderProps extends ThHTMLAttributes<HTMLTableCellElement> {
...
+          {children}
+          <span aria-hidden="true">
+            {sortDirection === 'none' ? (
+              <Icon type="sort-triangle" />
+            ) : sortDirection === 'ascending' ? (
+              <Icon type="triangle-up" />
+            ) : (
+              <Icon type="triangle-down" />
+            )}
+          </span>
```
thuey, line 65: For my own edification, would another way to solve this be to make this a live region with labels? Or would that wind up being too verbose when reading through it normally (do screen readers already say what the sort direction is?)

packages/react/src/components/Table/TableHeader.tsx
scurker: Ah, yeah `role="status"` is probably a better attribute

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -11,45 +12,69 @@ interface TableHeaderProps extends ThHTMLAttributes<HTMLTableCellElement> {
...
+  const [announcement, setAnnouncement] = useState('');
+  useEffect(() => {
+    if (sortDirection !== 'none') {
+      setAnnouncement(
+        sortDirection === 'ascending'
+          ? sortAscendingAnnouncement
+          : sortDescendingAnnoucnement
+      );
+    }
+  }, [sortDirection]);
```
thuey, line 40: I don't think there's any value here in storing this as state:

```suggestion
 const announcement = sortDirection === 'ascending' ? sortAscendingAnnouncement : sortDirection === 'descending' ? sortDescendingAnnouncement : ''; 
```

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -11,45 +12,69 @@ interface TableHeaderProps extends ThHTMLAttributes<HTMLTableCellElement> {
   sortDirection?: SortDirection;
   onSort?: () => void;
   className?: string;
+  sortAscendingAnnouncement?: string;
+  sortDescendingAnnoucnement?: string;
```
thuey, line 16: There's a typo here in `sortDescendingAnnoucnement`

thuey (dismissed):
Only blocker here is the typo

packages/react/src/components/Table/TableHeader.tsx
scurker: It might be too verbose.

When I was testing in JAWS, the behavior was something like:

> Table Column 4 Sorted Ascending, Email, Button

Unfortunately there's a bit of redundancy here, but we probably don't want want the current sort state to be part of the calculated accessible name of the button.

packages/react/src/components/Table/TableHeader.tsx
scurker: There is, because we don't want to announce the _initial_ state of the sort, only updates.

packages/react/src/components/Table/TableHeader.tsx
scurker: spelling is hard

packages/react/src/components/Table/TableHeader.tsx
thuey: Actually, I guess it shouldn't be any more verbose than the other live region. So if we changed this to something like:
```
<span aria-role="status">
  {sortDirection === 'none' ? (
    <Icon type="sort-triangle" />
  ) : sortDirection === 'ascending' ? (
    <Icon type="triangle-up"  aria-label={{sortAscendingAnnouncement}} />
  ) : (
    <Icon type="triangle-down" aria-label={{sortDescendingAnnouncement}} />
  )}
</span>
```

Could we do away with the `announcement` variable?

thuey: I replied before seeing your reply.

scurker: There's also the issue with the above being `aria-hidden` so the labels would not be visible to AT users anyway

packages/react/src/components/Table/TableHeader.tsx
thuey: `useEffect` runs on the initial render, so if the sort direction is something other than `none`, so wouldn't it suffer from that same problem?

packages/react/src/components/Table/TableHeader.tsx
thuey: In my example, I removed the `aria-hidden`. I don't think it would be any more verbose than what is currently here and might simplify the code some, but it's not that big a deal.

thuey (approved):

packages/react/src/components/Table/TableHeader.tsx
thuey: Actually, I don't _think_ it would read it either way. In any case, I think the behavior would be the same without the `useEffect` and useState`

Issue #629, opened 2022-04-20
[Pagination]Add method for `goToPage` or `onPageChange`.


okry:
The goal of adding this method would be to utilize some of the logic that is already in the Pagination component to reduce some of the work needed to implement the four `on{{First|Previous|Next|Last}}PageClick` callbacks and the math needed to pass to state logic.

This makes sense if you think of pagination in it's simplest form with only three parameters to handle a page's state:
- pageSize
- currentPage
- totalCount

And `totalCount` and `pageSize` are not currently handled by Cauldron which only leaves `currentPage`.

With a `goToPage` you replace implementations for the four "click" callbacks mentioned above with:
- firstPageClick === goToPage(0)
- previousPageClick === goToPage(currentPage - 1)
- nextPageClick === goToPage(currentPage + 1)
- lastPageClick === goToPage(ceil(totalCount / pageSize) - 1)

Doing this we can not only replace four methods with one, reducing the prop complexity of the component, but also reduce the amount of work that the user would need to do to manage their data.

okry:
Addressed with pagination hook feature.

PR #632, opened 2022-04-21
chore(pagination): use goToPage instead of four different functions

okry:
closes: #629

- Replaces the four `onPageChange` functions with `goToPage` to remove redundancy
- Makes props that are necessary for the component to function "required"
- Adds types for `sinon`
- Uses zero-indexed paging
- Cleans up paging logic

### DIFF
--- a/packages/react/src/components/Pagination/index.tsx
+++ b/packages/react/src/components/Pagination/index.tsx
@@ -9,16 +9,13 @@ import Icon from '../Icon';
 interface Props extends React.HTMLAttributes<HTMLDivElement> {
   totalItems: number;
   itemsPerPage?: number;
-  currentPage?: number;
+  currentPage: number;
+  goToPage: (page: number) => void;
   statusLabel?: React.ReactNode;
   firstPageLabel?: string;
   previousPageLabel?: string;
   nextPageLabel?: string;
   lastPageLabel?: string;
-  onNextPageClick?: () => void;
-  onPreviousPageClick?: () => void;
-  onFirstPageClick?: () => void;
-  onLastPageClick?: () => void;
   tooltipPlacement?: Placement;
   className?: string;
 }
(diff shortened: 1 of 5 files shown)


schne324:
> Replaces the four onPageChange functions with goToPage to remove redundancy

That sounds like a `BREAKING CHANGE`...breaking changes need to be declared in the commit message body (or description of the PR here)

schne324:
Why remove the existing 4 props? I think we can have it both ways

okry:
> Why remove the existing 4 props? I think we can have it both ways

Yeah, I can revert them, it just helped to show the reduction of code used with the `goToPage` function instead of the 4.

okry:
Although, looking at it some more, It would make the component easier to understand if we just do away with the four props since we're already making breaking changes to how it bases the page index.

With them gone, there's less room for confusion, easier maintainability, and less chance of some weird error happening if someone mixes the two conventions.

scurker (changes requested):
I don't think breaking changes should be taken lightly, and we need to have good justifications for breaking changes.

I do have several issues with this PR:

> Replaces the four onPageChange functions with goToPage to remove redundancy

This doesn't reduce redundancy. Each one of those functions has a specific purpose and allows for the author to decide how they want to implement the function. What if the component is rendered server side and the author needs to perform a redirect instead? What if the author needs to fetch the next set of results? What if the results are cached and the author needs to progressively fetch data? There's a number of use cases where by removing the implicit handling of functions that limits the flexibility of someone using this component.

In theory you could provide the same with a `onPageChange` function, but if the author doesn't control the page changes then they have to potentially determine the direction or numeration of the page change. This is a heavy ask when having `onNext/onPrev/onFirst/onLast` removes this ambiguity and allows for the author to implicitly handle changes if they choose to do so.

> Uses zero-indexed paging

I'm not sure what benefit this provides. We've already established pagination based on the actual page numbers, so this just seems like an unnecessary breaking change. 

If your concern is you can't just "use" the pagination component without implementing the high level functions yourself, I think there's ways to implement that without there being a breaking change. 

I've seen this pattern used with [`react-aria`](https://react-spectrum.adobe.com/react-aria/) in which `hooks` are provided that implement the high level functionality necessary. We could do something similar here:

```jsx
import { Pagination, usePagination } from '@deque/cauldron-react'

function PaginationComponent() {
  const { paginationProps } = usePagination()
  return <Pagination {...paginationProps} />
}
```

The `usePagination` hook could provide the `currentPage` and `onNextPageClick`, `onPreviousPageClick`, `onFirstPageClick`, and `onLastPageClick` functions.

This still allows for the flexibility of handling each function separately if needed, while providing a utility for authors who do not need that fine tuned control. This would also avoid unnecessarily breaking backwards compatibility and requiring existing implementors to rewrite their implementations.

okry:
@schne324 @scurker Can I get another review? I'll probably hold off on a hook implementation till we have a clear idea of what we want to do with it.

packages/react/src/components/Pagination/index.tsx
```diff
@@ -9,16 +9,17 @@ import Icon from '../Icon';
 interface Props extends React.HTMLAttributes<HTMLDivElement> {
   totalItems: number;
   itemsPerPage?: number;
-  currentPage?: number;
+  currentPage: number;
+  goToPage?: (page: number) => void;
```
scurker, line 13: If `currentPage` is author controlled, what benefit does `goToPage` provide?

packages/react/src/components/Pagination/index.tsx
```diff
@@ -9,16 +9,17 @@ import Icon from '../Icon';
 interface Props extends React.HTMLAttributes<HTMLDivElement> {
   totalItems: number;
   itemsPerPage?: number;
-  currentPage?: number;
+  currentPage: number;
```
scurker, line 12: This is a breaking change.

scurker (changes requested):
I'm not still entirely clear what this PR accomplishes. The PR title should also be `feat` instead of `chore` as it changes functionality for a component.

Issue #633, opened 2022-04-22
[PanelTrigger]: Add ability to override icons


longmatthewh:
I'd like to be able to use various icons with `PanelTrigger` - plus/minus, triangle-right/triangle-down - right now I'm limited to the chevron icons.

PR #634, opened 2022-04-22
feat(PanelTrigger): add icon override

longmatthewh:
Adds 2 new optional props:
- `expandedIcon`
- `collapsedIcon`

This enables devs to use alternatives to the `chevron-right` and `chevron-down` icons.

Closes: #633

### DIFF
--- a/packages/react/src/components/ExpandCollapsePanel/PanelTrigger.tsx
+++ b/packages/react/src/components/ExpandCollapsePanel/PanelTrigger.tsx
@@ -9,8 +9,8 @@ export interface PanelTriggerProps
   open?: boolean;
   fullWidth?: string;
   onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
-  expandedIcon?: IconType;
-  collapsedIcon?: IconType;
+  iconExpanded?: IconType;
+  iconCollapsed?: IconType;
 }
 
 function PanelTrigger({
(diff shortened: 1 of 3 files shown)


packages/react/src/components/ExpandCollapsePanel/PanelTrigger.tsx
```diff
@@ -17,6 +19,8 @@ function PanelTrigger({
   open,
   fullWidth,
   onClick,
+  expandedIcon = 'chevron-down',
+  collapsedIcon = 'chevron-right',
```
scurker, line 23: Very minor nitpick: `iconExpanded` and `iconCollapsed` would be a cleaner way to define this since you're defining the icon for the expanded/collapsed state.

scurker (changes requested):
This needs documentation.

I'm not opposed to this change, but is this something that is in @awpearlm or @bobbyomari's designs?

packages/react/src/components/ExpandCollapsePanel/PanelTrigger.tsx
longmatthewh: done

longmatthewh:
@scurker this was driven by a Monitor design where I'm using ExpandCollapsePanel for (notice the different icons)
Screenshot from 2022-04-22 11-39-58

scurker:
> @scurker this was driven by a Monitor design where I'm using ExpandCollapsePanel for (notice the different icons)

Thanks for the info! I just wanted to make sure this had designs behind it.

So the only thing missing is updating the docs page here. FYI I do think in the future the above pattern also will exist elsewhere so we will likely elevate it to its own pattern while the `ExpandCollapsePanel` provides the underlying functionality.

longmatthewh:
@scurker docs have been updated

scurker (approved):

PR #643, opened 2022-05-03
feat: allow a node to be passed in as an IconButton label

thuey:
This allows a node to be passed in as the label for an IconButton. Allowing a node to be passed in allows consumers to include things like screen reader-only content. (see the docs example in this PR)

### DIFF
--- a/docs/patterns/components/IconButton/index.js
+++ b/docs/patterns/components/IconButton/index.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import Demo from '../../../Demo';
-import { IconButton } from '@deque/cauldron-react/';
+import { IconButton, Offscreen } from '@deque/cauldron-react/';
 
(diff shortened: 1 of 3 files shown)


schne324 (approved):

scurker (changes requested):
I don't think this will work as implemented, see how we're also setting `aria-label` below:

https://github.com/dequelabs/cauldron/blob/9d2b7bb935e642a0d20b7aca769e01185556ecdc/packages/react/src/components/IconButton/index.tsx#L78-L83

I think you're on the right track, but we also need the button to have an accessible name as well. Perhaps we should be using `<Offscreen>` there instead of just using `aria-label`.

thuey:
@scurker Good catch. I missed the "disabled" scenario.

scurker (approved):

PR #649, opened 2022-05-11
feat: add ref forwarding to Panel component

thuey:
This PR allows a ref to be forwarded to the Panel's section, which makes it possible to sen programmatic focus to the Panel.


packages/react/src/components/Panel/index.tsx
```diff
@@ -1,67 +1,67 @@
...
 
 interface PanelProps extends HTMLAttributes<HTMLElement> {
   children: ReactNode;
   heading?:
-    | ReactElement<any>
+    | ReactNode
     | {
         id?: string;
-        text: ReactElement<any>;
+        text: ReactNode;
```
thuey, line 12: After adding forwarding, typescript didn't like how `ReactElement<any>` didn't match `PropTypes.node`

schne324 (approved):

PR #654, opened 2022-05-17
feat(icons): update star icon and add alt outline star icon

longmatthewh:
I received 2 new icons from @bobbyomari - an updated star icon and a new alt star icon that's an outlined version


scurker:
@longmatthewh I'm no legal expert, but I think we may be in violation of the CC license there. The icons I used in the past did not have an associated license embedded in the SVG and comments would get stripped out anyway in the bundle. Is there a license free version that we can use?

longmatthewh:
Good question @scurker. Let me follow up with @bobbyomari .

scurker:
Closing this issue since we may address this later with #869

PR #655, opened 2022-05-17
fix(BreadcrumbLink): TS would not allow href attribute on component

longmatthewh:
Having the props extend only `HTMLAttributes<HTMLLinkElement>` did not include things like `href`. Extending `LinkHTMLAttributes<HTMLLinkElement>` brings in some Link specific attributes and extends `HTMLAttributes<HTMLLinkElement>`.

### DIFF
--- a/packages/react/src/components/Breadcrumb/BreadcrumbLink.tsx
+++ b/packages/react/src/components/Breadcrumb/BreadcrumbLink.tsx
@@ -1,8 +1,8 @@
 import React, { forwardRef } from 'react';
 import classnames from 'classnames';
-import { Cauldron } from '../../types';
 
-interface BreadcrumbLinkProps extends React.HTMLAttributes<HTMLLinkElement> {
+interface BreadcrumbLinkProps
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Breadcrumb/BreadcrumbLink.tsx
```diff
@@ -1,8 +1,8 @@
 import React, { forwardRef } from 'react';
 import classnames from 'classnames';
-import { Cauldron } from '../../types';
 
-interface BreadcrumbLinkProps extends React.HTMLAttributes<HTMLLinkElement> {
+interface BreadcrumbLinkProps
+  extends Omit<React.LinkHTMLAttributes<HTMLLinkElement>, 'as'> {
```
scurker, line 5: Is the `Omit<..., 'as'>` necessary? Was there a conflict?

longmatthewh: I ended up doing this, because `as` is typed as a `string` from `HTMLAttributes` vs `React.ElementType` in `BreadcrumbLinkProps`

scurker (approved):

PR #656, opened 2022-05-17
fix(react): ensure that other elements get correctly hidden when setting focusTrap on overlay loader

scurker:
This fixes an issue where focus-trap handles keyboard trapping only and does not isolate the trapped component for SR users.

### DIFF
--- a/packages/react/src/components/LoaderOverlay/index.tsx
+++ b/packages/react/src/components/LoaderOverlay/index.tsx
@@ -35,8 +35,8 @@ const LoaderOverlay = forwardRef<HTMLDivElement, LoaderOverlayProps>(
         ? new AriaIsolate(overlayRef.current)
         : null;
 
-      if (focusTrap && isolator) {
-        isolator.activate();
+      if (isolator) {
+        focusTrap ? isolator.activate() : isolator.deactivate();
       }
 
       return () => {
--- a/packages/react/src/components/LoaderOverlay/index.tsx
+++ b/packages/react/src/components/LoaderOverlay/index.tsx
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Loader from '../Loader';
 import AxeLoader from './axe-loader';
+import AriaIsolate from '../../utils/aria-isolate';
 import useSharedRef from '../../utils/useSharedRef';
 
 interface LoaderOverlayProps extends React.HTMLAttributes<HTMLDivElement> {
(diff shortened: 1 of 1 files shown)


packages/react/src/components/LoaderOverlay/index.tsx
```diff
@@ -29,6 +30,20 @@ const LoaderOverlay = forwardRef<HTMLDivElement, LoaderOverlayProps>(
...
+    useEffect(() => {
+      const isolator = overlayRef.current
+        ? new AriaIsolate(overlayRef.current)
+        : null;
+
+      if (focusTrap && isolator) {
+        isolator.activate();
+      }
+
+      return () => {
```
schne324, line 42: this return function is only run on unmount...doesn't this assume callers are using conditional rendering (`condition && <LoaderOverlay />`) and not using a different technique such as css classes to toggle display (`<LoaderOverlay  className={shouldShow ? 'Loader--visible' : 'Loader--hidden'} />`)

scurker: Correct, but how would we know if they're using a class to conditionally display the element?

I did realize we do need to handle the deactivation if the `focusTrap` prop changes in addition to unmount.

schne324: >  but how would we know if they're using a class to conditionally display the element?

We currently couldn't. We would have to add something like a `show` prop for that.

What I'm worried about here is callers who realized that LoaderOverlay didn't do any aria-hidden management and rolled it themselves. I'm also concerned that callers who happened to be using the above className approach having completely broken apps (they use css to hide their overlay and once hidden, the entire app is `aria-hidden`).  Because of this, should we make this a `BREAKING CHANGE`?

thuey: I understand your concerns, @schne324, but I agree with @scurker on this. It was a bug in the implementation that should be corrected with a patch. I would reserve "BREAKING" for something that directly breaks intended functionality or functionality that has been in place for a long time such that it is widely relied on as the expected behavior. There's some room for subjectivity there, but since the feature with the bug was just released in the last release, I think it would be reasonable to treat it as a patch. I also agree that we should be more explicit about the usage of the LoaderOverlay component.

It would also be good to formalize our definition of a breaking change, since this has come up a few times (do I smell an RFC in the works?). For example, that definition could include that patches to minor releases aren't considered breaking.

schne324: Fair enough...the code change LGTM

schne324 (approved):

PR #662, opened 2022-05-31
feat(pagination): create usePagination hook

okry:
Creates Cauldron's first hook: `usePagination`! 🎉 
- Non-Breaking. Doesn't affect current implementations
- Controller that manages `currentPage` for the user to make it easier to implement pagination with their tables
- Test describes for hook and no-hook implementations
- Describe wrapper for entire test

How do we want to go documenting hooks in cauldron?


docs/patterns/components/Pagination/index.js
```diff
@@ -1,19 +1,16 @@
...
   const itemsPerPage = 25;
-  const [currentPage, setCurrentPage] = useState(3);
-  const itemStart = currentPage * itemsPerPage - itemsPerPage + 1;
-  const itemEnd = Math.min(itemStart + itemsPerPage - 1, totalItems);
 
-  const onNextPageClick = () => setCurrentPage(currentPage + 1);
-  const onFirstPageClick = () => setCurrentPage(1);
-  const onPreviousPageClick = () => setCurrentPage(currentPage - 1);
-  const onLastPageClick = () => setCurrentPage(5);
+  const [pagination, pageStatus] = usePagination({
```
scurker, line 10: Nitpick: I wonder if we should be more implicit in the naming of the return values here? I'd be concerned with someone confusing this to something like a `useState` which typically has `[state, setter]` in that order.

I think named values might be cleaner and more readable:

```ts
const { paginationState, paginationProps } = usePagination(...)
```

With named values, it would be more clear what's supposed to be spread onto the pagination component vs someone getting the order wrong for the prior example. What do you think?

okry: Yeah, that makes sense.

scurker (dismissed):
Just need to document this within the pagination docs and we should be good to go!

docs/patterns/components/Pagination/index.js
```diff
@@ -130,7 +124,56 @@ const PaginationDemo = () => {
...
+                  'The (1-based) number of the page that pagination will start on.',
+                required: false,
+                default: 1
+              }
+            }}
+            defaultProps={undefined}
+          />
+        </div>
+        <p>
+          <span>
```
scurker, line 164: Nit: This doesn't get converted to markdown, so you may want to wrap the variables with `<code>`. Additionally, are the opening `{` and `}` brackets needed?

docs/patterns/components/Pagination/index.js
```diff
@@ -130,7 +124,56 @@ const PaginationDemo = () => {
...
+            {
+              'This returns `pagination`, a series of controls that can be spread into the Pagination component and'
+            }
+          </span>
+          <br />
+          <span>
+            {
```
scurker, line 171: Same as above comment.

docs/patterns/components/Pagination/index.js
okry: Yeah, I felt like the `code` tag didn't stick out too well, but I can make that update.

scurker (approved):

PR #663, opened 2022-05-31
feat(components): RadioCardGroup

Zidious:
image

ref: https://github.com/dequelabs/cauldron/issues/666


packages/react/src/components/RadioCardGroup/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+            onChange={() => {
+              handleChange(radioValue);
+              onChange(radio, inputs.current?.[index]);
+            }}
+            disabled={disabled}
+            checked={isChecked}
+            {...other}
+          />
+          <CardContent>
+            <div className={classNames('Radio_card_icon_div')}>
```
scurker, line 109: I'm not sure this wrapper is necessary. We should be able to position the `Icon` with absolute styles:

```css
.RadioCardGroup {
  position: relative;
}

.RadioCardGroup .Icon {
  position: absolute;
  top: n; /*whatever the x value needs to be */
  right: n; /* whatever the y value needs to be */
}
```

packages/react/src/components/RadioCardGroup/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+              {isChecked && (
+                <Icon
+                  className={classNames('Radio_card--icon')}
+                  type="check-circle"
+                />
+              )}
+            </div>
+
+            <div className={classNames('Radio_card_base')}>
```
scurker, line 118: This doesn't follow BEM, it should be `RadioCardGroup__NAMEOFBLOCK`

packages/react/src/components/RadioCardGroup/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+              <img
+                className={classNames('Radio_card_img')}
```
scurker, line 120: This doesn't follow BEM, a more appropriate name would be something like `RadioCardGroup__Image`

packages/react/src/components/RadioCardGroup/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+  void defaultValue;
+
+  // reset the input refs array
+  // refs get clobbered every re-render anyway and this supports "dynamic" radios
+  // (changing the number of radio buttons for example)
+  inputs.current = [];
+
+  return (
+    <div
+      className={classNames('Radio_card_group')}
```
scurker, line 146:
```suggestion
      className="RadioCardGroup"
```

Following our BEM guidelines, this should match the name of the component.

packages/react/src/components/RadioCardGroup/index.tsx
Zidious: I found I need to have the wrapper to stop from this occurring when clicking a card, this way the div is always set with a min-height


before: 
image
after:
image

packages/styles/forms.css
```diff
@@ -323,6 +323,34 @@ textarea.Field--has-error:focus:hover,
   appearance: none;
 }
 
+.RadioCard {
+}
+
```
scurker: Let's put these styles in a `package/styles/radio-card.css` since these styles aren't strictly related to all form styles.

packages/styles/forms.css
```diff
@@ -323,6 +323,34 @@ textarea.Field--has-error:focus:hover,
...
+.RadioCard__overlay {
+  border-radius: 50%;
+}
+
+.RadioCard input[type='radio'] {
+  position: absolute;
+  opacity: 0;
+  top: 8px;
+  left: 8px;
```
scurker, line 334: The positioning of this element shouldn't matter if it's visually hidden.

docs/patterns/components/RadioCardGroup/index.js
```diff
@@ -0,0 +1,199 @@
...
+              ),
+              cardIcon: 'check-circle'
+            },
+            {
+              id: 'no',
+              value: 'no',
+              label: 'No',
+              disabled: true,
+              cardImg: (
+                <svg viewBox="0 0 100 100">
```
schne324, line 58: Should maybe aria-hide this so screen readers won't announce a generic nameless image

```suggestion
                <svg viewBox="0 0 100 100" aria-hidden="true">
```

schne324 (dismissed):
The footprint of the actual `input[type=radio]` spans outside of the "card"...as a result, there is an area outside of the card that, when clicked, will activate a radio button.


Screen Shot 2022-06-13 at 10 08 31 AM

Zidious:
> The footprint of the actual input[type=radio] spans outside of the "card"...as a result, there is an area outside of the card that, when clicked, will activate a radio button.

Updated input width / height: 
image

schne324 (approved):

Issue #650, opened 2022-05-11
Create new component for our 'issue view' expand/collapse widget


schne324:
<img width="462" alt="Screen Shot 2022-05-11 at 9 43 43 AM" src="https://user-images.githubusercontent.com/3180185/167903193-561c0c9a-e72f-447d-8451-e2287121b17b.png">

ping @schne324 / @scurker  for the uxpin link as an example

stephenmathieson:
Is this just a styled version of https://github.com/dequelabs/react-wai-accordion?

PR #669, opened 2022-06-09
feat: adds an accordion component to library

tbusillo:
ref: https://github.com/dequelabs/cauldron/issues/650

FYI, updated UX Pin for component was put in the pin for Cauldron: https://preview.uxpin.com/68df515d0023cc4b79f25d43f7663e24e363b1b9#/pages/150823418/simulate/sitemap


docs/patterns/components/Accordion/index.css
```diff
@@ -0,0 +1,11 @@
+.Accordion__container {
+  max-width: 600px;
+}
+
+.Accordion__panel {
+  min-height: 200px;
```
dequejosie, line 6: I forgot, do we usually convert px to rem?

tbusillo: I can definitely do that - this is only included for the demo purposes though, just as an FYI

dequejosie: Ah gotcha. Not a big deal.

packages/styles/variables.css
```diff
@@ -8,6 +8,8 @@
   /* color palette */
   --white: #ffffff;
   --dark-workspace-color: #53636e;
+  --stroke-dark: #5d676f;
+  --background-dark: #53636e;
```
tbusillo, line 12: This adds in styles not currently covered, but found in https://preview.uxpin.com/68df515d0023cc4b79f25d43f7663e24e363b1b9#/pages/141465411/simulate/sitemap

packages/react/src/components/ExpandCollapsePanel/PanelTrigger.tsx
```diff
@@ -11,45 +11,61 @@ export interface PanelTriggerProps
...
+  }
+);
 
 PanelTrigger.propTypes = {
   children: PropTypes.oneOfType([PropTypes.func, PropTypes.node]),
   open: PropTypes.bool,
   onClick: PropTypes.func,
-  className: PropTypes.string,
-  iconExpanded: PropTypes.string,
-  iconCollapsed: PropTypes.string
```
tbusillo, line 52: These prop types are causing a TypeScript issue. Not sure anything was changed that should have effected these. 

see: https://app.circleci.com/pipelines/github/dequelabs/cauldron/1429/workflows/2ec9949c-ab27-4a33-b535-1edf6531356e/jobs/6202

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+      {children}
+    </div>
+  );
+};
+
+interface AccordionContainerProps extends React.HTMLAttributes<HTMLDivElement> {
+  className?: string;
+  children: React.ReactElement[];
+  open?: boolean;
+  setIsOpen?: (open: boolean) => void;
```
scurker, line 28:
```suggestion
  open?: boolean;
```

I don't think we need both a prop and callback function to control the state of the component.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+  trigger?: React.ReactNode | React.ReactNode[];
+  iconExpanded?: IconType;
+  iconCollapsed?: IconType;
+  shouldHideIcon?: boolean;
+  isControlled?: boolean;
```
scurker, line 33: This seems like a bit of an anti-pattern, we should be able to devise how the component is controlled by the existence of other props.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+  animationTiming?: number | boolean;
+  props?: React.HTMLAttributes<HTMLDivElement>;
```
scurker, line 35: This prop isn't really self documenting.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+interface AccordionContainerProps extends React.HTMLAttributes<HTMLDivElement> {
+  className?: string;
+  children: React.ReactElement[];
+  open?: boolean;
+  setIsOpen?: (open: boolean) => void;
+  trigger?: React.ReactNode | React.ReactNode[];
+  iconExpanded?: IconType;
+  iconCollapsed?: IconType;
+  shouldHideIcon?: boolean;
```
scurker, line 32:
```suggestion
  hideIcon?: boolean;
```

The above reads a bit cleaner and is more consistent with how we name our props elsewhere. Additionally this should likely be on the trigger.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+    <div
+      className={classNames(className ? className : '', 'Accordion__container')}
+      {...props}
+    >
+      {children}
+    </div>
+  );
+};
+
+interface AccordionContainerProps extends React.HTMLAttributes<HTMLDivElement> {
```
scurker, line 24: If this is intended to be an extension of the `ExpandCollapsePanel` component, we should instead extend those props instead of recreating all of the props here.

```suggestion
interface AccordionContainerProps extends ExpandCollapsePanelProps {
```

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+  className?: string;
+  children: React.ReactElement[];
+  open?: boolean;
+  setIsOpen?: (open: boolean) => void;
+  trigger?: React.ReactNode | React.ReactNode[];
+  iconExpanded?: IconType;
+  iconCollapsed?: IconType;
```
scurker, line 31: Is there a use case where we need to support different icons for the expand/collapse state? If we do, this should probably be on the trigger and not the container.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+  isControlled = false,
+  animationTiming,
+  ...props
+}: AccordionContainerProps) => {
+  const [elementId, setElementId] = useState<string | null>(null);
+  const [notControlledOpen, setNotControlledOpen] = useState(false);
+
+  useEffect(() => {
+    setElementId(randomId());
+  }, []);
```
scurker, line 54: 2 things:

* The content author should be able to pass in their own id
* I think we would likely want this id under the underlying component (ExpandCollapsePanel) if we want to allow this id to be used somewhere.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+  if (!isControlled) {
+    return (
+      <ExpandCollapsePanel
+        id={`${elementId}-panel`}
+        open={notControlledOpen}
+        onToggle={() => setNotControlledOpen(!notControlledOpen)}
+        aria-labelledby={`${elementId}-trigger`}
```
scurker, line 62:
```suggestion
```

This shouldn't be needed unless we're making this a named region. `aria-labelledby` isn't supported for `generic` roles.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+      >
+        <PanelTrigger
+          id={`${elementId}-trigger`}
+          iconCollapsed={iconCollapsed}
+          iconExpanded={iconExpanded}
+          className={classNames(
+            children[0].props.className
+              ? children[0].props.className
+              : 'Accordion__trigger'
+          )}
```
scurker, line 74: I'm a bit confused about what's happening here. What are you trying to accomplish?

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+          aria-controls={`${elementId}-panel`}
+          shouldHideIcon={shouldHideIcon}
+        >
+          {children[0]}
```
scurker, line 78: What if I do 

```jsx
<AccordionContainer>
  <AccordionContent>...</AccordionContent>
  <AccordionTrigger>...</AccordionTrigger>
</AccordionContainer>
```

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+        </PanelTrigger>
+        {children[1]}
+      </ExpandCollapsePanel>
+    );
+  }
+
+  if (setIsOpen) {
+    return (
+      <>
+        <ExpandCollapsePanel
```
scurker, line 88: This seems repetitive and isn't very dry, we should try to render only once and pass only the necessary props.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,144 @@
...
+  trigger?: React.ReactNode | React.ReactNode[];
+  iconExpanded?: IconType;
+  iconCollapsed?: IconType;
+  shouldHideIcon?: boolean;
+  isControlled?: boolean;
+  animationTiming?: number | boolean;
+  props?: React.HTMLAttributes<HTMLDivElement>;
+}
+
+const AccordionContainer = ({
```
scurker, line 38: Why `Accordion` and `AccordionContainer`? Why not just `Accordion`?

packages/react/src/components/Code/index.tsx
```diff
@@ -41,7 +41,8 @@ Code.displayName = 'Code';
 Code.propTypes = {
   children: PropTypes.string.isRequired,
   language: PropTypes.oneOf(['javascript', 'css', 'html', 'yaml']),
-  className: PropTypes.string
+  className: PropTypes.string,
+  tabIndex: PropTypes.number
```
scurker, line 45: ?

packages/react/src/components/ExpandCollapsePanel/ExpandCollapsePanel.tsx
```diff
@@ -11,9 +11,10 @@ import {
 export interface ExpandCollapsePanelProps
   extends React.HTMLAttributes<HTMLDivElement> {
   open?: boolean;
-  children: React.ReactNode;
+  children?: React.ReactNode;
```
scurker, line 14: IMO children shouldn't be optional

packages/react/src/components/ExpandCollapsePanel/ExpandCollapsePanel.tsx
```diff
@@ -11,9 +11,10 @@ import {
...
   animationTiming?: number | boolean;
   onToggle: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  props?: React.HTMLAttributes<HTMLDivElement>;
```
scurker, line 17: ?

packages/react/src/components/ExpandCollapsePanel/ExpandCollapsePanel.tsx
```diff
@@ -198,16 +199,15 @@ export default class ExpandCollapsePanel extends React.Component<
             onClick: this.handleToggle
           })}
         <div
-          {...other}
           className={classnames(
-            'ExpandCollapse__panel',
+            className ? className : 'ExpandCollapse__panel',
```
scurker, line 203: This should _always_ have the `ExpandCollapse__panel` className, otherwise we're introducing a breaking change.

scurker (changes requested):
Comments below:

I also noticed the accordion container in light mode doesn't have a border, and in dark mode it looks like there's a 1px offset of the trigger so that it doesn't appear aligned with the content.

packages/react/src/components/Code/index.tsx
tbusillo: Yeah, this got stuck in from an old branch and not sure why (guessing not rebasing) - needed to remove this.

tbusillo:
> Comments below:
> 
> 
> 
> I also noticed the accordion container in light mode doesn't have a border, and in dark mode it looks like there's a 1px offset of the trigger so that it doesn't appear aligned with the content.

Thank you! Will take a look tomorrow and try to get everything addressed.

packages/react/src/components/ExpandCollapsePanel/ExpandCollapsePanel.tsx
tbusillo: Would this be breaking if it is always providing it as the default, but allowing for a different class name to be passed in, if the developer wants to? That's more or less what is trying to be accomplished here (among other areas with similar patterns).

packages/react/src/components/Accordion/Accordion.tsx
tbusillo: Definitely good with that! Was trying to keep some of the JSX a little less complicated, but you're right and will do this differently.

packages/react/src/components/Accordion/Accordion.tsx
tbusillo: Is that a pattern we want to be able to support? Because as it stands right now, there's a pretty specific ordering of children elements required by the underlying collapse panel/trigger. The way it implements things is slightly different, but it requires similar ordering of children.

packages/react/src/components/Accordion/Accordion.tsx
tbusillo: So this was to address the specific ordering of children that the underlying expand/collapse component requires. I tried several ways, including passing in a child component itself, but when the expand/collapse would not clone the elements in the correct order unless being explicit in this manner. 

It's also what ended up necessitating the container altogether, but perhaps this is something we could look at together? Could certainly use your input/advice on how to handle it!

packages/react/src/components/ExpandCollapsePanel/ExpandCollapsePanel.tsx
scurker: Additional class names should be appended onto the existing one.

packages/react/src/components/ExpandCollapsePanel/PanelTrigger.tsx
```diff
@@ -1,55 +1,69 @@
...
+      ...otherProps
+    }: PanelTriggerProps,
+    ref
+  ) => {
+    return (
+      <button
+        className={classnames(
+          className ? className : 'ExpandCollapse__trigger',
+          fullWidth ? 'fullWidth' : '',
+          open ? 'expanded' : ''
```
scurker, line 41: I know some of this was already there, but `classnames` can accept an object with boolean values, example below

```jsx
className={classnames('ExpandCollapse__trigger', className, {
  'fullWidth': fullWidth,
  ...
})
```

Also is the `expanded` class needed? We can use `[aria-expanded="true"]` attribute to achieve the same thing.

packages/styles/accordion.css
```diff
@@ -0,0 +1,89 @@
...
+  padding: var(--space-small);
+  border-left: 1px;
+  border-bottom: 1px;
+  border-right: 1px;
+  border-style: solid;
+  border-top: 0px;
+  border-color: var(--gray-40);
+  border-bottom-left-radius: 3px;
+  border-bottom-right-radius: 3px;
+  box-shadow: 0px 2px 3px 0px #00000015;
```
scurker, line 46: Nit: `rgba` is more widely supported. Also `0` values do not need a unit.

```suggestion
  box-shadow: 0 2px 3px 0 rgba(0, 0, 0, .15);
```

packages/styles/accordion.css
```diff
@@ -0,0 +1,89 @@
...
+.Accordion__panel {
+  background-color: var(--white);
+  color: var(--gray-90);
+  padding: var(--space-small);
+  border-left: 1px;
+  border-bottom: 1px;
+  border-right: 1px;
+  border-style: solid;
+  border-top: 0px;
+  border-color: var(--gray-40);
```
scurker, line 43: Nitpick: This could be shorter:

```css
border: 1px solid var(--gray-40);
border-bottom: none;
```

packages/styles/accordion.css
```diff
@@ -0,0 +1,89 @@
...
+  border-bottom-left-radius: 3px;
+  border-bottom-right-radius: 3px;
+  box-shadow: 0px 2px 3px 0px #00000015;
+}
+
+.Accordion__container div.ExpandCollapse__panel {
+  max-width: calc(100% - 0.5px);
```
scurker, line 50: Why the half pixel?

packages/react/src/components/ExpandCollapsePanel/PanelTrigger.tsx
tbusillo: Oh, cool - I never knew that about `classnames`! Will definitely give that a try and I don't have strong feelings one way or the other, re: the expanded class and you are right that it could use `aria-expanded`.

packages/styles/accordion.css
scurker: That's an indication that something else may be going on, but generally we want to avoid subpixel rendering because it is dependent from browser to browser.

tbusillo: Ended up being able to remove it - think it was from when the borders on the panel were not rendering for some reason and it looks good now.

packages/react/src/components/ExpandCollapsePanel/PanelTrigger.tsx
tbusillo: All great points and I agree 100%! Definitely like that approach better and I made that change!

tbusillo:
@scurker Think this is ready for another review and thank you again for the help earlier! I just realized that the [ARIA docs](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) mention something that isn't being incorporated here or in the `ExpandCollapse` components (below). Do you think that is something that should be supported internally in the component, or allow users to manage on their own? 

- The title of each accordion header is contained in an element with role [button](https://w3c.github.io/aria/#button).
- Each accordion header button is wrapped in an element with role [heading](https://w3c.github.io/aria/#heading) that has a value set for [aria-level](https://w3c.github.io/aria/#aria-level) that is appropriate for the information architecture of the page.
 - If the native host language has an element with an implicit heading and aria-level, such as an HTML heading tag, a native host language element may be used.
 - The button element is the only element inside the heading element. That is, if there are other visually persistent elements, they are not included inside the heading element."

packages/styles/accordion.css
```diff
@@ -0,0 +1,77 @@
...
+}
+
+.cauldron--theme-dark button.Accordion__trigger[aria-expanded='true'] {
+  border-bottom-right-radius: 0px;
+  border-bottom-left-radius: 0px;
+  background-color: var(--accent-dark);
+}
+
+.cauldron--theme-dark button.Accordion__trigger:hover {
+  box-shadow: inset 3.5px 0px var(--accent-info);
```
scurker, line 64: Nitpick: unitless

```suggestion
  box-shadow: inset 3.5px 0 var(--accent-info);
```

packages/styles/accordion.css
```diff
@@ -0,0 +1,77 @@
+.Accordion__trigger,
+.Accordion > .ExpandCollapse__trigger {
+  background-color: var(--white);
+  padding: 12px;
+  width: 100%;
+  display: flex;
+  align-items: center;
+  border-radius: 3px;
+  border: 1px var(--gray-40) solid;
+  background-size: cover;
```
scurker, line 10: This may be related to the border shorthand properties being out of order.

```suggestion
```

packages/styles/accordion.css
```diff
@@ -0,0 +1,77 @@
...
+  border-color: var(--gray-40);
+  border-bottom-left-radius: 3px;
+  border-bottom-right-radius: 3px;
+  box-shadow: 0 2px 3px 0 rgba(0, 0, 0, 0.15);
+}
+
+.cauldron--theme-dark button.Accordion__trigger {
+  color: var(--white);
+  background-color: var(--accent-medium);
+  border: 1px var(--stroke-dark) solid;
```
scurker, line 49:
```suggestion
  border: 1px solid var(--stroke-dark);
```

scurker (changes requested):
This is coming along, but some suggestions:

We should have variables assigned for all of the defined colors (in both light/dark themes):

Example:

```css
:root {
  --accordion-background-color: var(--white);
}

.cauldron--theme-dark {
  --accordion-background-color: var(--accent-dark);
}

.Accordion {
  background: var(--accordion-background-color);
}
```

This provides 2 distinct advantages:

* An author does not need to know everywhere that color is used
* It helps to future proof components

For the 1st, if an author wanted to redefine the color that is set via a property they would need to do so where the rule cascades and with an equal or higher specificity. 

For the 2nd, this allows us (the maintainers) greater flexibility to change styles while maintaining backwards compatibility. Say in the future we could change `background` to `background-color` and as long as an author set their own variable correctly it would automatically be inherited.

This also makes it easier to see where light/dark styles are used since you can define the rule using them once, and the variable will be utilized based on the current theme.

You can see an example of this with button:

https://github.com/dequelabs/cauldron/blob/cc78a93efb85d21e7d6491a483cf87f2fe4cd2f7/packages/styles/button.css#L1-L5

packages/styles/accordion.css
```diff
@@ -0,0 +1,77 @@
...
+}
+
+button.Accordion__trigger[aria-expanded='true'] {
+  border-bottom-right-radius: 0px;
+  border-bottom-left-radius: 0px;
+  background-color: var(--gray-20);
+}
+
+button.Accordion__trigger:hover {
+  box-shadow: inset 3.5px 0px var(--accent-primary);
```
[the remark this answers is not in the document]
tbusillo: Ah, both of those things were already verified with him (sorry, forgot to mention that). He only did the dark on during an ad-hoc call we had together and he was good with the box shadow/radius. I can double check, but he definitely was good with both.

scurker:
> * The title of each accordion header is contained in an element with role [button](https://w3c.github.io/aria/#button).
> * Each accordion header button is wrapped in an element with role [heading](https://w3c.github.io/aria/#heading) that has a value set for [aria-level](https://w3c.github.io/aria/#aria-level) that is appropriate for the information architecture of the page.

For a proper accordion, that's probably what we want.

> * If the native host language has an element with an implicit heading and aria-level, such as an HTML heading tag, a native host language element may be used.

This would be specified by the content author, so nothing we need to do here.

> * The button element is the only element inside the heading element. That is, if there are other visually persistent elements, they are not included inside the heading element."

There's a bit more nuance to this once we implement the advanced functionality so I think what we have for now is fine?

packages/react/src/components/ExpandCollapsePanel/PanelTrigger.tsx
```diff
@@ -13,35 +13,33 @@ export interface PanelTriggerProps
...
       type="button"
       aria-expanded={open}
       onClick={onClick}
+      {...otherProps}
     >
       <div className="ExpandCollapse__trigger-title">
         {typeof children === 'function' ? children({ open: !!open }) : children}
       </div>
-      <Icon type={open ? iconExpanded : iconCollapsed} />
+      {<Icon type={open ? iconExpanded : iconCollapsed} />}
```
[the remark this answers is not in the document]
tbusillo: Sorry about that - actually meant to remove a change had previously in there but forgot to remove the brackets.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,91 @@
...
+    child => (child as React.ReactElement<any>).type === AccordionTrigger
+  );
+  const panel = childrenArray.find(
+    child => (child as React.ReactElement<any>).type === AccordionContent
+  );
+
+  useEffect(() => {
+    setElementId(randomId());
+    return;
+  }, []);
```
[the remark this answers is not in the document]
tbusillo: Think I did it in the useEffect because it was re-calculating the random number repeatedly, but it might not be needed anymore after some of the other changes. Thank you for the reference and will try it out!

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,91 @@
...
+  className: string;
+}
+
+const AccordionContent = ({
+  children,
+  className,
+  ...otherProps
+}: AccordionContentProps) => {
+  return (
+    <div className={classnames('Accordion__panel', className)} {...otherProps}>
```
scurker, line 30: Nitpick: This one is a bit weird because it creates a double nested div:

```html
<div class="ExpandCollapse__panel" id="...">
  <div class="Accordion__panel">Here is some content</div>
</div>
```

If it's something we can easily solve it would be nice to not have div soup here, but I'm not going to die on that hill.

tbusillo: I tried doing this briefly, mainly through use of a fragment on the `<AccordionContent>`, but it ended up causing issues with the animation (it would no longer animate expand, only collapse). What would you think about leaving for now and adding it into the advance accordion issue as an item to address?

packages/styles/accordion.css
```diff
@@ -0,0 +1,72 @@
...
+}
+
+button.Accordion__trigger[aria-expanded='true'] {
+  border-bottom-right-radius: 0;
+  border-bottom-left-radius: 0;
+  background: var(--accordion-trigger-background-color-expanded);
+}
+
+button.Accordion__trigger:hover {
+  box-shadow: inset 3.5px 0px var(--accordion-trigger-box-shadow-hover);
```
scurker, line 50: I think a previous comment got wiped out, but this doesn't match the ux pin.

The hover border has a flat border where this has rounded edges.

scurker (changes requested):
This is _really_ close, good work on cleaning up the CSS!

The only other thing I see that may be missing is testing that the actual `AccordionContent` is hidden, not just `aria-expanded`.

packages/react/__tests__/src/components/Accordion/index.js
```diff
@@ -0,0 +1,273 @@
...
+    const accordion = mount(
+      <Accordion open={false}>
+        <AccordionTrigger>Testing 1 2 3</AccordionTrigger>
+        <AccordionContent>This is another test</AccordionContent>
+      </Accordion>
+    );
+
+    const triggerId = accordion
+      .find('button.Accordion__trigger')
+      .getDOMNode()
```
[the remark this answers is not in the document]
tbusillo: Changed all the others, but ended up removing this one since it's pretty much identical to the one right below it.

packages/react/src/components/Accordion/Accordion.tsx
tbusillo: Ended up being able to work it out using the fragment and a little more specificity with the CSS for the element!

docs/patterns/components/Accordion/index.js
```diff
@@ -0,0 +1,172 @@
...
+              default: 'function () {}'
+            }
+          }}
+        />
+
+        <h3>
+          <code>AccordionTrigger</code>
+        </h3>
+        <p>
+          The <code>AccordionTrigger </code> extends
```
scurker, line 124:
```suggestion
          The <code>AccordionTrigger </code> extends{' '}
```

Isn't React great?

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  const elementId = useId();
+
+  return (
+    <div className="Accordion">
+      <ExpandCollapsePanel
+        id={panelElement.props.id || `${elementId}-panel`}
+        className={classnames(panelElement.props.className)}
+        {...panelElement.props.otherProps}
```
scurker, line 81: I think this can just be 
```suggestion
        {...panelElement.props}
```

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,132 @@
...
+Accordion.propTypes = {
+  children: PropTypes.node
+};
+
+Accordion.defaultProps = {
+  children: null
+};
+
+AccordionTrigger.propTypes = {
+  children: PropTypes.oneOfType([element, string]),
```
scurker, line 112: This should probably be 

```suggestion
  children: PropTypes.node,
```

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,132 @@
...
+  headingLevel: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6', undefined])
+};
+
+AccordionTrigger.defaultProps = {
+  children: null,
+  headingLevel: undefined
+};
```
scurker, line 119: For functional components, the preferred way to set default values is to use the default parameters of the functional arguments.

```suggestion
```

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+AccordionContent.propTypes = {
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string
+};
+
+AccordionContent.defaultProps = {
+  children: null,
+  className: undefined
+};
```
scurker, line 129:
```suggestion
```

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,132 @@
...
+  children: PropTypes.node
+};
+
+Accordion.defaultProps = {
+  children: null
+};
+
+AccordionTrigger.propTypes = {
+  children: PropTypes.oneOfType([element, string]),
+  headingLevel: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6', undefined])
```
scurker, line 113: `headingLevel` isn't a prop, did you mean `heading`?

scurker (changes requested):
Noticed a few more things.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,132 @@
...
+        className={classnames(panelElement.props.className)}
+        {...panelElement.props.otherProps}
+      >
+        <PanelTrigger
+          iconCollapsed="triangle-right"
+          iconExpanded="triangle-down"
+          className={classnames('Accordion__trigger', trigger.props.className)}
+          aria-controls={panelElement.props.id || `${elementId}-panel`}
+          heading={trigger.props.heading}
+          {...trigger.props.otherProps}
```
scurker, line 89: Sorry, looks like this comment weirdly disappeared in my review.

This doesn't actually work as `otherProps` isn't a named prop on the component. What you actually probably want is to destructure the props to pull out `className` and spread the prop value onto the panel.

```jsx
const  { className: triggerClassName, ...triggerProps } = trigger.props

...

<PanelTrigger
  iconCollapsed="triangle-right"
  iconExpanded="triangle-down"
  className={classnames('Accordion__trigger', triggerClassName)}
  aria-controls={panelElement.props.id || `${elementId}-panel`}
  {...triggerProps}
>
  ...
</PanelTrigger>
```

docs/patterns/components/Accordion/index.js
tbusillo: Hah! Yeah, the spacing on the prop doc pages was getting really weird sometimes. I kept removing things, then sometimes prettier would add back in, sometimes I have no idea how it happened 😂

packages/react/src/components/Accordion/Accordion.tsx
tbusillo: Ugh, I ran into a weird merge conflict at one point and there was some really bizarre stuff happening. I initially did it like you see here with "headingLevel" where it was really declarative/limited (you could do H1-6 or undefined), then ended up doing the more dynamic "heading" one that's mostly in there. Will get this fixed!

packages/react/src/components/Accordion/Accordion.tsx
tbusillo: Ah, that's a good call! didn't even think about destructing from there!

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,128 @@
...
+AccordionContent.displayName = 'AccordionContent';
+AccordionTrigger.displayName = 'AccordionTrigger';
+
+Accordion.propTypes = {
+  children: PropTypes.node
+};
+
+Accordion.defaultProps = {
+  children: null
+};
```
scurker, line 115:
```suggestion
```

scurker (approved):

PR #673, opened 2022-06-10
feat(styles): add axe-core issue impact colors

AdnoC:
Required for https://github.com/dequelabs/walnut/pull/3320.
Colors taken from https://github.com/dequelabs/walnut/blob/develop/app/src/modules/app/components/RuleHelp.css#L95-L119

### DIFF
--- a/packages/styles/variables.css
+++ b/packages/styles/variables.css
@@ -35,6 +35,10 @@
   --accent-secondary-active: var(--gray-30);
   --focus-light: #c11bde;
   --focus-dark: #eb94ff;
+  --axe-issue-critical: #ed5959;
+  --axe-issue-serious: #f3826b;
+  --axe-issue-moderatee: #ffdd76;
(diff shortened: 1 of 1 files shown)


packages/styles/variables.css
```diff
@@ -35,6 +35,10 @@
   --accent-secondary-active: var(--gray-30);
   --focus-light: #c11bde;
   --focus-dark: #eb94ff;
+  --axe-issue-critical: #ed5959;
+  --axe-issue-serious: #f3826b;
+  --axe-issue-moderatee: #ffdd76;
+  --axe-issue-minor: #d3dde0;
```
straker, line 41: Maybe simplify and say `issue-cirtical` rather than specifying axe? Other than that LGTM

scurker: `issue-critical` gets my vote

PR #674, opened 2022-06-10
feat(button): add option to render as tag

AdnoC:
Required for https://github.com/dequelabs/walnut/pull/3320

### DIFF
--- a/packages/react/src/components/Button/index.tsx
+++ b/packages/react/src/components/Button/index.tsx
@@ -27,6 +27,7 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(
         'Button--error': variant === 'error',
         Link: variant === 'link',
         Tag: variant === 'tag',
+        'Button--tag': variant === 'tag',
         'Button--thin': thin
       })}
       ref={ref || buttonRef}
--- a/packages/react/src/components/Button/index.tsx
+++ b/packages/react/src/components/Button/index.tsx
@@ -3,7 +3,7 @@ import classNames from 'classnames';
 
 export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
   buttonRef?: Ref<HTMLButtonElement>;
-  variant?: 'primary' | 'secondary' | 'error' | 'link';
+  variant?: 'primary' | 'secondary' | 'error' | 'link' | 'tag';
   thin?: boolean;
 }
 
@@ -26,6 +26,7 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(
         'Button--secondary': variant === 'secondary',
         'Button--error': variant === 'error',
         Link: variant === 'link',
+        Tag: variant === 'tag',
         'Button--thin': thin
       })}
       ref={ref || buttonRef}
--- a/packages/styles/button.css
+++ b/packages/styles/button.css
@@ -41,6 +41,10 @@
   gap: 8px;
 }
 
+.Button--tag {
+  position: relative;
+}
+
 button.Link {
   cursor: pointer;
 }
@@ -55,7 +59,8 @@ button.Link {
 .Button--primary:before,
 .Button--secondary:before,
 .Button--clear:before,
-.Button--error:before {
+.Button--error:before,
+.Button--tag:before {
   content: '';
   position: absolute;
   top: -2px;
@@ -64,6 +69,10 @@ button.Link {
   left: -2px;
   border-radius: 3px;
   pointer-events: none;
+}
+
+.Button--tag:before {
+  border-radius: 11px;
 }
 
 .Button--primary:not([disabled]):not([aria-disabled='true']):hover:before {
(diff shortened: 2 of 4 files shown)


straker (dismissed):
Should add an example of the Tag use on the button component page

straker:
LGTM

thuey (changes requested):
We should provide a visual indicator to sighted users on hover that the tag button is interactive. Right now, there is no indication that the tag button is interactive

packages/react/src/components/Button/index.tsx
```diff
@@ -26,6 +26,8 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(
         'Button--secondary': variant === 'secondary',
         'Button--error': variant === 'error',
         Link: variant === 'link',
+        Tag: variant === 'tag',
+        'Button--tag': variant === 'tag',
```
thuey, line 30: Rather than introduce a new class name, can we just leverage `button.Tag` like we are doing for `button.Link`?

thuey (changes requested):

packages/react/src/components/Button/index.tsx
scurker: I could go either way on this, although we should be consistent on both. I think we can reutilize the styles for `.Tag` but it would be nice to specifically target our buttons with `.Button--variant` classname. If we want to use `.Button--tag`, I think `.Button--link` should exist as well.

packages/styles/button.css
```diff
@@ -41,6 +41,10 @@
   gap: 8px;
 }
 
+.Button--tag {
+  position: relative;
+}
```
scurker, line 46: Why `position: relative`?

AdnoC: That is needed for hover to work. Since the `:before` element has `position: absolute`.

packages/react/src/components/Button/index.tsx
thuey: @scurker My personal preference would be `.Button.Tag`, since it still targets our buttons without introducing another classname that needs to be included any time you want to do `className="Button Tag"`. However, like you said, I wanted to be consistent with what we had w/o introducing a breaking change (`button.Link` -> `.Button.Link`) 🤷

thuey: Thoughts?

scurker: @thuey The problem is css specificity, and is counter to [our guidelines](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#styles): 

> For modifiers, such as variants or states, these values would be delimited by two dashes --

```css
.Calendar--variant-large {
  font-size: var(--text-size-large);
}

.Calendar--variant-small {
}
```

thuey: @scurker 
I see what you are saying. I think part of the weirdness we are running into here is that `Tag` itself is an "element", but it's also being used as a variant.

In any case, it seems like our options are:
1. Leave it `button.Tag` and `button.Link` to have both consistency and not introduce a breaking change.
2. Change to `.Button--tag` and `.Button--link` to have both consistency, but introduce a breaking change.
3. Change to `.Button--tag` but keep `button.Link` to get the tag variant to be how we want it, not introduce a breaking change, but lose consistency.

I'm leaning toward option 1, but would be good with whichever you think is best.

scurker: I'm suggesting something completely different that doesn't introduce a breaking change:

* Use both `button.Tag` and `button.Button--tag`. `.Tag` inherits the styles from `.Tag` and `.Button--tag` keeps our conventions in place (and implementing the same pattern for `.Link`).

This allows someone to target either all tags (with `.Tag`) or just button tag variants (with `.Button--tag`) while keeping specificity simple.

thuey: Gotcha. That works for me 👍

thuey: Actually, that still doesn't answer the question of what happens to the current `button.Link` styles:

https://github.com/dequelabs/cauldron/blob/62c917271fd30df2e4e65c8eb421578ae20f373c/packages/styles/button.css#L44-L46

https://github.com/dequelabs/cauldron/blob/62c917271fd30df2e4e65c8eb421578ae20f373c/packages/styles/button.css#L231-L237

If we change those to `button.Button--link`, that is a breaking change. If we leave them as-is, that breaks consistency.

scurker: Since this is an unrelated change to the proposed work, I don't have an issue with this being a breaking change as part of a separate ticket so here's my suggestion:

* Create a separate issue to change the `.Link` class to `.Button--link`
* Make the change here for `.Button--tag` as a `feat`

thuey: Here's the ticket: https://github.com/dequelabs/cauldron/issues/681

Sorry for the confusion @AdnoC. We will want to do `Button--tag` after all

scurker (approved):

thuey (approved):

PR #676, opened 2022-06-14
fix(react): Support React 18 types

stephenmathieson:
This patch updates `@deque/cauldron-react` to play nicely with [React v18 types](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-typescript-definitions).

Part of https://github.com/dequelabs/walnut/issues/3349

### DIFF
--- a/packages/react/src/components/Tooltip/index.tsx
+++ b/packages/react/src/components/Tooltip/index.tsx
@@ -49,7 +49,7 @@ export default function Tooltip({
   hideElementOnHidden = false,
   className,
   ...props
-}: TooltipProps) {
+}: TooltipProps): JSX.Element {
   const [id] = propId ? [propId] : useId(1, 'tooltip');
   const hideTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const [showTooltip, setShowTooltip] = useState(!!initialShow);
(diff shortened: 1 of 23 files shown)


packages/react/src/components/Tooltip/index.tsx
```diff
@@ -178,34 +178,43 @@ export default function Tooltip({
...
-          })}
-          ref={setTooltipElement}
-          role="tooltip"
-          style={styles.popper}
-          {...attributes.popper}
-          {...props}
-        >
-          {variant !== 'big' && (
+  return (
+    <>
```
scurker, line 182: Why the change to fragment here?

packages/react/src/components/TwoColumnPanel/TwoColumnPanel.tsx
```diff
@@ -231,22 +232,24 @@ const TwoColumnPanel = forwardRef<HTMLDivElement, TwoColumnPanelProps>(
...
-          containerElements={[columnLeftRef.current as HTMLElement]}
-        />
-        <ClickOutsideListener
-          onClickOutside={handleClickOutside}
-          target={columnLeftRef.current as HTMLElement}
-        />
-        {isCollapsed ? null : skipLink}
-        {showPanel ? ColumnLeftComponent : null}
-        {ColumnRightComponent}
+        <>
```
scurker, line 235: ?

packages/react/src/utils/polymorphic-type/index.ts
```diff
@@ -51,7 +51,7 @@ interface ForwardRefComponent<
   <As = IntrinsicElementString>(
     props: As extends ''
       ? { as: keyof JSX.IntrinsicElements }
-      : As extends React.ComponentType<infer P>
+      : As extends React.ComponentType<React.PropsWithChildren<infer P>>
```
scurker, line 54: Is this a safe assumption to make? I don't know that the polymorphic type will always be on a component with children.

packages/react/src/components/TwoColumnPanel/TwoColumnPanel.tsx
stephenmathieson: I no longer have the exact error message, but it was something like...

> `<div>` is expecting a single child, but multiple were provided

This was the easiest/less intrusive fix, as it has 0 effect on anything other than TypeScript.

packages/react/src/utils/polymorphic-type/index.ts
stephenmathieson: This was done by the codemod (not by hand). I'll see what happens when I revert this change.

packages/react/src/components/Tooltip/index.tsx
scurker: Interesting, I thought React 18 allowed returning `null` but 🤷

scurker (approved):

PR #691, opened 2022-07-06
feat(styles): implement dark theme and updated design for options menu dropdown

scurker:


packages/styles/dropdown.css
```diff
@@ -1,44 +1,3 @@
...
-  box-shadow: 0px 0px 2px 1px rgba(0, 0, 0, 0.3);
-  border: 1px solid var(--field-border-color);
-  color: var(--text-color-base);
-  display: none;
-  list-style: none;
-}
-
-.Dropdown.Dropdown--active {
-  display: block;
-}
```
thuey, line 14: Should this be removed, too? https://github.com/dequelabs/cauldron/blob/7fd7d095bc8c85c5369371baaf74c7300c578fd2/packages/react/src/components/TopBar/TopBarMenu.tsx#L104

packages/styles/dropdown.css
```diff
@@ -1,44 +1,3 @@
-.Dropdown {
```
thuey, line 1: As far as I can tell, these styles weren't documented anywhere, but is there any concern about this being a breaking change if any code was using `.Dropdown`?

thuey (dismissed):

packages/styles/dropdown.css
scurker: No, this prevents a breaking change for someone using that class.

packages/styles/dropdown.css
scurker: I'm not 100% sure why those styles were even there. They probably should've exclusively been a part of the OptionsMenu styles, but I'm leaving them there for now.

thuey (approved):

PR #697, opened 2022-07-12
fix(react): fixes incorrect PropType and TypeScript interfaces

tbusillo:
This is to fix some issues that I noticed the other day as I was using it, as well as some TypeScript errors being thrown that needed to be fixed (the underlying functionality was fine, this just fixes PropType/TypeScript issues).

### DIFF
--- a/packages/react/src/components/Accordion/Accordion.tsx
+++ b/packages/react/src/components/Accordion/Accordion.tsx
@@ -111,7 +111,9 @@ Accordion.propTypes = {
 
 AccordionTrigger.propTypes = {
   children: PropTypes.node,
-  heading: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6', undefined])
+  heading: PropTypes.shape({
+    level: PropTypes.number
+  })
 };
 
 AccordionContent.propTypes = {
--- a/packages/react/src/components/Accordion/Accordion.tsx
+++ b/packages/react/src/components/Accordion/Accordion.tsx
@@ -9,7 +9,7 @@ import PropTypes from 'prop-types';
 
 export interface AccordionTriggerProps
   extends React.HTMLAttributes<HTMLButtonElement> {
-  children: React.ReactElement;
+  children: React.ReactNode;
   heading?:
     | React.ReactElement
     | {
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -9,7 +9,7 @@ import PropTypes from 'prop-types';
 
 export interface AccordionTriggerProps
   extends React.HTMLAttributes<HTMLButtonElement> {
-  children: React.ReactElement;
+  children: React.ReactNode;
```
tbusillo, line 12: Trigger was not accepting a standalone `string` (e.g., `<AccordionTrigger>Trigger title</AccordionTrigger>`), as I believe `React.ReactElement` is expecting a JSX element whereas `ReactNode` is more inclusive and can be a `string`.

packages/react/src/components/ExpandCollapsePanel/ExpandCollapsePanel.tsx
```diff
@@ -13,7 +13,7 @@ export interface ExpandCollapsePanelProps
   open?: boolean;
   children: React.ReactNode;
   animationTiming?: number | boolean;
-  onToggle: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  onToggle?: (e: React.MouseEvent<HTMLButtonElement>) => void;
```
scurker, line 16: Why is `onToggle` optional? It should be implicitly required.

tbusillo: Is it currently being required? Because you can definitely do an `ExpandCollapse` without passing in `onToggle` which seems to not cause issues since it is being set with a `defaultProp`. 

However, with the `Accordion` component, it is throwing a TypeScript error if you do not explicitly pass in the `onToggle` (below is the error when trying it as not being optional). I guess an alternative solution would be to always pass in the `onToggle` in the accordion component itself? 

`TS2741: Property 'onToggle' is missing in type '{ children: Element[]; }' but required in type 'Pick<AccordionProps, "slot" | "style" | "title" | "children" | "className" | "aria-label" | "defaultChecked" | "defaultValue" | "suppressContentEditableWarning" | ... 246 more ... | "animationTiming">'.`

tbusillo: @scurker Just pinging in case you did not see the above

scurker (approved):

PR #698, opened 2022-07-12
fix(react): only call onClose one per user event

### LINKED ISSUE #694: Escape doesn't close Dialogs/Modals
If you use the toggle syntax shown in the docs, <kbd>escape</kbd> will not close a `<Dialog>`/`<Modal>` because `onClose` fires twice. The close button still works because `onClose` fires three times.

<kbd>escape</kbd> does not close the modals [on the docs site](https://cauldron.dequelabs.com/components/Modal) and I've replicated this in 4 `<Dialog>`s in the `feature-upgrade-cauldron` branch in `axe-mobile-client`.

Code for this screen recording:

```
toggleHelpModal = () => {
    console.log("toggle modal!")
    this.setState(({ showHelpModal }) => {
      return { showHelpModal: !showHelpModal };
    });
  };
```

```
<MenuItem onClick={toggleHelpModal} >
```

```
<Dialog
      show={props.showHelpModal}
      heading={{ text: "Help" }}
      onClose={props.toggleHelpModal}
      className="no-footer"
    >
```

Screen recording showing "toggle modal!" logged in the console once when the menu item is clicked, twice when I hit <kbd>escape</kbd> while the modal remains open, and three times when I hit the close button as the modal closes:

https://user-images.githubusercontent.com/99676396/178051629-e0820344-bf05-4d24-a577-dc85b51370e3.mov

scurker:
BREAKING CHANGE: This makes it so that `onClose` is only called when the element is not controlled. If an author is currently depending on this behavior via a controlled component, `onClose` will no longer be closed when the `show` prop changes from `true` to `false`. 

closes #694


schne324 (dismissed):
the description needs to be updated to match [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) (`BREAKING CHANGE: {description of breaking chage}`)

schne324 (approved):

PR #702, opened 2022-07-19
feat(react,styles): adds toggleable Tag component variant

tbusillo:
ref: https://github.com/dequelabs/axe-extension/issues/2480 (additional comments in issue)

## Other references
- https://preview.uxpin.com/68df515d0023cc4b79f25d43f7663e24e363b1b9#/pages/141465421/simulate/no-panels

### DIFF
--- a/packages/react/src/components/Tag/index.tsx
+++ b/packages/react/src/components/Tag/index.tsx
@@ -24,13 +24,13 @@ const Tag = ({
   children,
   className,
   onToggle,
-  variant,
+  variant = 'default',
   ...other
 }: TagProps) => {
   if (variant === 'toggle') {
     return (
       <button
-        className={classNames('Tag Tag__toggle', className)}
+        className={classNames('Tag', className)}
         {...other}
         onClick={onToggle}
       >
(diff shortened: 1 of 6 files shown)


packages/react/src/components/Tag/index.tsx
```diff
@@ -18,14 +20,37 @@ TagLabel.propTypes = {
...
-  </div>
-);
+const Tag = ({
+  children,
+  className,
+  onToggle,
+  variant = 'default',
+  ...other
+}: TagProps) => {
+  if (variant === 'toggle') {
```
scurker, line 30: I'm wondering if this should instead be `<Button variant="tag">` instead of `<Tag variant="toggle">`, that's similar to how we've done other variants is they're a variant of the semantic component that it's supposed to be.

There's a bit more nuance here because if we're treating this like a switch, we may want to consider having `aria-pressed` or `aria-checked` with `role="switch"` depending on the thing we're toggling. Because the allowed props are going to be dependent on the type of variant, and we need to allow props to be spread to the component. If we use `<Button variant="toggle">` we already have some of that built in for us.

Issue #409, opened 2021-11-09
Update tooltips' colors (backgrounds/borders/shadows)


schne324:
[dark ux pin](https://preview.uxpin.com/4052c8b6f05ce03e28813f28deb3a682950c2ac6#/pages/140808220)
[light](https://preview.uxpin.com/68df515d0023cc4b79f25d43f7663e24e363b1b9#/pages/141465361)

Note: This just tackles the accessible name/text tooltip and the informational tooltip.

scurker:
Make all tooltips wrapped with a `<marquee>` tag.

somaalapati:
@Keerthi-Penukonda Please test this.

Keerthi-Penukonda:
Verified the tooltips in the axe DevTools pro extension. in both dark theme and light theme. Working as expected. 

Dark mode:
![image](https://user-images.githubusercontent.com/40262884/181180046-5b1fa9ee-2f2b-4234-874e-1fbad950457e.png)
Light mode:
![image](https://user-images.githubusercontent.com/40262884/181182095-7dbaa1b7-53d4-47b3-ac27-32474c6fc812.png)

PR #704, opened 2022-07-20
feat(styles): update tooltip styles to support dark theme

thuey:
Closes https://github.com/dequelabs/cauldron/issues/409

### DIFF
--- a/packages/styles/tooltip.css
+++ b/packages/styles/tooltip.css
@@ -1,21 +1,35 @@
 :root {
-  --tooltip-background-color: var(--gray-90);
-  --tooltip-border-color: #fff;
-  --tooltip-text-color: var(--gray-20);
-  --tooltip-info-background-color: var(--gray-20);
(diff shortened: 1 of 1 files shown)


awpearlm:
Looks good!

packages/styles/tooltip.css
```diff
@@ -156,7 +122,7 @@
 }
 
 .TooltipHead {
-  background: var(--tooltip-info-background-color);
+  background: var(--gray-20);
   font-size: 16px;
   color: var(--gray-90);
   padding: 16px 28px;
```
scurker, line 128: Not a blocker since it's not part of your PR:

Only commenting because I noticed, but `font-size` and `padding` should really be using our cauldron vars.

thuey: Created a ticket: https://github.com/dequelabs/cauldron/issues/707. We may want to just fix it when we update the big tooltip to support dark theme, since I'm guessing the designs will change a bit at that time anyway

PR #705, opened 2022-07-20
feat(react,styles): adds thin modifier to pagination

scurker:
this is needed for https://github.com/dequelabs/axe-extension/pull/2636

### DIFF
--- a/packages/react/src/components/Pagination/Pagination.tsx
+++ b/packages/react/src/components/Pagination/Pagination.tsx
@@ -20,6 +20,7 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
   onFirstPageClick?: () => void;
   onLastPageClick?: () => void;
   tooltipPlacement?: Placement;
+  thin?: boolean;
   className?: string;
 }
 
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -20,6 +20,7 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
   onFirstPageClick?: () => void;
   onLastPageClick?: () => void;
   tooltipPlacement?: Placement;
+  thin?: boolean;
```
thuey, line 23: For my own edification, why did you decide to go with a modifier prop, instead of `variant="thin"`?

thuey (approved):

packages/react/src/components/Pagination/Pagination.tsx
scurker: Imagine you could have both a variant AND thin, so you would want to be able to combine both so `thin` acts better as a modifier. This is how we've implemented it for Button, MenuBar, Tabs, etc. There's no variant for pagination right now, but this keeps it in sync with other components and allows for flexibility in the future if we do ever need it.

PR #709, opened 2022-07-21
feat(react): add `question-circle-clear` icon

Zidious:
image


scurker (changes requested):
Just a minor nitpick: 

It looks like we have `info-circle` for the solid icon, and `info-circle-alt` for the stroked component, so we should keep naming consistent here.

PR #713, opened 2022-07-26
fix(react): fix issue where accordion would not pass through props correctly

scurker:
This is needed for https://github.com/dequelabs/axe-extension/issues/2476 and addresses some outstanding issues.

### DIFF
--- a/docs/patterns/components/Accordion/index.js
+++ b/docs/patterns/components/Accordion/index.js
@@ -60,7 +60,7 @@ const AccordionDemo = () => {
           you can do this by setting the open prop and handling changes with
           onToggle.
         </p>
-        <ControlledAccordion label="Accordion #1" heading={{ level: 4 }} />
+        <ControlledAccordion label="Accordion #1" heading={{ level: 4 }} open />
         <ControlledAccordion label="Accordion #2" heading={{ level: 4 }} />
         <ControlledAccordion label="Accordion #3" heading={{ level: 4 }} />
         <Code role="region" tabIndex={0}>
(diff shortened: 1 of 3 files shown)


schne324 (approved):

docs/patterns/components/Accordion/index.js
```diff
@@ -71,15 +71,14 @@ const AccordionDemo = () => {
...
     <Accordion open={open} onToggle={() => setIsOpen(!open)}>
-    <Accordion>
-    <AccordionTrigger heading={{ level: 4 }}>
-      {label}
-    </AccordionTrigger>
+      <AccordionTrigger heading={{ level: 4 }}>
+        {label}
+      </AccordionTrigger>
     <AccordionContent>Here is some content</AccordionContent>
   </Accordion>
```
thuey, line 83: nit: Some of the alignment on this is off:

image

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -81,9 +87,12 @@ const Accordion = ({ children }: AccordionProps) => {
   const elementId = useId();
 
   return (
-    <div className="Accordion">
+    <div className="Accordion" {...props}>
       <ExpandCollapsePanel
         id={panelElement.props.id || `${elementId}-panel`}
+        open={open}
+        onToggle={onToggle}
+        animationTiming={animationTiming}
```
thuey, line 95: We are already spreading `{...panelElement.props}`. Why not have these props be part of the panel's props (`AccordionContentProps`)?

thuey: I realize that the `AccordionProps` extend from `ExpandCollapsePanel`, but since that wasn't actually implemented, I don't think it would be a breaking change to make `AccordionContentProps` extend from `ExpandCollapsePanel` instead

scurker: Because they're 2 different elements. The resulting HTML looks something like this:

```jsx
<div class="Accordion">
  <button>...</button>
  <div class="ExpandCollapse__Panel">...</div>
</div>
```

The accordion props should apply to the `.Accordion` element. The panel props should apply to the `.ExpandCollapse__Panel` element. They need to be able to set html attributes independently of one another.

docs/patterns/components/Accordion/index.js
scurker: I really hate our documentation here. I wish we used markdown or something we could run prettier over.

packages/react/src/components/Accordion/Accordion.tsx
scurker: That's an pre-existing issue, would you mind raising a new issue for that?

thuey (approved):

PR #717, opened 2022-07-27
fix(react): add types to Pagination onClick props

Zidious:
Assign types to the `onClick` props otherwise TS will shout: 

```
Type '(event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void' is not assignable to type '() => void'.ts(2322)
```

### DIFF
--- a/packages/react/src/components/Pagination/Pagination.tsx
+++ b/packages/react/src/components/Pagination/Pagination.tsx
@@ -15,10 +15,18 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
   previousPageLabel?: string;
   nextPageLabel?: string;
   lastPageLabel?: string;
-  onNextPageClick?: () => void;
-  onPreviousPageClick?: () => void;
-  onFirstPageClick?: () => void;
-  onLastPageClick?: () => void;
+  onNextPageClick?: (
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>
+  ) => void;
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -15,10 +15,18 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
   previousPageLabel?: string;
   nextPageLabel?: string;
   lastPageLabel?: string;
-  onNextPageClick?: () => void;
-  onPreviousPageClick?: () => void;
-  onFirstPageClick?: () => void;
-  onLastPageClick?: () => void;
+  onNextPageClick?: (
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>
```
scurker, line 19: You should just be able to use `React.MouseEvent<HTMLButtonElement>` here (and for the other functions). React defaults the [second generic] to `NativeMouseEvent` by default.

```suggestion
    event: React.MouseEvent<HTMLButtonElement>
```

scurker (changes requested):
This should be `fix` instead a `chore` since it contains a functional code change.

packages/react/src/components/Pagination/Pagination.tsx
Zidious: Whoops, will update

scurker (approved):

PR #724, opened 2022-08-03
feat(react): allow Toasts to not be dismissible

thuey:
Needed for https://github.com/dequelabs/axe-extension/issues/2621

### DIFF
--- a/packages/react/src/components/Toast/index.tsx
+++ b/packages/react/src/components/Toast/index.tsx
@@ -12,6 +12,7 @@ export interface ToastProps {
   toastRef: React.Ref<HTMLDivElement>;
   focus?: boolean;
   show?: boolean;
+  dismissible?: boolean;
   children: React.ReactNode;
 }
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Toast/index.tsx
```diff
@@ -105,6 +109,7 @@ export default class Toast extends React.Component<ToastProps, ToastState> {
       toastRef,
       focus,
       show,
+      dismissible,
```
stephenmathieson, line 112: It took me a minute to see that the default is `true`. I haven't seen/used `defaultProps` (or `class` components) in a really, really long time 😅

stephenmathieson (approved):

packages/react/src/components/Toast/index.tsx
thuey: Yep, haha. It's fun updating things in three different places.

PR #725, opened 2022-08-04
fix(styles): fix accordion background colors

scurker:
kudos to @tbusillo for noticing this

The recent cleanup with #723 missed that buttons needed a more specific selector because they also inherit from `.ExpandCollapsePanel__trigger`. Including the fallback selector will create a higher specificity while allowing the existing classname to be extended.

### DIFF
--- a/packages/styles/accordion.css
+++ b/packages/styles/accordion.css
@@ -29,8 +29,9 @@
   all: unset;
 }
 
-.Accordion__trigger {
-  background: var(--accordion-trigger-background-color);
+.Accordion__trigger,
+button.Accordion__trigger {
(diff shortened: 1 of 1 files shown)


thuey (approved):

packages/styles/accordion.css
```diff
@@ -29,8 +29,9 @@
   all: unset;
 }
 
-.Accordion__trigger {
-  background: var(--accordion-trigger-background-color);
+.Accordion__trigger,
+button.Accordion__trigger {
```
thuey, line 33: A comment would be nice for posterity

PR #728, opened 2022-08-09
fix(react): fix issue where ref property was getting clobberd on <Tab>

scurker:
This fixes an issue where `<Tab ref={tabRef}>` was always `null` because it was getting clobbered within `cloneElement`.

### DIFF
--- a/packages/react/src/components/Tabs/Tabs.tsx
+++ b/packages/react/src/components/Tabs/Tabs.tsx
@@ -32,7 +32,6 @@ const Tabs = ({
 }: TabsProps): JSX.Element => {
   const [activeIndex, setActiveIndex] = useState(initialActiveIndex);
   const tabsRef = useRef<HTMLDivElement>(null);
-  const tabsListRef = useRef<HTMLUListElement>(null);
 
   const tabs = React.Children.toArray(children).filter(
     child => (child as React.ReactElement<any>).type === Tab
--- a/packages/react/src/components/Tabs/Tabs.tsx
+++ b/packages/react/src/components/Tabs/Tabs.tsx
@@ -131,7 +131,7 @@ const Tabs = ({
 
   useDidUpdate(() => {
     const activeTab = tabsRef.current?.querySelector(
-      '[role="tablist"] > [aria-selected="true"]'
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -122,7 +122,6 @@ const Tabs = ({
       tabIndex: index === activeIndex ? 0 : -1,
       ['aria-controls']: target.current?.id,
       ['aria-selected']: selected,
-      ref: index === activeIndex ? focusedTabRef : null,
```
scurker, line 125: This was clobbering the ref, in reality we don't really need it when we can utilize `[aria-selected="true"]`

thuey: I believe what you have implemented here should work. However, I have found relying on query selectors in React components to be somewhat finicky. For example, sometimes the DOM isn't in the state you expect it to be in. Or, the structure changes and your query selectors are now incorrect.

fwiw, it should be possible to get this working closer to the way it was originally implemented. See: https://github.com/facebook/react/issues/8873#issuecomment-275423780 and https://github.com/facebook/react/issues/8873#issuecomment-489579878.

```suggestion
      ref: index === activeIndex ? (node) => {
        focusedTabRef.current = node;
        if (typeof child.ref === 'function') {
          child.ref(node)
        } else if (child.ref !== null && child.ref.hasOwnProperty('current')) {
          child.ref.current = node
        }
      } : child.ref
```

thuey (dismissed):
Unused variable should be removed. The other comment is just a suggestion

packages/react/src/components/Tabs/Tabs.tsx
scurker: The query selector shouldn't need to change unless something were to change with the WCAG spec respective of how tabs are implemented. I feel that it's much safer to rely on query here rather than trying to patch into whatever React uses for refs.

scurker: Additionally, we don't need to clone the ref for every item in the list - we only need the `[aria-selected="true"]` one and the query does exactly that.

thuey (approved):

PR #732, opened 2022-08-10
feat(Pagination): add overload of statusLabel prop to take in function exposing totalItems, itemStart, itemEnd

longmatthewh:
closes: #733

### DIFF
--- a/packages/react/src/components/Pagination/Pagination.tsx
+++ b/packages/react/src/components/Pagination/Pagination.tsx
@@ -6,27 +6,25 @@ import IconButton from '../IconButton';
 import TooltipTabstop from '../TooltipTabstop';
 import Icon from '../Icon';
 
+export type StatusLabeller = (
+  totalItems: number,
+  itemStart: number,
+  itemEnd: number
+) => React.ReactNode;
+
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -6,27 +6,25 @@ import IconButton from '../IconButton';
 import TooltipTabstop from '../TooltipTabstop';
 import Icon from '../Icon';
 
+export type StatusLabeller = (
+  totalItems: number,
+  itemStart: number,
+  itemEnd: number
+) => React.ReactNode;
```
scurker, line 13: My preference here would still be to allow for component props to be used here as opposed to a pure function.

```ts
type StatusLabeler = ({
  totalItems: number,
  itemStart: number,
  itemEnd: number
}) => React.ReactNode;
```

I feel that props gives greater flexibility since it's a little easier to destructure props by name as opposed to the property order in the function.

scurker:
Since this PR hasn't been updated in a bit, I'm closing it. Please re-open and update if the changes are still needed.

PR #736, opened 2022-08-16
fix(react): ensure check-circle icon meets color-constrast requirements

Zidious:
ensure the check-circle background color defaults to white. to avoid color-contrast issues, currently when used with `RadioCardGroup` we do not meet the 3 : 1 requirements: 

image

After: 
image

image

### DIFF
--- a/packages/react/src/components/Icon/icons/check-circle.svg
+++ b/packages/react/src/components/Icon/icons/check-circle.svg
@@ -1,1 +1,8 @@
-<svg version="1.2" xmlns="http://www.w3.org/2000/svg" overflow="visible" preserveAspectRatio="none" viewBox="0 0 24 24" width="20" height="20"><g><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm-2 15l-5-5 1.41-1.41L10 14.17l7.59-7.59L19 8l-9 9z" vector-effect="non-scaling-stroke" fill="currentColor" /></g></svg>
+<svg version="1.2" xmlns="http://www.w3.org/2000/svg" overflow="visible" preserveAspectRatio="none"
+	 viewBox="0 0 44 44" vector-effect="non-scaling-stroke" fill="currentColor">
+<style type="text/css">
+	.st1{fill:#FFFFFF;}
+</style>
+<circle id="body" cx="22" cy="22" r="18.3"/>
+<polygon id="check" class="st1" points="34.8,14.7 18.3,31.2 9.2,22 11.8,19.4 18.3,26 32.2,12.1"/>
+</svg>
--- a/packages/react/src/components/Icon/icons/check-circle.svg
+++ b/packages/react/src/components/Icon/icons/check-circle.svg
@@ -1,8 +1,1 @@
-<svg version="1.2" xmlns="http://www.w3.org/2000/svg" overflow="visible" preserveAspectRatio="none"
-	 viewBox="0 0 44 44" vector-effect="non-scaling-stroke" fill="currentColor">
-<style type="text/css">
-	.st1{fill:#FFFFFF;}
-</style>
-<circle id="body" cx="22" cy="22" r="18.3"/>
-<polygon id="check" class="st1" points="34.8,14.7 18.3,31.2 9.2,22 11.8,19.4 18.3,26 32.2,12.1"/>
-</svg>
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Icon/icons/check-circle.svg
```diff
@@ -1 +1,8 @@
-<svg version="1.2" xmlns="http://www.w3.org/2000/svg" overflow="visible" preserveAspectRatio="none" viewBox="0 0 24 24" width="20" height="20"><g><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm-2 15l-5-5 1.41-1.41L10 14.17l7.59-7.59L19 8l-9 9z" vector-effect="non-scaling-stroke" fill="currentColor" /></g></svg>
\ No newline at end of file
+<svg version="1.2" xmlns="http://www.w3.org/2000/svg" overflow="visible" preserveAspectRatio="none"
+	 viewBox="0 0 44 44" vector-effect="non-scaling-stroke" fill="currentColor">
+<style type="text/css">
+	.st1{fill:#FFFFFF;}
+</style>
```
thuey, line 5: This breaks dark mode: 
image

For backwards compatibility and dark mode, we should not fill the check by default. Instead, the radio card group should fill the check by targeting the `polygon` element

packages/react/src/components/Icon/icons/check-circle.svg
```diff
@@ -1 +1,8 @@
...
+<circle id="body" cx="22" cy="22" r="18.3"/>
+<polygon id="check" class="st1" points="34.8,14.7 18.3,31.2 9.2,22 11.8,19.4 18.3,26 32.2,12.1"/>
```
thuey, line 7: These ID's will not be unique on the page and should be removed

thuey (dismissed):

packages/react/src/components/Icon/icons/check-circle.svg
Zidious: We may need to think about giving this a default colour, removing all styling: 

image

image

thuey: Gotcha. Alternatively, we could  add a new Icon type "check-circle-solid" (similar to how we have other "solid" icon variants). We would then have light and dark theme stylings for it to fill in the checkmark appropriately.

Although I just noticed `cardIcon` is passed in to the RadioCardGroup as a string, which makes it tricker to do this in a forwards-compatible way.

I'm curious to get @scurker's thoughts on this.

scurker: I don't think the problem is inherent to the icon here. Cauldron icons don't tend to have a "default" color, but rather inherit the color from the text.

The underlying issue to me seems to be is in how the icon is used and the color. Green is notorious for being difficult to pass color contrast on light backgrounds, so my first suggestion would be to ask if we can apply a darker color for the icon in the RadioCardGroup component.

Zidious: Thoughts on this colour? 

`color: #4f990f` - I suspect we'll need a CSS variable if we go  down right route?

image
image

thuey: > my first suggestion would be to ask if we can apply a darker color for the icon in the RadioCardGroup component

Originally, I was trying to suggest a way to avoid introducing another color by making it so the checkmark could be made white. However, after becoming more familiar with how we handle iconography in cauldron, I'm inclined to agree that just using a darker color is the better option. Sorry to lead you astray, @Zidious

scurker: > Thoughts on this colour?

It looks like it's passing 3:1, although I would double check with @awpearlm to get him to sign off.

awpearlm: Its fine to change the color, just make it a one-off rather than some color that's gonna find it's way into Cauldron...at least for now. The Mobile team should be notified of the change so they can change their respective icon to be that color too @karnett @ryomtoob

ryomtoob: Thanks for bringing this to my attention @awpearlm. The mobile workflow is using the `<RadioCardGroup>` with `cardIcon=check-circle`. https://github.com/dequelabs/walnut/blob/develop/app/src/modules/axe-devtools-mobile/components/ChooseOS.tsx#L30-L53
I think that means when this this PR gets merged, released, and adopted into walnut, it should be fixed for us too.

packages/styles/radio-card.css
```diff
@@ -43,7 +43,7 @@
   min-height: 44px;
 }
 .RadioCardGroup__Icon.Icon svg {
-  color: var(--accent-success-dark);
+  color: #4f990f;
```
thuey, line 46: A comment here for posterity would be helpful to explain why this one-off exists.

thuey (approved):

Issue #739, opened 2022-08-19
Footer Hacktoberfest 22


abbeyperini-deque:
A footer, with link back to [deque.com](http://deque.com/) and a link to the open-source terms surrounding this project.

PR #745, opened 2022-08-19
docs: footer

abbeyperini-deque:
Marketing request for Hacktoberfest: A footer, with link back to [deque.com](http://deque.com/) and a link to the open-source terms surrounding this project.

- create component with footer tag
- add copyright, link to deque.com, link to terms of use, and link to MPL-2.0 License
- position sidebar and layout so they're not blocked by the footer

Issue #739

### DIFF
--- a/docs/Demo/Footer.js
+++ b/docs/Demo/Footer.js
@@ -2,11 +2,13 @@ import React from 'react';
 import { Link } from '@deque/cauldron-react/';
 
 const Footer = props => {
+  const year = new Date().getFullYear();
+
   return (
     <footer className="Footer">
       <ul className="Footer-links">
         <li>
-          Copyright © 2022{' '}
+          Copyright © {year}{' '}
           <Link href="https://deque.com/">Deque Systems Inc.</Link>
         </li>
         <li>
--- a/docs/Demo/Footer.js
+++ b/docs/Demo/Footer.js
@@ -13,7 +13,7 @@ const Footer = props => {
           <Link href="https://www.deque.com/terms-of-use/">Terms of Use</Link>
         </li>
         <li>
-          <Link href="https://www.mozilla.org/en-US/MPL/2.0/ ">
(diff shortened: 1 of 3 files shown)


docs/Demo/Footer.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import { Link } from '@deque/cauldron-react/';
+
+const Footer = props => {
+  return (
+    <footer className="Footer">
+      <ul className="Footer-links">
+        <li>
+          Copyright © 2022{' '}
```
scurker, line 9: Should we not calculate the year? i.e.

```tsx
const year = new Date().getFullYear()
return (
   ...
   <li>
     Copyright © {year}
   </li>
)
```

docs/index.css
```diff
@@ -46,3 +46,35 @@ code {
 .Demo ul {
   margin: var(--space-large);
 }
+
+.Footer {
+  position: fixed;
+  bottom: 0;
+  width: 100%;
+  height: var(--top-bar-height);
```
scurker, line 54: I'm not sure we need the footer visible at all times. Can it not just be a footer inside of the workspace section?

scurker (changes requested):

docs/index.css
abbeyperini-deque: `<footer>` can't go inside `<main>`, so I can't make it a child of `<Workspace>`. I can style it that way, but I thought it looked weird when I tried it initially.

Full width:
Screen Shot 2022-08-23 at 10 25 11 AM

Workspace width:
Screen Shot 2022-08-23 at 10 35 15 AM

docs/Demo/Footer.js
abbeyperini-deque: These changes have been pushed!

docs/index.css
karnett: eek lol - I like it full width personally

PR #746, opened 2022-08-19
docs: new homepage

abbeyperini-deque:
Cauldron docs updates requested by Marketing added in this PR:

Must do to start promoting Cauldron:
----
- New cauldron homepage, describing: 
  - The purpose of the resource, 
  - Guidelines for how to use it, 
  - Who made it (Deque), 
  - A path to submit contributions and bugs, 
  - A call to action to try Deque tools to test for a11y. 
- This homepage also needs a better page title and meta description, centered around "accessible components/component library/pattern library" to increase its findability in search engines.
- A footer, with link back to [deque.com](http://deque.com/) and a link to the open-source terms surrounding this project.

Nice to haves for '22
----
- Each component also needs a better page title and meta description to increase findability chances in search. Example: https://cauldron.dequelabs.com/components/Accordion should show up in search results when people look for "accessible accordion component" or "accessible accordion pattern."
- Updated favicon


docs/Home/index.js
```diff
@@ -1,14 +1,26 @@
...
 const Home = () => (
   <div>
-    <h1>Cauldron React</h1>
+    <h1>Cauldron React by Deque Systems</h1>
+    <h2>Purpose</h2>
+    <p>
+      These packages include everything from typography and colors to components
+      like custom form controls. The design and interactions are intended to
+      show how Deque would solve these problems. It's how we approach them in
+      our prouducts.
```
karnett, line 12:
```suggestion
      These packages include everything from typography and colors to components, like custom form controls. Considerations in design and interactions show how Deque would solve these problems in our products.
```

Also expand on 'these problems' -- what problems? :)

docs/Home/index.js
```diff
@@ -1,14 +1,28 @@
...
-import { Code } from '@deque/cauldron-react';
+import { Code, Link } from '@deque/cauldron-react';
 
 const Home = () => (
   <div>
-    <h1>Cauldron React</h1>
+    <h1>Cauldron React: Accessible Components Library</h1>
+    <h2>Purpose</h2>
+    <p>
+      Friends don't let friends ship inaccessible code! These accessible
```
stephenmathieson, line 9:
```suggestion
      Friends don’t let friends ship inaccessible code! These accessible
```

docs/Home/index.js
```diff
@@ -23,6 +37,25 @@ const Foo = () => (
...
+    </p>
+    <p>
+      Found a bug?{' '}
+      <Link href="https://github.com/dequelabs/cauldron/issues/new/choose">
+        Open an issue!
+      </Link>
+    </p>
+    <h2>More Accessibility Tools</h2>
+    <p>
+      Don't stop here! Check out{' '}
```
stephenmathieson, line 55:
```suggestion
      Don’t stop here! Check out{' '}
```

scurker (commented):
Is the footer supposed to be fixed or just be on the bottom of the page? It just looks kind of funky always being visible.

abbeyperini-deque:
@scurker I agree that it doesn't need to always be showing. We need a footer that isn't blocking content. The problem I can't figure out how to get around is the layout component is fixed, so I can't put the footer in normal flow below it, it also has to be fixed.

scurker:
@abbeyperini-deque Can the footer not just be included as a child of `<Workspace>`?

https://github.com/dequelabs/cauldron/blob/26021aeac4b4ded346950ae5e2564b22cd76bf54/docs/index.js#L244-L248

Example:

```jsx
<Workspace>
  <Route>...</Route>
  <Footer />
</Workspace>
```

abbeyperini-deque:
Yeah footer shouldn't be a child of main, and 3/4 of the page looks even weirder to me. 😬

stephenmathieson (changes requested):
We have a footer pattern in UX Pin. We should follow it rather than creating something else.

https://preview.uxpin.com/4052c8b6f05ce03e28813f28deb3a682950c2ac6#/pages/140808156/simulate/no-panels

abbeyperini-deque:
Styling updated and it looks *much* better!

abbeyperini-deque:
Is there a light mode UXPin? Is there a UXPin repository somewhere?

karnett:
Aaron signed off on the footer

docs/Demo/Footer.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Link } from '@deque/cauldron-react/';
+
+const Footer = props => {
+  const year = new Date().getFullYear();
+
+  return (
+    <footer
+      className={
+        props.theme === 'dark' ? 'cauldron--theme-dark Footer' : 'Footer'
```
scurker, line 10: Why is this needed? The body class should already have the `cauldron--theme-*` class

docs/index.css
```diff
@@ -46,3 +46,55 @@ code {
...
+  position: fixed;
+  bottom: 0;
+  width: 100%;
+  flex: none;
+  flex-direction: row;
+  box-sizing: border-box;
+  z-index: var(--z-index-sidebar);
+  display: flex;
+  align-items: center;
+  height: 50px;
```
scurker, line 60: Fixed heights can cause issues with zooming/variable text sizes. Instead, it would be better to only set the appropriate padding so the element scales with the text OR set the size relative to the font size.

docs/index.css
```diff
@@ -46,3 +46,55 @@ code {
...
+  border-width: 1px 0 0 0;
+  border-style: solid;
+  border-color: #b3bfc6;
+  background-color: var(--white);
```
scurker, line 64: I was confused why this was here for a second, and then I realized this was the background color for the light theme.

A cleaner way to do this would be to set vars for each theme and utilize them in the styles:

```css
.cauldron--theme-dark .Footer {
  --footer-background-color: var(--accent-medium);
}

.cauldron--theme-light .Footer {
  --footer-background-color: #fff;
}

.Footer {
  background-color: var(--footer-background-color);
}
```

The var will cascade down appropriately based on the current "theme". This would be recommended for any style that's theme specific.

docs/index.css
```diff
@@ -46,3 +46,55 @@ code {
...
 }
+
+.Footer {
+  position: fixed;
+  bottom: 0;
+  width: 100%;
+  flex: none;
+  flex-direction: row;
+  box-sizing: border-box;
+  z-index: var(--z-index-sidebar);
```
scurker, line 57: Do you know why this is needed?

docs/index.css
```diff
@@ -46,3 +46,55 @@ code {
...
+  border-width: 1px 0 0 0;
+  border-style: solid;
+  border-color: #b3bfc6;
+  background-color: var(--white);
+  background-size: cover;
+}
+
+.Footer-links {
+  display: flex;
+  flex-direction: row;
```
scurker, line 70: Nit: this is the default value
```suggestion
```

scurker (changes requested):

docs/Demo/Footer.js
abbeyperini-deque: Good catch!

docs/index.css
abbeyperini-deque: 😬 I agree, but this came from the UXPin.

docs/index.css
abbeyperini-deque: Leftover from when I was trying to avoid `position: fixed`. Good catch!

docs/index.css
abbeyperini-deque: Could this be a different PR? I was trying to avoid editing anything outside the docs, but now that we know there is a UXPin that includes a footer, we could make an issue to make it one of the Cauldron components.

docs/index.css
scurker: You can still achieve the same height relative to the text by using the appropriate top/bottom padding to still end up with 50px.

abbeyperini-deque: 🤔 The TopBar height is fixed too. I have been repeatedly told to follow the UXPin - Would it be worth working with Design to get a responsive design and work on that in another PR/issue?

docs/Home/index.js
```diff
@@ -23,6 +37,25 @@ const Foo = () => (
 );
       `}
     </Code>
+    <h2>Contribute</h2>
+    <p>
+      Check out our{' '}
+      <Link href="https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md">
+        contribution guide!
+      </Link>
```
anastasialanz, line 45: We could use an icon with `type="external-link"` here.

```suggestion
      <Link href="https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md">
        <Icon type="external-link" />
        contribution guide!
      </Link>
```

https://cauldron.dequelabs.com/components/Icon

docs/Home/index.js
```diff
@@ -23,6 +37,26 @@ const Foo = () => (
...
+    <p>
+      Found a bug?{' '}
+      <Link href="https://github.com/dequelabs/cauldron/issues/new/choose">
+        Open an issue!
+      </Link>
+    </p>
+    <h2>More Accessibility Tools</h2>
+    <p>
+      Don’t stop here! Check out{' '}
+      <Link href="https://www.deque.com/axe/">Deque Systems' axe Tools</Link>{' '}
```
anastasialanz, line 57: There are two more external links here.

docs/Demo/Footer.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { Link } from '@deque/cauldron-react/';
+
+const Footer = props => {
+  const year = new Date().getFullYear();
+
+  return (
+    <footer className="Footer">
+      <ul className="Footer-links">
```
anastasialanz, line 9: If we are going with BEM, since `links` is an element


```suggestion
      <ul className="Footer__links">
```

docs/Demo/Footer.js
```diff
@@ -0,0 +1,27 @@
...
+        </li>
+        <li>
+          <Link href="https://www.mozilla.org/en-US/MPL/2.0/">
+            MPL-2.0 License
+          </Link>
+        </li>
+      </ul>
+      <div className="Copyright">
+        Copyright © {year}{' '}
+        <Link href="https://deque.com/">Deque Systems Inc.</Link>
```
anastasialanz, line 21: On the main footer for deque.com, it says "Deque Systems, Inc." I'm not sure which way is right.

It seems like your way is right.
https://grammarhow.com/comma-before-or-after-inc-company-name/

anastasialanz: I think it depends on how the company is registered. I'm really into the details here. 😆

docs/index.js
```diff
@@ -245,6 +253,7 @@ const App = () => {
           id="main-content"
           workspaceRef={el => setWorkspace}
           tabIndex={-1}
+          className="Layout-with-footer"
```
anastasialanz, line 256: Since this a modifier of `Layout`, I suggest this:
```suggestion
          className="Layout--with-footer"
```

docs/index.js
```diff
@@ -225,7 +229,11 @@ const App = () => {
             </TopBarItem>
           </MenuBar>
         </TopBar>
-        <SideBar show={state.show} onDismiss={onTriggerClick}>
+        <SideBar
+          show={state.show}
+          onDismiss={onTriggerClick}
+          className="SideBar-with-footer"
```
anastasialanz, line 235: Since this a modifier of `SideBar`, I suggest this:
```suggestion
          className="SideBar--with-footer"
```

docs/index.css
anastasialanz: When the browser is used at a smaller screen size like `360` width, the height of the `.Footer-links` is taller than the `height: 50px` `.Footer` container. I think you could remove the explicit height and it would be fine?

docs/index.css
```diff
@@ -46,3 +46,51 @@ code {
...
+  position: fixed;
+  bottom: 0;
+  width: 100%;
+  box-sizing: border-box;
+  display: flex;
+  align-items: center;
+  height: 50px;
+  border-width: 1px 0 0 0;
+  border-style: solid;
+  border-color: #b3bfc6;
```
anastasialanz, line 60: Nit:
```suggestion
  border-top: 1px solid #b3bfc6;
```

docs/Home/index.js
```diff
@@ -23,6 +37,30 @@ const Foo = () => (
...
+      <Link href="https://github.com/dequelabs/cauldron/issues/new/choose">
+        Open an issue!
+        <Icon type="external-link" />
+      </Link>
+    </p>
+    <h2>More Accessibility Tools</h2>
+    <p>
+      Don’t stop here! Check out{' '}
+      <Link href="https://www.deque.com/axe/">
+        Deque Systems' axe Tools
```
anastasialanz, line 59:
```suggestion
        Deque Systems’ axe Tools
```

docs/index.css
```diff
@@ -46,3 +46,51 @@ code {
 .Demo ul {
   margin: var(--space-large);
 }
+
+.Footer {
+  position: fixed;
+  bottom: 0;
+  width: 100%;
+  box-sizing: border-box;
```
anastasialanz, line 54: This can be removed since `box-sizing` is inherited 
Screen Shot 2022-09-15 at 12 38 23 PM

docs/Demo/Footer.js
```diff
@@ -0,0 +1,27 @@
...
+
+const Footer = props => {
+  const year = new Date().getFullYear();
+
+  return (
+    <footer className="Footer">
+      <ul className="Footer__links">
+        <li>
+          <Link href="https://www.deque.com/terms-of-use/">Terms of Use</Link>
+        </li>
```
anastasialanz, line 12: I'm wondering if we should use external icons or the words "external" here, but I'm not sure how strict we need to be on this.

This resources says, "A link to an external site MAY indicate that it leads to an external site.": https://dequeuniversity.com/checklists/web/links

docs/Demo/Footer.js
abbeyperini-deque: I'll match the main website - that seems best.

docs/index.css
anastasialanz: I like @scurker's suggestion. I guess it's up to you if you wanted to do it here or put it in another PR.

docs/Demo/Footer.js
abbeyperini-deque: 🤔 I think the main thing to warn users about links is if it opens in a new tab, and these open in the same tab?

docs/index.css
anastasialanz: To get it to be `50px` height at a larger browser width, it would be `padding: 0.359375rem 0;.

docs/Demo/Footer.js
anastasialanz: [This example at the bottom](https://dequeuniversity.com/class/semantic-structure/links/external-windows-files#external-site) doesn't have it opening in a new tab but indicates that it's external.

Screen Shot 2022-09-15 at 12 59 35 PM

anastasialanz: And yeah that same URL above has "SHOULD" for "A link that opens in a new window or tab SHOULD indicate that it opens in a new window or tab."

docs/index.js
abbeyperini-deque: Unused class I left in - removing it

docs/index.css
abbeyperini-deque: 🤔 sidebar breaks into it at that width when using `0.36rem` padding

Screen Shot 2022-09-15 at 3 09 59 PM

docs/index.css
abbeyperini-deque: 🤣 I misunderstood. I thought this meant add variables to the Cauldron package, not in the same file.

docs/index.css
anastasialanz: The `bottom` property would need to adjust for the `.Sidebar-with-footer` since the footer needs to be taller at that screen width.

abbeyperini-deque:
- CSS naming
- removed unnecessary CSS
- moved Footer file into Home directory instead of Demo
- restyled links in homepage, because the `div` Icon creates can't be a child of `p`
- all the external link icons
- footer theme variables
- punctuation

docs/index.css
abbeyperini-deque: I was hoping I would get more math from you 😆

abbeyperini-deque: Added!

abbeyperini-deque:
OK that should be all the changes - let me know if I missed any

docs/Home/index.js
```diff
@@ -38,29 +38,37 @@ const Foo = () => (
...
-    </p>
+    <ul>
+      <li className="is--flex-row">
+        <p>Check out our</p>
+        <Link
+          className="Home__links"
+          href="https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md"
+        >
+          contribution guide!
+          <Icon type="external-link" />
```
anastasialanz, line 49: I think we also want to add a label so that screen reader users who can't see the icon know it's an external link.

`<Icon type="external-link" label="link leads to external site" />`

or 

`<Icon type="external-link" label="external site" />`

`

> It can be helpful to users to know when a link will take them away from the current web site to a different web site. Including a text indication (e.g. "link leads to an external site") or an icon with equivalent alt text are two of the most common ways to indicate this. Ensure the indication is available to both sighted users and blind screen reader users.

docs/index.css
anastasialanz: I don't see the 1px gray border on top of footer when in the sidebar area.

Screen Shot 2022-09-15 at 3 29 05 PM

Can we switch the bottom rem to `bottom: 3.5325rem;`?

Screen Shot 2022-09-15 at 3 31 02 PM

docs/index.css
```diff
@@ -99,3 +99,24 @@ code {
...
+    bottom: 5.7rem;
+  }
+}
+
+@media (max-width: 300px) {
+  .SideBar--with-footer,
+  .Layout {
+    bottom: 6.8rem;
+  }
+}
```
anastasialanz, line 122: What do you think about using a mobile-first approach and using `min-width` for each media query as the screen size increases?

Using `max-width` media queries is more desktop-first.

> A mobile-first can usually be spotted by it’s use of min-width instead of max-width. If we start with min-width, we’re essentially saying, “hey, browser, start here and work up.” On the flip side, max-width is sort of like prioritizing larger screens.

https://css-tricks.com/a-complete-guide-to-css-media-queries/

I tried this out and I think it works alright. This also removes one extra media query.

```
.SideBar--with-footer,
.Layout {
  bottom: 5.72rem;
}

@media (min-width: 371px) {
  .SideBar--with-footer,
  .Layout {
    bottom: 4.62625rem;
  }
}

@media (min-width: 591px) {
  .SideBar--with-footer,
  .Layout {
    bottom: 3.5325rem;
  }
}
```

https://user-images.githubusercontent.com/3081483/190518400-9a0c77b9-2b8e-4a10-acf2-57c1aba85018.mov

docs/Home/index.js
abbeyperini-deque: Used `aria-label="{link text} - opens an external site"`

docs/index.css
abbeyperini-deque: You right. You right. I should have done mobile-first.

I'm keeping below 300px because I'm that way and design mentioned responsive down to like 250px when I was working on a different project, though.

anastasialanz: Gotta watch out for those people on their apple watches using the pattern library site!

abbeyperini-deque: I'm here for our developers on the go.

docs/index.css
```diff
@@ -46,3 +46,77 @@ code {
...
+.cauldron--theme-dark .Footer {
+  --footer-background-color: var(--accent-medium);
+  --footer-border-color: var(--stroke-dark);
+}
+
+.Home__links {
+  margin: 2px 0 0 3px;
+}
+
+@media (min-width: 301px) {
```
anastasialanz, line 103: Love a min-width mq ❤️

anastasialanz (approved):
👏🏻 for having a homepage and better SEO! Thanks for bearing with all the comments on this one. I'm sure we can come back to it and improve some things but it's in a good enough spot for now.

Issue #726, opened 2022-08-05
Tooltip arrow offset when zoomed


scurker:
![cauldron tooltip with arrow slightly offset from the edge of the tooltip border](https://user-images.githubusercontent.com/1062039/183112028-b8550ac2-b157-410a-a347-d8cc3d7bccac.png)

When the user zooms, it appears that the arrow pointing back to the targeted element is offset incorrect (I noticed this specifically at 125% or 175% zoom). The tooltip arrow should be inline with the borders of the tooltip.

somaalapati:
@thuey Can you please provide some QA notes on how & where to test this?

scurker:
@somaalapati Take a look at the screenshot above, the tooltip arrow is very slightly misaligned when zoomed or on some retina devices. The tooltip arrow should always be in line with the borders of the tooltip no matter the zoom/pixel device ratio.

somaalapati:
@Keerthi-Penukonda Please test this.

Keerthi-Penukonda:
This issue is fixed.

At 125% Zoom
![image](https://user-images.githubusercontent.com/40262884/187851220-b1c7684f-0c73-4014-bcc0-b37b6dadf5c7.png)

At 175% Zoom
![image](https://user-images.githubusercontent.com/40262884/187851373-7acaacb7-da2b-4058-8a4c-3969f9917575.png)

At 250% Zoom
![image](https://user-images.githubusercontent.com/40262884/187851505-8b99b931-cc2f-4ce8-ae90-11ca3604f754.png)

PR #750, opened 2022-08-26
fix(styles): fix tooltip arrow positioning when zoomed

thuey:
Closes https://github.com/dequelabs/cauldron/issues/726

### DIFF
--- a/packages/styles/tooltip.css
+++ b/packages/styles/tooltip.css
@@ -53,16 +53,17 @@
   width: 0.1px;
 }
 
+/* Adjust position to try to center the arrow in the tooltip's border */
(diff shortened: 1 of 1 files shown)


packages/styles/tooltip.css
```diff
@@ -53,16 +53,17 @@
...
-  top: 0;
+  top: -1px;
 }
 
 [class*='Tooltip--left'] .TooltipArrow {
-  right: 0;
+  right: -1px;
 }
 
 [class*='Tooltip--right'] .TooltipArrow {
```
thuey, line 69: For some reason, Tooltip--right didn't need to shift at all

scurker (approved):

PR #770, opened 2022-09-16
fix(docs): Ensures alert buttons do not overlap

### LINKED ISSUE #769: Alert demo page: triggers overlap at reflow.
## Description
Content overlaps other content at reflow.
Note: For reflow, resize the browser viewport to 640px width by 512px height and zoom in to 200%

## Element: 
-Default Alert 
-Warning Alert

**Location**: Below Demo

**Note**: Button overlaps at the reflow checkpoint
Recommendation to fix
**RULE**: Content MUST be presented without loss of information or functionality when the viewport is set to 320 CSS pixels wide.

## HOW TO FIX: 

## REFERENCE: 
WCAG Understanding document:

## BACKGROUND: 
People with low vision often read content on a desktop computer by scaling content using the browser zoom - up to 400%. When zooming causes content to overlap other content, users with low vision may not be able to see or use that content. When content is created so that it reflows without losing information or functionality at screen widths down to 320px, users can view a web page in a 1280px width browser screen and zoom the content to 400% (320px) without having to scroll side to side to read and without losing access to content.

**selector**: #main-content > div > button:nth-child(3)
**source**: 

```html
<button type="button" class="Button--primary">Default Alert</button>
```

schne324:
Closes #769

## Before:

buttons overlapping

## After:

buttons not overlapping


scurker (dismissed):
This is fine to fix the accessibility issue, but I was hoping we could improve this experience within cauldron itself. 🤷

anastasialanz:
Can we left align both buttons at this size?

schne324:
> This is fine to fix the accessibility issue, but I was hoping we could improve this experience within cauldron itself. 🤷

Sure, that's fine. I was just trying to knock out as many vpat issues as I could in a 30 min span.

schne324:
> Can we left align both buttons at this size?

Yes, that would be quite easy. I'm not a huge fan of how that margin-left is defined as well (https://github.com/dequelabs/cauldron/blob/develop/packages/styles/button.css#L193-L195). If we solve this problem within cauldron, as @scurker suggested above, we might as well address the margin-left problem too.

What does everyone want to do? Should we merge this as it resolves a legit a11y issue? If so, could someone create a ticket for fixing the issue in cauldron itself? Or would we prefer to just close this one and let this get resolved when we fix the issue in cauldron css?

anastasialanz:
@schne324 I think merging this PR in and then making a new issue to fix it within Cauldron is best. I like how you tried to knock a bunch of them out in 30 mins! 😄

schne324:
ok, I will get these conflicts resolved soon and get this merged in then

ryomtoob:
I resolved the merge conflict, so we can get this one in.

ryomtoob (approved):

PR #793, opened 2022-10-03
fix(styles): clean up table sort button styles

anastasialanz:
BREAKING CHANGE: The TableHeader sort button text colors do not change when ascending or descending, per the UX Pin. The removed styles were overriding with the same hex value for light or dark mode.

https://preview.uxpin.com/4052c8b6f05ce03e28813f28deb3a682950c2ac6#/pages/140808126/simulate/sitemap

https://cauldron.dequelabs.com/components/Table

sort-button

### DIFF
--- a/packages/styles/table.css
+++ b/packages/styles/table.css
@@ -2,7 +2,6 @@
   --table-header-text-color: var(--gray-90);
   --table-header-background-color: var(--gray-20);
   --table-header-sorting-background-color: #caddec;
-  --table-header-sorting-text-color: var(--gray-90);
   --table-header-background-color-hover: #dddddd;
   --table-cell-text-color: var(--gray-60);
   --table-cell-background-color: var(--white);
@@ -16,7 +15,6 @@
   --table-header-text-color: var(--white);
   --table-header-background-color: var(--accent-dark);
   --table-header-sorting-background-color: #53636e;
-  --table-header-sorting-text-color: var(--white);
   --table-header-background-color-hover: #53636e;
   --table-cell-text-color: var(--accent-light);
   --table-cell-background-color: var(--accent-medium);
@@ -70,11 +68,6 @@
 .TableHeader--sort-descending {
   background: var(--table-header-sorting-background-color);
   border-bottom: 3px solid var(--table-header-sorting-border-color);
-}
-
-.TableHeader--sort-ascending .TableHeader__sort-button,
-.TableHeader--sort-decscending .TableHeader__sort-button {
-  color: var(--table-header-sorting-text-color);
 }
 
 /* compensate the gap in height between TableHeader and Icon */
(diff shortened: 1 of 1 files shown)


packages/styles/table.css
```diff
@@ -15,7 +14,6 @@
   --table-header-text-color: var(--white);
   --table-header-background-color: var(--accent-dark);
   --table-header-sorting-background-color: #53636e;
-  --table-header-sorting-text-color: var(--white);
```
anastasialanz, line 18: This is the same color as `--table-header-text-color`

packages/styles/table.css
```diff
@@ -2,7 +2,6 @@
   --table-header-text-color: var(--gray-90);
   --table-header-background-color: var(--gray-20);
   --table-header-sorting-background-color: #caddec;
-  --table-header-sorting-text-color: var(--gray-90);
```
anastasialanz, line 5: This is the same color as `--table-header-text-color`. The sorting color does not change.

thuey (commented):
These changes look good to me. However, it seems like there's still some disagreement about whether or not this should be considered a breaking change. Continuing the discussion elsewhere.

anastasialanz:
Based on https://github.com/dequelabs/cauldron/issues/881, do we agree that this is a breaking change? @dequelabs/cauldron-team

thuey:
> Based on #881, do we agree that this is a breaking change since classes are removed? @dequelabs/cauldron-team

Thoughts, @scurker ?

scurker:
That proposal hasn't yet been discussed/approved so I would lean towards this not being a breaking change.

anastasialanz:
https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#styles-1

Based on the breaking changes section for styles: 
> For cauldron's styles library, any of the following would be considered breaking changes:
> - Changing or removing a class name
> - Changing or removing a global css custom property name (see variables.css)
> - Changing or removing a component css custom property name


I think this change would be breaking because I am removing a component's css custom property name.

scurker (commented):
These are breaking changes, so per the recommendation of the [contributing guidelines](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#deprecating) we would want to mark these as deprecated, in this instance by including a comment above each deprecated item and warning to not use.

Once the 2 month period deprecation has passed we could remove them in a future PR.

anastasialanz:
> > Is there a certain commit format for adding deprecation comments?
> 
> Not specifically. It might be good to link back to the issue to remove the styles in the comments but I don't have strong opinions here aside from having some sort of comment.

I'm gonna set a reminder in my calendar to remove them.

scurker:
> I'm gonna set a reminder in my calendar to remove them.

We'll also review that ticket tomorrow in our review and stick it in some milestone so it won't be forgotten.

PR #795, opened 2022-10-04
fix(react): change div element to span in Icon component

### LINKED ISSUE #668: Switch `<Icon>` components to use `span` instead of `div` for semantic HTML
When using an `<Icon type="external-link">` component in a `Link` component that is inside of a `<p>` tag, you will see an error for `Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.`

```
<p>
  <Trans>
    Sample Text.
    <Link
      target="_blank"
      rel="noreferrer noopener"
      className={styles.finalLink}
      href="https://github.com"
    >
      External Link <Icon type="external-link"></Icon>
    </Link>
  </Trans>
</p>
```

Since an `Icon` could be used inside of `<p>` tags in this link example, I think the safest thing to do would be to change the `div` to a `span` for the `Icon` component.

The `p` tag only accepts phrasing content. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p

This is similar to https://github.com/dequelabs/cauldron/issues/224.

villet2:
fix(react): change div element to span in Icon component for better semantics.

When Icon component is used inside p tag it is not valid HTML, because Icon component uses div element. Replace div element with span, so the Icon can be used inside p tags and it is considered valid HTML.

Closes #668

### DIFF
--- a/packages/react/src/components/Icon/index.tsx
+++ b/packages/react/src/components/Icon/index.tsx
@@ -51,10 +51,10 @@ const Icon = forwardRef<HTMLDivElement, IconProps>(
     };
 
     return (
-      <div ref={ref} {...data}>
+      <span ref={ref} {...data}>
         {label && <Offscreen>{label}</Offscreen>}
         {IconSVG && <IconSVG />}
-      </div>
+      </span>
     );
   }
 );


anastasialanz:
The `<span>` is `inline-block` because of the `.Icon` class so nothing changes visually and the error message is gone. Looks good!

Screen Shot 2022-10-09 at 11 31 23 AM

anastasialanz (approved):

anastasialanz:
Reviewed for security.

Issue #627, opened 2022-04-20
Checkbox wrapping issue


scurker:
![checkbox with label above the visual checkbox](https://user-images.githubusercontent.com/1062039/164276980-eac78d11-90a5-4f7c-8170-37bd4b6ffc4f.png)

Checkboxes have an issue when the label needs to wrap, the label content appears before the checkbox.

abbeyperini-deque:
Goal is to keep the label on the same line as the checkbox.

YuriDevAT:
Hi. I would like to work on this issue.

YuriDevAT:
Hi 🙂 I wanted to work on my open issues over the weekend but I am facing problems with yarn since then which I cannot overcome. Since I will not have time to work on it for the next few weeks, I would unassign me, and come back later if the issue is still open. Sorry for the inconvenience.

abbeyperini-deque:
Not a problem! Thanks for all your contributions!

PR #798, opened 2022-10-08
fix: prevent label to appear before checkbox on smaller screens

YuriDevAT:
## Problem
Label appears before checkbox on smaller screen.

Labels appear before checkboxes on smaller screens.

## Goal

Goal is to keep the label on the same line as the checkbox.

## Solution

- [x] Prevent flex container to wrap on smaller screens by setting `flex-wrap: nowrap`.
- [x] Add `overlfow-wrap: break-word` to label to fit the screen.

Label stays in same line with checkbox.

Fixes issue #627 

## Running Tests

All 64 test suits have passed.
All 64 test suites have passed.


docs/patterns/components/Checkbox/index.css
```diff
@@ -0,0 +1,7 @@
+.Checkbox {
+  flex-wrap: nowrap;
+}
+
+.Checkbox .Field__label {
+  overflow-wrap: break-word;
+}
```
dequejosie, line 7: This is the right approach, but we would like to make this change to the Checkbox component itself. Apologies that our ticket did not provide enough context.

dequejosie (changes requested):
Please see the comment above.

docs/patterns/components/Checkbox/index.css
YuriDevAT: Thank you so much for your feedback. I noticed that changing `wrap` to `nowrap` breaks the layout of demo checkboxes 5 to 7. So I'm looking for another solution (since :has (which was my first thought) is not yet [supported by Firefox automatically](https://developer.mozilla.org/en-US/docs/Web/CSS/:has), I have to find another way).

dequejosie: Ah, good catch. I see the broken ones you mentioned.

scurker:
@YuriDevAT Are there changes missing from this PR? I went to review and didn't see any files changed.

YuriDevAT:
My first approach fixed the bug for the mentioned problem but causes another one. Still looking for another approach to fix the issue.

PR #808, opened 2022-10-14
fix: GitHub icon invisible on smaller screens

### LINKED ISSUE #566: Cauldron on GitHub menu item is not visible properly when page is Re flowed
size your browser down to ~320px width...

image

<dl><dt>Description</dt><dd>Content disappearing and not being available when page has re-flowed at 320px. 

Element : Cauldron on GitHub Menuitem on header is not visible when page is Re flowed.</dd><dt>Impact</dt><dd>Serious</dd><dt>Recommendation</dt><dd>RULE : 

HOW TO FIX: 

REFERENCE: 
WCAG Understanding document:

BACKGROUND: 
People with low vision often read content on a desktop computer by scaling content using the browser zoom - up to 400%. When zooming causes content to overlap other content, users with low vision may not be able to see or use that content. When content is created so that it reflows without losing information or functionality at screen widths down to 320px, users can view a web page in a 1280px width browser screen and zoom the content to 400% (320px) without having to scroll side to side to read and without losing access to content.</dd><dt>Source code</dt><dd>

```html
<li class="MenuItem--separator" tabindex="0" role="menuitem"><a href="https://github.com/dequelabs/cauldron" class="fa fa-github" aria-label="Cauldron on GitHub" tabindex="-1"></a></li>
```
</dd><dt>Checkpoint</dt><dd>Reflow (1.4.10.a)</dd><dt>URL</dt><dd>https://cauldron.dequelabs.com/</dd></dl>

YuriDevAT:
## Problem

GitHub Icon is invisible on smaller screen, starting at a screen width under 348px.
Showing invisible GitHub Icon on a screen smaller than 348px.

Fixes #566 

## Solution

My thought about the bug was to hide the apps name at a certain viewport width, because this is common on mobile version to only show the apps icon (or a smaller/simplified version of the logo).
Showing solution with GitHub Icon now visible.

In the `index.js` I added an `alt` text to the image by setting `alt="Cauldron"`, which is the apps name, and set the aria attribute  `aria-hidden="true"` to the `span` with the name as a content to hide it from screen reader users (to prevent the screen reader saying the name twice: once from the `alt` attribute, once from the span content).

Code snippet from Logo with Name with aria for right screen reader users

In the `index.css` I added a media query for the `span` element in order to be not displayed when the screen is under 348px, which is still needed for example an iPhone 5. This solution works fine to a width of 290px.
Screenshot 2022-10-14 at 18 00 55
I realized (to late - sorry) that the other media queries have an empty line between the media query and the first css attribute. Let me know if I should fix it.

## Tests
All tests passed.
Showing all 64 tests were passing.

### DIFF
--- a/docs/index.css
+++ b/docs/index.css
@@ -26,6 +26,12 @@ code {
   width: 38px;
 }
 
+@media (max-width: 348px) {
+  .MenuItem__logo span {
+    display: none;
+  }
+}
+
 [role='menubar'] .OptionsMenu__list {
   width: 140px;
   top: calc(var(--top-bar-height) - 6px);
--- a/docs/index.js
+++ b/docs/index.js
@@ -191,7 +191,7 @@ const App = () => {
             <TopBarItem>
               <Link to="/" className="MenuItem__logo" tabIndex={-1}>
                 <img src={theme === 'dark' ? logo : darkLogo} alt="" />{' '}
-                <span>Cauldron</span>
+                <span aria-hidden="true">Cauldron</span>
               </Link>
             </TopBarItem>
 
--- a/docs/index.js
+++ b/docs/index.js
@@ -190,7 +190,7 @@ const App = () => {
             </TopBarTrigger>
             <TopBarItem>
               <Link to="/" className="MenuItem__logo" tabIndex={-1}>
-                <img src={theme === 'dark' ? logo : darkLogo} alt="" />{' '}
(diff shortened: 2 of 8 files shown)


abbeyperini-deque (approved):
Great idea and very descriptive PR! Thank you!

young girl pointing with both hands at the camera captioned 'you're a rockstar'

abbeyperini-deque:
🚨 reviewed for security 🚨

Issue #707, opened 2022-07-21
The "big" tooltip uses non-standard text and padding sizes


thuey:
The "big" tooltip uses non-standard text and padding sizes: https://github.com/dequelabs/cauldron/blob/0ff6cd9464be2743b5f42536b939c9909871d75f/packages/styles/tooltip.css#L126-L128

We could either fix this separately, or fix it when we update the "big" tooltip to support the dark theme.

PR #810, opened 2022-10-14
fix(styles): use css variables for big tooltip font size and padding

ryomtoob:
Closes: #707 

I also removed the `font-size` property for `.TooltipContent` because it was using a css variable that didn't exist.

### DIFF
--- a/packages/styles/tooltip.css
+++ b/packages/styles/tooltip.css
@@ -132,7 +132,6 @@
 
 .TooltipContent {
   background: #fff;
-  font-size: var(--font-size-smaller);
   text-align: left;
 }
 
--- a/packages/styles/tooltip.css
+++ b/packages/styles/tooltip.css
@@ -124,9 +124,9 @@
 
 .TooltipHead {
   background: var(--gray-20);
-  font-size: 16px;
+  font-size: var(--text-size-small);
   color: var(--gray-90);
-  padding: 16px 28px;
+  padding: var(--space-small) var(--space-large);
   text-align: center;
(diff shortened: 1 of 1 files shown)


packages/styles/tooltip.css
```diff
@@ -124,15 +124,14 @@
...
+  font-size: var(--text-size-small);
   color: var(--gray-90);
-  padding: 16px 28px;
+  padding: var(--space-small) var(--space-large);
   text-align: center;
 }
 
 .TooltipContent {
   background: #fff;
-  font-size: var(--font-size-smaller);
```
anastasialanz, line 135: Seems like it was a typo and should have been `--text-size-smaller`.

PR that added this CSS:
https://github.com/dequelabs/cauldron/pull/134/files

ryomtoob: Good point, however the parent class `.Tooltip` has `font-size: --text-size-smaller`, so it is inherited from there:
https://github.com/dequelabs/cauldron/blob/60782888daab263a83724252e8891c1317da2c89/packages/styles/tooltip.css#L22-L27

Screen Shot 2022-10-19 at 9 42 20 AM

I decided to remove it because doing so had no change on the appearance. And assuming that it was a typo, the presumably desired font size is applied. Does that make sense or do you think correcting instead of removing would be better?

schne324 (approved):
Thank you @anastasialanz / @ryomtoob for fixing my `font-size-smaller` typo (by removing the line of css)!

PR #812, opened 2022-10-17
fix(LoaderOverlay): export LoaderOverlayProps interface

dbowling:
Add missing export so users can extend the `LoaderOverlay` component.


stephenmathieson (approved):

scurker (commented):
Just as a heads up, assuming that our types are exported correctly, you can also infer the interface via typescript without an implicit type:

```tsx
import { Component } from '@deque/cauldron-react'
type ComponentProps = React.ComponentProps<typeof Component>
```

Issue #803, opened 2022-10-14
Props table for Modal component docs page 


abbeyperini-deque:
Add a [PropDocs component](https://github.com/dequelabs/cauldron/blob/develop/docs/Demo/PropDocs.js) at the bottom of the [Modal component docs page](https://github.com/dequelabs/cauldron/blob/develop/docs/patterns/components/Modal/index.js) with the [Modal component props](https://github.com/dequelabs/cauldron/blob/15273d742a4bc2359dd9e757d1c0277637b1d2fc/packages/react/src/components/Dialog/index.tsx#L44).

PR #815, opened 2022-10-19
docs: props table for Modal component docs page

abbeyperini-deque:
Add PropDocs component with Modal props to Modal docs page.

Screen Shot 2022-10-20 at 9 51 49 AM

Closes: #803

### DIFF
--- a/docs/patterns/components/Modal/index.js
+++ b/docs/patterns/components/Modal/index.js
@@ -169,7 +169,7 @@ class Demo extends Component {
               },
               portal: {
                 type: 'any',
-                description: 'The parent element to place the Modal in.',
+                description: 'The parent element to render the Modal within.',
                 default: 'document.body'
               },
               show: {
@@ -188,7 +188,8 @@ class Demo extends Component {
               },
               closeButtonText: {
                 type: 'string',
-                description: 'The desired accessible name of the modal's close button.',
+                description:
+                  "The desired accessible name of the modal's close button.",
                 default: 'Close'
               }
             }}
--- a/docs/patterns/components/Modal/index.js
+++ b/docs/patterns/components/Modal/index.js
@@ -6,6 +6,8 @@ import {
   ModalFooter,
   Code
 } from '@deque/cauldron-react';
+import PropDocs from '../../../Demo/PropDocs';
+import { children, className } from '../../../props';
 
 export default class Demo extends Component {
   constructor() {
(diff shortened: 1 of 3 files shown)


docs/patterns/components/Modal/index.js
```diff
@@ -144,6 +146,54 @@ class Demo extends Component {
...
+                description: 'Whether or not to show the Modal'
+              },
+              dialogRef: {
+                type: 'function or function.current',
+                description: 'Pass a ref to the Modal.'
+              },
+              forceAction: {
+                type: 'boolean',
+                description:
+                  'If true, user must interact with Modal content before closing it.',
```
schne324, line 186:
```suggestion
                  'If true, no close button will be rendered and the user must take action with Modal content.',
```

docs/patterns/components/Modal/index.js
```diff
@@ -144,6 +146,54 @@ class Demo extends Component {
...
+                  'Displayed in the heading at the top of the Modal. Optional to pass heading level.',
+                required: true
+              },
+              onClose: {
+                type: 'function',
+                description: 'Function called when the Modal is closed'
+              },
+              portal: {
+                type: 'any',
+                description: 'The parent element to place the Modal in.',
```
schne324, line 172:
```suggestion
                description: 'The parent element to render the Modal within.',
```

schne324 (approved):
LGTM - left a few nitpicky comments on the wording that you can choose to accept or ignore

docs/patterns/components/Modal/index.js
abbeyperini-deque: Updated this one in Modal, Tooltip, and TooltipTabstop since I copied and pasted 😁

docs/patterns/components/Modal/index.js
```diff
@@ -144,6 +146,55 @@ class Demo extends Component {
...
+              dialogRef: {
+                type: 'function or function.current',
+                description: 'Pass a ref to the Modal.'
+              },
+              forceAction: {
+                type: 'boolean',
+                description:
+                  'If true, no close button will be rendered and the user must take action with Modal content.',
+                default: 'false'
+              },
```
scurker, line 188:
```suggestion
```

This isn't true, modals are implicitly `forceAction={false}`: 
https://github.com/dequelabs/cauldron/blob/fa8e54c533a55ae2c5164802675a6a3fbf8fa985/packages/react/src/components/Modal/index.tsx#L15

scurker (changes requested):

docs/patterns/components/Modal/index.js
abbeyperini-deque: 🤔 I have default as false - is the description confusing?

scurker: `forceAction` isn't a valid prop for the modal component, see the interface:

https://github.com/dequelabs/cauldron/blob/3ba9dc2a80bf52567c00361f9bddbe4ab6afa154/packages/react/src/components/Modal/index.tsx#L5-L7

We omit the `forceAction` prop, so we don't need to include that property for documentation.

scurker:
Closing this, we're going to work soon on updating the documentation on all of our components and this will be outdated.

Issue #680, opened 2022-06-21
Code component with taxindex=0 should have role and accessible name


straker:
The `Code` component allows adding `tabindex=0` to support keyboard only scrolling of the code. When tabindex is applied, it should also [have a role and accessible name](https://adrianroselli.com/2022/06/keyboard-only-scrolling-areas.html) so screen readers know what the element in the tab order is doing. `role=region` is the typical role to add, but doing so also requires that the element [have an accessible name](https://www.w3.org/TR/wai-aria-1.2/#region). 

When `tabindex` is used on the `Code` component, it should add `role=region` to the element and require that the element have an accessible name (either `aria-label` or `aria-labelledby`) as `region` role is not named from contents.

PR #817, opened 2022-10-21
fix: focusable Code component needs role and accessible name

abbeyperini-deque:
- add `focusable` prop
- when true, `tabIndex=0` and `role='region'` will be added. If an `aria-label` or `aria-labelledby` is not added, there will be a console error.
- borrowed the [hasLabel validator from RadioGroup](https://github.com/dequelabs/cauldron/blob/3e730f66df86cecedfa022fce6a00ede3c837ded/packages/react/src/components/RadioGroup/index.tsx#L154)
- added 2 tests

Issue #680

All tests pass

Screen Shot 2022-10-27 at 3 51 32 PM

### DIFF
--- a/docs/patterns/components/Code/index.js
+++ b/docs/patterns/components/Code/index.js
@@ -4,7 +4,7 @@ import Demo from '../../../Demo';
 
 const CodeDemo = () => (
   <>
-    {/* <Demo
+    <Demo
       component={Code}
       componentDescription={
         'Displays text styled to indicate that it is computer code.'
@@ -67,19 +67,12 @@ div[class="foo"] {
         },
         ariaLabelledBy: {
           type: 'string',
-          description: 'Pass an id for an element to ariaLabelledBy instead of ariaLabel',
+          description:
+            'Pass an id for an element to ariaLabelledBy instead of ariaLabel',
           default: 'undefined'
         }
       }}
-    /> */}
-    <div>
-      <h1 id="heading">Javascript code snippet</h1>
-      <Code
-        language="javascript"
-        tabIndex={0}
-        ariaLabelledBy="heading"
-      >{`var some = "javascript"`}</Code>
-    </div>
+    />
   </>
 );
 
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Code/index.tsx
```diff
@@ -23,25 +23,32 @@ interface Props extends SyntaxHighlighterProps {
...
+  if (!ariaLabel) ariaLabel = 'Code snippet';
+
+  return (
+    <>
+      <Highlighter
+        {...props}
+        useInlineStyles={false}
+        className={classNames('Code', className)}
+        tabIndex={tabIndex}
+        {...(tabIndex === 0 && { role: 'region', 'aria-label': ariaLabel })}
```
schne324, line 45: This could cause problems in cases where the accessible name for the code block is added via `aria-labelledby`

```js
<h3 id="foo">Some heading</h3>
<Code aria-labelledby="foo">some code</Code> // cauldron adds an `aria-label` which will override/conflict with the aria-labelledby value
```

I don't think cauldron should enforce how an accessible name is added to a code block

abbeyperini-deque: Good suggestion! Updated.

docs/patterns/components/Code/index.js
```diff
@@ -4,7 +4,7 @@ import Demo from '../../../Demo';
 
 const CodeDemo = () => (
   <>
-    <Demo
+    {/* <Demo
```
schne324, line 7: Accidentally commented out code?

schne324 (commented):
I'm thinking certain code blocks wouldn't need the region role, nor would they need to be named via `aria-label` or `aria-lablledby` and instead could be named by their contents. For example, short/1-liner code blocks that will never have scroll overflow...

The region role with accessible name is really only "required" in situations where the code block is scrollable so keyboard users can scroll it with <kbd>⬆️ </kbd> and <kbd>⬇️ </kbd>.

So my real concern here is that slapping `role=region` could create a page with excessive amounts of regions which could be overwhelming for screen reader users.

docs/patterns/components/Code/index.js
abbeyperini-deque: 😢 fixed

scurker:
Closing this ticket because code components now handle tab index automatically.

Issue #661, opened 2022-05-19
Add "border" variant for Table component


scurker:
![cauldron table with border variant](https://user-images.githubusercontent.com/1062039/169328364-fdbd25fc-8e1b-43d0-ad77-e992fb708098.png)

Please contact @awpearlm or @scurker or @schne324 if you need access to the UX pin.

villet2:
I could work on this one.

villet2:
@scurker Is there a spec for the border styles?

scurker:
Here's some example of tables in different configurations:

<details>
  <summary>Light theme w/o header row</summary>

![cauldron light themed table](https://user-images.githubusercontent.com/1062039/196498176-b170981f-4d09-473a-86a4-f964de20146b.png)

</details>

<details>
  <summary>Light theme w/ header row</summary>

![cauldron light themed table](https://user-images.githubusercontent.com/1062039/196498210-e23f3449-1b6a-4792-935c-5f099d49f831.png)

</details>

<details>
  <summary>Dark theme w/o header row</summary>
  
![cauldron dark themed table](https://user-images.githubusercontent.com/1062039/196498338-00e85c77-cff3-44ec-83f9-8ece50cd1186.png)

</details>

<details>
  <summary>Dark theme w/header row</summary>

![cauldron dark themed table](https://user-images.githubusercontent.com/1062039/196498376-c89fb408-b883-4bb6-8f63-8632f4a93b4e.png)
</details>

The border for the light table should be `#cccccc` utilizing the css variable `var(--gray-40)`.

The border for the dark table should be `#5d676f` utilizing the css variable `var(--stroke-dark)`.

There does appear to be a drop shadow for the bordered variant, I think it may be `box-shadow: 0 2px 3px rgba(0, 0, 0, 0.15)` for both themes.

scurker:
This has been implemented

PR #820, opened 2022-10-25
feat: add border variant for table component.

villet2:
Enable border variant for Table by introducing variant prop to Table.

#661


packages/react/src/components/Table/Table.tsx
```diff
@@ -2,20 +2,27 @@ import React, { TableHTMLAttributes } from 'react';
...
+interface TableProps extends TableHTMLAttributes<HTMLTableElement> {
+  variant: 'border' | undefined;
+}
+
+const Table = ({ children, className, variant, ...other }: TableProps) => (
+  <table
+    className={classNames(
+      variant === 'border' ? 'Table Table--variant-border' : 'Table',
+      className
+    )}
```
scurker, line 14: We've typically just been using the `--variantName` for our variant classes (see [Button](https://github.com/dequelabs/cauldron/blob/ded5fdafa720d19cc47b2e7b51f61437a37a3e6b/packages/styles/button.css#L23-L26) or [Tooltip](https://github.com/dequelabs/cauldron/blob/ded5fdafa720d19cc47b2e7b51f61437a37a3e6b/packages/styles/tooltip.css#L112) for examples).

```suggestion
    className={classNames(
      'Table',
      'Table--border': variant === 'border',
      className
    )}
```

Just be sure to remember to update the class name in css as well.

packages/react/src/components/Table/Table.tsx
```diff
@@ -2,20 +2,27 @@ import React, { TableHTMLAttributes } from 'react';
...
 import classNames from 'classnames';
 
-const Table = ({
-  children,
-  className,
-  ...other
-}: TableHTMLAttributes<HTMLTableElement>) => (
-  <table className={classNames('Table', className)} {...other}>
+interface TableProps extends TableHTMLAttributes<HTMLTableElement> {
+  variant: 'border' | undefined;
```
scurker, line 6: Alternatively, we can make this optional:

```suggestion
  variant?: 'border';
```

scurker (changes requested):
Thank you for your contribution @villet2! Just a few small tweaks.

packages/react/src/components/Table/Table.tsx
villet2: Ah, okay. I'll make the changes.

packages/react/src/components/Table/Table.tsx
villet2: Indeed, this is better.

packages/react/src/components/Table/Table.tsx
villet2: This exact syntax could not be used because `'Table--border': variant === 'border',` is not valid. Used `variant === 'border' && 'Table--border',` instead.

scurker: Yup, that's my bad. What you have works but for FYI classnames also accepts object parameters and will apply the classnames based on the boolean values, so the following would also work:

```js
classNames(
  'Table',
  { 'Table--border': variant === 'border' },
  className
)
```

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

PR #821, opened 2022-10-25
feat(icon): add link icon

dbowling:


scurker (approved):
Wow, we didn't already have a link icon? 🤯

dbowling:
I was convinced I was just not seeing it, but we really don't until now!

Issue #828, opened 2022-11-14
Accordion should respect `prefers-reduced-motion`


scurker:
When this media feature is set to `reduce`, we should not animate the panels.

PR #832, opened 2022-11-25
feat(react): respect prefers-reduced-motion with Accordion

tbusillo:
closes: https://github.com/dequelabs/cauldron/issues/828

## Summary
- Adds utility for detecting whether a system has `prefers-reduced-motion: reduce` enabled.
- If it is `prefers-reduced-motion` is enabled, the `<Accordion>` animations/transitions are disabled.
- Unrelated , added a few basic properties to the prettier config in `package.json`. Was running into config issues/inconsistencies in things like trailing commas, parens, etc.


packages/react/src/utils/prefers-reduced-motion/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ * A utility to check whether a user has 'prefers-reduced-motion' enabled
+ *
+ * @return {Boolean}
+ */
+const prefersReducedMotion = (): Boolean => {
+  const reducedMotionEnabled =
+    window.matchMedia && window.matchMedia('(prefers-reduced-motion: reduce)');
+
+  if (!reducedMotionEnabled) {
+    return true;
```
tbusillo, line 13: If `window` (e.g., apps using SSR) is not available or `window.matchMedia` is otherwise undefined (e.g., in tests), this will err on the side of caution and assume the user has it enabled. I recommend defaulting to `true`, but we could possibly improve this by adding a `useEffect` to check `window` once the component is mounted/window is available.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -85,6 +86,11 @@ const Accordion = ({
   const { className: triggerClassName, ...triggerProps } = trigger.props;
 
   const elementId = useId();
+  const prefersReducedMotionEnabled = prefersReducedMotion();
+
+  if (prefersReducedMotionEnabled) {
+    animationTiming = false;
+  }
```
tbusillo, line 93: When `animationTiming` is falsy, the `ExpandCollapse` component disables the animation for expanding/collapsing the panel.

packages/styles/accordion.css
```diff
@@ -75,3 +75,10 @@ button.Accordion__trigger {
   border-radius: 0 0 3px 3px;
   box-shadow: 0 2px 3px 0 var(--accordion-box-shadow-color);
 }
+
+@media (prefers-reduced-motion: reduce) {
+  .Accordion__trigger:hover,
+  .Accordion .Icon {
+    transition: none;
+  }
+}
```
tbusillo, line 84: An alternative to using this media query would be to add/remove a class from the Accordion. However, using the CSS media query seems safer/cleaner to use here.

packages/react/src/utils/use-media-query/index.ts
```diff
@@ -0,0 +1,31 @@
+import { useState, useEffect } from 'react';
+
+function useMediaQuery(query: string, initialValue = false): boolean {
```
tbusillo, line 3: Took the `useMediaQuery` hook from Walnut, my thought being the hook could be used elsewhere in Cauldron/exported, then imported into Walnut (it's already a common module).

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -52,6 +53,14 @@ const Accordion = ({
   animationTiming,
   ...props
 }: AccordionProps) => {
+  const elementId = useId();
+  const reducedMotionQuery = '(prefers-reduced-motion: reduce)';
+  const reducedMotionEnabled = useMediaQuery(reducedMotionQuery, true);
+
+  if (reducedMotionEnabled) {
+    animationTiming = false;
+  }
```
tbusillo, line 62: If window (e.g., apps using SSR) is not available or window.matchMedia is otherwise undefined (e.g., in tests), this will err on the side of caution and assume the user has `prefer-reduce-motion: reduce` enabled. 

I recommend defaulting to true, as the impact of animations being mistakenly disabled is less of an issue than the alternative.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -52,6 +53,14 @@ const Accordion = ({
   animationTiming,
   ...props
 }: AccordionProps) => {
+  const elementId = useId();
```
tbusillo, line 56: Reorganized a few variables to clean things up a bit in the component.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -52,6 +53,14 @@ const Accordion = ({
...
+  const reducedMotionQuery = '(prefers-reduced-motion: reduce)';
+  const reducedMotionEnabled = useMediaQuery(reducedMotionQuery, true);
```
scurker, line 58: We're just using `matchMedia` inside of a `useLayoutEffect` elsewhere, see: https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/TwoColumnPanel/TwoColumnPanel.tsx#L181

If we want to consolidate, that's fine but we should try to be consistent.

packages/react/src/utils/use-media-query/index.ts
```diff
@@ -0,0 +1,31 @@
+import { useState, useEffect } from 'react';
```
scurker, line 1: This doesn't have any tests.

packages/styles/accordion.css
scurker: I'm not sure we need `prefers-reduced-motion` here. Setting this value doesn't mean "no motion", but removing non-essential motion. I would argue that the slight motion that happens here (the rotation of caret and the highlight of the trigger) are still subtile enough where it isn't needed.

scurker (changes requested):
The bulk of the animation happens inside of `ExpandCollapsePanel` in the `animateOpen`/`animateClose` functions. I would think we would instead want to be sure our reduced motion handling code happens there instead of `Accordion`.

Additionally, I think you were close to the right idea of adding the `reduce-motion` media query in both spots, but we should be very specific with what we're targeting. For example:

```css
@media (prefers-reduced-motion: reduce) {
  .ExpandCollapse__panel {
    animation: none !important;
  }
}
```

Targeting ExpandCollapseComponent I think is better because it's the component that's actually controlling the animation and is composed with the Accordion component. Also we can safely use `!important` here because we _always_ want the animation of the section to be turned if there's more specific/cascading selectors.

packages/styles/accordion.css
bobbyomari: From a user experience perspective, the accordion animation can also be helpful for users to notice the change of content and what is occurring on the screen.

CharlesBelov:
@scurker :  re: "I'm not sure we need prefers-reduced-motion here. Setting this value doesn't mean "no motion", but removing non-essential motion. I would argue that the slight motion that happens here (the rotation of caret and the highlight of the trigger) are still subtile enough where it isn't needed."

There is some confusion around prefers-reduced-motion. First, the name is a historical artifact of what Apple called it when it created the feature, "Reduce motion".  Windows calls it "show animation in windows" and Android uses a similar all-or-nothing label. Second, there are at least a couple conditions which are addressed by this setting: vestibular motion disorders (the motion makes the viewer feel ill) and hypervigilance (the motion attracts the attention of the viewer out of proportion to its importance). Either can disrupt the end-user. This makes it difficult to determine a safe amount of animation which will serve everyone implementing this non-granular setting on their operating system. A choice of no animation is the safest.

That said, "the slight motion that happens here (the rotation of caret and the highlight of the trig yesger)" is probably sufficiently subtle.

@bobbyomari : re: "From a user experience perspective, the accordion animation can also be helpful for users to notice the change of content and what is occurring on the screen."

People for whom the accordion animation would be helpful would be unlikely to set prefers-reduce-motion to reduce.  I set prefers-reduced-motion to reduce because the motion pretty much always distracts me and sometimes makes my eyes feel ill.

tbusillo:
Looks like was able to accomplish pretty much everything via the css media query and simplified everything as a result. 

@CharlesBelov would you be willing/interested to take a look at the [preview](https://expand-collapse-prefer-reduced-motion.d1gko6en628vir.amplifyapp.com/) and let us know what you think? The two components likely relevant are the [Accordion](https://expand-collapse-prefer-reduced-motion.d1gko6en628vir.amplifyapp.com/components/Accordion) and the [Expand Collapse Panel](https://expand-collapse-prefer-reduced-motion.d1gko6en628vir.amplifyapp.com/components/ExpandCollapsePanel). We're looking to get this into an upcoming release, so any feedback you have would be greatly valued!

packages/react/src/components/ExpandCollapsePanel/index.css
```diff
@@ -31,3 +31,9 @@
   --expandcollapse-trigger-color: var(--white);
   --expandcollapse-panel-color: var(--accent-light);
 }
+
+@media (prefers-reduced-motion: reduce) {
+  .ExpandCollapse__panel {
+    animation: unset !important;
```
scurker, line 37: `unset` behaves like `inherit` in some instances, so we should implicitly set it to `none`.

scurker (changes requested):

scurker (approved):

scurker:
I created https://github.com/dequelabs/cauldron/issues/1062 to make the UI immediately responsive, but this addresses the immediate accessible need.

CharlesBelov:
@tbusillo Thank you for the fast follow up. I would be happy to take a look at the preview; however I am not knowledgeable about dealing with pull requests on github. When I go to those pages that you have linked, for example, https://expand-collapse-prefer-reduced-motion.d1gko6en628vir.amplifyapp.com/, I get a blank page, and Firefox's network inspector says "no requests".

tbusillo:
@CharlesBelov Ah, it looks like the link(s) expired when the PR was merged. Here are some new ones! 

Accordion: https://cauldron.dequelabs.com/components/Accordion

Expand Collapse Panel: https://cauldron.dequelabs.com/components/ExpandCollapsePanel

Thank you for being willing to take a look!

scurker:
@CharlesBelov This will likely land in the axe extension with the next update (which normally happens around every ~2 weeks) so keep a lookout for that release!

Issue #747, opened 2022-08-22
LoaderOverlay triggering focus multiple times


scurker:
https://github.com/dequelabs/cauldron/blob/b6bfee024f465ff4c1270aa07aeb8a24f90658e6/packages/react/src/components/LoaderOverlay/index.tsx#L47-L54

I noticed the above code can be problematic for `<LoaderOverlay>`. Depending on how `<LoaderOverlay>` is rendered, this can cause focus to be rendered multiple times causing some jank in the UI. Ideally, we only want this logic to trigger _once_ and not be dependent on `overlayRef.current` changing.

PR #833, opened 2022-11-26
fix(react): makes LoaderOverlay focusOnInitialRender work

tbusillo:
closes: https://github.com/dequelabs/cauldron/issues/747

## Summary
- Removes `overlayRef.current` as a dependency and forcing of focus only triggers when the component initially mounts.
- Removes the `setTimeout` that wrapped `overlayRef.current.focus()`, which previously existed due to the focus indicator styling not appearing on the focused element (it was a timing/race condition). 
- Updates tests/docs.


scurker:
Sorry for the delay on this @tbusillo but is this something you want to try to get up to date on or just defer?

tbusillo:
> Sorry for the delay on this @tbusillo but is this something you want to try to get up to date on or just defer?

@scurker I'd be open to doing it. Is the original issue still valid/have there been any changes to the requirement(s), outside of adding test coverage maybe?

scurker:
> @scurker I'd be open to doing it. Is the original issue still valid/have there been any changes to the requirement(s), outside of converting test coverage maybe?

Yes, I believe it's still valid. `useEffect`ing off of a ref isn't really super reliable.

tbusillo:
> > @scurker I'd be open to doing it. Is the original issue still valid/have there been any changes to the requirement(s), outside of converting test coverage maybe?
> 
> Yes, I believe it's still valid. `useEffect`ing off of a ref isn't really super reliable.

IIRC, this originally had to be done due to animation timing(s) causing the active element to be the loader, but the UI not painting the focus indicator. 

I am not sure what changed (could be any number of things) but I don’t mind cleaning this up and finishing it when I get some down time.

tbusillo:
Closed in favor of https://github.com/dequelabs/cauldron/pull/1486.

Issue #845, opened 2023-01-10
link as a thin and primary button has incorrect padding and font weight


stephenmathieson:
### Steps to reproduce

Steps to reproduce

```html
<a href="#" class="Button--primary Button--thin">
  Hello
</a>
```

### Expected Behavior

It has the correct padding and font weight.

### Actual Behavior

It doesn't have the correct padding and font weight,

### Version

latest/next

### Anything else we should know?

**Expected**

<img width="265" alt="image" src="https://user-images.githubusercontent.com/571265/211578950-686c2119-2061-4b15-9573-4adb2502d5cc.png">

**Actual**

<img width="241" alt="image" src="https://user-images.githubusercontent.com/571265/211579018-98191999-ca85-4e1e-97e5-4ddae214a9f9.png">

gayathrit12:
This is not working as expected. So, reopening it.

## Screenshot:
![image](https://user-images.githubusercontent.com/102580833/212324914-9cea2a8a-ce41-49b0-96c2-38eb113c22e4.png)

dequejenn:
@gayathrit12 - I just found out when something is updated in Cauldron it needs additional steps before the change shows up in walnut. To QA, just check it in Cauldron with this link: https://cauldron.dequelabs.com/components/Link

We will have then Projects Page issue checked during the Design Approval to ensure the Cauldron changes came through to the Projects page, so you don't need to worry about it.

gayathrit12:
## Retesting details:
Working as expected.

## Screenshots:
![image](https://user-images.githubusercontent.com/102580833/213140558-53cac2f3-1f9a-42a9-ac81-63a4e8808daa.png)

## Environment:
| **Label**                         | **Value** |
| -------------------------- | ------ |
| **Server URL**                | _https://axe.dequelabs.com/_ |
| **Server version**           | _4.62.0_ |
| **OS**                              | _Windows 10_ |
| **Tested Browsers**       | _Chrome Version 108.0.5359.125 (Official Build) (64-bit)_ |

gayathrit12:
Reopening this ticket as the button name is not matching with the UX Pin.

## UX Pin: "Add New project" 
![image](https://user-images.githubusercontent.com/102580833/213174191-fc8ac8fe-085a-4c52-abb8-c00c8573eab7.png)

## Web App:  "Add a new project" 
[...]

PR #849, opened 2023-01-12
fix: allow users to use thin link button

michael-siek:
Closes: #845 
image

### DIFF
--- a/packages/react/src/components/Link/index.css
+++ b/packages/react/src/components/Link/index.css
@@ -4,5 +4,5 @@ a.Button--secondary {
   padding: 9px 18px;
 }
 a.Button--thin {
-  padding: 0px 16px;
+  padding: 0 16px;
 }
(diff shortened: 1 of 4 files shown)


stephenmathieson (approved):
LGTM but someone from @dequelabs/cauldron-team should take a look

packages/react/src/components/Link/index.css
```diff
@@ -3,3 +3,6 @@ a.Button--secondary {
   text-decoration: none;
   padding: 9px 18px;
 }
+a.Button--thin {
+  padding: 0px 16px;
```
scurker, line 7: Nit: Values of 0 in css can usually be unitless

```suggestion
  padding: 0 16px;
```

scurker (dismissed):

scurker (approved):

Issue #858, opened 2023-01-20
Update Impact colors in Cauldron


dequejenn:
Per a discussion with Bobby 

```
  --issue-serious: #ed9a55;
  --issue-moderate: var(--accent-warning);
```
should be changed to

```
  --issue-serious: var(--accent-warning);
  --issue-moderate: #f0c4f8;


```

PR #859, opened 2023-01-20
fix: update css variable for moderate/serious

michael-siek:
Closes: https://github.com/dequelabs/cauldron/issues/858
image


scurker (approved):
Should probably be `fix` so issue will show up in changelog.

Issue #861, opened 2023-01-20
Bug: Pagination component responsiveness is off


dequejenn:
### Steps to reproduce

To Do:

- remove the left margin from .Pagination ul so that the <Pagination> component can be properly centered
- in smaller viewports, remove the margin from .Pagination ul [role=log], enabling the text to fit in smaller viewports

Examples:
1. Without removing left margin from `ul`
<img width="324" alt="image (31)" src="https://user-images.githubusercontent.com/112327890/213795345-09acaa15-6f44-4e61-9e2b-34020423601f.png">

2. Left margin removed from `ul`
<img width="332" alt="image (32)" src="https://user-images.githubusercontent.com/112327890/213795490-be7ed43e-b1be-4419-8e52-16e564d022c9.png">

3. Without margin removed from `role=log`
<img width="370" alt="image (33)" src="https://user-images.githubusercontent.com/112327890/213795642-15df1d00-9b04-4e56-b9c2-81d25d34b6d6.png">

4. Margin removed from `role=log`
<img width="348" alt="image (34)" src="https://user-images.githubusercontent.com/112327890/213795689-e31ac1f0-d9ac-46e1-aa25-eade92475cd0.png">

### Expected Behavior

As screen shrinks and grows the pagination component should stay centered

### Actual Behavior

See above

### Version

I dunno

### Anything else we should know?

Additional information

Internal: Slack Jenn for UX Pin

PR #865, opened 2023-01-21
fix(pagination): remove unnecessary margins

michael-siek:
Closes: #861 

<details>
	<summary>Pagination</summary>
	320px:
	
image
Ï
</details>

### DIFF
--- a/packages/styles/pagination.css
+++ b/packages/styles/pagination.css
@@ -67,7 +67,7 @@
   color: var(--accent-light);
 }
 
-@media (max-width: 450px) {
+@media (max-width: 28.125rem) {
   .Pagination [role='log'] {
     margin: 0;
   }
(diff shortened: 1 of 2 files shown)


packages/styles/pagination.css
```diff
@@ -66,3 +66,13 @@
 .cauldron--theme-dark .Pagination [role='log'] {
   color: var(--accent-light);
 }
+
+@media (max-width: 450px) {
```
scurker, line 70: We tend to use rems:
```suggestion
@media (max-width: 28.125rem) {
```

Example: 
https://github.com/dequelabs/cauldron/blob/f7f6318a79b59471b15f518f3825a86604d06fe9/packages/styles/two-column-panel.css#L242

scurker (dismissed):

scurker (approved):

PR #867, opened 2023-01-24
build: Change dev server port to avoid conflicts

stephenmathieson:
When running the axe server's dependencies locally (`dc.sh up -d`), `:8000` is already taken by DynamoDB. This patch changes the Cauldron dev server's port to 8003 to avoid this conflict.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ $ yarn build
 $ yarn dev
 ```
 
-navigate browser to http://localhost:8000
+navigate browser to http://localhost:8003/
 
 ### run tests
 
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -64,7 +64,7 @@ const config = {
     }
   },
   devServer: {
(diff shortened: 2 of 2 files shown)


tbusillo:
This is fine IMO and not to over complicate this, but maybe worth using something like https://github.com/unjs/get-port-please?

stephenmathieson:
@tbusillo I'm fine with that too. We could also just not specify a port, and Webpack will pick one on its own. Will wait to hear what others say.

scurker:
> @tbusillo I'm fine with that too. We could also just not specify a port, and Webpack will pick one on its own. Will wait to hear what others say.

I don't really care as long as we communicate the change somewhere

scurker:
> @scurker I added a note in the README. Is there somewhere else you think it belongs?

Works for me!

scurker (approved):

PR #868, opened 2023-01-24
fix: Integrate design review feedback

### LINKED ISSUE #866: Changes needed per Design Review with Bobby
Make all changes in Cauldron that Bobby and Stephen went through at today's Design Review (too many to list out here, just going to copy changes Stephen made to Walnut during the meeting into Cauldron)

stephenmathieson:
Ref https://github.com/dequelabs/walnut/pull/4952
Closes #866

BREAKING CHANGE: various variables have changed values (eg `--space-largest`).


packages/react/src/components/Icon/index.tsx
```diff
@@ -14,10 +14,9 @@ export interface IconProps extends React.HTMLAttributes<HTMLDivElement> {
 const Icon = forwardRef<HTMLDivElement, IconProps>(
   ({ label, className, type, ...other }: IconProps, ref) => {
     const isMounted = useRef(true);
-    const [, name, direction] = type.match(/(.*)-(right|left|up|down)$/) || [
-      '',
-      type
-    ];
+    const [, name, direction] = (type !== 'table-sort-down' &&
+      type !== 'table-sort-up' &&
+      type.match(/(.*)-(right|left|up|down)$/)) || ['', type];
```
thuey, line 19: Although this works, it would be nice to not have this override here. Instead of doing this, could we just change the name of the icon to not have "up" or "down"? For example, `table-sort-ascending` and `table-sort-descending`

packages/styles/icon-button.css
```diff
@@ -61,11 +61,13 @@
...
   vertical-align: middle;
+  line-height: unset;
 }
 
 .IconButton .Icon {
-  width: 100%;
-  height: 100%;
+  height: var(--space-small);
+  width: var(--space-small);
+  margin-top: 2px;
```
thuey, line 70: Can you help me understand the motivation behind this change? I'm also wondering if this will mess up anyone's usage of IconButton if they have overridden the IconButton's height and width.

thuey (changes requested):
I don't think we had a resolution on what constitutes a "breaking change", but I'm wondering if the nature of some of these changes justifies labeling this a breaking change. Thoughts?

packages/styles/icon-button.css
stephenmathieson: @bobbyomari said the icon button's SVGs are too big. He wanted them to be this size.

stephenmathieson:
> I don't think we had a resolution on what constitutes a "breaking change", but I'm wondering if the nature of some of these changes justifies labeling this a breaking change. Thoughts?

IMO no, nothing here is breaking. Code that worked previously will still work. There are visual changes, but those are being made intentionally.

packages/styles/icon-button.css
thuey: Instead of setting the constraints on the Icon, can we solve this by adding padding to the IconButton container?

packages/styles/table.css
```diff
@@ -102,6 +103,10 @@
   box-shadow: var(--table-box-shadow);
 }
 
+.TableCell {
+  color: var(--gray-90);
+}
```
stephenmathieson, line 108: This needs to only apply in light-mode. There are tons of color contrast issues otherwise.

image

packages/styles/variables.css
```diff
@@ -69,7 +69,7 @@
   --space-small: 16px;
   --space-medium: 18px;
   --space-large: 24px;
-  --space-largest: 29px;
+  --space-largest: 32px;
```
scurker, line 72: I'm not sure if this constitutes a breaking change or not, and I could be convinced either direction but I think warrants discussion.

packages/styles/icon-button.css
scurker: I agree this change might cause issues as @thuey is suggesting. Also, if we want them to be a fixed size, we should instead be using something like `align-self: center` to let the browser do the centering instead approximating it with margins.

packages/styles/variables.css
thuey: Yeah, this is the one that jumped out at me. I _think_ it's fine from a breaking change perspective. In some ways, it's no worse than other changes we make.

stephenmathieson: Who needs to discuss it? This is what Bobby said the largest space should be.

packages/styles/icon-button.css
stephenmathieson: That's a good idea, I'll experiment with that.

packages/styles/variables.css
scurker: I'm not sure, it just means anything utilizing this var outside of cauldron would shift. I'm okay with changing it, just trying to figure out the impact. It's only used in a few spots in the extension/walnut so the impact would be minimal there and we will detect any visual changes via the extension screenshots.

thuey (approved):

PR #876, opened 2023-01-31
feat(app): add new spacing and colour variables

Zidious:
This PR adds new variables based from new UX pin. 

\cc @bobbyomari

### DIFF
--- a/packages/styles/variables.css
+++ b/packages/styles/variables.css
@@ -50,6 +50,12 @@
   --header-text-color: var(--gray-80);
   --header-text-color-dark: var(--gray-90);
 
+  /* stroke colours */
+  --stroke-light: #b3bfc6;
+
+  /* background colours */
+  --background-light: #f0f2f5;
+
   /* accents */
   --error: var(--accent-error);
   --disabled: var(--gray-40);
@@ -66,6 +72,7 @@
 
   /* spacing (padding/margin) */
   --space-smallest: 8px;
+  --space-smaller: 12px;
   --space-small: 16px;
   --space-medium: 18px;
   --space-large: 24px;
@@ -81,6 +88,7 @@
   --text-size-larger: 56px;
   --text-size-large: 45px;
   --text-size-large-medium: 34px;
+  --text-size-medium-small: 32px;
   --text-size-medium: 24px;
   --text-small-medium: 20px;
   --text-size-normal: 18px;
--- a/packages/styles/variables.css
+++ b/packages/styles/variables.css
@@ -8,6 +8,7 @@
   /* color palette */
   --white: #ffffff;
   --dark-workspace-color: #53636e;
+  --stroke-light: #b3bfc6;
   --stroke-dark: #5d676f;
   --gray-10: #fdfdfe;
   --gray-20: #f2f2f2;
@@ -50,9 +51,6 @@
   --header-text-color: var(--gray-80);
   --header-text-color-dark: var(--gray-90);
 
-  /* stroke colours */
-  --stroke-light: #b3bfc6;
-
   /* background colours */
   --background-light: #f0f2f5;
 
@@ -88,7 +86,7 @@
   --text-size-larger: 56px;
   --text-size-large: 45px;
   --text-size-large-medium: 34px;
-  --text-size-medium-small: 32px;
+  --text-size-medium-large: 32px;
   --text-size-medium: 24px;
   --text-small-medium: 20px;
   --text-size-normal: 18px;


packages/styles/variables.css
```diff
@@ -50,6 +50,12 @@
   --header-text-color: var(--gray-80);
   --header-text-color-dark: var(--gray-90);
 
+  /* stroke colours */
+  --stroke-light: #b3bfc6;
```
thuey, line 54: There's already a `--stroke-dark` above. Could we collocate them?

packages/styles/variables.css
```diff
@@ -50,6 +50,12 @@
...
+
+  /* background colours */
+  --background-light: #f0f2f5;
```
thuey, line 57: @bobbyomari Just wondering why one of the existing "gray"'s isn't sufficient?

packages/styles/variables.css
```diff
@@ -81,6 +88,7 @@
   --text-size-larger: 56px;
   --text-size-large: 45px;
   --text-size-large-medium: 34px;
+  --text-size-medium-small: 32px;
```
thuey, line 91: A more appropriate name for this would be `--text-side-large-small` or `--text-size-medium-large`, since it is a step _up_ from `--text-size-medium` (the `-small` makes it sound like a step down)

thuey (changes requested):

packages/styles/variables.css
bobbyomari: We have had these colors in our designs that appear to have never been translated into variables (and use them quite frequently throughout light theme components/styles). Aaron and I developed these colors `stroke-dark` and `stroke-light` because the colors for both light and dark themes vary in contrast. It is not a 1:1 match between both themes.

packages/styles/variables.css
scurker: In retrospect, I wish we had gone with a different verbiage because these text sizes are all confusing and don't scale well when adding new ones. That said, I agree with @thuey here.

bobbyomari: Here are my observations:

Frankly, these are bad naming conventions because it's pretty ambiguous what each variable means without having to remember or go look at documentation.

My recommendation in the future is to use naming conventions such as `--space-2`, `--space-4`, `--space-6`, `--space-8`, etc. as I've outlined in this doc that's currently a work in progress: https://github.com/dequelabs/cauldron/blob/documentation/documentation/home.md

This would apply to text size classes as well as in the future, in my opinion. (e.g. `--text-size-56`,`--text-size-45`, etc. 

Additionally, we should have more standardization in text size as these sizes are all over the place for no rhyme or reason, which is also outlined in the document I provided for future consideration.

scurker: > Frankly, these are bad naming conventions because it's pretty ambiguous what each variable means without having to remember or go look at documentation.

I don't disagree that the naming convention was a poor choice in the past, but right now we don't have the resources to allocate towards making these changes. I do believe there is a time and place to have this discussion, but unfortunately feel that right now isn't the time to be making these sort of changes.

packages/styles/variables.css
thuey: @bobbyomari Sorry, was specifically trying to ask about the `--background-light` color. Typically, our backgrounds are either white or one of the grays

packages/styles/variables.css
anastasialanz: Could we have both the old and new variable names together and slowly migrate over to the new ones when we have time?

stephenmathieson: I think @anastasialanz's suggestion is spot-on. Let's deprecate "old" variables and create new ones with names that make sense.

bobbyomari: > 

I agree, this is definitely a long-term issue we need to think about, so I don't have a bone to pick if we are just arbitrarily choosing names right now and will have an effort in the future to change the var names. You guys are the ones to have to figure out what naming convention works today 😄

packages/styles/variables.css
bobbyomari: > @bobbyomari Sorry, was specifically trying to ask about the --background-light color. Typically, our backgrounds are either white or one of the grays

That is not entirely true - there are layouts and designs that do use `--background-light`. It's likely it just was never implemented correctly and/or the color variable was never implemented. In fact, our navbar, topbar, and footer components are not using white `#ffffff` and using `#fdfdfe` for light theme. It's subtle, but there are use cases where `--background-light` is being used in our applications such as codeblocks and various background colors that only apply to light theme styles. 

I do have a color palette included on that documentation I previously included, but I think it will help to separate them into Dark and Light theme categories for clarity and examples of where they are used.

thuey: Thanks for the explanation.

thuey (approved):

scurker (approved):

PR #882, opened 2023-02-03
fix: remove base TableHeader from sort styles

anastasialanz:
This PR fixes a bug that was causing the table header to use `--table-header-sorting-background-color` by default even when the table header wasn't sorted.

[UXPin - Tables](https://preview.uxpin.com/68df515d0023cc4b79f25d43f7663e24e363b1b9#/pages/141465418/simulate/sitemap?mode=i)
Screenshot from UXPin Light Theme
Screen Shot 2023-02-02 at 5 46 57 PM

Screenshot from UXPin Dark Theme
Screen Shot 2023-02-02 at 5 47 34 PM

This change is fixing a breaking change in 5.0.0. Would that mean that the version bumps to 5.0.1?

### Reproduce the bug
- [ ] Go to https://cauldron.dequelabs.com/components/Table
- [ ] See that the table headers are using `--table-header-sorting-background-color` by default for the light and dark theme
Screen Shot 2023-02-02 at 5 46 32 PM
Screen Shot 2023-02-02 at 5 55 27 PM

## Verify the fix
- [ ] Go to https://fix--table-header-css.d1gko6en628vir.amplifyapp.com/
- [ ] Scroll down to the Sortable Table and verify the headers are set to `--table-header-background-color` for the light and dark theme
Screen Shot 2023-02-02 at 5 40 42 PM
Screen Shot 2023-02-02 at 5 53 57 PM

### DIFF
--- a/packages/styles/table.css
+++ b/packages/styles/table.css
@@ -66,7 +66,6 @@
   outline-offset: unset;
 }
 
-.TableHeader,
 .TableHeader--sort-ascending,
 .TableHeader--sort-descending {
   background: var(--table-header-sorting-background-color);


stephenmathieson (approved):
Thanks!

stephenmathieson:
IIRC the issue I attempted to fix was the background color wasn’t being applied to the selected header on the initial render. May need to circle back after this lands to fix that issue (since we don’t have a way to test it here).

anastasialanz:
> IIRC the issue I attempted to fix was the background color wasn’t being applied to the selected header on the initial render. May need to circle back after this lands to fix that issue (since we don’t have a way to test it here).

Is this something that happens in one of your apps? We could make a new issue for it.

PR #887, opened 2023-02-12
feat(IconImpact): add new IconImpact component

longmatthewh:


scurker:
There's a new RFC to tag along with this: https://github.com/dequelabs/cauldron/issues/1141

scurker:
Closing this for now since the proposed changes are a bit different from what's here now.

PR #894, opened 2023-02-14
feat(styles): Replace variable focus light with a value that has higher color contrast

dequejosie:
@bobbyomari and I replaced the current value of `--focus-light` to solve color contrast issues with a grayer background. We tested this color with all the background colors we use and it passed.


scurker (dismissed):
These two changes shouldn't be grouped together, this should be changing the `--focus-light` color only.

dequejosie:
Because we are actually pulling `--focus-light` from our styles package, not from within?

scurker (approved):

PR #895, opened 2023-02-15
fix(styles): overwrote base button font-family (#888)

### LINKED ISSUE #888: Bug: ExpandCollapse component should use font-family: roboto
### Steps to reproduce

https://cauldron.dequelabs.com/components/ExpandCollapsePanel is not forcing all font within it to be `font-family: roboto`

### Expected Behavior

All font should be `font-family: roboto`

### Actual Behavior

It is not the right font

### Version

I dunno

_No response_

### Relevant log output

_No response_

Additional information

Bracciata:
Added styling to prevent the font-family from being overwritten for elements within a button. 

Closes #888

### DIFF
--- a/packages/styles/base.css
+++ b/packages/styles/base.css
@@ -22,7 +22,7 @@ html {
 }
 
 body {
-  font-family: 'Roboto', Helvetica, Arial, sans-serif;
+  font-family: var(--base-font-family);
   font-style: normal;
   font-weight: 400;
   color: var(--text-color-base);
@@ -59,7 +59,7 @@ ul {
 }
 
 button {
-  font-family: 'Roboto', Helvetica, Arial, sans-serif;
+  font-family: var(--base-font-family);
 }
 
 button,
--- a/packages/styles/base.css
+++ b/packages/styles/base.css
@@ -58,6 +58,10 @@ ul {
   padding: 0;
 }
 
+button {
+  font-family: 'Roboto', Helvetica, Arial, sans-serif;
+}
+
 button,
 [role='button'] {
   display: inline-block;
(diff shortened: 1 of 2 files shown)


scurker:
This is on hold for a moment, @bobbyomari and I are going to discuss the change here to ensure we're implementing it correctly.

packages/styles/base.css
```diff
@@ -58,6 +58,10 @@ ul {
   padding: 0;
 }
 
+button {
+  font-family: 'Roboto', Helvetica, Arial, sans-serif;
```
scurker, line 62: Since this is inherited/shared, we should use a css variable in with the font-family in [`variables.css`](https://github.com/dequelabs/cauldron/blob/develop/packages/styles/variables.css), something like `--base-font-family: 'Roboto', Helvetica, Arial, sans-serif`. 

That can then get applied for the existing `body` selector, and here for the `button`.

scurker (changes requested):

packages/styles/base.css
Bracciata: This change is made

packages/styles/variables.css
```diff
@@ -127,6 +127,9 @@
   /* radio card */
   --radio-card-width: 255px;
   --radio-card-height: 229px;
+
+  /* Font family for the project*/
+  --base-font-family: 'Roboto', Helvetica, Arial, sans-serif;
```
scurker, line 132: Minor nit:
```suggestion
  /* Fonts */
  --base-font-family: 'Roboto', Helvetica, Arial, sans-serif;
```

...and move it near the top since it's a shared global thing.

scurker (approved):

scurker:
🚨 REVIEWED FOR SECURITY 🚨

Issue #850, opened 2023-01-13
Bug: <Breadcrumb> with several items on small layout renders inconsistently


dbowling:
### Steps to reproduce

Add more content to the Breadcrumbs example on the cauldron site and resize viewport to 320px.

### Expected Behavior

Expected the item text to align consistently, perhaps to change the layout to support the given items.

### Actual Behavior

Text does not consistently align because separator is pushed down.

![CleanShot 2023-01-13 at 11 32 50](https://user-images.githubusercontent.com/46435/212393694-b02ae24e-b89a-4025-aa2b-30bba519a4f2.png)

### Version

4.7.0

### Anything else we should know?

Additional information

PR #897, opened 2023-02-16
fix(styles): updated breadcrumbs to wrap at low content consistently

Bracciata:
Updated the breadcrumbs so the separators will stay inline, while also still wrapping for the cases where there is a large number of breadcrumbs. 

Closes: #850 

Below is an example where it is still multiple columns and one where it is shrunk down to one column due to screen width.

breadcrumbs where it is not fully one column due to small screen width
breadcrumbs where it is one column due to small screen width

### DIFF
--- a/packages/styles/breadcrumb.css
+++ b/packages/styles/breadcrumb.css
@@ -4,6 +4,7 @@
   list-style-type: none;
   padding: 0;
   margin: 0;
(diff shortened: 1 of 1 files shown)


scurker (approved):
🚨 REVIEWED FOR SECURITY 🚨

PR #914, opened 2023-02-21
fix: forced tooltips below if they will overflow expected width

### LINKED ISSUE #905: Tooltip: Horizontal scrolling at 320px width
Original Auditor URL: https://axeauditor.dequecloud.com/test-run/f6620faa-6d1e-11ec-a0b7-97f3ccf70959/issue/62507eac-da86-11ec-80f9-57a989c2c515?sortField=identifier&sortDir=asc&page=0&activeTab=dt-issue&pageSize=50&viewMode=issues&row=4&issuesCount%5B0%5D=&issuesCount%5B1%5D=&filter%5Bgroup_id%5D%5B0%5D=42b8f3a2-ac6e-11ed-adb1-3bfa602e0928&filter%5Bgroup_id%5D%5B1%5D=df2b20c0-1317-11ed-acb1-c7cbc98656eb

URL: https://cauldron.dequelabs.com/components/Tooltip

## Description
Content does not reflow to fit horizontally within the viewpoint when page width is adjusted to 320px.

Element 1: Left button- tooltip content cut off

Location: Below Info-tooltip

Element 2: Right button-tooltip content is cut off and it has a horizontal scroll bar.

Location: Below Info-tooltip

Note: For reflow, resize the browser viewport to 640px width by 512px height and zoom in to 200%
### RULE : 
The content MUST be viewable without scrolling horizontally when the viewport is set to 320 CSS pixels wide.

### HOW TO FIX: 
Fix this issue by using ANY of the following techniques:
1. Use CSS techniques such as media queries, grids, or flexbox to reflow content down to 320px width.
2. Provide options within the content to switch to a layout that does not require the user to scroll horizontally to read a line of text.

### REFERENCE: 
WCAG Understanding document:

### BACKGROUND: 
People with low vision often read content on a desktop computer by scaling content using the browser zoom - up to 400%. When zooming content creates the need to scroll in the direction of reading in order to reveal lines that are cut off by the viewport, it significantly increases the effort required to read text and sometimes causes content hidden off-screen with no way to scroll or pan to it. When content is created so that it reflows at screen widths down to 320px, users can view a web page in a 1280px width browser screen and zoom the content to 400% (320px) without having to scroll side to side to read and without losing access to content.

tooltip not positioning itself correctly within cauldron documentation

Bracciata:
Closes #905 

A couple of discussion points for this would be to determine if we can calculate a minimum expected width and if we want to allow the user to define the minimum expected width.


scurker (changes requested):
I'm not sure this is the change that we actually want. 

The auto-placement calculation comes from [`popper`](https://popper.js.org/) so I suspect there may be a bug in placement from that 3rd party library, but it _should_ be trying to doing a best placement via the `preventOverflow` modifier. It might mean we need to upgrade the version if there is an existing bug that has been fixed, or raise an issue with popper but we need to do a little digging first to determine how we want to handle this.

anastasialanz:
Is this PR still relevant?

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -177,6 +180,25 @@ export default function Tooltip({
...
+        setActualPlacement('bottom');
+      }
+    } else if (initialPlacement === 'left') {
+      const offset = elemRect.left;
+      if (offset < tooltipExpectedWidth && actualPlacement !== 'bottom') {
+        placement = 'bottom';
+        setActualPlacement('bottom');
+      }
+    }
+  }
```
anastasialanz, line 201: A fun exercise, how would you write this code now over a year later?

PR #917, opened 2023-02-23
fix: made radio inputs show checked when checked (872)

### LINKED ISSUE #872: Bug: Utilising `<RadioGroup />` swapping radio buttons does not update the `checked` attribute on each `input` element
### Steps to reproduce

Steps to reproduce: 

1. Use the example Radio Group
2. Swap radio buttons
3. Inspect the element `checked` only appears on the first `input` of the group

### Expected Behavior

`checked` attribute to swap to each `input` element on click

### Actual Behavior

 The element `checked` only appears on the first `input` of the group

### Version

latest

Chrome

### Relevant log output

_No response_

Additional information

Bracciata:
Closes #872 
Made radio inputs show checked by adding the current value to the key so the check to see if they are the currently checked element will run after the change is made.


anastasialanz:
Is this PR still relevant?

PR #919, opened 2023-02-23
fix: updated radio and checkboxes so labels wrap below

### LINKED ISSUE #627: Checkbox wrapping issue

Bracciata:
Closes #627 

I updated it so the label will wrap below when needed. I noticed somewhere we mentioned keeping it on the same line and that can be accomplished by disabling wrap. I set it up this way because I want to consider the cases where the label is very long.

checkbox on small screen width with the label wrapping below it


scurker (changes requested):
This just inverts the problem. We want the label text to be on the same line as the checkbox and ideally wrap the label correctly.

Bracciata:
Corrected: 
image

packages/styles/forms.css
```diff
@@ -187,7 +187,7 @@ textarea.Field--has-error:focus:hover,
   margin-bottom: 0;
   flex-direction: row-reverse;
   justify-content: flex-end;
-  flex-wrap: wrap;
+  flex-wrap: nowrap;
```
scurker, line 190: This breaks checkboxes/radios with descriptions along with error states:

checkbox with description on single line before checkbox

We may need to take a more holistic approach into how we construct the checkbox as there's numerous issues when the checkbox wraps here.

scurker (changes requested):

packages/styles/forms.css
Bracciata: The change's made here do not introduce the issue above. I removed them and the issue still exists. I will try to look at a way to incorporate the fix to the above issue into this PR.

Bracciata:
The original issue no longer exists due to a redesign. There is still now the above-discovered issue.

packages/styles/forms.css
```diff
@@ -187,7 +188,15 @@ textarea.Field--has-error:focus:hover,
...
+}
+.column {
+  display: flex;
+  flex-direction: column;
+  width: fit-content;
+}
+.with-label {
+  flex-wrap: nowrap;
 }
```
scurker, line 200: Cauldron styles are global meaning that it is extremely likely that these stills will conflict with already implemented styles in an app, especially given an ambiguous name like `.column` and doesn't really follow our [style guidelines](https://cauldron.dequelabs.com/style-guidelines).

packages/styles/forms.css
```diff
@@ -187,7 +188,15 @@ textarea.Field--has-error:focus:hover,
...
   justify-content: flex-end;
-  flex-wrap: wrap;
+  flex-wrap: nowrap;
+}
+.column {
+  display: flex;
+  flex-direction: column;
+  width: fit-content;
+}
+.with-label {
```
scurker, line 198: Where is this being used?

packages/react/src/components/Checkbox/index.tsx
```diff
@@ -72,58 +72,62 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
...
-            id={id}
-            ref={typeof refProp === 'function' || !refProp ? checkRef : refProp}
-            type="checkbox"
-            checked={isChecked}
-            disabled={disabled}
-            onFocus={(e): void => {
-              setFocused(true);
-              if (typeof onFocus === 'function') {
-                onFocus(e);
+        <div className="column">
```
scurker, line 75: I _think_ this looks like you're trying to wrap the checkbox to style it separately from the description, correct?

We have the odd ordering of elements here to support being able to target the label field following a checkbox. I'm wondering if we can use invert the order to allow for better text wrapping and instead use `:has` to target the appropriate elements. We would need to see how this would look in Firefox given that it doesn't yet fully support `:has` but we might be okay with some minor differences.

scurker (changes requested):

scurker:
This was fixed with #1110

PR #920, opened 2023-02-23
fix: added icon size variable and applied to current icons

### LINKED ISSUE #515: Set `--icon-size` as a css var
This should be the default icon size (currently `24px`)

Bracciata:
Closes #515

### DIFF
--- a/packages/styles/dialog.css
+++ b/packages/styles/dialog.css
@@ -80,8 +80,8 @@
 }
 
 .Dialog__close svg {
-  height: 24px;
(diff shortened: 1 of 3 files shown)


scurker (approved):
LGTM.

In the future, would you mind creating the PR from a branch in the repo? Right now, we don't currently support build previews via forked branches so it makes it more difficult to review.

scurker:
🚨 REVIEWED FOR SECURITY 🚨

PR #923, opened 2023-02-28
docs: add breaking changes definition

### LINKED ISSUE #881: Define a breaking change
# What is a Breaking Change?

Cauldron provides a public API that consists of two parts, React components and CSS styles, each of which has its own package. In principle, any change to these two parts of API that is backward incompatible should be marked as a breaking change, and we should increment the major version number whenever a breaking change takes place (see [Semantic Versioning Specification (SemVer)](https://semver.org/)).

However there is a debate on whether to mark as breaking certain changes to Cauldron CSS styles with no visual impact (such as removing or renaming a classname), or whether to increment the major version number assuming that those CSS styles changes are breaking. There are a few reasons behind the hesitance to mark these changes as breaking and bump the major version:

1. we might break our API too frequently and that would decrease Cauldron's stability and risk fatiguing our users

2. we assume that the majority of Cauldron users do not consume CSS styles separately from their React components (although we allow our users to do so)

3. we have not been marking the changes to remove or rename a classname as breaking

The counterpoint to the first concern is that major version numbers are not sacred, as the creator of `SemVer` spec [defends releasing a new major version only about one month after the previous one](https://tom.preston-werner.com/2022/05/23/major-version-numbers-are-not-sacred.html). The proponents of `SemVer` spec also point out that we do not have any data to show how exactly our users utilize Cauldron. Therefore the second assumption could be right, but we have no proof for it and we have no control over how our users consume Cauldron. The last point is dictated by how we agree on the first two.

One key to the debate is whether we would want to (or continue to) support the consumption of Cauldron CSS styles separately from Cauldron React components, although that was the original intention for Cauldron. The following definitions and examples of breaking and non-breaking changes are based on the original intention. At the same time, the notes explore different scenarios if we decide against the original intention.

## Cauldron React components: breaking changes

1. Remove, replace, rename a component, as well as modify the relations between components

   - renamed Panel to IssuePanel
   - TabPanels are no longer TabList's children

2. Add, remove, replace, or rename required properties of a React component

   - removes the light/dark variant options for IconButton and TopBar.
   - The panel implementation has been replaced, including removing props that are not used in the new implementation
   - Loader component now set role, aria-valuetext, aria-busy, aria-valuemin, and aria-valuemax instead of making a generic role (div) have an aria-label

3. Modify the elements of which a React component consists, such as HTML tags and properties
   - selects now have parity with `<TextField />`s in that they always render an .Error div. This causes a slight layout difference with more space below each `<Select />`. The text "Required" will now show up with any `<Select />` who is passed a true required prop
   - topbar no longer renders role=menubar. New component `<MenuBar />` must now be used if menubar is desired
   - The api for for implementing a tooltip has changed. Instead of `<Tooltip><button /><Tooltip>`, the tooltip takes a target ref or html element. (This is a combination of breaking change #2 and #3)

## Cauldron CSS styles: breaking changes

1. Remove or rename a CSS classname

2. Change the value of a CSS classname that results in visual difference (see examples in [Visual Breaking Change in Design Systems](https://medium.com/eightshapes-llc/visual-breaking-change-in-design-systems-1e9109fac9c4))

   **Note**: if we decide not to support the consumption of CSS styles separately from the React components, then we should at least mark the shared classnames as breaking (`packages/styles/variables`)

## What is not a Breaking Change?

1. Add a React component or CSS classname

2. Add optional properties to a React component

3. Change the value of a CSS classname that does not result in visual difference

4. Change the areas that are not a part of Cauldron's public API, such as docs, demos, and tests

scurker:
closes #881

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,6 +120,20 @@ For cauldron's component library, any of the following would be considered break
 * Changing a prop that would break usage of an existing component
 * Adding a new required prop to an existing component
 
+Some examples of implementing non-breaking changes for components:
+
+#### Aliasing component name in export, supporting both component names
+
+```js
+export { NewName, OldName: NewName }
+```
+
+#### Aliasing a component prop, supporting both property names
+
+```js
+function Component({ oldProp: newProp, newProp }) { ... }
+```
+
 ### Styles
 
 For cauldron's styles library, any of the following would be considered breaking changes:
@@ -127,6 +141,26 @@ For cauldron's styles library, any of the following would be considered breaking
 * Changing or removing a class name
 * Changing or removing a global css custom property name (see `variables.css`)
 * Changing or removing a component css custom property name
+
+Some examples of implementing non-breaking changes for styles:
+
+#### Aliasing CSS class, supporting both class names
+
+```css
+/* Usage of .Foo is deprecated, please use .Bar */
+.Foo,
+.Bar {
+  ...
+}
+```
+
+#### Utilizing new css custom variable for new property name
+
+```css
+/* Usage of --red is deprecated, please use --better-red */
+--red: var(--better-red);
+--better-red: red;
+```
 
 In addition, changing a css custom property may be a breaking change for the following:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,7 +137,7 @@ For the latter, the change must be reviewed by a member of the cauldron team to 
 
 ### Deprecating
 
-Before a breaking change can be released, the breaking change should be documented with the component, property, or style being changed getting flagged as _Deprecated_. This could meaning warning the consumer that a component or property is now deprecated, or including a deprecated comment next to a css class name. This deprecation must exist for at least two months.
+Before a breaking change can be released, the breaking change should be documented with the component, property, or style being changed getting flagged as _Deprecated_. This could mean warning the consumer that a component or property is now deprecated, or including a deprecated comment next to a css class name. This deprecation must exist for at least two months.
 
 ### Removal
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,3 +107,38 @@ Icons are located in [packages/react/src/components/Icon/icons](./packages/react
 - Remove any namespace attributes (such as `xmlns:*`) from the SVG.
 - Save your icon in the above location with the proposed name of your icon, e.g. `[your-name].svg`. The `your-name` portion will be how the icon is used in the React Component, e.g. `<Icon type="your-name" />`.
 - Submit a new `feat` PR with your changes.
+
+## Breaking Changes
+
+Cauldron is constantly evolving, and it's inevitable that breaking changes will need to be introduced. Before a breaking change is released, other alternatives should be considered and allow for a deprecation period before the breaking change is released.
+
+### Components
+
+For cauldron's component library, any of the following would be considered breaking changes:
+
+* Renaming (exported name) or removing a Component
+* Changing a prop that would break usage of an existing component
+* Adding a new required prop to an existing component
+
+### Styles
+
+For cauldron's styles library, any of the following would be considered breaking changes:
+
+* Changing or removing a class name
+* Changing or removing a global variable (see `variables.css`)
+* Changing or removing a component variables
+
+In addition, changing a css property may be a breaking change for the following:
+
+* Changing a global variable property (see `variables.css`)
+* Changing a component variable property
+
+For the latter, the change must be reviewed by a member of the cauldron team to ensure there are no breaking changes.
+
+### Deprecating
+
+Before a breaking change can be released, the breaking change should be documented with the component, property, or style being changed getting flagged as _Deprecated_. This could meaning warning the consumer that a component or property is now deprecated, or including a deprecated comment next to a css class name. This deprecation must exist for at least two months.
+
+### Removal
+
+Once the deprecation period has passed, the breaking change can be implemented by removing the deprecation and including `BREAKING CHANGE:` in the body of your commit message with the description of the breaking change.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,19 +125,19 @@ For cauldron's component library, any of the following would be considered break
 For cauldron's styles library, any of the following would be considered breaking changes:
 
 * Changing or removing a class name
-* Changing or removing a global variable (see `variables.css`)
-* Changing or removing a component variable
+* Changing or removing a global css custom property name (see `variables.css`)
+* Changing or removing a component css custom property name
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -107,3 +107,38 @@ Icons are located in [packages/react/src/components/Icon/icons](./packages/react
...
+* Changing a prop that would break usage of an existing component
+* Adding a new required prop to an existing component
+
+### Styles
+
+For cauldron's styles library, any of the following would be considered breaking changes:
+
+* Changing or removing a class name
+* Changing or removing a global variable (see `variables.css`)
+* Changing or removing a component variables
```
thuey, line 129: nit: It might be good to include an example of what a "component variable" is.

```suggestion
* Changing or removing a component variable
```

CONTRIBUTING.md
```diff
@@ -107,3 +107,38 @@ Icons are located in [packages/react/src/components/Icon/icons](./packages/react
...
+
+In addition, changing a css property may be a breaking change for the following:
+
+* Changing a global variable property (see `variables.css`)
+* Changing a component variable property
```
thuey, line 134: This is a little unclear to me what you are referring to.

> In addition, changing a css property may be a breaking change for the following:

Sounds like a CSS property value is changed maybe? (e.g. `color: blue; -> color: red;`)

> Changing a global variable property

This sounds like changing the value of a CSS variable (e.g. `--background-color: blue; -> --background-color: red;`. And if that's the case, how is that different than what is called out above "Changing or removing a global variable". Or is what is called out above more about "changing the name of" instead of "changing the value of"?

CONTRIBUTING.md
```diff
@@ -107,3 +107,38 @@ Icons are located in [packages/react/src/components/Icon/icons](./packages/react
...
+
+For the latter, the change must be reviewed by a member of the cauldron team to ensure there are no breaking changes.
+
+### Deprecating
+
+Before a breaking change can be released, the breaking change should be documented with the component, property, or style being changed getting flagged as _Deprecated_. This could meaning warning the consumer that a component or property is now deprecated, or including a deprecated comment next to a css class name. This deprecation must exist for at least two months.
```
thuey, line 140:
```suggestion
Before a breaking change can be released, the breaking change should be documented with the component, property, or style being changed getting flagged as _Deprecated_. This could mean warning the consumer that a component or property is now deprecated, or including a deprecated comment next to a css class name. This deprecation must exist for at least two months.
```

thuey: Also, this section seems to make the assumption that breaking changes entail a deprecation, but sometimes the deprecation is additive (e.g. a new required prop). What is the process there?

thuey (commented):
Great to see this moving forward!

CONTRIBUTING.md
scurker: It's the property portion of a css variable. Changing the name will always be a breaking change because someone consuming cauldron could be using that variable name in their code, so changing a name from `--foo` to `--bar` would mean that css property using that variable would become unset.

Changing the property value of a css variable isn't as clear cut which is why the additive section is slightly different and requires a more in-depth review from a member of the cauldron team. A good example of this is that we recently changed the global `--focus` property value to better contrast against all of our standard background colors. This was not a breaking change as we already identified that we has some low contrast accessibility issues within cauldron. But that's not to say there aren't other property values that would be a breaking change, thus requiring a more in-depth review before allowing.

CONTRIBUTING.md
scurker: Good question. In this instance you would need to add the new prop as non-required and warn that in the future that property will become required. Alternatively, don't make the property required but set a default value if not set (if possible).

CONTRIBUTING.md
```diff
@@ -107,3 +107,38 @@ Icons are located in [packages/react/src/components/Icon/icons](./packages/react
...
+
+### Removal
```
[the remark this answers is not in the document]
scurker: This is intentional, deprecating/removal are 2 different stages of the process.

CONTRIBUTING.md
thuey: > Good question. In this instance you would need to add the new prop as non-required and warn that in the future that property will become required. Alternatively, don't make the property required but set a default value if not set (if possible).

Can we document that?

CONTRIBUTING.md
thuey: If I understood correctly, then IMO, this would be more clearly written as:

```suggestion
In addition, changing a css variable value may be a breaking change for the following:

* Changing a global variable value (see `variables.css`)
* Changing a component variable value
```

Since a variable IS a custom property, but it's the value of the variable/custom property that is being referred to here.

scurker: Official documentation calls them _css custom properties_, but I can be more specific here.

thuey: I'm just trying to distinguish between the _name_ of the custom property/variable (e.g. `--focus`) and the _value_ of that variable (e.g. `blue`)

scurker: Yes, I updated the words here 🙃

CONTRIBUTING.md
```diff
@@ -107,3 +107,38 @@ Icons are located in [packages/react/src/components/Icon/icons](./packages/react
...
+In addition, changing a css custom property may be a breaking change for the following:
+
+* Changing a global css custom property value (see `variables.css`)
+* Changing a component css custom property value
+
+For the latter, the change must be reviewed by a member of the cauldron team to ensure there are no breaking changes.
+
+### Deprecating
+
+Before a breaking change can be released, the breaking change should be documented with the component, property, or style being changed getting flagged as _Deprecated_. This could mean warning the consumer that a component or property is now deprecated, or including a deprecated comment next to a css class name. This deprecation must exist for at least two months. If a change is additive (e.g. making a new property required), the new property must be optional until the deprecation period has passed.
```
thuey, line 140: > If a change is additive (e.g. making a new property required), the new property must be optional until the deprecation period has passed.

nit: It seems odd to find this under a heading for "Deprecating" and calling it a "deprecation period", when a new required property isn't a deprecation

thuey:
I left the nit, but otherwise, LGTM. I can't approve, though :)

thuey (dismissed):

CONTRIBUTING.md
scurker: Technically it is, because if you want to add a required property you're deprecating usage of the component without the required prop.

I think we should be able to avoid this scenario for the most part, I'm trying to envision the scenarios where we would add a new required prop and struggled to come up with good examples unless we completely missed something during the initial development of a component.

thuey: > I think we should be able to avoid this scenario for the most part, I'm trying to envision the scenarios where we would add a new required prop and struggled to come up with good examples unless we completely missed something during the initial development of a component.

Yeah, I think this scenario is unusual.

Issue #926, opened 2023-03-03
Tag component restricts the content width


thuey:
The Tag component restricts the width of the content to 90px: https://github.com/dequelabs/cauldron/blob/97d52033e140d2428c2c37768f1ad9cc46d68d93/packages/styles/tag.css#L25

This can create accessibility issues when text is scaled or with translations. Width should adjust to the content.

PR #927, opened 2023-03-03
fix(styles): allow Tag to adjust width based on content

thuey:
Closes https://github.com/dequelabs/cauldron/issues/926


scurker (approved):
LGTM

Issue #928, opened 2023-03-03
Two column layout should allow allow horizontal scrolling


thuey:
When using elements like `Code`, which have horizontal scrolling, inside a `TwoColumnLayout`, horizontal scrolling doesn't actually kick in and the `Code` element overflows the container. In order for the horizontal scrolling to kick in, the `TwoColumnLayout` should set `overflow-x: auto`

scurker:
The desired behavior should be:

The right container should have an x overflow if items extend past the width of that container.

thuey:
I had to revert the changes. See: https://github.com/dequelabs/cauldron/pull/943#issuecomment-1464217352

tldr; the focus indicators were getting cut off

PR #929, opened 2023-03-03
fix(styles): allow two column panel content to scroll

thuey:
Closes https://github.com/dequelabs/cauldron/issues/928

### DIFF
--- a/docs/patterns/components/TwoColumnPanel/index.js
+++ b/docs/patterns/components/TwoColumnPanel/index.js
@@ -126,6 +126,7 @@ function TwoColumnPanelDemo() {
               Lorem ipsum dolor sit amet, consectetur adipiscing elit.
               Pellentesque efficitur sollicitudin ornare.
             </p>
+            <Code>{`<div>example of an element that has horizontal scrolling</div>`}</Code>
           </div>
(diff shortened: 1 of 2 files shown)


docs/patterns/components/TwoColumnPanel/index.js
```diff
@@ -126,6 +126,10 @@ function TwoColumnPanelDemo() {
               Lorem ipsum dolor sit amet, consectetur adipiscing elit.
               Pellentesque efficitur sollicitudin ornare.
             </p>
+            <Code
+              role="region"
+              tabIndex="0"
+            >{`<div>example of an element that has horizontal scrolling</div>`}</Code>
```
scurker, line 132: I don't think we necessarily need an example here of an element that scrolls (although it did make it easier for me to verify).

scurker (dismissed):

docs/patterns/components/TwoColumnPanel/index.js
thuey: Yeah, I had a similar thought, but wanted to leave it in for verification. Happy to pull out, though.

scurker (approved):

PR #936, opened 2023-03-08
fix(NavBar): adjust styling for nav items

longmatthewh:
As requested by @bobbyomari I made the following adjustments to the NavItem css:
 - use `--font-weight-normal` for nav items unless they're active, in which case, use `--font-weight-medium`

### DIFF
--- a/packages/styles/navbar.css
+++ b/packages/styles/navbar.css
@@ -50,6 +50,18 @@
   transition: background-color 0.3s ease;
 }
 
+.NavBar > ul > .NavItem {
+  font-weight: var(--font-weight-normal);
+}
+
+.NavBar > ul > .NavItem--active {
+  font-weight: var(--font-weight-medium);
+}
+
+.NavBar > ul > .NavItem a {
+  font-weight: inherit;
+}
+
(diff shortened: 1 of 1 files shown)


JoshMK (dismissed):
👏

packages/styles/navbar.css
```diff
@@ -48,6 +48,19 @@
   border-bottom: 1px solid var(--top-nav-border-bottom-color);
   border-right: 1px solid var(--top-nav-box-shadow-color);
   transition: background-color 0.3s ease;
+  text-transform: uppercase;
```
scurker, line 51:
```suggestion
```

scurker (changes requested):
I discussed this with @bobbyomari. General research has suggested that some people have more difficulty reading all capital letters vs not:

- https://www.a11yproject.com/posts/how-to-accessible-heading-structure/#all-caps
- https://www.w3.org/TR/low-vision-needs/#capitalization

While not officially a wcag violation, we should probably err on the side of caution and try to follow best practices as best as we can.

@bobbyomari said we will likely make updates in UX pin to match this. Everything else looks good.

scurker (approved):

Issue #703, opened 2022-07-20
Pagination component loses focus when start or end of pagination is reached


straker:
The bowser focus jumps back to the `body` element when the user is using the keyboard to navigate the pagination and reaches either the start or the end (using any of the buttons). Focus should remain on the button when the start / end of the pagination is reached.

ryomtoob:
This component uses a combination of `<IconButton>` and `<TooltipTabstop>`. This is so that the disabled pagination buttons can still be focused and display a tooltip indicating direction. When the start or end is reached, the `<IconButton>` is swapped for a `<TooltipTabstop>` which cause focus to be lost. To fix this, focus should be placed on the appropriate `<TooltipTabstop>` once the user reaches the start or end of pagination.

joanjeremiah:
I would like to solve this issue. Kindly assign this task to me.

anastasialanz:
@joanjeremiah I assigned this one to you. Thanks!

VaniChinnam:
Verified in Cauldron - https://cauldron.dequelabs.com/components/Pagination
The focus order is proper. On activating any of the pagination links, the focus stays on that button itself.

PR #937, opened 2023-03-09
fix(react): maintain focus on pagination button when last or first page is reached

thuey:
Closes https://github.com/dequelabs/cauldron/issues/703

Trying to manage focus when child components are replaced seems error-prone. Instead, I got rid of the tabstop component and made the button component behave like the tabstop component when disabled.

### DIFF
--- a/docs/Demo/index.js
+++ b/docs/Demo/index.js
@@ -52,16 +52,18 @@ const Demo = props => {
               DEMO_renderAfter,
               DEMO_renderBefore,
               DEMO_hide_renderAfterBefore = false,
+              DEMO_key,
               ...thinState
             } = state;
             const componentMarkup = renderState(thinState);
+            const key = DEMO_key || componentMarkup;
             const afterMarkup =
               DEMO_renderAfter &&
               !DEMO_hide_renderAfterBefore &&
               jsxStringify(DEMO_renderAfter, stringifyConfig);
 
             return (
-              <div key={componentMarkup}>
+              <div key={key}>
                 {DEMO_renderBefore}
                 <Component {...thinState} />
                 {DEMO_renderAfter}
--- a/packages/react/__tests__/src/components/Pagination/index.js
+++ b/packages/react/__tests__/src/components/Pagination/index.js
@@ -28,29 +28,39 @@ describe('Pagination', () => {
           wrapper
             .find('button')
             .at(0)
-            .is('[aria-disabled="true"]')
-        ).toBe(true);
+            .getDOMNode()
+            .getAttribute('aria-disabled')
+        ).toBe('true');
 
         expect(
           wrapper
             .find('button')
             .at(1)
-            .is('[aria-disabled="true"]')
-        ).toBe(true);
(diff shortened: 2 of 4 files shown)


docs/Demo/index.js
```diff
@@ -52,16 +52,18 @@ const Demo = props => {
...
             const componentMarkup = renderState(thinState);
+            const key = DEMO_key || componentMarkup;
             const afterMarkup =
               DEMO_renderAfter &&
               !DEMO_hide_renderAfterBefore &&
               jsxStringify(DEMO_renderAfter, stringifyConfig);
 
             return (
-              <div key={componentMarkup}>
+              <div key={key}>
```
thuey, line 66: Focus wasn't being maintained at all when paging, whether you were on the last page or not. That's because the DOM was getting remounted every time the markup changed. I added the ability to provide a key

packages/react/__tests__/src/components/Pagination/index.js
```diff
@@ -28,29 +28,39 @@ describe('Pagination', () => {
           wrapper
             .find('button')
             .at(0)
-            .is('[aria-disabled="true"]')
-        ).toBe(true);
+            .getDOMNode()
+            .getAttribute('aria-disabled')
+        ).toBe('true');
```
thuey, line 33: Enzyme struggles when you set `aria-disabled={true | false}` (boolean). Before, the TooltipTabstop was setting `aria-disabled="true"`.

packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -184,7 +130,8 @@ Pagination.propTypes = {
   onLastPageClick: PropTypes.func,
   // @ts-expect-error
   tooltipPlacement: PropTypes.string,
-  className: PropTypes.string
+  className: PropTypes.string,
+  thin: PropTypes.bool
```
thuey, line 134: Not part of this PR, but a linting issue was raised with `thin` not being defined on prop types

scurker (approved):
It looks like the styles did change slightly, but I don't think that's a flaw with what you did. It looks like the active state has a background color while in dark mode for disabled pagination elements which probably shouldn't be happening.

packages/react/__tests__/src/components/Pagination/index.js
```diff
@@ -28,29 +28,39 @@ describe('Pagination', () => {
...
           wrapper
             .find('button')
             .at(1)
-            .is('[aria-disabled="true"]')
-        ).toBe(true);
+            .getDOMNode()
+            .getAttribute('aria-disabled')
+        ).toBe('true');
 
+        console.log(
```
stephenmathieson, line 43: 👀

thuey: Good 👀 . I'll fix it. At least it was just a test file

thuey: https://github.com/dequelabs/cauldron/pull/941

PR #942, opened 2023-03-10
fix: fix overflow behavior of Select component

### LINKED ISSUE #579: Select control is not properly visible at 320px width
<dl><dt>Description</dt><dd>Content or functionality is lost due to overlapping content when page width is adjusted to 320px.
1. Select control caret key is not visible at 320px width for all such controls on page.
2. List popup does not fit to screen at 320px.</dd><dt>Impact</dt><dd>Moderate</dd><dt>Recommendation</dt><dd>RULE : 

HOW TO FIX: 

REFERENCE: 
WCAG Understanding document:

BACKGROUND: 
People with low vision often read content on a desktop computer by scaling content using the browser zoom - up to 400%. When zooming causes content to overlap other content, users with low vision may not be able to see or use that content. When content is created so that it reflows without losing information or functionality at screen widths down to 320px, users can view a web page in a 1280px width browser screen and zoom the content to 400% (320px) without having to scroll side to side to read and without losing access to content.</dd><dt>Source code</dt><dd><pre><code><select id="x_80_6128"><option class="Field__option" value="Yes">Yes</option><option class="Field__option" value="No">No</option><option class="Field__option" value="Maybe">Maybe</option></select></code></pre></dd><dt>Checkpoint</dt><dd>Reflow (1.4.10.a)</dd><dt>URL</dt><dd>https://cauldron.dequelabs.com/components/Select</dd></dl>

yhafez:
Closes #579

### DIFF
--- a/packages/styles/select.css
+++ b/packages/styles/select.css
@@ -83,3 +83,9 @@
 .Field__select--wrapper.Field--has-error select:focus:hover {
   border-color: var(--field-border-color-error-hover);
 }
+
+@media (max-width: 23rem) {
+  .Field__select--wrapper {
+    min-width: auto;
+  }
+}
--- a/packages/styles/select.css
+++ b/packages/styles/select.css
@@ -16,7 +16,8 @@
 
 .Field__select--wrapper {
   position: relative;
-  min-width: var(--select-width);
+  width: var(--select-width);
+  max-width: 100%;
 }
 
 .arrow-down {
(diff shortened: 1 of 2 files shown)


packages/styles/select.css
```diff
@@ -83,3 +83,9 @@
 .Field__select--wrapper.Field--has-error select:focus:hover {
   border-color: var(--field-border-color-error-hover);
 }
+
+@media (max-width: 23rem) {
```
scurker, line 87: I don't think this is the change that we need.

What if the component is inside of a smaller container on a larger viewport? It could still overflow. I believe the desired behavior should be we need to change `min-width` to `width` and then set a `max-width: 100%` so the select element can responsively flex.

scurker (changes requested):

packages/styles/select.css
yhafez: That works!

scurker (approved):
🎉

PR #944, opened 2023-03-10
docs: start implementing V2 of cauldron documentation

scurker:
This is a V2 version of the docs with the following enhancements:

- Support documentation using [mdx](https://mdxjs.com/)
- Support using typescript in docs components
- Additional pages can be included in docs under `docs/pages/*.mdx`
- Enhanced component documenation
  - V2 documentation will use mdx under `docs/pages/components/*.mdx`
  - V1 documentation will continue to exist under `docs/components` until replaced by V2 documentation
- Components & pages will automatically get included in collections (no more manual adding component names to an array 🎉 )
- Live rendering of examples with special markdown code fences, no separate `Demo/Code` components needed
- Refactoring of docs layout to better support some of the above features


docs/collections.ts
```diff
@@ -0,0 +1,88 @@
...
+
+interface Collections {
+  pages: Collection[];
+  components: Collection<{ deprecated?: boolean }>[];
+  componentsV1: Collection[];
+}
+
+// Collections
+const collections: Collections = (require as any)
+  .context('./', true, /\.(mdx|jsx?)$/)
```
isner, line 18: Neat 👍

docs/components/CopyToClipboardButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  useEffect(() => {
+    // We don't know what context this button will be included in, so providing
+    // a minimal accessible name of "example, x of y"
+    const elements = Array.from(
+      document.querySelectorAll('[data-copy-example]')
+    );
+    const index = elements.findIndex(element => element === ref.current);
+    if (index !== -1 && elements.length) {
+      setAccessibleName(`${label} example, ${index + 1} of ${elements.length}`);
```
schne324, line 56: The accName of these buttons will be

> copy to clipboard example, 1 of 5

Isn't that a bit confusing? Shouldn't it be something like "copy code example to clipboard"?

docs/components/CopyToClipboardButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+  });
```
schne324, line 58: There aren't any dependencies we could add here to avoid an effect on every render?  Maybe `value`?

schne324 (changes requested):
Just one question about the accessible name of the `<CopyToClipboardButton />` component

docs/components/CopyToClipboardButton.tsx
scurker: Then they would all have the same accessible name, isn't that something that we try to avoid as a best practice? I'm open to alternatives.

scurker: Are you asking about the phrasing or the index numbers?

docs/components/CopyToClipboardButton.tsx
scurker: I realize updating every render is probably overkill, but I wanted to be sure that the _x of y_ value was always up to date with whatever other components were rendered on the page. It seems to work fine with `[value]` though 🤷

docs/components/ComponentProps.tsx
```diff
@@ -0,0 +1,231 @@
...
+              </DescriptionListItem>
+              <DescriptionListItem>
+                <DescriptionTerm>Description</DescriptionTerm>
+                <DescriptionDetails>{description}</DescriptionDetails>
+              </DescriptionListItem>
+            </DescriptionList>
+          </li>
+        )
+      )}
+      {children && (
```
thuey, line 169: I'm guessing this is supposed to be:
```suggestion
      {ref && (
```

thuey (dismissed):

docs/collections.ts
```diff
@@ -0,0 +1,96 @@
...
+  componentsV1: Collection[];
+}
+
+/*
+ * Maps specific components/mdx to collections:
+ *  - Pages collection is for generic documentation (like getting started, or principles)
+ *  - Components collection is for all component MDX documentation
+ *  - ComponentsV1 collection is for all current (js) documentation
+ *
+ * These collections are can be used to automatically import/sort components and display
```
dequejosie, line 22: nit: no `are` here.

docs/pages/intro.mdx
```diff
@@ -0,0 +1,55 @@
...
+#### Peer Dependencies
+
+There are a few additional packages that you will need to install:
+
+- `react` at `16.x` or higher
+- `react-dom` at `16.x` or higher
+
+### Setup
+
+Once installed, you will need to add include cauldron's styles somewhere in the top level of your application. You will also need to include the required font family of Roboto with something like [google fonts](https://fonts.google.com/specimen/Roboto) or [fontsource](https://fontsource.org/fonts/roboto):
```
kethomas, line 40: could we show an example of how we generally do this? (aka the npm package and importing specific weight CSS)

docs/components/ComponentProps.tsx
scurker: Yup, good catch!

docs/pages/intro.mdx
scurker: I didn't want to be _too_ descriptive since it's dependent on how your application is setup and may differ, but I updated this to at least include a baseline example of using fontsource.

docs/components/CopyToClipboardButton.tsx
schne324: @scurker on slack:

> Jason Wilson: I changed it to “copy code example to clipboard, x of y” if that makes more sense
> Harris: Yes that makes more sense

schne324 (approved):

Issue #947, opened 2023-03-14
Wordwrap and width to Tooltip


michael-siek:
In the event that the there is a long text with no spaces associated to the tooltip the tooltip should have word-break and also have a max-width so that it doesn't span the entire page

PR #948, opened 2023-03-14
fix: allow word break and add max width to tooltip

michael-siek:
Closes: #947


docs/patterns/components/Tooltip/index.js
```diff
@@ -42,6 +42,13 @@ const DemoTooltip = () => {
             description:
               'Manually control the show state of the tooltip. (Note: there is a known issue where only the initial value of `show` is respected.)'
           },
+          wordwrap: {
```
scurker, line 45: The one issue with this is that not every component has direct control over its tooltip(s):

- `IconButton`
- `TooltipTabstop`
- `TwoColumnPanel`
- `Stepper`
- `Pagination`

If I'm an author using an `IconButton` that has a long name that text overflows, I have no way of controlling this property unless we make it pass-though for component that uses it as well. I would rather err on the side of caution and make the tooltip break word for everything to prevent accessibility issues / overflow and we could add a prop later to make it opt-out if needed.

scurker (changes requested):

docs/patterns/components/Tooltip/index.js
michael-siek: The problem with adding `word-break: break-all;` set by default is in this instance 
image

scurker: I think we want to use `break-word` here instead of `break-all`. That way it'll force a break on unbreakable words, but introduce natural breaks where necessary.

scurker (approved):

Issue #786, opened 2022-09-20
TwoColumnPanel: content overlap at 200% zoom


schne324:
https://cauldron.dequelabs.com/components/TwoColumnPanel

<img width="517" alt="two column panel component's content overlapping with other content at 200% zoom" src="https://user-images.githubusercontent.com/3180185/191329693-4dad32b6-5dd1-4437-a302-098cc0984125.png">

scurker:
This is partially fixed by #897, we need to adjust to allow for the header not not have a fixed height when the text in the column wraps. This should be taken into consideration when #889 is addressed.

PR #952, opened 2023-03-15
fix(react): Make TwoColumnPanel wrap to prevent content from getting cut out

dequejosie:
closes https://github.com/dequelabs/cauldron/issues/786

### DIFF
--- a/packages/styles/two-column-panel.css
+++ b/packages/styles/two-column-panel.css
@@ -33,6 +33,7 @@
 }
 
 .TwoColumnPanel__Header {
+  /* The -2px is to account for the borders that are outside of the element and will make it line up with the left panel when there's only one line of text */
   padding: calc(var(--space-small) - 2px);
   border-bottom: 1px solid var(--two-column-panel-border-color);
   min-height: var(--two-column-panel-header-height);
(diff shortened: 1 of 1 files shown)


packages/styles/two-column-panel.css
```diff
@@ -176,12 +176,16 @@
   margin-bottom: 0;
 }
 
+.TwoColumnPanel__ButtonToggle {
+  display: flex;
+}
```
dequejosie, line 181: make ButtonToggle the same height as header

scurker (dismissed):
Great work @dequejosie!

Just one small tweak is needed to make the padding consistent when it wraps. Since the line-height was removed, we need to set the padding for `.TwoColumnPanel__Header` so it doesn't look squished when content wraps:

```css
padding: calc(var(--space-small) - 2px);
```

The `-2px` is to account for the borders that are outside of the element and will make it line up with the left panel when there's only one line of text.

packages/styles/two-column-panel.css
```diff
@@ -176,12 +177,16 @@
...
+
 .TwoColumnPanel__ButtonToggle button,
 .TwoColumnPanel__Close button {
   background-color: transparent;
   border: none;
   border-bottom: 1px solid var(--two-column-panel-border-color);
-  height: var(--two-column-panel-header-height);
+  min-height: var(--two-column-panel-header-height);
```
dequejosie, line 189: This does not change Close button, so I just added here to make Toggle button to expand.

dequejosie:
Screen Shot 2023-03-15 at 11 31 59 AM

scurker (approved):

PR #953, opened 2023-03-15
feat(icon): add hashtag icon

Zidious:
This PR adds a new hashtag icon - 
image


scurker (changes requested):

scurker:
Also FYI, cauldron doesn't like `/` in the branch name - you can't see deploy previews.

scurker (approved):

Issue #955, opened 2023-03-15
Allow Pagination label props to take a ReactNode


thuey:
Currently, the `Pagination` component's label props (e.g. `firstPageLabel`, `previousPageLabel`, `nextPageLabel`, and `lastPageLabel`) require a string to be passed in: https://github.com/dequelabs/cauldron/blob/8d67f3f39d75af259bea0e88437da019358af74e/packages/react/src/components/Pagination/Pagination.tsx#L12C2-L15

It should allow `React.ReactNode` to be passed in, like `statusLabel` does so that more complex elements can be passed in. This would allow you to include screen reader specific text, for example:

```tsx
previousPageLabel={<span>Previous <Offscreen>Issue</Offscreen></span>}
```

PR #956, opened 2023-03-15
feat(react): allow pagination labels to take react nodes

thuey:
Closes https://github.com/dequelabs/cauldron/issues/955

### DIFF
--- a/packages/react/src/components/Pagination/Pagination.tsx
+++ b/packages/react/src/components/Pagination/Pagination.tsx
@@ -9,10 +9,10 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
   itemsPerPage?: number;
   currentPage?: number;
   statusLabel?: React.ReactNode;
-  firstPageLabel?: string;
-  previousPageLabel?: string;
-  nextPageLabel?: string;
-  lastPageLabel?: string;
+  firstPageLabel?: React.ReactNode;
+  previousPageLabel?: React.ReactNode;
+  nextPageLabel?: React.ReactNode;
+  lastPageLabel?: React.ReactNode;
   onNextPageClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Pagination/Pagination.tsx
```diff
@@ -9,10 +9,10 @@ interface Props extends React.HTMLAttributes<HTMLDivElement> {
   itemsPerPage?: number;
   currentPage?: number;
   statusLabel?: React.ReactNode;
-  firstPageLabel?: string;
-  previousPageLabel?: string;
-  nextPageLabel?: string;
-  lastPageLabel?: string;
+  firstPageLabel?: React.ReactNode;
```
scurker, line 12: I think `React.ReactNode` is more inclusive of what we actually want:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/59c89cc048de1e795eb3c88108fcb982ee4674ae/types/react/v16/index.d.ts#L235

I don't think we want people to be able to pass in `boolean`, `null`, or `undefined`. What do you think?

scurker: Hmm... that would be a breaking change for that component. I think `Omit<ReactNode, boolean | null | undefined>` is probably what we want here (and elsewhere). If we're going to use a more loose type, we should ensure that it's helping us catch edge case issues. Do you mind creating a ticket for `IconButton` and see if we can implement the constrained type here?

thuey: Sure. Do you want a separate ticket for [statusLabel](https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/Pagination/Pagination.tsx#L11), too?

scurker (approved):

packages/react/src/components/Pagination/Pagination.tsx
thuey: https://github.com/dequelabs/cauldron/issues/958
https://github.com/dequelabs/cauldron/issues/957

Issue #954, opened 2023-03-15
Bug: non dismissable toasts clip content


qazwsxedcrfvtgb1111:
### Steps to reproduce

Any content close to the top of the toast container will be clipped, and since it's not dismissible there's no way to see it.

### Expected Behavior

Toast will push content down

### Actual Behavior

Toast clips content

### Version

4.7.0-canary.c168abdb

### Anything else we should know?

Additional information

PR #959, opened 2023-03-16
feat(Toast): make non-dismissible toasts positioned statically

qazwsxedcrfvtgb1111:
This pr makes non-dismissible toasts static so they don't cover up content.

image

Closes: https://github.com/dequelabs/cauldron/issues/954

### DIFF
--- a/docs/index.css
+++ b/docs/index.css
@@ -126,3 +126,7 @@ code {
     bottom: 3.45rem;
   }
 }
+
+.Toast:not(.Toast--dismissible) {
+  margin-bottom: 1rem;
+}
--- a/docs/index.css
+++ b/docs/index.css
@@ -126,7 +126,3 @@ code {
     bottom: 3.45rem;
   }
 }
-
-.Toast:not(.Toast--dismissible) {
-  margin-bottom: 1rem;
-}
--- a/docs/patterns/components/Toast/index.js
+++ b/docs/patterns/components/Toast/index.js
@@ -131,6 +131,12 @@ export default class Demo extends Component {
               </>
             ),
             show: type === 'not-dismissible',
+            DEMO_renderBefore: (
+              <p>
+                A current limitation is that non-dismissible Toasts are shown
+                where they are rendered to prevent them from clipping content.
+              </p>
+            ),
             DEMO_renderAfter: (
               <Button
                 variant="error"
--- a/packages/styles/toast.css
+++ b/packages/styles/toast.css
@@ -1,6 +1,4 @@
 .Toast {
-  top: var(--top-bar-height);
-  position: fixed;
   color: #0b0e11;
   font-size: var(--text-size-small);
   z-index: var(--z-index-toast);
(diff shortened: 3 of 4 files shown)


docs/index.css
```diff
@@ -126,3 +126,7 @@ code {
     bottom: 3.45rem;
   }
 }
+
+.Toast:not(.Toast--dismissible) {
+  margin-bottom: 1rem;
```
qazwsxedcrfvtgb1111, line 131: Not sure if I should add this to the Toast styles themselves or leave it up to the implementer to set a margin

docs/patterns/components/Toast/index.js
```diff
@@ -131,6 +131,12 @@ export default class Demo extends Component {
               </>
             ),
             show: type === 'not-dismissible',
+            DEMO_renderBefore: (
+              <p>
+                A current limitation is that non-dismissible Toasts are shown
+                where they are rendered to prevent them from clipping content.
```
qazwsxedcrfvtgb1111, line 137: The text here might need to be different

packages/styles/toast.css
```diff
@@ -1,13 +1,16 @@
...
   z-index: var(--z-index-toast);
   align-items: center;
   box-sizing: border-box;
   display: flex;
   padding: 4px 15px;
+}
+
+.Toast.Toast--dismissible {
+  top: var(--top-bar-height);
+  position: fixed;
```
scurker, line 13: This is a breaking change for a couple of reasons. A fixed toast is the standard implementation, while a non-dismissible toast is the variant. In order to make this not a breaking change, we should flip the implementation.

Maybe use `Toast--static`? I don't necessarily like describing the visual state of a component in its class, but I'm struggling to come up with a better option.

scurker (changes requested):

packages/styles/toast.css
qazwsxedcrfvtgb1111: would doing something like this work so I don't have to set the prop in one class and unset it in another?
```
.Toast:not(.Toast--static) {
  top: var(--top-bar-height);
  position: fixed;
}
```

scurker: It would technically work, but it would also change specificity:

```css
.Toast { ... } /* specificity of [0, 1, 0] */
.Toast:not(.Toast--static) { ... } /* specificity of [0, 2, 0] */
```

Meaning that if someone was overriding cauldron's styles there's a possibility that we could be clobbering the styles that someone else added. If this were a new component it wouldn't necessarily be an issue but changing the specificity to an existing component could cause issues.

While specificity isn't implicitly called out in our breaking changes guidelines, we might need to consider documenting it. For now, we should follow the guiding principle:

> Before a breaking change is released, other alternatives should be considered

qazwsxedcrfvtgb1111:
Talked to @bobbyomari about the bottom margin, we decided to add it to the Toast

scurker (approved):

Issue #961, opened 2023-03-17
IconButton renders the label twice


thuey:
When `IconButton` is given a `label`, it renders it as offscreen button content: https://github.com/dequelabs/cauldron/blob/37cf874919277b1373e143418ea95073d8bf368a/packages/react/src/components/IconButton/index.tsx#L94 and in the tooltip: https://github.com/dequelabs/cauldron/blob/37cf874919277b1373e143418ea95073d8bf368a/packages/react/src/components/IconButton/index.tsx#L105

This causes an issue when the label has an `id`, since id's need to be unique on a page.

The `label` prop should only be rendered as the button content when the tooltip is disabled, since that is when it is needed. Otherwise, the button gets its name from the tooltip.

PR #962, opened 2023-03-17
fix(react): only render icon button label where it is needed

thuey:
Closes https://github.com/dequelabs/cauldron/issues/961


scurker (approved):
Tested with VO/NVDA. This seems to be fine.

PR #984, opened 2023-03-22
docs: fix markdown intermingling with react components

scurker:
I noticed that some of the react examples were getting intermingled with the [mdx components](https://github.com/dequelabs/cauldron/blob/develop/docs/mdx-components.tsx) causing some weird layout issues. Wrapping the rendered code in a named function seems to resolve that problem.

The following markdown:

````markdown
```jsx example
<Button>Hello</Button>
```
````

Will now end up transformed into something like:

```mdx
export function CodeExample1234() {
  return <><Button>Hello</Button></>
}

{
  <Example>
    <CodeExample1234 />
  </Example>
}
```

### DIFF
--- a/docs/remark-plugins.mjs
+++ b/docs/remark-plugins.mjs
@@ -10,6 +10,8 @@ import { mdxjs } from 'micromark-extension-mdxjs';
 import { mdxFromMarkdown } from 'mdast-util-mdx';
 import { mdxExpressionFromMarkdown } from 'mdast-util-mdx-expression'
 
+const index = 0
+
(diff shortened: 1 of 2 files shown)


docs/remark-plugins.mjs
```diff
@@ -52,10 +63,6 @@ const exampleFromMDX = () => {
         options
       );
 
-      // Note: From markdown will strip extra whitespace from raw code, which
```
scurker, line 55: This appears to no longer be needed with the above change.

PR #986, opened 2023-03-23
fix: fix aria-hidden not toggling when Alert closes

yhafez:
Closes Walnut-[#5222](https://github.com/orgs/dequelabs/projects/64/views/1?pane=issue&itemId=23408990)

### DIFF
--- a/packages/react/src/components/Dialog/index.tsx
+++ b/packages/react/src/components/Dialog/index.tsx
@@ -170,8 +170,8 @@ export default class Dialog extends React.Component<DialogProps, DialogState> {
   }
 
   close() {
+    this.state.isolator?.deactivate();
     if (this.props.show) {
-      this.state.isolator?.deactivate();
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Dialog/index.tsx
```diff
@@ -170,8 +170,8 @@ export default class Dialog extends React.Component<DialogProps, DialogState> {
   }
 
   close() {
+    this.state.isolator?.deactivate();
     if (this.props.show) {
-      this.state.isolator?.deactivate();
```
yhafez, line 174: Not sure why this behavior was only happening conditionally based on if `this.props.show` is truthy, or if there's a compelling reason to keep it that way, but moving this line out of the conditional fixes the problem in the ticket where aria-hidden wasn't being removed from the top-level app component when `Alert` is closed

PR #988, opened 2023-03-23
fix: Fix tooltip overflowing viewport bug

yhafez:
Relates to [Walnut #5344](https://github.com/dequelabs/walnut/issues/5344)

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -27,6 +27,7 @@
     "focus-trap-react": "8",
     "focusable": "^2.3.0",
     "keyname": "^0.1.0",
+    "popper-max-size-modifier": "^0.2.0",
     "prop-types": "^15.6.0",
     "react-id-generator": "^3.0.1",
     "react-popper": "^2.2.4",
--- a/packages/react/src/components/Tooltip/index.tsx
+++ b/packages/react/src/components/Tooltip/index.tsx
@@ -6,6 +6,7 @@ import { useId } from 'react-id-generator';
 import { Placement } from '@popperjs/core';
 import { usePopper } from 'react-popper';
 import { isBrowser } from '../../utils/is-browser';
+import maxSize from 'popper-max-size-modifier';
 
 const TIP_HIDE_DELAY = 100;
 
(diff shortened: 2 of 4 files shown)


scurker (changes requested):
I'm hesitant to add a new dependency on a package that is marked as deprecated. It should be auto repositioning here, do we know why that isn't happening?

this package has been deprecated

yhafez:
> I'm hesitant to add a new dependency on a package that is marked as deprecated. It should be auto repositioning here, do we know why that isn't happening?
> 
> this package has been deprecated



> I'm hesitant to add a new dependency on a package that is marked as deprecated. It should be auto repositioning here, do we know why that isn't happening?
> 
> this package has been deprecated

Hmmm, that's fair. This is a community modifier that's highlighted in the React Popper docs https://popper.js.org/docs/v2/modifiers/community-modifiers/, but it looks like the author of the modifier recommends users to switch to floating-ui to handle poppers instead of react popper, and that's the deprecation note that they left in the npm repo too (https://www.npmjs.com/package/popper-max-size-modifier). Figured that would be a bigger conversation if we were to try doing that.

While I was working on this, the repositioning behavior was pretty sporadic and inconsistent. I know it partially has to do with the `max-width`, but I also don't trust the default repositioning behavior of react popper. We also have to account for strings that are longer than the viewport, regardless of whether it gets repositioned or not.

yhafez:
> > I'm hesitant to add a new dependency on a package that is marked as deprecated. It should be auto repositioning here, do we know why that isn't happening?
> > this package has been deprecated
> 
> Hmmm, that's fair. This is a community modifier that's highlighted in the React Popper docs https://popper.js.org/docs/v2/modifiers/community-modifiers/, but it looks like the author of the modifier recommends users to switch to floating-ui to handle poppers instead of react popper, and that's the deprecation note that they left in the npm repo too (https://www.npmjs.com/package/popper-max-size-modifier). Figured that would be a bigger conversation if we were to try doing that.
> 
> While I was working on this, the repositioning behavior was pretty sporadic and inconsistent. I know it partially has to do with the `max-width`, but I also don't trust the default repositioning behavior of react popper. We also have to account for strings that are longer than the viewport, regardless of whether it gets repositioned or not.

Implemented a custom modifier instead of using the package

scurker:
I didn't realize looking at this initially, but it looks like [`floating-ui`](https://floating-ui.com/docs/migration) is the replacement for `popperjs` and it may have resolved some of the issues that we're seeing here. I suggested to @yhafez that we should investigate upgrading to see if that solves some of the issues that we're seeing as opposed to trying to create a custom modifier that might have issues.

yhafez:
> I didn't realize looking at this initially, but it looks like [`floating-ui`](https://floating-ui.com/docs/migration) is the replacement for `popperjs` and it may have resolved some of the issues that we're seeing here. I suggested to @yhafez that we should investigate upgrading to see if that solves some of the issues that we're seeing as opposed to trying to create a custom modifier that might have issues.

Going to do this work in a separate brach, so closing this PR!

PR #996, opened 2023-03-29
feat(react): allow scrollability of Code component to be configured, auto-add tab-index to scrollable Code elements

### LINKED ISSUE #932: Code component is not accessible when scrolled
When the `Code` component scrolls because it does not have keyboard access. Example issue: https://axe.deque.com/issues/a01c8247-0945-4313-afc4-07e04ec38b3e

The `Code` component should wrap lines instead of scrolling by default.

It would also be nice if you could opt-in to scrolling, and when it needs to scroll that it would turn into a focusable region. This could maybe be implemented by using a `ResizeObserver`. However, the `Code` component would need to be wrapped in something else that scrolls, since `react-syntax-highlighter` does not forward refs: https://github.com/react-syntax-highlighter/react-syntax-highlighter/issues/335.

scurker:
closes #932


packages/react/src/components/Code/index.tsx
```diff
@@ -18,30 +18,64 @@ const Highlighter = SyntaxHighlighter as React.ComponentType<
...
     <Highlighter
       {...props}
+      PreTag={PreWithRef}
       useInlineStyles={false}
-      className={classNames('Code', className)}
-      tabIndex={tabIndex}
+      className={classNames('Code', className, {
+        'Code--scrollable': scrollable
+      })}
+      tabIndex={scrollableRegion ? 0 : undefined}
```
thuey, line 73: This is a breaking change, right? Should we either make sure to label the commit as a breaking change or try to preserve the current functionality by continuing to allow `tabIndex` to be passed in? If we did the latter, this code could be (but maybe pulled out into a separate variable:

```suggestion
      tabIndex={tabIndex !== undefined ? tabIndex : scrollableRegion ? 0 : undefined}
```

packages/react/src/components/Code/index.tsx
```diff
@@ -18,30 +18,64 @@ const Highlighter = SyntaxHighlighter as React.ComponentType<
...
+        }
+
+        const element = ref.current;
+        setScrollableRegion(element.clientWidth < element.scrollWidth);
+      };
+      const observer = new ResizeObserver(listener);
+      observer.observe(ref.current.parentElement);
+    }
+
+    return () => observer?.disconnect();
```
thuey, line 62:
```suggestion
    return () => {
      setScrollableRegion(false);
      observer?.disconnect();
    }
```

packages/react/src/components/Code/index.tsx
```diff
@@ -18,30 +18,64 @@ const Highlighter = SyntaxHighlighter as React.ComponentType<
...
+  }, []);
+
+  return (
     <Highlighter
       {...props}
```
thuey, line 67: It looks like we're spreading `...props` on both `Highlighter` and the `pre` element.  What is the result of doing that?

scurker: They were differently scoped vars, I renamed pre's props to clarify this.

packages/react/src/components/Code/index.tsx
scurker: Why is this needed? If the component is getting unmounted we don't need to update the state of the component.

thuey: I assumed you had added `scrollable` and `ref.current` to the list of things for `useEffect` to watch for. I see now, though, that you left the list empty.  Is there a reason why you didn't add those to the list?

thuey: > Ref objects don't always work as expected inside of useEffect because they're updated outside of the flow of a normal render, see comment https://github.com/facebook/react/issues/14387#issuecomment-503616820

You're right. I always forget about that.

> I hadn't considered there actually being a use case for allowing the scrollable prop to be changed after the initial render. Do you think that's a scenario that someone would use?

I think it's possible, such as if the layout changes to be responsive 🤷 But it's probably not a super common use case

thuey: Sounds good. So should we apply this suggestion now?

scurker: Whoops, look like I forgot to push that change 😅

thuey (approved):

PR #997, opened 2023-03-29
docs: update code documentation to v2

scurker:
I need to update documentation for #996, and would like to get this in first.


JoshMK (changes requested):
Looks good! The only thing I noticed was that [className](https://github.com/dequelabs/cauldron/blob/e2ffe2bbcda12dcc1f1101686dd5fd66cb891c59/packages/react/src/components/Code/index.tsx#L24) wasn't listed in the props table.

JoshMK (approved):

PR #1024, opened 2023-04-11
docs: update ClickOutsideListener to v2 docs

scurker:


docs/pages/components/ClickOutsideListener.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+## Props
+
+<ComponentProps
+  children={true}
+  props={[
+    {
+      name: 'onClickOutside',
+      type: ['MouseEvent', 'TouchEvent'],
+      description: 'Callback function for event type.'
```
kethomas, line 29: looks like this prop is required but I don't see that explicitly in the new docs

scurker: Yup, good catch. This should be required.

thuey (approved):

Issue #987, opened 2023-03-23
Update Panel component


scurker:
We want to support 2 new primitive components to use as composition inside of the existing `<Panel>` component:

`Panel` should support `padding={false}` to render with no padding.

* `PanelHeader` - this would be in addition to the existing "heading" property, you shouldn't use both at the same time.

![image](https://user-images.githubusercontent.com/1062039/227268110-b3c2e6d5-537e-425c-96a7-004185cc753c.png)

* `PanelContent` - content sections that when stacked would have an adjacent border between them

![image](https://user-images.githubusercontent.com/1062039/227268884-46d285ff-f496-410f-9715-e4229947b3a3.png)

The author should also be able disable the padding for the panel content for more advanced layouts.

```
<PanelContent padding={false} />
```

For the existing heading prop, it should match the style of the `PanelHeader` with a border divider below the heading.

Note: We should enhance the documentation showing different examples of usages with both simple and more complex usages, with a note that both the `heading` prop and `PanelHeader` child should not both be used.

Related #712

Amandeque:
Estimates: Small

Keerthi-Penukonda:
@tbusillo can you please provide some QA notes to test this ticket?

tbusillo:
> @tbusillo can you please provide some QA notes to test this ticket?

Hey Keerthi, I had just moved this to our board for tracking purposes last sprint. I think between Jason's review and design review with Shawn we can just QA once the updated component is implemented in some of the upcoming tickets we have. 

Will remove pull this off the board unless there's any objections!

PR #1025, opened 2023-04-11
feat(Panel): add composable components and revise docs 

tbusillo:
~Note: getting with @shawnsharpDQ to get feedback on dark-mode color scheme.~

closes: https://github.com/dequelabs/cauldron/issues/987


docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+
+## Examples
+
+### Panel
+
+At its simplest, a `Panel` is a styled semantic HTML `<section>` element with with content inside.
+
+<Note>
+If a `Panel` does not have a heading, <Link href="https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/general-principles.html" rel="noopener noreferrer" target="_blank">it is recommended</Link> to add an `aria-label` to the `Panel` component.
+</Note>
```
scurker, line 31: I'm not sure this section is needed or may need to be reworded. Overused landmarks [can be problematic](https://webaim.org/blog/aria-cause-solution/) so we want to tread lightly in how strongly we want to encourage it. It might be better to say, "If your section needs to have an accessible name, be sure to include an `aria-label` or `aria-labelledby` pointed to the id of your label."

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+
+```jsx example
+<Panel aria-label="Panel Heading">
+  <p>Panel Content</p>
```
scurker, line 35: Nit, we could show this without a `<p>`.

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+```
+
+### Panel (Composed)
+
+In addition to using the `heading` prop, you can also compose the `Panel` component with the `PanelHeader` and 
+`PanelContent` components.
+
+<Note>
+  The `PanelHeader` and `PanelContent` should not be used with the `Panel`
+  heading prop at the same time. Choose one or the other to use.
```
scurker, line 60:
```suggestion
  heading prop at the same time. Choose one or the other. 
```

Nit: Wording seems a little odd?

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+
+In addition to using the `heading` prop, you can also compose the `Panel` component with the `PanelHeader` and 
+`PanelContent` components.
+
+<Note>
+  The `PanelHeader` and `PanelContent` should not be used with the `Panel`
+  heading prop at the same time. Choose one or the other to use. 
+  
+  In addition, if you use the `PanelHeader`, you are responsible for adding `aria-labelledby` to
+  the `Panel` and an ID to the heading.
```
scurker, line 63: See initial note, we just need to think about this as it's not absolutely required.

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+
+For greater control over the styling of the `Panel`'s content, you can disable the default padding by setting 
+the `padding` prop to `false`.
+
+```jsx example
+<Panel
+  heading={{ id: 'example', text: 'Panel Heading', level: 3 }}
+  padding={false}
+>
+  <p style={{ paddingLeft: '1rem' }}>Panel Content</p>
```
scurker, line 87:
```suggestion
  <p style={{ padding: '1rem' }}>Panel Content</p>
```

Nit: This should have padding all around.

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+  <PanelHeader
+    style={{
+      display: 'flex',
+      justifyContent: 'space-between',
+      alignItems: 'center',
+      flexWrap: 'wrap'
+    }}
+  >
+    <h3>Panel Header</h3>
+    <Button variant="primary" thin>
```
scurker, line 110: This button has a contrast issue in light mode. (I also don't think it's needed for this particular example.

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+    <div style={{ padding: 'var(--space-small)' }}>Column 1</div>
+    <div style={{ padding: 'var(--space-small)' }}>Column 2</div>
+  </PanelContent>
+  <PanelContent>
+    <p>Footer</p>
+  </PanelContent>
+</Panel>
+```
+
+### Header With Flexbox Content
```
scurker, line 132: I'm not 100% sure this section is needed as we haven't really suggested this in other documentation as other components can render however you put stuff in them as well. Maybe it is useful here though since this is more of a container component. What do you think?

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+      justifyContent: 'space-between',
+      alignItems: 'center',
+      flexWrap: 'wrap'
+    }}
+  >
+    <h3>Panel Header</h3>
+    <Button variant="primary" thin>
+      <Icon type="plus" />
+      Button
+    </Button>
```
scurker, line 151: Not sure this is needed here for this example.

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+  </PanelHeader>
+  <PanelContent>
+    <p>Panel Content</p>
+  </PanelContent>
+</Panel>
+```
+
+## Props
+
+<ComponentProps
```
scurker, line 228: Can we have subheadings here with the component names?

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
...
+  name="Panel"
```
scurker, line 229: `name` isn't a valid prop, I think you may need a heading (see above).

packages/react/__tests__/src/components/Panel/index.js
```diff
@@ -84,4 +90,123 @@ describe('Panel', () => {
...
+  test('renders with composed heading', () => {
+    const panel = mount(
+      <Panel>
+        <PanelHeader>
+          <h1>Panel Heading</h1>
+        </PanelHeader>
+        Content
+      </Panel>
+    );
+    console.log(panel);
```
scurker, line 111: 🚨

packages/react/src/components/Panel/PanelContent.tsx
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from 'prop-types';
+
+export interface PanelContentProps
+  extends React.HTMLAttributes<HTMLDivElement> {
+  children: ReactNode;
+  className?: string;
+  padding?: boolean;
+}
+
+const PanelContent = ({
```
scurker, line 12: This should include a ref (via `React.forwardRef`) according to our [component standards](https://cauldron.dequelabs.com/component-guidelines#component-standards).

packages/react/src/components/Panel/index.tsx
```diff
@@ -69,7 +83,10 @@ Panel.propTypes = {
   children: PropTypes.node.isRequired,
   // @ts-expect-error
   heading: PropTypes.oneOfType([PropTypes.object, PropTypes.node]),
-  className: PropTypes.string
+  className: PropTypes.string,
+  padding: PropTypes.bool
 };
 
+export { PanelHeader, PanelContent };
```
scurker, line 90: Just curious, why import/re-export here? We've typically used just an index component for this.

scurker (changes requested):
Good work so far! I've reviewed most of the documentation, I will need to review the styles tomorrow. FYI there's some color contrast accessibility issues here with the buttons.

packages/react/src/components/Panel/PanelContent.tsx
```diff
@@ -0,0 +1,38 @@
...
+  children,
+  className,
+  padding = true,
+  ...otherProps
+}: PanelContentProps) => {
+  return (
+    <div
+      className={classNames('Panel__Content', className, {
+        ['Panel__Content--padding']: padding
```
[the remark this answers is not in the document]
tbusillo: Ah, I think I'm understanding, but just to make sure so I do it right - are you saying should do something like `Panel__Content--no-padding`, with the default `.Panel` styling being with the padding?

packages/react/src/components/Panel/index.tsx
tbusillo: I think I originally had everything in this single file, then broke it out into separate files. Are you thinking/do you prefer I do rename this to `Panel.tsx`, then import/export everything from a `index.tsx`?

docs/pages/components/Panel.mdx
tbusillo: Only reason I was thinking it might be a good example to show is because it follows more or less the pattern in the [Cauldron Pattern Library](https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161871161/simulate/no-panels?mode=cvidm). Was also thinking that we have a lot of designs that use similar patterns (e.g., heading with button/justified with space between). 

I am cool with it either way, just thought people may find it useful/save them some time!

docs/pages/components/Panel.mdx
tbusillo: Commented below, but think it would definitely be useful to include at least one example like this (either this one or the one below) since its used in a lot of the designs we have/is a pretty common use case generally. Also demonstrates some of the real usefulness of using composition over the heading prop, IMO.

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,261 @@
+---
```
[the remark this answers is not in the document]
tbusillo: I was actually wondering about this - do we actually need the collapsed variant any longer? I ask, because the padding on the top element is already `12px`, and collapsed is only taking things from `16px` -> `12px`. I'll add it back in either way, but was something meant to check with you on.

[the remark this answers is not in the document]
scurker: Let me check with @bobbyomari and @shawnsharpDQ, it might not be needed but of course we'd still have to deprecate it if not.

packages/react/src/components/Panel/index.tsx
scurker: The 2 ways we've done it:

- Export everything via the `src/index.ts` file
- Export via an `component/${name}/index.ts` file.

I don't have a preference, just didn't want to introduce a new variant 🙃

docs/pages/components/Panel.mdx
tbusillo: Not sure why this change didn't get committed - unresolving and fixing now.

packages/styles/panel.css
```diff
@@ -3,6 +3,22 @@
   --panel-background-color: var(--white);
   --panel-heading-color: var(--gray-90);
   --panel-padding: var(--space-small);
+  --panel-content-padding: var(--space-small);
+  --panel-content-margin: 0 calc(var(--space-small) * -1);
+  --panel-header-margin: calc(var(--space-smaller) * -1)
+    calc(var(--space-small) * -1);
+  --panel-header-padding: var(--space-smaller) var(--space-small);
```
scurker, line 10: The padding variables are tricky in part due to how the heading padding is different. Ideally I should be able to set `--panel-padding` and have it cascade everywhere where it's needed. So minimally, I think we need something like this:

```css
  --panel-padding: var(--space-small);
  --panel-content-padding: var(--panel-padding);
  --panel-header-padding: var(--panel-padding);
```

This allows for one value to be set vs 3 (and still allows for individual overrides). Alternatively, we could use a fallback value as well instead of defining them at the root which might be a little more robust:

```css
.PanelContent {
  padding: var(--panel-content-padding, --panel-padding);
}
```

I'm leaning towards the latter, what do you think?

scurker, line 10: Talking with @shawnsharpDQ, we can keep the top/bottom padding of the header relative to the standard padding. So this value can just be for the left/right padding and have the following in the class styles:

```css
padding: max(calc(var(--panel-header-padding, --panel-padding) - 4px), 0) var(--panel-header-padding, --panel-padding))
```

Note: The format depends on what we decide for the comment prior to this.

packages/styles/panel.css
```diff
@@ -3,6 +3,22 @@
   --panel-background-color: var(--white);
   --panel-heading-color: var(--gray-90);
   --panel-padding: var(--space-small);
+  --panel-content-padding: var(--space-small);
+  --panel-content-margin: 0 calc(var(--space-small) * -1);
+  --panel-header-margin: calc(var(--space-smaller) * -1)
```
scurker, line 8: These should inherit from appropriate variable value, otherwise they won't match up if the padding changes.

scurker (changes requested):
Docs are good, this is primarily regarding styles.

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,237 @@
...
+  Code
+} from '@deque/cauldron-react';
+
+```js
+import { Panel, PanelHeader, PanelContent } from '@deque/cauldron-react';
+```
+
+## Examples
+
+### Panel
```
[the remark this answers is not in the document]
tbusillo: What if I were to change this one to `### Panel Component`?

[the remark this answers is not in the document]
scurker: I think that would be fine.

packages/styles/panel.css
```diff
@@ -3,23 +3,90 @@
   --panel-background-color: var(--white);
   --panel-heading-color: var(--gray-90);
   --panel-padding: var(--space-small);
+  --panel-header-margin: 0 calc(var(--panel-padding) * -1);
+  --panel-divider-border: var(--gray-40);
```
scurker, line 7:
```suggestion
  --panel-divider-border-color: var(--gray-40);
```

Make it more self documenting.

packages/styles/panel.css
```diff
@@ -33,18 +100,11 @@ fieldset.Panel {
...
 }
 
-.Panel--collapsed {
-  --panel-padding: 12px;
-}
-
-.Panel--collapsed .Panel__Heading {
-  margin: 0 0 12px 0;
-  font-size: var(--text-size-normal);
-}
```
scurker, line 43: This can't be removed as it would make it a breaking change. Even though we do plan on removing these values, we need to go through a deprecation period first.

scurker (changes requested):

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,223 @@
...
+
+At its simplest, a `Panel` is a styled semantic HTML `<section>` element with with content inside.
+
+<Note>
+  If your section needs to have an accessible name, be sure to include an
+  `aria-label` or `aria-labelledby` pointed to the `id` of your label.
+</Note>
+
+```jsx example
+<Panel aria-label="Panel Heading">Panel Content</Panel>
```
scurker, line 31: We should make this something unique since _Panel Heading_ is used below, maybe _Panel Content_? This would get rid of a duplicate accessible name.

docs/pages/components/Panel.mdx
```diff
@@ -0,0 +1,223 @@
...
+the elements is it automatically adds an `id` to the heading and `aria-labelledby` to the `Panel` component.
+
+Inspect the below example and you will see that the `Panel` component has an `aria-labelledby` attribute with the
+value of the `id` of the heading.
+
+```jsx example
+<Panel
+  heading={{
+    id: 'panel-heading-prop-example',
+    text: 'Panel Heading',
```
scurker, line 46: This is also duplicated below in the disabled panel heading example. We should choose one to make it unique.

scurker (changes requested):
Docs, styles, component all good, just need to address a few minor accessibility issues:

PR #1027, opened 2023-04-12
docs: convert Loader to v2 docs

scurker:


docs/pages/components/Loader.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+<ComponentProps
+  children={true}
+  className={true}
+  refType="HTMLDivElement"
+  props={[
+    {
+      name: 'label',
+      type: 'string',
+      description: '  The desired label will be rendered offscreen. If not provided, aria-hidden="true" will be applied to the element.',
```
thuey, line 29: I'm not sure why there would be spaces here...

```suggestion
      description: 'The desired label will be rendered offscreen. If not provided, aria-hidden="true" will be applied to the element.',
```

scurker: I thought I pushed that fix 🤔

PR #1045, opened 2023-04-27
fix(style): fix aggressive styles getting applied to nested elements in Panel

scurker:


packages/styles/panel.css
```diff
@@ -7,15 +7,6 @@
...
 }
 
-/* Resets aggressive Chrome user agent styles */
-.Panel > :is(.Panel__Header, .Panel__Content, .Panel__Heading) > *,
-.Panel:first-child > :not(.Panel__Header, .Panel__Heading, .Panel__Content) {
-  -webkit-margin-before: 0;
-  -webkit-margin-after: 0;
-  margin-block-start: 0;
-  margin-block-end: 0;
-}
```
scurker, line 17: These styles were getting applied way too aggressively meaning that every nested element was getting its margins wiped leading to almost no spacing between elements. The added selectors below are more targeted and should better gracefully handle nested elements.

tbusillo (approved):
LGTM!

Issue #620, opened 2022-04-05
Text wrapping on `<Button thin>`


scurker:
![thin button with "change axe devtools pro plan" with little to no vertical spacing](https://user-images.githubusercontent.com/1062039/161836517-99845ee7-5930-4716-b4fb-21e70232b4ea.png)

It looks like there's a few more tweaks we need to make [here](https://github.com/dequelabs/cauldron/blob/61754b20d7a4ca320d0600047c0b959dd1427a4d/packages/styles/button.css#L185). We should remove `line-height` and instead rely on top/bottom padding to properly set the text alignment within the button so that when the button wraps it has the proper spacing.

Bracciata:
Tackled this one in PR #1046

PR #1046, opened 2023-04-28
fix(react): removed line height styling from thin button

Bracciata:
Closes: #620 

Please let me know if there is any additional styling that needs to be added or if there is a prototype the styling should be based on.

Before:

three buttons one being the thin one with line height compressing it

After:

three buttons with the thin button no longer being compressed


scurker (approved):

scurker:
🚨 REVIEWED FOR SECURITY 🚨

PR #1049, opened 2023-05-01
feat(RadioGroup): add inline modifier for RadioGroup

### LINKED ISSUE #1048: Add inline modifier for RadioGroup
We need to support an inline layout for radio items according to this design:

radio group with field label with the radio items displayed inline below the label

This is a fairly small change and is a modifier and not a variant so it should look something like this:

`<RadioGroup inline />`

scurker:
fixes #1048


packages/styles/forms.css
```diff
@@ -190,6 +190,13 @@ textarea.Field--has-error:focus:hover,
   flex-wrap: wrap;
 }
 
+.Radio--inline {
+  display: flex;
+  align-items: baseline;
+  flex-wrap: wrap;
```
thuey, line 196: When it wraps, the spacing seems a little big. Would you agree?

image

thuey (approved):

PR #1065, opened 2023-05-05
docs: update Link component to v2 docs

tbusillo:
## Description 
- Updates the `Link` documentation to use the new MDX format


docs/components/example.css
```diff
@@ -66,6 +66,10 @@ body.cauldron--theme-dark {
   position: relative;
 }
 
+.Component__example__react .Link--lowercase {
+  margin: 0 var(--space-smaller);
```
tbusillo, line 70: Added this in so there would be space between the examples where buttons are adjacent to a link. It was looking crowded without this. 

CleanShot 2023-05-05 at 01 35 32@2x

packages/react/src/components/Button/index.tsx
```diff
@@ -28,7 +30,8 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(
         Link: variant === 'link',
         Tag: variant === 'tag',
         'Button--tag': variant === 'tag',
-        'Button--thin': thin
+        'Button--thin': thin,
+        'Link--lowercase': lowercase
```
dequejosie, line 34: `Button--lowercase`?

docs/pages/components/Link.mdx
```diff
@@ -0,0 +1,88 @@
+---
+title: Link
+description: A link component with accessible styling.
```
dequejosie, line 3: Do we need to define what `link` is like what we do with `button`?

docs/pages/components/Link.mdx
```diff
@@ -0,0 +1,88 @@
...
+    {
+      name: 'lowercase',
+      type: 'boolean',
+      description:
+        'When `true`, adds style to make text contents all lowercase.'
+    },
+    {
+      name: 'buttonRef',
+      type: 'function',
+      description: 'Ref function for the button element'
```
dequejosie, line 81: nit: Period at the end.

docs/pages/components/Link.mdx
```diff
@@ -0,0 +1,88 @@
...
+The default variant for the `Link` component.
+
+```jsx example
+<Link>I am a link!</Link>
+```
+
+### Button
+
+When you need a semantic link with the visual appearance of a button.
+
```
dequejosie, line 26: nit: a primary button?

docs/pages/components/Link.mdx
```diff
@@ -0,0 +1,85 @@
...
+<ComponentProps
+  children={{
+    required: true
+  }}
+  props={[
+    {
+      name: 'variant',
+      type: ['button', 'button-secondary'],
+      defaultValue: 'button',
```
scurker, line 65: This is incorrect, there's no default value for variant.

scurker (changes requested):
One small change needed.

scurker (approved):

PR #1075, opened 2023-05-09
fix(styles): fix inaccessibile syntax highlight colors

### LINKED ISSUE #930: Color contrast issue with codeblock orange color on light theme
### The orange used here does not pass color contrast. 

image

image

scurker:
closes #930

### DIFF
--- a/packages/styles/code.css
+++ b/packages/styles/code.css
@@ -28,7 +28,7 @@
 }
 
 .Code .hljs-quote {
-  color: #969896;
+  color: #666666;
   font-style: italic;
 }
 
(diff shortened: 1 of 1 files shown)


packages/styles/code.css
```diff
@@ -88,11 +88,11 @@
 .Code .hljs-subst,
 .Code .hljs-meta,
 .Code .hljs-link {
-  color: #f99157;
+  color: #8a3500;
```
schne324, line 91: 🥴 the difference is jarring
Screenshot 2023-05-09 at 1 33 32 PM

scurker: I know, but that's that @dequelabs/cauldron-design wanted 🤷

PR #1078, opened 2023-05-16
docs: update Select to v2 docs

scurker:


docs/pages/components/Select.mdx
```diff
@@ -0,0 +1,263 @@
+---
+title: FieldWrap
```
dequejosie, line 2: Select instead of FieldWrap?

docs/pages/components/Select.mdx
```diff
@@ -0,0 +1,263 @@
...
+    },
+    {
+      name: 'defaultValue',
+      type: 'string',
+      description: 'The default element value. Use when the component is not controlled.'
+    },
+    {
+      name: 'onChange',
+      type: 'function',
+      description: ''
```
dequejosie, line 256: We are missing description for `onChange` function here.

dequejosie (approved):

PR #1080, opened 2023-05-23
docs: update TwoColumnPanel to v2 docs

scurker:


docs/pages/components/TwoColumnPanel.mdx
```diff
@@ -0,0 +1,310 @@
...
+  const selectItem = id => event => {
+    event.preventDefault();
+    setSelected(id);
+  };
+
+  return (
+    <TwoColumnPanel>
+      <ColumnLeft aria-labelledby="sidebar">
+        <ColumnHeader id="sidebar">Items</ColumnHeader>
+        <nav aria-label="Sidebar navigation">
```
schne324, line 70: "navigation" seems redundant with the implicit role of "navigation" for `<nav />` elements

```suggestion
        <nav aria-label="Sidebar">
```

Sidebar itself doesn't seem like great name here either but these are contrived examples so I don't think it is a huge deal. Ideally there would be some kind of descriptive heading above the two column panel (like they always do when we use this component in our apps) that we could make the nav `aria-labelledby`

scurker: Yeah I think I agree with your reasoning. I've updated the label to point to the header for both instances as that's more likely how we'd use it in practice.

dequejosie (approved):

PR #1089, opened 2023-06-08
chore(deps): upgrade eslint

### LINKED ISSUE #451: Upgrade eslint and plugins
Our version of eslint and eslint plugins are behind (sometimes years out of date). We should upgrade them.

This would allow us to do `import type` and would also might help address other issues we are seeing, such as https://github.com/yannickcr/eslint-plugin-react/issues/2269

jmetev1:
Hi folks. I see you're hiring. Does this qualify me for an interview for https://www.deque.com/careers/senior-full-stack-developer/ ?
closes #451 

I upgraded eslint and plugins. Of course this comes with some new rules and changes. To deal with that I disabled some rules that you were not using previously and made a few other small changes. Let me know if you want any changes. Or just push them.


.eslintignore
```diff
@@ -2,4 +2,5 @@
 **/lib/
 **/coverage/
 **/node_modules/
-docs/
\ No newline at end of file
+docs/
+.eslintrc.js
```
jmetev1, line 6: https://typescript-eslint.io/linting/troubleshooting/#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file

.eslintrc.js
```diff
@@ -13,7 +13,6 @@ module.exports = {
     'plugin:react/recommended',
     'plugin:jsx-a11y/recommended',
     'prettier',
-    'prettier/react',
```
jmetev1, line 16: was rolled into core prettier

packages/react/package.json
```diff
@@ -60,10 +60,6 @@
     "concurrently": "^5.3.0",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.2",
-    "eslint": "^6.8.0",
```
jmetev1, line 63: no need for these

jmetev1: assuming the plan was just to lint from top level which it appears

tsconfig.json
```diff
@@ -13,8 +13,7 @@
     "strictFunctionTypes": true,
     "typeRoots": ["node_modules/@types"],
     "jsx": "react",
-    "types": ["node"],
-    "esModuleInterop": true
```
jmetev1, line 17: was in here twice

tsconfig.json
```diff
@@ -13,8 +13,7 @@
...
+    "types": ["node"]
```
jmetev1, line 16: line 17 was a repeat

scurker:
Thanks for contributing @jmetev1!

Normally we would have you apply via our careers page, but I'm not sure where our active listings are. Would you mind submitting your resume to hr@deque.com ccing jason.wilson@deque.com? We'd love to have you apply!

.eslintrc.js
```diff
@@ -26,7 +25,10 @@ module.exports = {
       }
     ],
     '@typescript-eslint/camelcase': 'off',
-    '@typescript-eslint/explicit-function-return-type': 'off'
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    'react/prop-types': 'off',
```
scurker, line 29: Unfortunately, we haven't migrated away from using `prop-types` yet, even though we currently use typescript. This is something we would likely consider doing in the future.

.eslintrc.js
```diff
@@ -26,7 +25,10 @@ module.exports = {
...
+    '@typescript-eslint/no-empty-function': 'off',
```
scurker, line 30: Our preferred style has to keep this in and disable when necessary:

```js
// eslint-disable-next-line @typescript-eslint/no-empty-function
```

.eslintrc.js
```diff
@@ -26,7 +25,10 @@ module.exports = {
...
+    '@typescript-eslint/ban-ts-comment': 'off'
```
scurker, line 31: Preferably we wouldn't turn this off but rather use `@ts-expect-error` with a description.

packages/react/src/utils/polymorphic-type/index.ts
```diff
@@ -7,11 +7,13 @@ import * as React from 'react';
 /* -------------------------------------------------------------------------------------------------
  * Utility types
  * -----------------------------------------------------------------------------------------------*/
+// eslint-disable-next-line @typescript-eslint/ban-types
 type Merge<P1 = {}, P2 = {}> = Omit<P1, keyof P2> & P2;
```
scurker, line 11: I'm not 100% clear why this eslint rule is getting flagged here or below.

scurker (changes requested):
Thanks for contributing to cauldron! We are admittedly a little behind on updating some of our dependencies, so this is a big help. Requested changes below:

.eslintrc.js
jmetev1: i changed the rule. but now linting fails. is there someone else more qualified to add description? i can see that thuey added some of them

.eslintrc.js
jmetev1: fixed

.eslintrc.js
jmetev1: fixed

.eslintrc.js
scurker: Alternatively to get this over the finish line, we could change it to `warn` instead of off.

jmetev1: done

jmetev1:
this is one of those times where the ui shows "all conversations resolved", but also "scurker requests changes". Is there anything else i need to do here?

scurker:
> this is one of those times where the ui shows "all conversations resolved", but also "scurker requests changes". Is there anything else i need to do here?

Our process normally has been to have the submitter dismiss the changes, but I'm taking a look right now to ensure we're good.

scurker (approved):
Thank you for your help here!

Issue #1090, opened 2023-06-09
Allow select to take a node as a label


thuey:
The `Select` component's types only allow `label` to be a `string`: https://github.com/dequelabs/cauldron/blob/14d052ed352a6d6a7e72e2e8250fe0dc8cd98364/packages/react/src/components/Select/index.tsx#L15

It should allow the label to be a `ReactNode` to allow for greater flexibility, including being able to include screen-reader-only content.

PR #1091, opened 2023-06-09
feat(react): allow Select label to be a ReactNode

thuey:
Closes https://github.com/dequelabs/cauldron/issues/1090

### DIFF
--- a/packages/react/__tests__/src/components/Select/index.js
+++ b/packages/react/__tests__/src/components/Select/index.js
@@ -215,6 +215,5 @@ test('fires onChange when change occurs', () => {
 
 test('renders a ReactNode as a label', () => {
   const wrapper = withCustomOptions({ label: <span>Foo</span> });
-  console.log(wrapper.debug());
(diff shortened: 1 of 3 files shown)


packages/react/__tests__/src/components/Select/index.js
```diff
@@ -29,10 +29,9 @@ const withCustomOptions = (otherProps = {}) => {
 test('renders the expected UI', () => {
   const wrapper = withCustomOptions();
 
-  expect(wrapper.find('.Field__select--label')).toBeTruthy();
-  expect(wrapper.find('.Field__select--required')).toBeTruthy();
-  expect(wrapper.find('.Field__select')).toBeTruthy();
-  expect(wrapper.find('.Field__option')).toBeTruthy();
```
thuey, line 35: These assertions weren't actually doing anything before, since the enzyme wrapper is technically "truthy".

The --required in particular actually should _not_ exist, and we are asserting as much on line 35

schne324 (approved):

Issue #1094, opened 2023-06-20
Allow Checkbox to take `onFocus` and `onBlur` props


thuey:
## Context

If `onFocus`/`onBlur` props are passed to the `Checkbox` component, then checkbox focus state no longer gets updated. Because spreading `{...other}` props overrides the `onFocus` and `onBlur` handlers defined here: https://github.com/dequelabs/cauldron/blob/2d2f1f30d33d78e9d65d8fcb0e08b0f3ac70f516/packages/react/src/components/Checkbox/index.tsx#L80-L81

## Requirements

Similar to how `onChange` calls an `onChange` prop in addition to internal functions, `onFocus` and `onBlur` should not be overridden, but instead call those props if passed in: https://github.com/dequelabs/cauldron/blob/2d2f1f30d33d78e9d65d8fcb0e08b0f3ac70f516/packages/react/src/components/Checkbox/index.tsx#L85-L87

That way, focus state continues to be managed properly while allowing consumers to define their own handlers.

PR #1095, opened 2023-06-20
fix(react): allow onBlur and onFocus to be passed to Checkbox without messing up focus state

thuey:
Closes https://github.com/dequelabs/cauldron/issues/1094

### DIFF
--- a/packages/react/src/components/Checkbox/index.tsx
+++ b/packages/react/src/components/Checkbox/index.tsx
@@ -31,6 +31,8 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
       checkboxRef,
       className,
       onChange,
+      onFocus,
+      onBlur,
       'aria-describedby': ariaDescribedby,
       disabled = false,
       checked = false,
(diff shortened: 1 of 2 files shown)


Zidious (approved):

packages/react/src/components/Checkbox/index.tsx
```diff
@@ -77,8 +79,18 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
             type="checkbox"
             checked={isChecked}
             disabled={disabled}
-            onFocus={(): void => setFocused(true)}
-            onBlur={(): void => setFocused(false)}
+            onFocus={(e): void => {
+              setFocused(true);
+              if (onFocus) {
```
scurker, line 84: We typically do `typeof` checks for safeguarding:

```suggestion
              if (typeof onFocus === 'function') {
```

packages/react/src/components/Checkbox/index.tsx
```diff
@@ -77,8 +79,18 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
...
+                onFocus(e);
+              }
+            }}
+            onBlur={(e): void => {
+              setFocused(false);
+              if (onBlur) {
```
scurker, line 90:
```suggestion
              if (typeof onBlur === 'function') {
```

scurker (changes requested):

scurker (approved):

PR #1099, opened 2023-06-22
fix: update pagination statusLabel type to ContentNode

### LINKED ISSUE #958: Pagination statusLabel prop should be more restrictive
Currently, the Pagination `statusLabel` prop accepts `ReactNode`: https://github.com/dequelabs/cauldron/blob/655da6e04d8fd20a817a8325c5c2f2ce7aa93449/packages/react/src/components/Pagination/Pagination.tsx#L12

This is more permissive than it should be. It's important that the label contain valid content and not `null | undefined | boolean`. We should update the `label` prop type to be `ContentNode`: https://github.com/dequelabs/cauldron/blob/655da6e04d8fd20a817a8325c5c2f2ce7aa93449/packages/react/src/types.ts#L9-L18

Note: This would be a breaking change

Bracciata:
BREAKING CHANGE: update pagination statusLabel type to ContentNode
This closes #958 

We need to discuss how to handle this change because  as mentioned in the issue it is a breaking change.


scurker (changes requested):
This cannot land without a deprecation period according to our contributing guidelines:

> Before a breaking change can be released, the breaking change should be documented with the component, property, or style being changed getting flagged as Deprecated. This could mean warning the consumer that a component or property is now deprecated, or including a deprecated comment next to a css class name. This deprecation must exist for at least two months. If a change is additive (e.g. making a new property required), the new property must be optional until the deprecation period has passed.

I'm not 100% sure how we address this given it's a type change but it cannot land with our current guidelines.

Bracciata:
I am also unsure how to handle it. Maybe we could adjust the name of the prop and recommend future uses utilize the new name? I know that is likely not ideal especially if it is used in other parts of the repository.

scurker:
After further consideration, I'm going err on the side of this _not_ being a breaking change. Someone who is passing in `null` or `undefined` for a property intended to display a label is very likely a mistake and not a use case we would want to support for backwards compatibility anyway.

Issue #1069, opened 2023-05-09
Remove `text-uppercase` from button styles


scurker:
This should be removing the following property from button styles: 

https://github.com/dequelabs/cauldron/blob/a71901f1b08bae4f4b2248f847921a8cb1226a45/packages/styles/button.css#L34

PR #1102, opened 2023-06-26
feat(styles): removed unwanted text transform property

Bracciata:
Removed the css property making all buttons text appear as upper case as described by issue #1069


Bracciata:
I will say we can make all of the buttons full uppercase, however, I believe that causes issues with some screen readers (I may be misremembering).

PR #1109, opened 2023-07-12
docs: do not use headings for RadioGroup group label

### LINKED ISSUE #1104: Update RadioGroup documentation to not use headings
The `RadioGroup` component currently uses headings in documentation:

```jsx
<h3 id="gyros-label">Do you like gyros?</h3>
<RadioGroup
  aria-labelledby="gyros-label"
  name="gyros"
  radios={...}
  value="yes"
/>
```

While not specifically a problem, it's not really a best practice as this could lead to an overuse of headings. We should instead recommend a generic div label inside of a group container for the examples.

scurker:
closes #1104

### DIFF
--- a/docs/pages/components/RadioGroup.mdx
+++ b/docs/pages/components/RadioGroup.mdx
@@ -20,7 +20,7 @@ import { RadioGroup } from '@deque/cauldron-react'
 
 ```jsx example
 <FieldWrap>
-  <h3 id="gyros-label">Do you like gyros?</h3>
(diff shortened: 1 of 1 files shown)


thuey (approved):

thuey:
The appearance of the labels changed, too, is that intentional/acceptable?

This branch:
image

Develop:
image

scurker:
> The appearance of the labels changed, too, is that intentional/acceptable?

Yes.

thuey (approved):

Issue #1034, opened 2023-04-25
Sync RadioGroup radio items with UXPin


scurker:
[Pattern Library](https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161871161/simulate/no-panels?mode=i)

![RadioGroup radio items in different states, light mode](https://user-images.githubusercontent.com/1062039/234369022-ac01597a-60c3-4a16-a919-a915b63329bc.png)

![RadioGroup radio items in different states, dark mode](https://user-images.githubusercontent.com/1062039/234368961-349fd9e6-5ec7-409c-a26e-8a6571b996f0.png)

Light: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6764532,light-form-radio-regular
Dark: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6771800,dark-form-radio-regular (note: hard to see, but you can get the styles)

Pattern Library: https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161666306/simulate/no-panels?mode=i

The focus ring has changed, the disabled state has a more muted color in dark mode. The icons can be pulled from the above uxpins.

bobbyomari:
@scurker let's be sure to set each radio group to have a `margin-top: 0;` and `margin-bottom: 8px;` (currently they are set to be the opposite for margin-top and margin-bottom).

yhafez:
Implemented focus state (without slight glow), active state, disabled state, and font weight on label, as well as other refactors and styles. Pending UI for focus state glow.

scurker:
This needs to set the active background color for the radio on "active" state of the label and set the input width to 1px to allow the label to set hover correctly.

scurker:
Not doing this anymore. Thanks @shawnsharpDQ

Issue #949, opened 2023-03-14
Sync checkboxes with UXpin


scurker:
[Pattern Library](https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161871161/simulate/no-panels?mode=i)
[Checkbox UXPin](https://preview.uxpin.com/design-system/c933c74928e2195cc540/6764519,light-form-checkbox-regular)

![image](https://user-images.githubusercontent.com/1062039/225104930-76c636fa-e951-438f-85b0-593d40788f53.png)

The style appears to be mostly the same, we need to update the icon (which may depend on #869) and the following:

- Focus state (slight glow)
- Active state
- Disabled state
- Font weight to label

scurker:
For now when this gets initially implemented, ignore the icon changes and keep this issue open.

yhafez:
Implemented focus state (without slight glow), active state, disabled state, and font weight on label, as well as other refactors and styles. Pending UI for focus state glow and icons.

scurker:
Solution here might be similar to radio for the remaining fixes: see https://github.com/dequelabs/cauldron/issues/1034#issuecomment-1745444943

scurker:
Not doing this anymore. Thanks @bobbyomari

PR #1110, opened 2023-07-16
fix(styles): fix overflow styling on Radio Group and Checkbox and align styles with Pattern Library

### LINKED ISSUE #627: Checkbox wrapping issue

yhafez:
Closes #627 
Relates to #949 
Relates to #1034 

Relates to Walnut [#5739](https://github.com/dequelabs/walnut/issues/5739)

Current behavior:
https://github.com/dequelabs/cauldron/assets/60084578/b94f6133-0c7c-4368-9277-6736618301f9

---

With .Checkbox, .Radio `flexwrap` changed to `nowrap`:
https://github.com/dequelabs/cauldron/assets/60084578/ee399c8d-88c7-4b37-be61-2e2b4403bb21
no align-item self

w/ 4th option under wrapping line:
https://github.com/dequelabs/cauldron/assets/60084578/4dbaa581-63a3-41e5-9a02-d0033e09c6e0

---

With .Checkbox, .Radio `flexwrap` changed to `nowrap` and .Radio__overlay given `align-self` of `flex-start` (implemented):
https://github.com/dequelabs/cauldron/assets/60084578/6142d258-e011-451f-9427-bff88b6bcee9
align-item self

w/ 4th option under wrapping line:
https://github.com/dequelabs/cauldron/assets/60084578/fbe68121-6973-4831-af4b-12316976020d

### DIFF
--- a/docs/pages/components/Checkbox.mdx
+++ b/docs/pages/components/Checkbox.mdx
@@ -120,9 +120,9 @@ import { Checkbox } from '@deque/cauldron-react'
   />
   <Checkbox
     error="You must check this checkbox to continue"
-    id="checkbox-error"
+    id="checkbox-error-list"
     label="Error Checkbox"
-    name="error-checkbox"
+    name="error-checkbox-list"
     value="1"
   />
 </FieldWrap>
--- a/docs/pages/components/Checkbox.mdx
+++ b/docs/pages/components/Checkbox.mdx
@@ -90,6 +90,7 @@ import { Checkbox } from '@deque/cauldron-react'
 
 ```jsx example
 <FieldWrap>
+<div className="Field__label" id="checkbox-list-label">Checkbox List Label</div>
   <Checkbox
     id="checkbox-list-unchecked"
     label="Unchecked Checkbox"
@@ -106,7 +107,7 @@ import { Checkbox } from '@deque/cauldron-react'
   <Checkbox
     disabled
     id="checkbox-list-checkbox-disabled-unchecked"
-    label="Checked Disabled Checkbox"
+    label="Unchecked Disabled Checkbox"
     name="disabled-checkbox-list-unchecked"
     value="1"
   />
@@ -114,7 +115,7 @@ import { Checkbox } from '@deque/cauldron-react'
     disabled
     checked
     id="checkbox-list-checkbox-disabled-checked"
-    label="Unchecked Disabled Checkbox"
+    label="Checked Disabled Checkbox"
     name="disabled-checkbox-list-checked"
     value="1"
   />
@@ -161,6 +162,7 @@ import { Checkbox } from '@deque/cauldron-react'
 
 ```jsx example
 <FieldWrap>
+<div className="Field__label" id="checkbox-list-label">Checkbox with Description List Label</div>
   <Checkbox
     id="description-checkbox-list-unchecked"
     label="Description Checkbox Unchecked"
--- a/docs/pages/components/Checkbox.mdx
+++ b/docs/pages/components/Checkbox.mdx
@@ -86,6 +86,48 @@ import { Checkbox } from '@deque/cauldron-react'
 </FieldWrap>
 ```
 
+### Checkbox List
+
+```jsx example
+<FieldWrap>
+  <Checkbox
+    id="checkbox-list-unchecked"
+    label="Unchecked Checkbox"
+    name="unchecked-checkbox-list"
+    value="1"
+  />
+  <Checkbox
+    checked
+    id="checkbox-list-checked"
+    label="Checked Checkbox"
+    name="checked-checkbox-list"
+    value="1"
+  />
+  <Checkbox
+    disabled
+    id="checkbox-list-checkbox-disabled-unchecked"
+    label="Checked Disabled Checkbox"
+    name="disabled-checkbox-list-unchecked"
+    value="1"
+  />
+  <Checkbox
+    disabled
+    checked
+    id="checkbox-list-checkbox-disabled-checked"
+    label="Unchecked Disabled Checkbox"
+    name="disabled-checkbox-list-checked"
+    value="1"
+  />
+  <Checkbox
+    error="You must check this checkbox to continue"
+    id="checkbox-error"
+    label="Error Checkbox"
+    name="error-checkbox"
+    value="1"
+  />
+</FieldWrap>
+```
+
 ### Checkbox with Description
 
 ```jsx example
--- a/docs/pages/components/Checkbox.mdx
+++ b/docs/pages/components/Checkbox.mdx
@@ -162,7 +162,7 @@ import { Checkbox } from '@deque/cauldron-react'
 
 ```jsx example
 <FieldWrap>
-<div className="Field__label" id="checkbox-list-label">Checkbox with Description List Label</div>
+<div className="Field__label" id="checkbox-list-with-description-label">Checkbox with Description List Label</div>
   <Checkbox
     id="description-checkbox-list-unchecked"
     label="Description Checkbox Unchecked"
--- a/docs/pages/components/Checkbox.mdx
+++ b/docs/pages/components/Checkbox.mdx
@@ -157,6 +157,53 @@ import { Checkbox } from '@deque/cauldron-react'
 </FieldWrap>
 ```
 
+### Checkbox with Description List
+
+```jsx example
+<FieldWrap>
+  <Checkbox
+    id="description-checkbox-unchecked"
+    label="Description Checkbox Unchecked"
+    labelDescription="Additional text to describe the checkbox"
+    name="description-checkbox- unchecked"
+    value="1"
+  />
+  <Checkbox
+  checked
+    id="description-checkbox-checked"
+    label="Description Checkbox Checked"
+    labelDescription="Additional text to describe the checkbox"
+    name="description-checkbox-checked"
+    value="1"
+  />
+  <Checkbox
+    disabled
+    id="description-checkbox-disabled-unchecked"
+    label="Description Checkbox Disabled Unchecked"
+    labelDescription="Additional text to describe the checkbox"
+    name="description-checkbox-disabled-unchecked"
+    value="1"
+  />
+  <Checkbox
+    disabled
+    checked
+    id="description-checkbox-disabled-checked"
+    label="Description Checkbox Disabled Checked"
+    labelDescription="Additional text to describe the checkbox"
+    name="description-checkbox-disabled-checked"
+    value="1"
+  />
+  <Checkbox
+    error="You must check this checkbox to continue"
+    id="description-error-checkbox"
+    label="Description Error Checkbox"
+    labelDescription="Additional text to describe the checkbox"
+    name="description-error-checkbox"
+    value="1"
+  />
+</FieldWrap>
+```
+
 ## Props
 
 <ComponentProps
--- a/docs/pages/components/Checkbox.mdx
+++ b/docs/pages/components/Checkbox.mdx
@@ -162,43 +162,43 @@ import { Checkbox } from '@deque/cauldron-react'
 ```jsx example
 <FieldWrap>
   <Checkbox
-    id="description-checkbox-unchecked"
+    id="description-checkbox-list-unchecked"
     label="Description Checkbox Unchecked"
     labelDescription="Additional text to describe the checkbox"
-    name="description-checkbox- unchecked"
+    name="description-checkbox-list-unchecked"
     value="1"
   />
   <Checkbox
   checked
-    id="description-checkbox-checked"
+    id="description-checkbox-list-checked"
     label="Description Checkbox Checked"
     labelDescription="Additional text to describe the checkbox"
-    name="description-checkbox-checked"
+    name="description-checkbox-list-checked"
     value="1"
   />
   <Checkbox
     disabled
-    id="description-checkbox-disabled-unchecked"
+    id="description-checkbox-list-disabled-unchecked"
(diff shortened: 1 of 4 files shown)


yhafez:
> Checkbox/radio share styles, we cannot make a change to one without changing both, see below screenshot:
> 
> cauldron checkbox with description to the left of the checkbox label and checkbox

Addressed!

scurker:
After a quick zoom meeting with @yhafez and @dequelabs/cauldron-design, we settled on the following:

- Make sure the final designs for RadioGroup and Checkbox match the [Pattern Library](https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161871161/simulate/no-panels?mode=i)
  - This should partially address #949 and fully address #1034 
- Change the outer wrapper to use a class with something like `Checkbox__wrap` and `RadioGroup__wrap`

packages/styles/forms.css
```diff
@@ -16,6 +16,7 @@
   --field-label-text-color: var(--gray-90);
   --field-label-description-text-color: var(--gray-60);
   --input-min-width: 250px;
+  --focus-glow: rgba(181, 26, 209, 1);
```
scurker, line 19: We want to pull from global vars here:
https://github.com/dequelabs/cauldron/blob/720bc9bd5db3eab975d0c6d80639c8614f3899a4/packages/styles/variables.css#L41-L42

We also already have `focus-glow` defined in vars so we probably need to rescope the var here. Ideally following our guidelines it should be descriptive something like:

```css
--field-focus-ring-color: var(--focus-light);
```

With the alternative set for dark mode as well.

packages/styles/forms.css
```diff
@@ -159,10 +161,12 @@ textarea.Field--has-error:focus:hover,
 
 .Error {
   color: var(--error);
+  border-top: 1px solid var(--error);
   text-align: left;
   font-size: var(--text-size-smallest);
-  font-weight: var(--font-weight-normal);
-  padding-left: var(--space-half);
+  margin-top: var(--space-half);
+  margin-left: calc(24px + 2px + var(--space-half));
```
scurker, line 168: I know this was copy pasta, but we should update this to use `--icon-size`:

```suggestion
  margin-left: calc(var(--icon-size) + 2px + var(--space-half));
```

packages/styles/forms.css
```diff
@@ -395,5 +397,11 @@ textarea.Field--has-error:focus:hover,
 .cauldron--theme-dark .Radio__overlay--disabled,
 .cauldron--theme-dark .Checkbox__overlay--disabled.Icon--checkbox-checked,
 .cauldron--theme-dark .Checkbox__overlay--disabled.Icon--checkbox-unchecked {
-  color: #4c99a9;
+  color: #5d676f;
```
scurker, line 400:
```suggestion
  color: var(--stroke-dark);
```

packages/styles/forms.css
```diff
@@ -395,5 +397,11 @@ textarea.Field--has-error:focus:hover,
...
+}
+
+.Checkbox__overlay.Checkbox__overlay--focused,
+.Radio__overlay.Radio__overlay--focused {
+  border-radius: 5px;
+  box-shadow: 0 0 0 2px var(--focus-glow);
```
scurker, line 406: It looks like this was defined twice.

packages/styles/forms.css
```diff
@@ -31,6 +32,7 @@
   --field-required-text-color: var(--white);
   --field-label-text-color: var(--white);
   --field-label-description-text-color: rgba(255, 255, 255, 0.8);
+  --focus-glow: rgba(235, 148, 255, 1);
```
scurker, line 35: Same thing as above.

packages/styles/forms.css
```diff
@@ -197,13 +206,9 @@ textarea.Field--has-error:focus:hover,
...
-.Checkbox {
-  margin-bottom: var(--space-smallest);
-}
-
-.Checkbox + .Checkbox,
-.Radio + .Radio {
-  margin-top: var(--space-smallest);
+.Checkbox__wrap:last-of-type,
+.Radio__wrap:last-of-type {
+  margin-bottom: 0;
```
scurker, line 211: Clever!

packages/styles/forms.css
```diff
@@ -364,7 +363,10 @@ textarea.Field--has-error:focus:hover,
   color: #3c7aae;
 }
 
-.Checkbox__overlay--disabled.Icon--checkbox-checked,
+.Checkbox__overlay--disabled.Icon--checkbox-checked {
+  color: #78a6d8;
```
scurker, line 367: Following our [variable guidelines](https://cauldron.dequelabs.com/style-guidelines#variables) this should be either inheriting from a var or specifying a fallback value. I couldn't find this defined anywhere else so we should probably specify that as fallback value:

```css
color: var(--field-checkbox-disabled-color, #78a6d8)
```

packages/styles/forms.css
```diff
@@ -364,7 +363,10 @@ textarea.Field--has-error:focus:hover,
...
+}
+
 .Checkbox__overlay--disabled.Icon--checkbox-unchecked {
   color: #ccc;
```
scurker, line 371: While we're here, we should update this to use our global variable (similar to above):

```css
color: var(--field-checkbox-unchecked-disabled-color, var(--gray-40))
```
```

**Review by scurker — CHANGES_REQUESTED** [review]
Good work! Just a few nits to fix.

**Comment by scurker** [comment]
Update from @dequelabs/cauldron-design 

> There should be a global var `accent-primary-disabled` set to `#78a6d8` as we will be using it elsewhere.

**Comment by bobbyomari** [comment]
@yhafez here are my observations...

### checkbox dark theme:
- the active fill on touch down should be `accent-light` @ 25% opacity
- the active fill on touch down in error state should be `accent-danger` @ 25% opacity
- the `.Field_labelDescription` should be color `accent-light`
- Set `margin-top: var(--space-smallest);` for `.error` on checkbox with description. I think it strikes a better visual balance with how the code is structured (see attached screenshots, the one with more spacing is 8px). Looks fine with what it's at on
- If label is being used with checkboxes, can we set the field margin-bottom to `var(--space-smallest)`? This also comes to play with radios.

### checkbox light theme:
- the active fill on touch down should be `accent-primary` @ 25% opacity
- the active fill on touch down in error state should be `error` @ 25% opacity
- same as dark theme... set `margin-top: var(--space-smallest);` for `.error` on checkbox with description

### radio dark theme:
- it appears the radio focus is not circular, it's a rectangle
- `.Field_label` `padding-bottom` should be `var(--space-smallest)` 
- related to above: should we use `margin-bottom` instead of `padding-bottom`?
- like dark checkboxes, the active fill on touch down should be the same
- on inline layout, remove `margin-bottom` from `.Radio_wrap`

### radio light theme:
- same radio focus issue as dark theme
- same `.Field_label` request as dark theme
- like light checkboxes, the active fill on touch down should be the same
- same inline layout `margin-bottom` change as dark theme radios

### current implementation with `margin-top: 4px;`
image

### updated implementation with `margin-top: 8px;`
image

File: packages/styles/forms.css (hunk 4 above)
Comment (yhafez) [reply packages/styles/forms.css:406]: Ah, this was originally supposed to include `.cauldron--theme-dark`, but forgot to add that, though it looks like the way the CSS variables work, it's not necessary after all. Removed!

**Comment by bobbyomari** [comment]
I already sent these changes to @yhafez but wanted to document them here as well for a 2nd PR review...

### Checkboxes
- I noticed on the error state, when the checkbox is empty, the active fill is correct, but when the error checkbox is checked, the active fill stays red. Is there way to make this change to red active fill **only when the checkbox is unchecked and in an error state?** This applies to light and dark theme. 

### Radios
- The label for "do you like pizzas?" is now no longer `font-weight: 500;`
- Some discrepancies between Cauldron in UXPin vs development...
-- The space between the radio label and description should be `var(--space-half);`
-- The space between radio groups should be `var(--space-small);`
-- This helps create some better visual balance between what description belongs to what radio field. Below are screenshots to share what that change looks like.

### Radio with long description
image

### Radio with short description
image

File: docs/pages/components/Checkbox.mdx [hunk 9]
```diff
@@ -86,6 +86,48 @@ import { Checkbox } from '@deque/cauldron-react'
...
+    value="1"
+  />
+  <Checkbox
+    disabled
+    checked
+    id="checkbox-list-checkbox-disabled-checked"
+    label="Unchecked Disabled Checkbox"
+    name="disabled-checkbox-list-checked"
+    value="1"
+  />
```
scurker, line 120: It looks like these 2 are swapped

packages/styles/forms.css
```diff
@@ -14,7 +14,21 @@
...
+    var(--accent-primary) 20%,
+    transparent
+  );
+  --icon-error-active-color: color-mix(in srgb, var(--error) 20%, transparent);
+  --icon-checked-color: var(--accent-primary);
+  --icon-checked-disabled-color: var(--accent-primary-dsiabled, #78a6d8);
+  --icon-unchecked-disabled-color: var(--gray-40);
+  --icon-focus-color: var(--focus-light);
+  --error-text-color: var(--error);
+  --error-border-color: var(--error);
```
scurker, line 31: Can we prefix these with `field`?

packages/styles/forms.css
```diff
@@ -30,7 +44,20 @@
...
+    in srgb,
+    var(--accent-danger) 20%,
+    transparent
+  );
+  --icon-checked-color: var(--accent-light);
+  --icon-checked-disabled-color: var(--stroke-dark);
+  --icon-unchecked-disabled-color: var(--stroke-dark);
+  --icon-focus-color: var(--focus-dark);
+  --error-text-color: var(--error);
+  --error-border-color: var(--error);
```
scurker, line 60: See comment above.

packages/styles/forms.css
```diff
@@ -14,7 +14,21 @@
...
   --field-required-text-color: var(--gray-60);
   --field-label-text-color: var(--gray-90);
+  --field-label-error-text-color: var(--error);
   --field-label-description-text-color: var(--gray-60);
+  --icon-inactive-color: var(--gray-90);
+  --icon-active-color: color-mix(
+    in srgb,
+    var(--accent-primary) 20%,
+    transparent
+  );
```
scurker, line 24: GIven that this was only recently introduced we should define a static value so we don't accidentally introduce color contrast issues to users who do not yet support this feature.

scurker (changes requested):
Looking good! Just a last few code changes needed.

docs/pages/components/Checkbox.mdx
yhafez: Fixed

packages/styles/forms.css
yhafez: Just went with the rgba values for now as per [our discussion on slack](https://deque.slack.com/archives/C5MUFQ7UM/p1689889583970209)

packages/styles/forms.css
```diff
@@ -197,29 +236,41 @@ textarea.Field--has-error:focus:hover,
...
   color: var(--field-label-text-color);
+  margin-bottom: var(--space-half);
   font-size: var(--text-size-small);
   font-weight: var(--font-weight-medium);
-  padding-bottom: var(--space-half);
+  cursor: default;
+}
+
+.Field__label + .Checkbox__wrap,
+.Field__label + div[role='radiogroup'] {
```
scurker, line 262: What if `role="radiogroup"` isn't a div?

packages/styles/forms.css
```diff
@@ -118,6 +137,10 @@ textarea.Field--has-error,
   color: var(--field-border-color-error);
 }
 
+.Icon--checkbox-unchecked.Checkbox__overlay:active.Field--has-error:not(.Checkbox__overlay--disabled) {
```
scurker, line 140:
```suggestion
.Checkbox__overlay:active.Field--has-error:not(.Checkbox__overlay--disabled) {
```

This feels like an overly specific selector, can we not just use `.Checkbox__overlay`?

scurker (commented):
Code LGTM, 2 questions. Let's get @dequelabs/cauldron-design to give a thumbs up too.

packages/styles/forms.css
yhafez: Used `.Field__label + [role^='radiogroup']` instead~

packages/styles/forms.css
yhafez: Ah in this case, @bobbyomari mentioned to me that it should only show the red active fill state when checking the unchecked box, not when unchecking the checked box

scurker: If that is the case, we should be checking the checked state and not the class.

scurker: Something like `.Checkbox__overlay:not(:checked):active...`

scurker (approved):
Good work @yhafez!

packages/styles/forms.css
```diff
@@ -158,11 +181,14 @@ textarea.Field--has-error:focus:hover,
...
+  color: var(--field-error-text-color);
+  border-top: 1px solid var(--field-error-border-color);
   text-align: left;
-  font-size: var(--text-size-smallest);
   font-weight: var(--font-weight-normal);
-  padding-left: var(--space-half);
+  font-size: var(--text-size-smallest);
+  margin-top: var(--space-half);
+  margin-left: calc(var(--icon-size) + 2px + var(--space-half));
+  padding: var(--space-half) 0;
```
tbusillo, line 191: @yhafez This is the change that is doing this.

PR #1117, opened 2023-07-19
docs: update Tooltip to v2 docs

scurker:


docs/pages/components/Tooltip.mdx
```diff
@@ -0,0 +1,183 @@
...
+}
+```
+
+## Props
+
+<ComponentProps
+  children={{
+    required: true
+  }}
+  props={[
```
thuey, line 148: There are a couple of undocumented props: `association` and `hideElementOnHidden`. Is it intentional to leave these undocumented?

scurker: It wasn't intended as I was porting over the existing documentation but this can be a quick fix here.

thuey (approved):

PR #1125, opened 2023-07-25
feat: support refs for radioGroup component

yhafez:
Relates to [Walnut #5739](https://github.com/dequelabs/walnut/issues/5739)

### DIFF
--- a/packages/react/src/components/RadioGroup/index.tsx
+++ b/packages/react/src/components/RadioGroup/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect } from 'react';
+import React, { Ref, useState, useRef, useEffect, forwardRef } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Icon from '../Icon';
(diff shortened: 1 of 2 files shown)


packages/react/src/components/RadioGroup/index.tsx
```diff
@@ -157,22 +165,23 @@ RadioGroup.propTypes = {
...
     if (!props['aria-label'] && !props['aria-labelledby']) {
       return new Error(
         `${componentName} must have an "aria-label" or "aria-labelledby" prop`
       );
     }
+    return null;
   },
   className: PropTypes.string,
   defaultValue: PropTypes.string,
-  onChange: PropTypes.func
```
scurker, line 175: This shouldn't be required because we do set a default.

packages/react/src/components/RadioGroup/index.tsx
```diff
@@ -157,22 +165,23 @@ RadioGroup.propTypes = {
       id: PropTypes.string.isRequired,
       label: PropTypes.string.isRequired,
       labelDescription: PropTypes.string
-    })
```
scurker, line 160: Name does appear to be optional from the typescript types.

packages/react/src/components/RadioGroup/index.tsx
```diff
@@ -17,137 +17,145 @@ export interface RadioGroupProps {
...
+            </span>
+          )}
+        </div>
+      );
+    });
+
+    // Hack to prevent ESLint from erroring about this variable not
+    // being used. We want to pull it from `props` to ensure it's
+    // not passed through to the radiogroup element.
+    void defaultValue;
```
scurker, line 140: I think what we can do instead is set `// eslint-disable-next-line no-unused-vars` for `defaultValue`. I'm not sure why we did this.

scurker (changes requested):

packages/react/src/components/RadioGroup/index.tsx
yhafez: Isn't it being used [on this line](https://github.com/dequelabs/cauldron/blob/423688b873a3da251bc7e2ea4ab874acf654f63a/packages/react/src/components/RadioGroup/index.tsx#L38-L40)?

I tried taking this line out completely, and eslint didn't complain, is that okay?

packages/react/src/components/RadioGroup/index.tsx
yhafez: I think GitHub might have collapsed the relevant line. This change isn't for `name` but for the object in the `radios` array

packages/react/src/components/RadioGroup/index.tsx
yhafez: Since we pass a default, shouldn't we make the [TypeScript type here optional as well](https://github.com/dequelabs/cauldron/blob/82aade9999dc85bc103c2db51539c645fccc8969/packages/react/src/components/RadioGroup/index.tsx#L19). Otherwise, my linter is complaining on this line:

```
Type 'Requireable<(...args: any[]) => any>' is not assignable to type 'Validator<(radio: RadioItem, input: HTMLElement) => void>'.
  Types of property '[nominalTypeHack]' are incompatible.
    Type '{ type: ((...args: any[]) => any) | null | undefined; } | undefined' is not assignable to type '{ type: (radio: RadioItem, input: HTMLElement) => void; } | undefined'.
      Type '{ type: ((...args: any[]) => any) | null | undefined; }' is not assignable to type '{ type: (radio: RadioItem, input: HTMLElement) => void; }'.
        Types of property 'type' are incompatible.
          Type '((...args: any[]) => any) | null | undefined' is not assignable to type '(radio: RadioItem, input: HTMLElement) => void'.
            Type 'undefined' is not assignable to type '(radio: RadioItem, input: HTMLElement) => void'.ts(2322)
```

packages/react/src/components/RadioGroup/index.tsx
scurker: Ah yeah, that's fine. I think that was added after we put in this hack so we should be good to remove.

packages/react/src/components/RadioGroup/index.tsx
scurker: Yup, my bad 🤦

scurker (approved):

PR #1128, opened 2023-07-26
feat: add support for tabIndex and other div element props to radiogroup

yhafez:
Relates to [Walnut #5739](https://github.com/dequelabs/walnut/issues/5739)


scurker (approved):
Updated to `fix` instead of `feat` but looks good 🎉

PR #1131, opened 2023-07-27
fix(styles): fix bug where Error class changes are applied globally

yhafez:
Relates to bug noted in [this Slack thread](https://deque.slack.com/archives/G017VP5SCUB/p1690478003393319)

Reverts [changes to Error class styles](https://github.com/dequelabs/cauldron/pull/1110/files#r1276602901), and scopes new changes to their respective elements

### DIFF
--- a/packages/styles/forms.css
+++ b/packages/styles/forms.css
@@ -188,8 +188,6 @@ textarea.Field--has-error:focus:hover,
   padding-left: var(--space-half);
 }
 
-.Field .Error,
-.Field__select .Error,
 .Checkbox__wrap .Error {
   border-top: 1px solid var(--field-error-border-color);
   margin-top: var(--space-half);
--- a/packages/styles/forms.css
+++ b/packages/styles/forms.css
@@ -185,12 +185,11 @@ textarea.Field--has-error:focus:hover,
   text-align: left;
   font-size: var(--text-size-smallest);
   font-weight: var(--font-weight-normal);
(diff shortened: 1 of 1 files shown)


packages/styles/forms.css
```diff
@@ -182,10 +182,16 @@ textarea.Field--has-error:focus:hover,
 
 .Error {
   color: var(--field-error-text-color);
-  border-top: 1px solid var(--field-error-border-color);
   text-align: left;
-  font-weight: var(--font-weight-normal);
   font-size: var(--text-size-smallest);
```
tbusillo, line 186: Think this was done in your previous PR as well, per the `git blame`. Was it left in intentionally?

scurker (dismissed):
There's a couple of other spots that we also need to fix:

cauldron select component with invalid padding/border

cauldron textfield component with invalid padding/border

packages/styles/forms.css
scurker: I believe the font-weight does need to stay out as we were trying to standardize on that.

packages/styles/forms.css
```diff
@@ -182,10 +182,16 @@ textarea.Field--has-error:focus:hover,
...
+  font-weight: var(--font-weight-normal);
```
yhafez, line 187: @tbusillo @scurker, the font-weight is still here, just got moved around

scurker (approved):

Issue #1031, opened 2023-04-19
Implement Notice Component


scurker:
This differs from toast where it's static content that can be stacked and/or placed anywhere and does not have a "fixed" position. It could still be dismissible or not, depending on its usage.

# Designs and functionality

**Things to note:**

- This design applies to both Light and Dark themes, similar to our `Toast` component.
- This should support the following variants similar to Toast, other variants are out of scope:  
  - `info` (blue)
  - `caution` (yellow)
  - `danger` (red)
- This component can be displayed anywhere where the content sits above or below using `display: block;`.
- Use `width: 100%;` to ensure this component stretches to the width of its container. 
- The author can change the icon to represent something other than the default icons used.
- Two different examples (not implemented as actual variants, this can be controlled via author content):
  - Thick variant: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6783863,notice-thick
  - Thin variant: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6783867,notice-thin

Component would look something like:

```jsx
<Notice type={type} title={title}>
  Optional content to be displayed below the title...
</Notice>
```

**How to view multiple states:**
1. Click on `Layers` on top left
4. Select the `Notice` layer group
5. Change the state from `Success` to view all other styles (`warning`, `danger`, `info`)

PR #1136, opened 2023-07-31
feat(react,styles): add Notice component

tbusillo:
closes: https://github.com/dequelabs/cauldron/issues/1031


packages/react/src/components/Notice/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * The cauldron Notice notification component
+ */
+const Notice = forwardRef<HTMLDivElement, NoticeProps>(
+  ({ type, title, icon, children, ...otherProps }: NoticeProps, ref) => {
+    const validType =
+      type && Object.keys(iconTypeMap).includes(type) ? type : 'info';
+    const validIcon =
+      icon && iconTypes.includes(icon)
+        ? icon
+        : (iconTypeMap[validType] as IconType);
```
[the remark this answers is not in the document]
scurker: Alternatively, we could skip the lookup completely here for `iconTypes` as we could rely on typescript to do the validation for us. What do you think?

docs/components/example.css
```diff
@@ -48,6 +48,10 @@ body.cauldron--theme-dark {
   margin: 0;
 }
 
+.Component__example__react .Notice {
+  line-height: 1.618;
+}
```
scurker, line 53: Why `line-height`, shouldn't we be rendering the component as is?

docs/pages/components/Notice.mdx
```diff
@@ -0,0 +1,192 @@
...
+but can be used in more flexible ways.
+
+For example, the `Notice` component can be used to display a notification banner within a [Panel](./Panel) component
+or anywhere else in the DOM without needing to be absolutely/statically positioned.
+
+In addition, the `Notice` component has only limited dynamic functionality built in, allowing developers to
+have full control over the underlying element via its `forwardRef`, while still having useful base styling..
+
+```jsx example
+<Notice type="info" title='Information'>
```
scurker, line 27:
```suggestion
<Notice type="info" title="Information">
```

scurker, line 27: Small nit: We typically have the variants listed out separately. Having each variant broken out allows for a copying of a single variant and breaks out each variant so it can appear/get linked via the table of contents.

docs/pages/components/Notice.mdx
```diff
@@ -0,0 +1,192 @@
...
+  Aliquam erat volutpat. Donec placerat nisl magna, et faucibus arcu condimentum sed.
+</Notice>
+
+<Notice type="caution" title="Caution">
+  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam at porttitor sem.
+  Aliquam erat volutpat. Donec placerat nisl magna, et faucibus arcu condimentum sed.
+</Notice>
+```
+
+### Notice With Icon
```
scurker, line 38: Is this the right title for this section? It looks like the heading is what changed here. Additionally, It feels like we should be able to pass in arbitrary headings and have the styles match the expected title style without modifying the heading styles.

docs/pages/components/Notice.mdx
```diff
@@ -0,0 +1,192 @@
...
+The `Notice` component will support the spreading of any props normally available to a `div` element. This includes
+properties such as `role`, `aria-label`, `aria-labelledby`, `id`, etc.
+
+```jsx example
+<Notice type="info" title="Alert!" role="alert">
+  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam at porttitor
+  sem. Aliquam erat volutpat. Donec placerat nisl magna, et faucibus arcu
+  condimentum sed.
+</Notice>
+```
```
scurker, line 157: We haven't typically been documenting additional html properties. I'm not opposed to doing so, but I wonder if there's a generic way we could maybe think about doing this via component props for other components?

docs/pages/components/Notice.mdx
```diff
@@ -0,0 +1,192 @@
...
+      You are a star!
+    </h2>
+  </div>
+  <p style={{ margin: 'var(--space-half) 0 0 var(--space-half)' }}>
+    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam at
+    porttitor sem. Aliquam erat volutpat. Donec placerat nisl magna, et faucibus
+    arcu condimentum sed.
+  </p>
+</Notice>
+```
```
scurker, line 122: What's this?

scurker (changes requested):
Looking good so far! Just some comments/thoughts after reviewing below:

packages/styles/notice.css
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.Notice .Notice__title + .Notice__content {
+  margin-top: var(--space-quarter);
+  margin-left: var(--space-quarter);
+}
+
+.Notice .Icon > svg {
+  height: var(--space-medium);
+  width: var(--space-medium);
```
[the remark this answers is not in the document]
tbusillo: I'm seeing what you're saying re: 16x16 when you select the element in the canvas, but the CSS on the righthand side is 18px. I'm fine with either one and I'll defer to your judgement if you're 100% sure, just want to make sure the one you're going off of is the one we want. . 

CleanShot 2023-08-01 at 13 58 49@2x

packages/styles/notice.css
```diff
@@ -0,0 +1,66 @@
...
+.Notice--info {
+  --notice-background-color: var(--notice-info-color);
+}
+
+.Notice--caution {
+  --notice-background-color: var(--notice-caution-color);
+}
+
+.Notice {
+  display: flex;
```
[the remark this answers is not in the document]
tbusillo: So I had tried with `block` and it created a ton of hassle when it came to vertically aligning the icon, text, and broken up blocks of text. IMO, we'd just be kicking the can down the develop by using something like block, as when I was experimenting with it, it required quite a bit of weirdness to get things aligning properly in various configurations.

packages/react/src/components/Notice/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const iconTypeMap = {
+  caution: 'caution',
+  info: 'info-circle'
+};
+
+export interface NoticeProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, 'title'> {
+  type?: keyof typeof iconTypeMap;
+  title?: React.ReactNode;
```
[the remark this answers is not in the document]
tbusillo: Out of curiosity, why? Couldn't we just allow them to compose the entire thing, gaining admittedly minimal benefit of the built-in styling, icons, and defaults?

docs/components/example.css
tbusillo: It's specifically an issue when a string gets passed in without an HTML element, e.g., 

```ts

<Notice type='info' title='Information'>A bunch of text that doesn't have a parent element such as p, span, etc.</Notice>

```

The line height for the text is super condensed - much more so than the line height we apply to paragraph elements - and really tough to target with styles outside of doing something potentially too generic that effects elements we don't want it to. 

I could try to hide it by just keeping the content blocks short instead of lengths of text that would wrap, then let the end-user determine the appropriate solution. 

I went with adding the style here as this is a problem now run into twice (with the Panel as well) and not having a reset for margins (e.g., the ones in Chrome for paragraphs) or line height caused similar issues. Seemed a low impact way to solve it, without strictly requiring those styles in the component itself.

docs/pages/components/Notice.mdx
tbusillo: Hmm, I would argue that they can match the styling by using the same class name. I know we did a few things that were somewhat smart with the Panel, but not sure that makes sense here, especially since my impression of this component was it's intended to allow for a fair amount of flexibility. 

I guess if we were to do that, how do you envision that would work? Cloning the element and adding an arbitrary class during render? Aside from that, what if the title is not intended to be style the way we expect it to? 

IMO, the shared settings beta notification banner is a good example of what I'm referring to.

docs/pages/components/Notice.mdx
tbusillo: Yeah, I honestly don't see a reason to typically, and other component libraries take a similar approach (e.g., MUI specifically calls out in a few places they don't generally describe HTML props). 

Where I've seen exceptions made, is when a component is pretty much guaranteed to be used with certain HTML props and the usage is expected to be common enough to warrant an example.

A good example where MUI makes an exception is with its Button component, specifically usage of the `onClick` prop. I'm fine with removing it, but I saw this example as being similar to that. 

To your point, I think it would be worth calling this out elsewhere like you mentioned regardless of what we do here.

docs/pages/components/Notice.mdx
tbusillo: It's a usage where just the base styles are being used and the component's contents are composed. People don't have to do this if the built in props get them where they want to go, but figured it would be nice to have that flexibility.

packages/react/src/components/Notice/index.tsx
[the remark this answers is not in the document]
tbusillo: I'm fine with either one, although without narrowing the value in some way we likely would need to either assert the type or allow for a generic. 

I'm hesitant to do that since if we make the interface overly generic, TS may not be as helpful as it typically would be.

packages/styles/notice.css
```diff
@@ -0,0 +1,66 @@
...
+
+.Notice p {
+  margin: 0;
+}
+
+.Notice .Notice__title,
+.Notice .Notice__title > :is(h1, h2, h3, h4, h5, h6) {
+  display: flex;
+  align-items: center;
+  justify-content: flex-start;
```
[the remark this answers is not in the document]
tbusillo: I'm fine removing it, but I will say the icon and title seemed to vertically align with each other better as opposed to using `center`.

packages/styles/notice.css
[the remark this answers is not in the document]
scurker: I'm double checking with @dequelabs/cauldron-design, but you might be right this is a different size than our standard icons. It looks like the text size is `18px` here but given that, we should be inheriting the icon from that size and not from space.

```css
:root {
  --notice-icon-size: var(--text-size-normal);
}

.Notice .Icon > svg {
  height: var(--notice-icon-size);
  width: var(--notice-icon-size);
}
```

packages/react/src/components/Notice/index.tsx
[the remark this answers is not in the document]
scurker: All of the designs show the presence of a title, and this is something we want to enforce via the component's interface.

docs/pages/components/Notice.mdx
scurker: I'm not sure this is a pattern we want to encourage in documentation vs. something like `Panel`. `Panel` is intended to be a generic container component, where this should have a more restrictive interface.

packages/react/src/components/Notice/index.tsx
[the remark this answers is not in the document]
scurker: I also confirmed with @dequelabs/cauldron-design that title is required.

packages/styles/notice.css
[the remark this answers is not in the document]
scurker: I removed all the `flex` properties and didn't see any layout changes with the `Notice` component. You might need to show me what you're seeing.

packages/styles/notice.css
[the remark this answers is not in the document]
scurker: Just curious, are you referring to `align-items: center`? `justify-content: flex-start` is the default value for flex containers so it should be redundant, where `align-items: center` aligns items along the vertical axis here.

docs/components/example.css
scurker: This seems like an implementation issue and shouldn't be something that's just fixed in documentation. I would expect the component structure that you passed in to work out of the box without any modifications on the developer side.

docs/pages/components/Notice.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+## Props
+
+<ComponentProps
+  children={{ required: false }}
+  props={[
+    {
+      name: 'type',
+      type: ["'info'", "'caution'"],
+      defaultValue: '"info"',
```
scurker, line 167:
```suggestion
      defaultValue: "info",
```

docs/pages/components/Notice.mdx
```diff
@@ -0,0 +1,192 @@
...
+      description:
+        'Notice preset variant that sets both the background color and Icon type (optionally overwritable)',
+      required: false
+    },
+    {
+      name: 'title',
+      type: 'React.ReactNode',
```
scurker, line 174: This should be [`ContentNode`](https://github.com/dequelabs/cauldron/blob/82de49c0307ba11266fe348884852b039e3ead9b/packages/react/src/types.ts#L10-L19) as we specifically use that type to avoid things like `boolean` or `null`.

packages/react/src/components/Notice/index.tsx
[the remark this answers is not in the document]
scurker: Sorry I missed this earlier, but this should also be [`ContentNode`](https://github.com/dequelabs/cauldron/blob/82de49c0307ba11266fe348884852b039e3ead9b/packages/react/src/types.ts#L10-L19)

packages/styles/notice.css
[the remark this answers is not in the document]
scurker: After checking with design here, it sounds like we want the icon to match the size of the title. I think using `em` approach may provide the best results.

docs/components/example.css
tbusillo: I'm fine with that, but if that's the case, I recommend adding more universal line-height for text containing elements than just `<p>`. This is not just an issue here, it is an issue with most other components as well, the docs just don't include long enough copy for it to be really seen (case in point, the `Panel` example copy is one word, the `ExpandCollapse` and `Accordion` have similar issues. 

[Here is where](https://github.com/dequelabs/cauldron/blob/10f53c90fe2466f913bb00c49fc0b8e1aad501b1/packages/styles/base.css#L84-L86) the `line-height` for the `<p>` comes from (although you probably already know that...)

Here's a comparison:

## Text content
CleanShot 2023-08-01 at 16 27 42@2x

## Paragraph
CleanShot 2023-08-01 at 16 36 06@2x

docs/pages/components/Notice.mdx
tbusillo: Hmm, not sure I'm fully understanding what you're saying/can't tell if we're saying the same thing or not. I'm going address as much as I can/shouldn't need much discussion and maybe we can talk through a few of the others tomorrow morning?

docs/pages/components/Notice.mdx
tbusillo: That's fair/reasonable and I'll go ahead and remove it.

packages/styles/notice.css
```diff
@@ -0,0 +1,66 @@
...
+  --notice-title-text-color: var(--accent-dark);
+  --notice-title-font-weight: var(--font-weight-medium);
+  --notice-background-color: var(--notice-info-color);
+  --notice-border-color: var(--accent-dark);
+}
+
+.cauldron--theme-dark {
+  --notice-text-color: var(--accent-dark);
+  --notice-title-text-color: var(--accent-dark);
+}
```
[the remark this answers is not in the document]
tbusillo: Hmm, are you sure? Running into issues with links and headings in particular. Or are you just saying can use the same variables for both since they're not changing?

packages/styles/notice.css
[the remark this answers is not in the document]
tbusillo: That works for me, but fair warning, IIRC some of the icon viewboxes won't play very nicely with this.

docs/pages/components/Notice.mdx
scurker: Yeah we can pair on this in the morning and get it knocked out.

packages/styles/notice.css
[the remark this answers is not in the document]
tbusillo: FYI, I just did this and realizing that `1em` is actually 15px...I'm not sure it looks right like that, or at the very least, it's a bit inconsistent with all of our other uses of the same pattern (e.g., the toast and even the Note component used in the docs). 

Notice UXPin with CSS styles listed

Here's a comparison between this and the Note: 

Notice and Note component comparison

scurker (approved):
LGTM!

shawnsharpDQ:
I am noticing one discrepancy, @tbusillo looks like the top and bottom padding should be 12px not 16px

scurker (dismissed):

docs/pages/components/Notice.mdx
```diff
@@ -76,14 +87,39 @@ Icon Type](/components/icon).
...
+```jsx example
+function DismissableNotice() {
+  const [dismissed, setDismissed] = useState(false);
+  const handleDismiss = () => {
+    setDismissed(true);
+    setTimeout(() => setDismissed(false), 2000);
+  };
+
+  return dismissed ? null : (
+    <Notice type="info" title="Information" dismissable>
```
scurker, line 101:
```suggestion
    <Notice type="info" title="Information">
```

Dismissable doesn't appear to be a valid prop.

scurker (changes requested):
Last change and then we should be good to go!

docs/pages/components/Notice.mdx
tbusillo: Oof, I forgot Copilot added that in when tabbing through 🤣

scurker (approved):

PR #1143, opened 2023-08-01
fix: fixes overflow bug on text-field and text-area

### LINKED ISSUE #1138: Bug: Text-field overflows its container
### Steps to reproduce

Steps to reproduce

Find a text field in a sufficiently narrow container, such as that in `/signup`. Increase zoom to 200% and shrink viewport width.

### Expected Behavior

Text-field shrinks to fit enclosing container.

### Actual Behavior

Text-field overflows out of container as a result of its [min-width property](https://github.com/dequelabs/cauldron/blob/82de49c0307ba11266fe348884852b039e3ead9b/packages/styles/forms.css#L313).

### Version

5.6.2

Chrome

### Relevant log output

_No response_

Additional information

Screenshot:
Screenshot 2023-07-29 at 6 32 05 PM

yhafez:
Closes #1138

### DIFF
--- a/packages/styles/forms.css
+++ b/packages/styles/forms.css
@@ -310,7 +310,7 @@ textarea.Field--has-error:focus:hover,
   box-sizing: border-box;
   font-size: var(--text-size-small);
   color: var(--field-content-color);
-  min-width: var(--input-min-width);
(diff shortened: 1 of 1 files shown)


scurker (approved):
🎉

PR #1144, opened 2023-08-01
feat(react): allow for NavBar to include additional props for html div elements.

scurker:
I was updating documentation for NavBar and noticed that it did not currently accept additional props which might be needed in some instances.


packages/react/src/components/NavBar/NavBar.tsx
```diff
@@ -9,7 +9,7 @@ import PropTypes from 'prop-types';
 import Icon from '../Icon';
 import { useId } from 'react-id-generator';
 
-interface NavBarProps {
+interface NavBarProps extends React.HTMLAttributes<HTMLDivElement> {
   children: React.ReactNode;
   initialActiveIndex?: number;
```
thuey, line 14: I realize this wasn't part of your PR, but while you're doing some clean-up, this does not appear to be used anywhere

```suggestion
```

scurker: I had noticed, but I guess we can include this too.

thuey (approved):

PR #1153, opened 2023-08-07
fix(breadcrumbs): amend `/` separator colours in light and dark mode

Zidious:
This PR updates the colours used for the `/` in light and dark mode. See https://github.com/dequelabs/walnut/issues/6291#issuecomment-1668485052 for reference.

### DIFF
--- a/packages/styles/breadcrumb.css
+++ b/packages/styles/breadcrumb.css
@@ -7,8 +7,13 @@
   margin: 0;
 }
 
-.Breadcrumb__Separator {
-  color: var(--gray-40);
+.cauldron--theme-light .Breadcrumb__Separator {
+  color: var(--gray-60);
+  padding: 0 var(--space-half);
+}
+
+.cauldron--theme-dark .Breadcrumb__Separator {
+  color: var(--accent-light);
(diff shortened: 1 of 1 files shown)


packages/styles/breadcrumb.css
```diff
@@ -7,8 +7,13 @@
   margin: 0;
 }
 
-.Breadcrumb__Separator {
-  color: var(--gray-40);
+.cauldron--theme-light .Breadcrumb__Separator {
```
scurker, line 10: Mostly looks good. We need to split out the changes so that `padding` is only listed once.

```css
.BreadCrumb__Separator {
  padding: ...
}

.cauldron--theme-dark { ... }
```

The default mode is light, so the color doesn't have to be set explicitly there.

scurker (changes requested):
Mostly looks good. We need to split out the changes so that `padding` is only listed once.

packages/styles/breadcrumb.css
Zidious: Yep - good catch!

scurker (approved):

Issue #1053, opened 2023-05-02
Sync Table with UXPin


scurker:
Changes:

- Thicker border between TableHeader / TableBody (vertical and horizontal) for bordered variant
- Fix the sizing of the sort iconography (when sorted, the icon is slightly displaced)
- For sortable table headers, change the icon from `space-between` to `flex-start`
- Update sort up / down icon so it's not as large and matches UXPin
- Add a note to documentation to recommend that authors indicate the current sorted column by default:
  - > As a best practice, when a table has sorted columns an author should indicate which column is currently sorted by setting the sort direction for the active sorted column.

UXPin Design System: https://app.uxpin.com/design-system/c933c74928e2195cc540/ui-patterns/213536

bobbyomari:
This has been completed and is part of #1154

PR #1154, opened 2023-08-07
fix(table): adjust border table thickness and sortable table icon

JoshMK:
- adds thickness to bordered tables to match UXPin designs
- adds a note for recommended best practices when sorting tables
- adjusts the table-sort-ascending and table-sort-descending icons to match UXPin designs

Closes Issue: #1053 

Testing:

Confirm bordered table borders and sort icons [now match their UXPIN counterparts](https://app.uxpin.com/design-system/c933c74928e2195cc540/ui-patterns/213536)

### DIFF
--- a/packages/styles/table.css
+++ b/packages/styles/table.css
@@ -117,6 +117,10 @@
   border: 1px solid var(--gray-40);
 }
 
+.Table--border .TableHeader {
+  border-bottom: 2px solid var(--gray-40);
+}
+
 .cauldron--theme-dark .Table--border,
 .cauldron--theme-dark .Table--border .TableHeader,
 .cauldron--theme-dark .Table--border .TableFooter,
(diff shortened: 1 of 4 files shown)


packages/styles/table.css
```diff
@@ -110,11 +110,34 @@
...
-.Table--border,
+.Table--border {
+  border: 1px solid var(--gray-40);
+}
+
 .Table--border .TableHeader,
 .Table--border .TableFooter,
 .Table--border .TableCell {
-  border: 1px solid var(--gray-40);
+  border: 2px solid var(--gray-40);
```
scurker, line 120: The 2px border should only apply to the border between whatever is the header/body. The border between individual cells should still be 1px.

scurker (changes requested):

scurker (approved):

PR #1166, opened 2023-08-15
fix(react): fix options menu size and update documentation to v2 docs

scurker:


packages/styles/options-menu.css
```diff
@@ -18,7 +18,6 @@
 
 .OptionsMenu {
   position: relative;
-  width: 34px;
```
scurker, line 21: This is an erroneous value and should not be there. It doesn't appear to effect other products from testing.

docs/pages/components/OptionsMenu.mdx
```diff
@@ -0,0 +1,100 @@
...
+    description: 'Function called when the OptionMenu component is closed.'
+  },
+  {
+    name: 'onSelect',
+    type: 'function',
+    description: 'Function called when a child of the OptionMenu component is selected.'
+  },
+  {
+    name: 'trigger',
+    type: 'function',
```
thuey, line 72: Nothing to do with your PR, but it seems weird to me that this takes a function instead of a `ReactElement`

thuey (approved):

docs/pages/components/OptionsMenu.mdx
scurker: It's a [render function/prop](https://react.dev/reference/react/cloneElement#passing-data-with-a-render-prop) because it needs to spread certain properties onto the component.

If we were to rewrite this today, we would likely use a hook instead to provide the same functionality.

thuey: Ah, that makes sense

Issue #1088, opened 2023-06-06
Combobox Component (Single Select)


scurker:
Follow general ARIA practices guidelines: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

This implementation addresses only single select comboboxes.

- for multiselect see #1103.
- for custom form values see #1184 

List items can be provided via a child component, `ComboboxOption`:

```jsx
<Combobox label="Combobox">
  <ComboboxOption value="...">
    Item
    Longer Description
  </ComboboxOption>
</Combobox>
```

...or via a list with item definition:

```jsx
<Combobox label="Combobox" options={[{ label: 'Foo', value: 'foo' }]} />
```

The label provided for Combobox should be of type [`ContentNode`](https://github.com/dequelabs/cauldron/blob/d9725cdd30faf9005ac53809d829e5b80c8983e8/packages/react/src/types.ts#L14-L19) to allow for additional values beyond strings:

```jsx
<Combobox label={`<Icon type="apple"/> Fruit`}>
  ...
</Combobox>
```

To allow for grouping of items, an additional `ComboboxGroup` component can be used to wrap combobox items:

```jsx
<Combobox label="Combobox">
  <ComboboxGroup label="Group A">
   <ComboboxItem>1</ComboboxItem>
   <ComboboxItem>2</ComboboxItem>
   <ComboboxItem>3</ComboboxItem>
  </ComboboxGroup>
  <ComboboxGroup label="Group B">
   <ComboboxItem>4</ComboboxItem>
   <ComboboxItem>5</ComboboxItem>
   <ComboboxItem>6</ComboboxItem>
  </ComboboxGroup>
</Combobox>
```

For the list of items and `ComboboxOption`, providing the `value` property is optional and will reflect the value of the label associated with that item. Additionally, a `selected` prop can be provided to both to indicate that item is currently selected.

Under the hood:

- The input component should behave similarly to `<TextField />` (some different accommodations may need to be made to help support #1103)
  - Include inner right padding to prevent text overflowing the down arrow
  - General input props should be supported such as `disabled`, `required`, etc.
- Use the same styles as `<OptionsMenuList>` to display the list of items
- Allow custom filtering function to display matching items on input via prop 
  - Defaults to a text matching based on the the value or label of the item
- Support both auto selection and manual selection during autocomplete (see `#2` and `#3` list items in the autocomplete section of the aria authoring practices)
- When selected, an item should show it's selected state in the popup
[...]

PR #1167, opened 2023-08-16
feat(react): add Listbox component

scurker:
see #1088 

`Listbox` will be one of the driving behaviors behind `Combobox` to provide that set of functionality for the combobox list. Eventually, we can also refactor `OptionMenu` to also utilize this component as well to help consolidate these types of behaviors within a single component.


docs/pages/components/Listbox.mdx
```diff
@@ -0,0 +1,212 @@
+---
+title: Listbox
+description: An unstyled component to provide a keyboard navigable list of options.
+source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/Listbox/Listbox.tsx
```
anastasialanz, line 4:
```suggestion
description: An unstyled component to provide a keyboard-navigable list of options.
source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/Listbox/Listbox.tsx
```

docs/pages/components/Listbox.mdx
```diff
@@ -0,0 +1,212 @@
...
+]} />
+
+### ListboxGroup
+
+<ComponentProps className={true} refType="HTMLUListElement" props={[
+  {
+    name: 'label',
+    required: true,
+    type: ['string', 'number'],
+    description: 'Value to be applied to the listbox option. When omitted, value will default to the text content of the option.'
```
dequejosie, line 204: Is this `description` correct? It seems that this prop is required, how is it omitted?

packages/styles/forms.css
```diff
@@ -58,7 +60,6 @@ textarea,
 [role='menuitemcheckbox'],
 [role='menuitemradio'],
 [role='textbox'],
-[aria-haspopup='listbox'],
```
dequejosie, line 61: What were these styles for?

docs/pages/components/Listbox.mdx
scurker: Whoops! Bad copy pasta, fixed.

packages/styles/forms.css
scurker: These styles were overly aggressive and getting applied incorrectly. The prior styles were assuming that anything that had `[aria-haspopup='listbox']` should be styled like an input - which is incorrect.

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,216 @@
...
+    target: T;
+    value: ListboxValue;
+  }) => void;
+  onSelectionChange?: ({
+    value
+  }: {
+    previousValue: ListboxValue;
+    value: ListboxValue;
+  }) => void;
+  onActiveChange?: (option: ListboxOption) => void;
```
dequejosie, line 33: Seems this prop is not documented.

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,216 @@
...
+    }, [activeOption]);
+
+    const handleSelect = useCallback(
+      (option: ListboxOption) => {
+        setActiveOption(option);
+        onSelect?.({ target: option.element, value: option.value as string });
+        // istanbul ignore else
+        if (!isControlled) {
+          setSelectedOption(option);
+          onSelectionChange?.({
```
dequejosie, line 98: It seems that `onSelectionChange` only applies to uncontrolled `listbox`. It makes sense, cause controlled `listbox` has `onSelect`, and uncontrolled `listbox` has to rely on `onSelectionChange` if the user wants to do something whenever the selection changes. it could be confusing if a controlled listbox user accidentally passes in `onSelectionChange` and it does not do anything.

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,216 @@
...
+        }
+      },
+      [isControlled, selectedOption]
+    );
+
+    const handleKeyDown = useCallback(
+      (event: React.KeyboardEvent<HTMLElement>) => {
+        onKeyDown?.(event);
+
+        if (!keys.includes(event.key)) return;
```
dequejosie, line 111: nit:
```suggestion
        if (!keys.includes(event.key)) { 
          return
        };
```

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,216 @@
...
+        event.preventDefault();
+        const enabledOptions = options.filter(
+          (option) => !isDisabledOption(option)
+        );
+
+        // istanbul ignore next
+        if (!enabledOptions.length) return;
```
dequejosie, line 118: The same.

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,216 @@
...
+          ({ element }) => element === currentOption.element
+        );
+        const allowCyclicalNavigation = navigation === 'cycle';
+
+        switch (event.key) {
+          case up:
+            const previousOption =
+              currentIndex === 0 && allowCyclicalNavigation
+                ? lastOption
+                : enabledOptions[Math.max(currentIndex - 1, 0)];
```
dequejosie, line 140: this always returns 0, right?

packages/react/src/components/Listbox/ListboxContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+} & ListboxContext<T>;
+
+/* istanbul ignore next */
+const ListboxContext = createContext({
+  options: [],
+  active: null,
+  selected: null,
+  setOptions: () => null,
+  onSelect: () => null
+});
```
dequejosie, line 29: Why is typescript not happy here?

packages/react/src/components/Listbox/ListboxGroup.tsx
```diff
@@ -0,0 +1,37 @@
...
+interface ListboxGroupProps extends React.HTMLAttributes<HTMLElement> {
+  as?: React.ElementType | string;
+  groupLabelProps?: React.HTMLAttributes<HTMLLIElement>;
+  label: ContentNode;
+}
+
+const ListboxGroup = forwardRef<HTMLElement, ListboxGroupProps>(
+  (
+    {
+      as: Component = 'ul',
```
dequejosie, line 14: I wonder what if a user changes `as` into a `div`? The we will have `li` under `div`?

packages/react/src/components/Listbox/ListboxOption.tsx
```diff
@@ -0,0 +1,121 @@
...
+    useEffect(() => {
+      const element = listboxOptionRef.current;
+
+      setOptions((options) => {
+        const option = { element, value: optionValue };
+        // istanbul ignore next
+        if (!element) return options;
+
+        // Elements are frequently appended, so check to see if the newly rendered
+        // element follows the last element element first before any other checks
```
dequejosie, line 56: `last element first`?

packages/react/src/components/Listbox/ListboxContext.tsx
scurker: What do you mean?

packages/react/src/components/Listbox/ListboxGroup.tsx
scurker: Correct. We don't have control over if the author changes the element to something that might not be valid markup, we just provide the hook to do so.

packages/react/src/components/Listbox/ListboxOption.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+        // istanbul ignore next
+        // this should never happen, but just in case fall back to options
+        return options;
+      });
+
+      return () => {
+        setOptions((opts) => opts.filter((opt) => opt.element !== element));
+      };
+    }, [optionValue]);
```
dequejosie, line 86: Took me some time to understand this block, but I think I get it now. It handles appending `listOptions` at the end of the options or in the middle of it.

packages/react/src/components/Listbox/ListboxContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+  selected: T | null;
+  setOptions: React.Dispatch<React.SetStateAction<T[]>>;
+  onSelect: (option: T) => void;
+};
+
+type ListboxProvider<T extends ListboxOption> = {
+  children: React.ReactNode;
+} & ListboxContext<T>;
+
+/* istanbul ignore next */
```
dequejosie, line 22: > Why is typescript not happy here?
Sorry, I meant here.

packages/react/src/components/Listbox/Listbox.tsx
scurker: `onSelect` happens before the selection happens. `onSelectionChange` happens only after the the change has actually happened.

I suppose we could simplify this by only having a single event that responds when a selection changes, but only actually set the selection when uncontrolled:

```tsx
const handleSelectionChange = useCallback(
  (option: ListboxOption) => {
    onSelectionChange?.({ 
      target: option.element, 
      value: option.value as string 
      previousValue: selectedOption?.value
    });

    if (!isControlled) {
      setActiveOption(option);
      setSelectedOption(option);
    }
  },
  [isControlled, selectedOption]
);
```

Would that help make it more clear?

packages/react/src/components/Listbox/ListboxContext.tsx
scurker: I'm still not clear what you're referring to. I do not have any typescript errors:

vscode showing no errors on the referenced line

packages/react/src/components/Listbox/ListboxOption.tsx
scurker: Yeah, it's a bit tricky. We need to pass the rendered options back up the context but need to sort the list based on dom order here.

I went this route because passing down props via a set of unknown components is gnarly, and this abstracts that away completely.

packages/react/src/components/Listbox/Listbox.tsx
thuey: Personally, I like the suggested alternative. I do feel like it would make it clearer

packages/react/src/components/Listbox/ListboxOption.tsx
dequejosie: When I saw the context, I also compared it with passing down the props in my mind and I definitely think context is better.

packages/react/src/components/Listbox/ListboxContext.tsx
scurker: nyc was complaining about none of the options getting invoked but I don't think we really care about these lines for coverage since it's stubbing out an empty context so I ignored these lines. ¯\\\_(ツ)_/¯

packages/react/src/components/Listbox/Listbox.tsx
scurker: I just pushed that change.

packages/react/src/components/Listbox/ListboxOption.tsx
```diff
@@ -0,0 +1,121 @@
...
+          isElementPreceding(
+            options[options.length - 1].element,
+            option.element
+          )
+        ) {
+          return [...options, option];
+        }
+
+        for (const opt of options) {
+          if (isElementPreceding(element, opt.element)) {
```
thuey, line 68: It seems like this might not work properly when elements are added. I did some testing and `Node.DOCUMENT_POSITION_PRECEDING` is not the same as "immediately preceding", but this code seems to assume that's how that works.

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,226 @@
...
+          return;
+        }
+
+        const [up, down, home, end, enter, space] = keys;
+        const firstOption = enabledOptions[0];
+
+        if (!activeOption) {
+          setActiveOption(firstOption);
+          return;
+        }
```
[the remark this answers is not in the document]
scurker: Focusing the component should set the first option as active for listbox, but that's not always necessarily the case for composed components that would set the active descendent and need to manage state in some other way (such as combobox).

packages/react/src/components/Listbox/ListboxContext.tsx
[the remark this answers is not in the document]
scurker: Smaller files / separation of concerns?? If it's just a nit I'd like to leave this as is.

packages/react/src/components/Listbox/Listbox.tsx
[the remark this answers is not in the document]
thuey: Gotcha. I think the question is still applicable with regards to what the `space` or `enter` behavior should be, though--should it also select the first option?

packages/react/src/components/Listbox/ListboxOption.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const listboxOptionRef = useSharedRef<HTMLElement>(ref);
+    const [id] = propId ? [propId] : useId(1, 'listbox-option');
+    const isActive =
+      active !== null && active.element === listboxOptionRef.current;
+    const isSelected =
+      selected !== null && selected.element === listboxOptionRef.current;
+    const optionValue =
+      typeof value !== 'undefined'
+        ? value
+        : listboxOptionRef.current?.innerText;
```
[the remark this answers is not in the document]
scurker: My thinking here was to mirror what native options do by reflecting the label to the text value when the value is omitted. An empty value/label is also valid, so making it required adds verbosity that I don't think offers much. So I'm opposed to changing this.

I did a bunch of testing around this and the value always gets updated due to the `useEffect` callback that sets options. The value does initially start out as `undefined` but gets updated when the hook re-renders as a side effect. That would be an issue if we removed the use effect, but that also fundamentally changes how the component works.

packages/react/src/components/Listbox/ListboxOption.tsx
scurker: Correct, it's not immediately proceeding.

From testing this was always triggered with a single node at a time (even when mounting/changing multiple nodes). Even when nodes got sorted, the options gets removed from the list and re-added in order so I didn't encounter issues from testing.

packages/react/src/components/Listbox/Listbox.tsx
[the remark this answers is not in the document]
scurker: No, I don't think it should select the first option.

packages/react/src/components/Listbox/ListboxOption.tsx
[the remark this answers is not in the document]
thuey: That's fair. I was also concerned about multiple options having an empty value, but I suppose that could still happen even if you made `value` required without some kind of validation that the values were unique.

tldr; I think this is fine as is.

packages/react/src/components/Listbox/ListboxOption.tsx
thuey: Even when dynamically adding a new option to the end of the list after the others had already rendered?

thuey (approved):
Assuming https://github.com/dequelabs/cauldron/pull/1167#discussion_r1304729388 is okay, this LGTM. Nice job! This is pretty cool.

packages/react/src/components/Listbox/ListboxOption.tsx
scurker: Yes, that's what lines 57-65 are covering. This is with the assumption that the elements are already in the correct dom order, if the last element in the array precedes the newly inserted element it should get tacked on at the end.

PR #1168, opened 2023-08-16
docs: prevent propName from wrapping in component props

scurker:
Component props were sometimes wrapping in the middle of of the prop name. Updating that column/item to not wrap should help make these look a little bit cleaner and should not introduce a11y overflow issues as prop names are generally fairly short.

### DIFF
--- a/docs/components/ComponentLayout.css
+++ b/docs/components/ComponentLayout.css
@@ -125,6 +125,15 @@ ul.Component__list li {
   text-align: center;
 }
 
+table.Component__Props tr td:first-child,
+ul.Component__Props
+  li
+  dl
+  .DescriptionList__item:first-child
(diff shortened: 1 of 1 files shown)


docs/components/ComponentLayout.css
```diff
@@ -125,6 +125,15 @@ ul.Component__list li {
   text-align: center;
 }
 
+table.Component__Props tr td:first-child,
+ul.Component__Props
+  li
+  dl
+  .DescriptionList__item:first-child
+  .DescriptionList__details {
```
scurker, line 133: This is here to apply the same properties to both values equilaterally if we wanted to apply styles to the prop name specifically in the future.

thuey (approved):

PR #1169, opened 2023-08-16
docs: update IconButton to v2 docs

scurker:


thuey (approved):
I noticed that the "disabled" colors are the same for the error variant in the dark theme:
image

Do we need a ticket for this?

scurker:
> I noticed that the "disabled" colors are the same for the error variant in the dark theme: image
> 
> Do we need a ticket for this?

I'll bring this up with @dequelabs/cauldron-design, it's likely not intentional.

PR #1171, opened 2023-08-17
feat(Popover): add new component

### LINKED ISSUE #203: (Tooltip dialog) Popover alert component
## Popover (Light) 

popover (light mode)

## Popover (Dark)

popover (dark mode)

This can probably be an extension of the existing tooltip with some defaults for buttons.

Light: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6764581,light-popover
Dark: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6771667,dark-popover

orest-s:
closes #203

### DIFF
--- a/docs/pages/components/Popover.mdx
+++ b/docs/pages/components/Popover.mdx
@@ -1,0 +1,87 @@
+---
+title: Popover
+description: A popover component is a user interface element designed to provide contextual information or actions in a compact and non-intrusive manner. 
+source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/Popover/index.tsx
+---
+
+import { useRef, useState } from 'react'
+import { Popover, Button } from '@deque/cauldron-react'
+
+```js
+import { Popover, Button } from '@deque/cauldron-react'
+```
+
+## Examples
+
+Cauldron's tooltip relies on [Popper](https://popper.js.org/) to position tooltips dynamically. Popover can be triggered from any focusable element via a `target` attribute pointed to an HTMLElement or React ref object.
+
+### Alter Popover
+
+Alert Popovers are designed to display a popup with two buttons, the first to apply some action, the second to cancel, 
+also popup contains information text to show some message to the user.
+Canceling the action will only close the popup window.
+
+```jsx example
+function TextTooltipExamples() {
+  const [show, setShow] = useState(false);
+  const onTogglePopover = () => setShow(!show);
+  const buttonRef = useRef();
+
+  return (
+    <>
+      <Button 
+        onClick={onTogglePopover} 
+        variant="secondary" 
+        buttonRef={buttonRef}
+      >
+        Alert!
+      </Button>
+      <Popover
+        variant="alert"
+        target={buttonRef} 
+        placement="auto" 
+        show={show} 
+        onClose={onTogglePopover}
+        infoText="Are you sure you want to exit this wizard? All changes will be lost."
+      />
+    </>
+  )
+}
+```
+
+### Custom Popover
+
+For custom purposes you can use "custom" variant, which is default.
+
+```jsx example
+function BigTooltipExample() {
+  const [show, setShow] = useState(false);
+  const onTogglePopover = () => setShow(!show);
+  const buttonRef = useRef();
+
+  return (
+    <>
+      <Button 
+        onClick={onTogglePopover} 
+        variant="secondary" 
+        buttonRef={buttonRef}
+      >
+        Custom!
+      </Button>
+      <Popover
+        target={buttonRef} 
+        placement="auto" 
+        show={show} 
+        onClose={onTogglePopover}
+      >
+        <div>
+          <p>Popover content</p>
+          <Button>Custom button</Button>
+          <Button>Custom button</Button>
+          <p>Popover content</p>
+        </div>
+      </Popover>
+    </>
+  )
+}
+```
--- a/docs/pages/components/Popover.mdx
+++ b/docs/pages/components/Popover.mdx
@@ -15,14 +15,14 @@ import { Popover, Button } from '@deque/cauldron-react'
 
 Cauldron's tooltip relies on [Popper](https://popper.js.org/) to position tooltips dynamically. Popover can be triggered from any focusable element via a `target` attribute pointed to an HTMLElement or React ref object.
 
-### Alert Popover
+### Prompt Popover
 
-Alert Popovers are designed to display a popup with two buttons, the first to apply some action, the second to cancel, 
+Prompt Popovers are designed to display a popup with two buttons, the first to apply some action, the second to cancel, 
 also popup contains information text to show some message to the user.
 Canceling the action will only close the popup window.
 
 ```jsx example
-function TextTooltipExamples() {
+function PromptPopoverExample() {
   const [show, setShow] = useState(false);
   const onTogglePopover = () => setShow(!show);
   const buttonRef = useRef();
@@ -34,10 +34,10 @@ function TextTooltipExamples() {
         variant="secondary" 
         buttonRef={buttonRef}
       >
-        Alert!
+        Prompt!
       </Button>
       <Popover
-        variant="alert"
+        variant="prompt"
         target={buttonRef} 
         placement="auto" 
         show={show} 
@@ -54,7 +54,7 @@ function TextTooltipExamples() {
 For custom purposes you can use "custom" variant, which is default.
 
 ```jsx example
-function BigTooltipExample() {
+function CustomPopoverExample() {
   const [show, setShow] = useState(false);
   const onTogglePopover = () => setShow(!show);
   const buttonRef = useRef();
@@ -73,8 +73,10 @@ function BigTooltipExample() {
         placement="auto" 
         show={show} 
         onClose={onTogglePopover}
+        aria-labelledby={"popover-title"}
       >
         <div>
+          <h2 id="popover-title">Popover Title</h2>
           <p>Popover content</p>
           <Button>Custom button</Button>
           <Button>Custom button</Button>
@@ -116,13 +118,13 @@ function BigTooltipExample() {
       name: 'infoText',
       required: true,
       type: 'string',
-      description: 'Information text to show some warning or useful information in Alert variant of the component. Please note that required is only for "Alert" variant.',
+      description: 'Information text to show some warning or useful information in Prompt variant of the component. Please note that required is only for "Prompt" variant.',
     },
     {
       name: 'onApply',
       required: true,
       type: 'function',
-      description: 'Callback, which is used for Apply button in Alert variant of the component. Please note, that is only required for Alert variant',
+      description: 'Callback, which is used for Apply button in Prompt variant of the component. Please note, that is only required for Prompt variant',
     },
     {
       name: 'placement',
@@ -132,7 +134,7 @@ function BigTooltipExample() {
     },
     {
       name: 'variant',
-      type: ['custom', 'alert'],
+      type: ['custom', 'prompt'],
       defaultValue: 'custom',
       description: 'The style of Popover to display.'
     },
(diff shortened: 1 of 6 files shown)


packages/styles/popover.css
```diff
@@ -0,0 +1,117 @@
+:root {
+  --popover-background-color: var(--gray-10);
+  --popover-border-color: var(--gray-90);
+  --popover-text-color: var(--gray-90);
+  --popover-box-shadow: rgba(0, 0, 0, 0.35) 0px 0px 10px 0px;
+  --popover-button-height: 24px;
+  --popover-alert-max-width: 192px;
+  --popover-alert-min-height: 156px;
```
scurker, line 8: We generally avoid fixed width/height in px since that can cause issues if users are resizing only font sizes. These should be converted to `rem` instead.

packages/styles/popover.css
```diff
@@ -0,0 +1,117 @@
...
+  gap: var(--space-smallest);
+  align-items: center;
+  justify-content: center;
+  max-width: var(--popover-alert-max-width);
+  min-height: var(--popover-alert-min-height);
+}
+
+.Popover--alert > button {
+  width: 100%;
+  margin-left: 0 !important;
```
scurker, line 115: Why `!important` here?

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,208 @@
...
+        {closeButtonText}
+      </Button>
+      <Button className="ApplyBtn" onClick={onApply}>
+        {applyButtonText}
+      </Button>
+    </>
+  );
+};
+
+export default function Popover({
```
scurker, line 53: This needs to include ref via `forwardRef`.

scurker (changes requested):
Solid work here! Lots of comments but mostly just trying to sync up on some of the standards/practices that we have in cauldron. There's some additional comments from @bobbyomari 

> Some notes:
> - There is a 4px gap at the top and bottom - is there a way to remove that so the gap and make the elements inside the popover set the spacing (the popover doesn't define the spacing between elements)
> - I think the min-height should be removed or set to auto so that the height can just form to the contents inside?
> - Set the close button to be secondary button and flip the positions so Apply is first
> - spacing between the text and the button should be 8px
> - spacing between both buttons should be 8px

packages/styles/popover.css
```diff
@@ -0,0 +1,118 @@
+:root {
+  --popover-background-color: var(--gray-10);
+  --popover-border-color: var(--gray-90);
+  --popover-text-color: var(--gray-90);
+  --popover-box-shadow: rgba(0, 0, 0, 0.35) 0px 0px 10px 0px;
+  --popover-font-size: var(--text-size-smaller);
+  --popover-line-height: var(--text-size-normal);
+  --popover-padding: 0 var(--space-small);
+  --popover-alert-max-width: 12rem;
+  --popover-alert-min-height: auto;
```
scurker, line 10: This can be omitted. Elements have an implicit height/min-height of `auto` so I don't think we need to set a variable for that.

packages/react/__tests__/src/components/Popover/index.js
```diff
@@ -0,0 +1,295 @@
...
+// // eslint-disable-next-line react/prop-types, react/display-name
+// const WrapperWithParentRef = React.forwardRef(({}, ref) => {
+//   const buttonRef = useRef();
+//   const onClose = jest.fn();
+//   return (
+//     <Popover ref={parentRef} target={ref} show onClose={onClose}>
+//       Hello Word
+//     </Popover>
+//   );
+// });
```
scurker, line 59: Is this supposed to be commented out?

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,250 @@
...
+    useEffect(() => {
+      const attrText = targetElement?.getAttribute('aria-controls');
+      const hasPopupAttr = targetElement?.getAttribute('aria-haspopup');
+
+      if (!attrText?.includes(id) && show) {
+        targetElement?.setAttribute('aria-controls', id);
+      }
+
+      if (!hasPopupAttr) {
+        targetElement?.setAttribute('aria-haspopup', Boolean(true).toString());
```
scurker, line 186: Very minor nit.

```suggestion
        targetElement?.setAttribute('aria-haspopup', 'true');
```

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,250 @@
...
+      }
+    }, [targetElement, id, show]);
+
+    const handleClickOutside = () => {
+      if (show) {
+        onClose();
+      }
+    };
+
+    if (!show || !isBrowser()) return <></>;
```
scurker, line 196: Very minor nit.

```suggestion
    if (!show || !isBrowser()) return null;
```

scurker (changes requested):
This is really close, just a few nits and one last thought.

With `role="dialog"` we also need to ensure we're setting `aria-hidden` correctly on elements outside of the popover. For Dialog, we have a utility for that ([see example](https://github.com/dequelabs/cauldron/blob/bf5d0e59ebc27001261322d52df8976603ce6026/packages/react/src/components/Dialog/index.tsx#L90-L97)) that we could pull in here.

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+              'Popover',
+              `Popover--${placement}`,
+              className,
+              {
+                'Popover--hidden': !show,
+                'Popover--alert': variant === 'alert'
+              }
+            )}
+            ref={popoverRef}
+            role="dialog"
```
scurker, line 251: Sorry I missed this earlier, but we need to set an accessible name for popover since it's a dialog. It's a bit tricky based on the current structure since we don't have direct access to the popover content.

One way we could do this is by passing an id into `AlertPopoverContent` which would get labelled by the `infoText`. For the custom side of things, we can just show an example of how to do this via documentation.

scurker (changes requested):
Last couple of things! We should be ready to go once we get these few things addressed.

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+        const firstFocusableElement = popoverRef.current.querySelector(
+          focusableSelector
+        );
+
+        if (firstFocusableElement instanceof HTMLElement) {
+          firstFocusableElement.focus();
+        }
+      }
+
+      targetElement?.setAttribute('aria-expanded', Boolean(show).toString());
```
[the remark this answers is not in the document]
scurker: Whoops, I missed the show. Dismissing.

docs/pages/components/Popover.mdx
```diff
@@ -160,5 +160,15 @@ function BigTooltipExample() {
       description: 'Label for button, which is used to close popover in Alert variant of the component',
       defaultValue: 'Close'
     },
+    {
+      name: 'aria-label',
+      type: 'string',
+      description: 'A label for tabs is required. This means you must provide either an aria-label or aria-labelledby prop. Please note that for alert variant we use infoText to get labelled by'
```
scurker, line 166:
```suggestion
      description: 'A label for Popover is required for non-custom variants. This means you must provide either an aria-label or aria-labelledby prop.'
```

docs/pages/components/Popover.mdx
```diff
@@ -160,5 +160,15 @@ function BigTooltipExample() {
...
+    },
+    {
+      name: 'aria-labelledby',
+      type: 'string',
+      description: 'A label for tabs is required. This means you must provide either an aria-label or aria-labelledby prop. Please note that for alert variant we use infoText to get labelled by'
```
scurker, line 171:
```suggestion
      description: 'A label for Popover is required for non-custom variants. This means you must provide either an aria-label or aria-labelledby prop.'
```

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+  portal?: React.RefObject<HTMLElement> | HTMLElement;
+};
+
+type CustomProps = BaseProps & {
+  variant: 'custom';
+  applyButtonText?: string;
+  onApply?: () => void;
+  closeButtonText?: string;
+  infoText?: string;
+  chidlren: ReactNode;
```
scurker, line 34:
```suggestion
  children: ReactNode;
```

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+} & Cauldron.LabelProps;
+
+type AlertProps = BaseProps & {
+  variant: 'alert';
+  applyButtonText?: string;
+  onApply: () => void;
+  closeButtonText?: string;
+  infoText: string;
+  chidlren?: ReactNode;
```
scurker, line 43:
```suggestion
  children?: ReactNode;
```

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+type CustomProps = BaseProps & {
+  variant: 'custom';
+  applyButtonText?: string;
+  onApply?: () => void;
+  closeButtonText?: string;
+  infoText?: string;
+  chidlren: ReactNode;
+} & Cauldron.LabelProps;
+
+type AlertProps = BaseProps & {
```
scurker, line 37: `infoTextId` needs to be included in props here.

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+  placement?: Placement;
+  portal?: React.RefObject<HTMLElement> | HTMLElement;
+};
+
+type CustomProps = BaseProps & {
+  variant: 'custom';
+  applyButtonText?: string;
+  onApply?: () => void;
+  closeButtonText?: string;
+  infoText?: string;
```
scurker, line 33:
```suggestion
  infoText?: ReactNode;
```

docs/pages/components/Popover.mdx
```diff
@@ -0,0 +1,174 @@
...
+        buttonRef={buttonRef}
+      >
+        Custom!
+      </Button>
+      <Popover
+        target={buttonRef} 
+        placement="auto" 
+        show={show} 
+        onClose={onTogglePopover}
+      >
```
scurker, line 76: This example needs to show how to set the accessible name for the popover either with `aria-label` or `aria-labelledby`.

packages/styles/popover.css
```diff
@@ -0,0 +1,116 @@
+:root {
+  --popover-background-color: var(--gray-10);
+  --popover-border-color: var(--gray-90);
+  --popover-text-color: var(--gray-90);
+  --popover-box-shadow: rgba(0, 0, 0, 0.35) 0px 0px 10px 0px;
```
scurker, line 5: Very minor nit: 0 values can be unitless in css.

```suggestion
  --popover-box-shadow: rgba(0, 0, 0, 0.35) 0 0 10px 0;
```

packages/react/src/index.ts
```diff
@@ -120,6 +120,7 @@ export {
   ColumnList
 } from './components/TwoColumnPanel';
 export { default as Notice } from './components/Notice';
+export { default as Popover, PopoverProps } from './components/Popover';
```
scurker, line 123: We don't typically export props here.

```suggestion
export { default as Popover } from './components/Popover';
```

packages/react/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,287 @@
...
+  chidlren: ReactNode;
+} & Cauldron.LabelProps;
+
+type AlertProps = BaseProps & {
+  variant: 'alert';
```
scurker, line 38: I'm thinking we should call this something else, since "alert" has a different connotation with regards to accessibility. I wonder if something like `prompt` might be better? What do you think?

packages/react/src/components/Popover/index.tsx
orest-s: Do we allow to pass `infoTextId` from outside of the component?
We generate id by ourself in main component, 
that's why i added type on a component level instead of the type?

scurker: Ah, I missed that it was set below.

Since we don't export the props including it here would be a best practice since an author would still see that prop set via typescript. We can leave it as is for now.

scurker (commented):
This one is almost good to go! Just need to fix the merge conflict and looks like the layout of the prompt popover doesn't match the design now:

Apply/close buttons are showing on the same line as the info text

scurker (dismissed):

scurker (approved):

scurker:
@orest-s If everything is good here, feel free to squash and merge when you're ready.

PR #1172, opened 2023-08-18
chore: upgrade babel/typescript to support latest features

scurker:
General cleanup of typescript/babel dependencies. Removed a few unused babel plugins, but there were some issues with `import type...` on #1167 that I was trying to resolve here separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "release": "./scripts/release.sh"
   },
   "prettier": {
-    "singleQuote": true
+    "singleQuote": true,
+    "trailingComma": "none"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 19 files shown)


package.json
```diff
@@ -128,7 +127,7 @@
     "style-loader": "^0.19.0",
     "terser-webpack-plugin": "^5.3.6",
     "ts-node": "^10.9.1",
-    "typescript": "^3.7.5",
+    "typescript": "~5.0.0",
```
scurker, line 130: prettier yells at you if it's anything greater than 5.1.0 🤷

packages/react/.babelrc
```diff
@@ -11,7 +11,6 @@
         }
       ],
       "transform-export-extensions",
-      "@babel/plugin-proposal-class-properties",
```
scurker, line 14: now included as part of preset-env

packages/react/package.json
```diff
@@ -34,18 +34,15 @@
     "tslib": "^2.4.0"
   },
   "devDependencies": {
-    "@babel/cli": "^7.8.4",
```
scurker, line 37: unused

packages/react/src/components/IconButton/index.tsx
```diff
@@ -115,7 +115,6 @@ IconButton.propTypes = {
   as: PropTypes.elementType,
   // @ts-expect-error
   icon: PropTypes.string.isRequired,
-  // @ts-expect-error
```
scurker, line 118: apparently none of these error anymore

package.json
```diff
@@ -23,7 +23,8 @@
     "release": "./scripts/release.sh"
   },
   "prettier": {
-    "singleQuote": true
+    "singleQuote": true,
+    "trailingComma": "none"
```
scurker, line 27: I don't really care that much, but this keeps the existing behavior that was present before the upgrade to keep things consistent

package.json
```diff
@@ -43,14 +44,12 @@
   },
   "devDependencies": {
     "@axe-core/puppeteer": "^4.3.1",
-    "@babel/core": "^7.4.3",
-    "@babel/plugin-proposal-class-properties": "^7.4.0",
-    "@babel/plugin-proposal-export-default-from": "^7.2.0",
-    "@babel/preset-env": "^7.8.4",
-    "@babel/preset-react": "^7.0.0",
-    "@babel/preset-stage-0": "^7.0.0",
```
scurker, line 51: not really needed anymore

packages/react/src/components/IconButton/index.tsx
dequejosie: The others still error?

dequejosie:
Looks good to me!

dequejosie (approved):

PR #1177, opened 2023-08-22
docs: update Icon to v2 docs

scurker:


docs/components/ComponentLayout.css
```diff
@@ -75,6 +75,10 @@ ul.Component__list li {
   overflow: hidden;
 }
 
+.Component .Offscreen {
+  top: 0;
+}
```
scurker, line 80: I noticed that Offscreen was creating double scrollbars on this page as it was positioned outside of the scrollable area creating some weird jank.

thuey (approved):
Search!!! 🎉

PR #1180, opened 2023-08-23
feat(react): add Combobox component

### LINKED ISSUE #1088: Combobox Component (Single Select)

scurker:
closes #1088 

Reviewers should make themselves familiar with Comboboxes, the aria authoring practices guide is particularly useful as a primer:
https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

### DIFF
--- a/docs/pages/components/Combobox.mdx
+++ b/docs/pages/components/Combobox.mdx
@@ -1,6 +1,6 @@
 ---
 title: Combobox
-description: An input component to display a list of suggested options matching the current text input.
+description: An input component to display a list of suggested options.
 source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/Combobox/Combobox.tsx
 ---
 
--- a/docs/pages/components/Combobox.mdx
+++ b/docs/pages/components/Combobox.mdx
@@ -1,0 +1,132 @@
+---
+title: Combobox
+description: An input component to display a list of suggested options matching the current text input.
+source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/Combobox/Combobox.tsx
+---
+
+import { Combobox, ComboboxItem, ComboboxGroup } from '@deque/cauldron-react'
+
+```js
+import { 
+  Combobox,
+  ComboboxItem,
+  ComboboxGroup
+} from '@deque/cauldron-react'
+```
+
+## Examples
+
+### With Items
+
+Options can be rendered from an array of items with a label and an optional value by setting the `items` prop:
+
+```jsx example
+function ComboboxWithItemPropExample() {
+  const items = [
+    { label: 'Red', value: 'Red' },
+    { label: 'Orange', value: 'Orange' },
+    { label: 'Yellow', value: 'Yellow' },
+    { label: 'Green', value: 'Green' },
+    { label: 'Blue', value: 'Blue' }
+  ]
+  return (
+    <Combobox label="Favorite Color" items={items} />
+  )
+}
+```
+
+### With Children
+
+For more advanced scenarios where the rendering of the item label needs to be customized, `ComboboxItem` can be passed in to `Combobox` as `children`.
+
+```jsx example
+<Combobox label="Fruit" defaultValue="Peach">
+  <ComboboxItem value="Apple">🍎 Apple</ComboboxItem>
+  <ComboboxItem value="Banana">🍌 Banana</ComboboxItem>
+  <ComboboxItem value="Cucumber">🥒 Cucumber</ComboboxItem>
+  <ComboboxItem value="Orange">🍊 Orange</ComboboxItem>
+  <ComboboxItem value="Peach">🍑 Peach</ComboboxItem>
+  <ComboboxItem value="Pear">🍐 Pear</ComboboxItem>
+</Combobox>
+```
+
+### With Descriptions
+
+Combobox items can optionally include a description, which can be set via a property on the array of items, or as a property on the `ComboboxItem` component.
+
+```jsx example
+<Combobox label="Select a Drink">
+  <ComboboxItem 
+    description="A monster of a barleywine, with a footprint of caramel and toffee."
+    value="Bigfoot's Barleywine"
+  >
+    Bigfoot's Barleywine
+  </ComboboxItem>
+  <ComboboxItem 
+    description="A hoppy explosion of citrus and pine, a true hopocalypse."
+    value="Hopocalypse IPA"
+  >
+    Hopocalypse IPA
+  </ComboboxItem>
+  <ComboboxItem 
+    description="A pot of liquid gold at the end of the rainbow."
+    value="Leprechaun's Gold Lager"
+  >
+    Leprechaun's Gold Lager
+  </ComboboxItem>
+
+  <ComboboxItem 
+    description="Delicate, effervescent, with a hint of mythical magic in every sip."
+    value="Unicorn Tears Saison"
+  >
+    Unicorn Tears Saison
+  </ComboboxItem>
+  <ComboboxItem 
+    description="Revive your taste buds with this undead burst of citrusy hops."
+    value="Zombie Zest Pale Ale"
+  >
+    Zombie Zest Pale Ale
+  </ComboboxItem>
+</Combobox>
+```
+
+### Required
+
+```jsx example
+<Combobox label="Fruit" defaultValue="Peach" required>
+  <ComboboxItem description="Apple is a fruit" value="Apple">🍎 Apple</ComboboxItem>
+  <ComboboxItem value="Banana">🍌 Banana</ComboboxItem>
+  <ComboboxItem value="Cucumber">🥒 Cucumber</ComboboxItem>
+  <ComboboxItem value="Orange">🍊 Orange</ComboboxItem>
+  <ComboboxItem value="Peach">🍑 Peach</ComboboxItem>
+  <ComboboxItem value="Pear">🍐 Pear</ComboboxItem>
+</Combobox>
+```
+
+## Grouping
+
+```jsx example
+<Combobox label="Countries">
+  <ComboboxGroup label="North America">
+    <ComboboxItem>Canada</ComboboxItem>
+    <ComboboxItem>Mexico</ComboboxItem>
+    <ComboboxItem>United States</ComboboxItem>
+  </ComboboxGroup>
+  <ComboboxGroup label="South America">
+    <ComboboxItem>Argentina</ComboboxItem>
+    <ComboboxItem>Bolivia</ComboboxItem>
+    <ComboboxItem>Brazil</ComboboxItem>
+    <ComboboxItem>Chile</ComboboxItem>
+    <ComboboxItem>Columbia</ComboboxItem>
+    <ComboboxItem>Ecuador</ComboboxItem>
+    <ComboboxItem>Falkland Islands</ComboboxItem>
+    <ComboboxItem>French Guiana</ComboboxItem>
+    <ComboboxItem>Guyana</ComboboxItem>
+    <ComboboxItem>Paraguay</ComboboxItem>
+    <ComboboxItem>Peru</ComboboxItem>
+    <ComboboxItem>Suriname</ComboboxItem>
+    <ComboboxItem>Uruguay</ComboboxItem>
+    <ComboboxItem>Venezuela</ComboboxItem>
+  </ComboboxGroup>
+</Combobox>
+```
--- a/docs/pages/components/Combobox.mdx
+++ b/docs/pages/components/Combobox.mdx
@@ -233,7 +233,7 @@ When autocomplete is set to "none" the listbox will provide the full list of opt
 
 ### Automatic
 
-When autocomplete is set to "automatic" the listbox will provide a filtered list of options with the first value selected by default. The selected value will become the value of the input when the listbox closes.
+When autocomplete is set to "automatic" the listbox will provide a filtered list of options with the first value active by default. The active value will become the value of the input when the listbox closes.
 
 ```jsx example
 <FieldWrap>
--- a/docs/pages/components/Combobox.mdx
+++ b/docs/pages/components/Combobox.mdx
@@ -178,7 +178,11 @@ To set the initial value for uncontrolled Comboboxes, use `defaultValue`.
 
 ### Controlled
 
-Comboboxes can also be controlled, by setting the `value` prop and updating the value from `onSelectionChange`.
+Comboboxes can also be controlled, by setting the `value` prop and updating the value from event callbacks.
+
+<Note>
+  `onChange` and `onSelectionChange` have different behaviors and special consideration needs need to be made when `Combobox` is controlled. `onChange` gets called when the value of the input field changes from user input or selections while `onSelectionChange` is only called when a selection changes from the listbox.
+</Note>
 
 ```jsx example
 function ComboboxControlledExample() {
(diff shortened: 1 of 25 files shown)


packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,216 @@
...
+    const isControlled = typeof value !== 'undefined';
+
+    useLayoutEffect(() => {
+      const listboxValue = isControlled ? value : defaultValue;
+      const selectedOption = options.find((option) =>
+        optionMatchesValue(option, listboxValue)
+      );
+      setSelectedOption(selectedOption || null);
+      setActiveOption(selectedOption || null);
+    }, [options, value]);
```
thuey, line 83: It seems like a bug would happen here if:
1. The component was uncontrolled
2. The options changed for some reason

We would reset the selected value to the default value.

scurker: This PR isn't ready, you're probably looking for #1167

packages/react/src/components/Combobox/ComboboxGroup.tsx
```diff
@@ -0,0 +1,52 @@
...
+        setShowGroup(true);
+        return;
+      }
+
+      const elements = Array.from(matchingOptions.keys());
+      const groupHasChildren = !!elements.find((element) =>
+        comboboxGroupRef.current?.contains(element)
+      );
+      setShowGroup(groupHasChildren);
+    }, [inputValue, autocomplete, matchingOptions]);
```
thuey, line 30: nit: It's unclear to me why this is a `useState + useLayoutEffect`. It seems like it could just be a `useMemo`

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,441 @@
...
+      ));
+
+    useEffect(() => {
+      if (!isAutoComplete) {
+        return;
+      }
+
+      if (!open && selectedValue && value !== selectedValue) {
+        setValue(selectedValue);
+      }
```
thuey, line 129: Playing around with it and looking at this, if I understand this correctly, the input's value is set to the value of the combobox option value. However, if the combobox option `value` is something like "option-1" and the option's `children` is "Pizza", it seems like we should display "Pizza"--not "option-1"

docs/pages/components/Combobox.mdx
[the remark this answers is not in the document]
scurker: Ah, I see the issue. It should close on blur but the input isn't gaining focus because you're clicking on the trigger. That's a simple fix.

packages/react/src/components/Combobox/Combobox.tsx
scurker: My original thinking is that `value` is the value you want as text in the resulting Combobox input and is the value that gets submitted to the server, while `label` is the text that gets displayed in the listbox. Text matching happens based on the label, not the value, and from looking at other comboboxes this is generally how they work with a few exceptions.

Although you bring up a good point that for non-text values that are intended to be internal we might want something different here. For an example, [Adobe's combobox has a `formValue` property](https://react-spectrum.adobe.com/react-spectrum/ComboBox.html#:~:text=space%20is%20limited.-,formValue,-%27text%27%20%7C%C2%A0) to specify the value that gets submitted as part of a form. We might could do something similar here, but I wonder if it's critical enough to be a part of this PR or we can tackle that separately.

I would really only consider this blocking if the change would be breaking from the existing pattern. Let me mull on this for a bit.

scurker: Proposal/RFC: https://github.com/dequelabs/cauldron/issues/1184

qazwsxedcrfvtgb1111:
Is there a way to change the default icon on a selected option? Not sure about what use cases we have in mind for this but as a general use ui library I'd definitely want that.

docs/pages/components/Combobox.mdx
```diff
@@ -0,0 +1,347 @@
...
+    { label: 'Green', value: 'Green' },
+    { label: 'Blue', value: 'Blue' }
+  ]
+  return (
+    <FieldWrap>
+      <Combobox label="Favorite Color" options={options} />
+    </FieldWrap>
+  )
+}
+```
```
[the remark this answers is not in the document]
qazwsxedcrfvtgb1111: Ah, nevermind, saw the discussion about this on the ticket

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,458 @@
...
+
+      if (open && autocomplete === 'automatic' && !selectedValue) {
+        // Fire an Home keydown event on listbox to ensure the first item is selected
+        listboxRef.current?.dispatchEvent(
+          new KeyboardEvent('keydown', {
+            key: 'Home',
+            bubbles: true,
+            cancelable: true
+          })
+        );
```
qazwsxedcrfvtgb1111, line 143: nit: I'd move this to a function since it's used multiple times

scurker:
> Is there a way to change the default icon on a selected option? Not sure about what use cases we have in mind for this but as a general use ui library I'd definitely want that.

That's more of a @dequelabs/cauldron-design question than something I can answer here. They have mentioned wanting to have more alignment and consistency between teams, so we might not need this white labeled out of the box until we identify the need to do so.

docs/pages/components/Combobox.mdx
```diff
@@ -0,0 +1,347 @@
...
+    name: 'autocomplete',
+    type: ['none', 'manual', 'automatic'],
+    defaultValue: 'manual',
+    description: 'Sets the autocompletion type of the Combobox listbox.'
+  },
+  {
+    name: 'onSelectionChange',
+    type: 'function',
+    description: 'Callback function that gets invoked when the selected value is changed.'
+  },
```
thuey, line 284: Should we document `onChange` here, too?

docs/pages/components/Combobox.mdx
```diff
@@ -0,0 +1,347 @@
...
+  </Combobox>
+</FieldWrap>
+```
+
+### Controlled
+
+Comboboxes can also be controlled, by setting the `value` prop and updating the value from event callbacks.
+
+<Note>
+  `onChange` and `onSelectionChange` have different behaviors and special consideration needs need to be made when `Combobox` is controlled. `onChange` gets called when the value of the input field changes from user input or selections while `onSelectionChange` is only called when a selection changes from the listbox.
```
dequejosie, line 184: The wording of the first sentence is a bit confusing to me. Did you mean "When being used in a uncontrolled `Combobox`,`onChange` and `onSelectionChange` have different behaviors and special consideration needs"

docs/pages/components/Combobox.mdx
dequejosie: ☝️

packages/styles/combobox.css
```diff
@@ -0,0 +1,246 @@
...
+  --combobox-input-error-focus-border-color: var(
+    --field-border-color-error-hover
+  );
+  --combobox-input-error-focus-border-glow-color: var(
+    --field-border-color-focus-glow
+  );
+  --combobox-input-font-size: var(--field-font-size);
+  --combobox-input-font-color: var(--field-content-color);
+  --combobox-listbox-border-color: var(--gray-40);
+  --combobox-listbox-background-color: #fff;
```
dequejosie, line 19: I think we have the variable of `--white`.
```suggestion
  --combobox-listbox-background-color: var(--white);
```

packages/styles/combobox.css
```diff
@@ -0,0 +1,246 @@
...
+  --combobox-option-hover-background-color: var(--gray-20);
+  --combobox-option-description-font-color: var(--gray-60);
+  --combobox-option-description-font-size: var(--text-size-smaller);
+  --combobox-option-min-height: 2.625rem;
+  --combobox-group-label-background-color: var(--gray-20);
+  --combobox-group-label-font-size: var(--text-size-smaller);
+  --combobox-option-selected-icon-color: var(--accent-success-dark);
+}
+
+.cauldron--theme-dark {
```
dequejosie, line 30: Haven't been in `Cauldron` land for a while, why are we assigning the same values for many of the variables under `theme-dark`?

packages/styles/combobox.css
```diff
@@ -0,0 +1,246 @@
...
+.ComboboxOption:is(:hover, [role='option'][aria-selected='true']:hover) {
+  background-color: var(--combobox-option-hover-background-color);
+}
+
+.ComboboxOption .Icon {
+  color: var(--combobox-option-selected-icon-color);
+}
+
+.ComboboxOption .Icon svg {
+  padding: 4px;
```
dequejosie, line 212:
```suggestion
  padding: var(--space-half);
```

packages/styles/forms.css
```diff
@@ -26,6 +26,7 @@
   --field-error-text-color: var(--error);
   --field-error-border-color: var(--error);
   --field-listbox-selected-background-color: var(--accent-secondary-active);
+  --field-font-size: var(--text-size-small);
   --input-min-width: 250px;
```
dequejosie, line 30: not a part of your changes, but shall we use `rem` for `--input-min-width`?

packages/styles/combobox.css
scurker: `#fff` is shorter than `--white`. We can't deprecate it because its in use, but there's little to no advantage to defining white as a var because that definition will never change.

packages/styles/combobox.css
scurker: Good question!

This one trips people up because the global var gets set based on its inherited scope. Take `--field-border-color` for example. If we do not set the same value again for `.cauldron-theme-dark`, the inherited value will come from `:root` and not from `.cauldron-theme-dark`, so we need to redefine it again for the value to cascade properly.

packages/styles/forms.css
scurker: Ideally yes, but that's unrelated to this PR here.

packages/styles/combobox.css
scurker: This should _not_ use the spacing vars here since this padding is a fixed padding relative to the icon size and not the design system spacing.

docs/pages/components/Combobox.mdx
scurker: Looks like I had an extra word there in there, I removed it - does that help?

thuey (approved):

packages/styles/combobox.css
dequejosie: Is this something like best practice? Whether redefine it in `.cauldron-theme-dark` or not, the value will be the same, right?

packages/react/src/utils/useIntersectionRef.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  useEffect(() => {
+    // istanbul ignore else
+    if (
+      'IntersectionObserver' in globalThis &&
+      typeof IntersectionObserver === 'function'
+    ) {
+      if (
+        !element &&
+        typeof element === 'object' &&
```
dequejosie, line 31: Trying to understand the logic here. How could `!element` be together with `typeof element === 'object'`?

packages/styles/combobox.css
scurker: Best practice would be for it to not be defined twice, but based on how we're inheriting the values here it has to be defined twice. Let me see if I can explain better:

Take the following markup:

```html
<div class="theme">
  <div class="component">Here is some text</div>
</div>
```

...and css:

```css
:root {
  --theme-color: red;
}

.theme {
  --theme-color: green;
}

:root {
  --inherited-theme-color: var(--theme-color);
}

.component {
  color: var(--inherited-theme-color);
}
```

You might think that the color of the text should be green, but it's actually red in this instance.

When the value of `--inherited-theme-color` is getting set, it is using the value of `:root` based on the scope of the current css rule. By the time we reach `.component`, it doesn't know that the variable has been redefined under a different scope because it's using the value from `--inherited-theme-color` which is currently `red`. If we add:

```css
.theme {
  --inherited-theme-color: var(--theme-color);
}
```

...the `--inherited-theme-color` gets set based on the current value for that scope, and correctly displays the value `green`.

packages/react/src/utils/useIntersectionRef.ts
scurker: `null` values, although it could be simplified as the 3rd statement is redundant and we cover non HTMLElement values below.

dequejosie (approved):

PR #1182, opened 2023-08-25
fix(styles): fix color contrast issue for required text on TextField with error

scurker:
Found a color contrast issue in dark mode when TextField was an an error state and required. Matching the color with what's currently in our design system.

## Before

required text above text field with poor color contrast

## After

required text above text field with correct color contrast

### DIFF
--- a/docs/pages/components/TextField.mdx
+++ b/docs/pages/components/TextField.mdx
@@ -53,6 +53,7 @@ import { TextField } from '@deque/cauldron-react'
   <TextField
     label="Error"
     error="This field has an error!"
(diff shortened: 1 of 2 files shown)


tbusillo (approved):

scurker:
@tbusillo why are you reviewing PRs at 1AM???? 🙃

tbusillo:
> @tbusillo why are you reviewing PRs at 1AM???? 🙃 

Dang it, weren't supposed to see this until Monday and hoped you'd miss the time stamp 🤣

PR #1191, opened 2023-09-08
feat(TagButton): add new tag button component

### LINKED ISSUE #1161: Create TagButton component for light and dark theme
This is an extension of the existing `<Button variant="tag">` that requires specific properties to help indicate its interactiveness.

UXPin Component (light): https://preview.uxpin.com/design-system/c933c74928e2195cc540/6769360,light-tag-actionable

Image

UXPin Component (dark): https://preview.uxpin.com/design-system/c933c74928e2195cc540/6771812,dark-tag-actionable

Image

### Configurable: 

- Label
- Value
- Icon
- Action (link to external URL, open things, remove things, etc.) 

The disabled state is not implicitly supported, the author should instead use `Tag` in absence of the disabled state.

orest-s:
closes #1161

### DIFF
--- a/docs/pages/components/TagButton.mdx
+++ b/docs/pages/components/TagButton.mdx
@@ -1,62 +1,73 @@
 ---
 title: TagButton
-description: Component that is an extension of the existing <Button variant="tag"> that requires those properties listed in the Props Section
+description: An interactive element that performs a programmable action when clicked by the user, displayed as a label and value with symbol image.
 source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/TagButton/index.tsx
 ---
 
-import { TagButton } from '@deque/cauldron-react'
+import { TagButton } from '@deque/cauldron-react';
 
 ```js
-import { TagButton } from '@deque/cauldron-react'
+import { TagButton } from '@deque/cauldron-react';
 ```
 
 ## Example
 
 ```jsx example
-<div style={{ display: "flex", gap: "8px" }}>
-  <TagButton label={"Label"} value={"value"} action={() => console.log("test")} icon={"pencil"} />
-
-  <TagButton label={"Label"} value={"value"} action={() => console.log("test")} icon={"pencil"} disabled={true} />
-</div>
+<TagButton
+  label={'Label'}
+  value={'value'}
+  onClick={() => console.log('test')}
+  icon={'pencil'}
+/>
 ```
-
 
 ## Props
 
 ### TagButton
 
-<ComponentProps 
-  className={true} 
+<ComponentProps
+  className={true}
   props={[
     {
       name: 'label',
       required: true,
-      type: ['string', 'number', 'ReactElement', 'ReactFragment', 'ReactPortal'],
+      type: [
+        'string',
+        'number',
+        'ReactElement',
+        'ReactFragment',
+        'ReactPortal'
+      ],
       description: 'Label for the tag button.'
     },
     {
       name: 'value',
       required: true,
-      type: ['string', 'number', 'ReactElement', 'ReactFragment', 'ReactPortal'],
+      type: [
+        'string',
+        'number',
+        'ReactElement',
+        'ReactFragment',
+        'ReactPortal'
+      ],
       description: 'Value for the tag button.'
     },
     {
-      name: 'action',
+      name: 'onClick',
       required: true,
       type: '(e: React.MouseEvent<HTMLButtonElement>) => void',
-      description: 'Label for the tag button.'
+      description: 'a function that gets executed when a user clicks on an element.'
     },
     {
       name: 'icon',
       type: 'string',
       required: true,
       description: 'Icon to use for the TagButton'
-    },
-    {
-      name: 'disabled',
-      required: false,
-      type: 'boolean',
-      description: 'Disabled state of the element'
-    },
+    }
   ]}
 />
+
+## Related Components
+
+- [Button](./Button)
+- [Tag](./Tag)
--- a/docs/pages/components/TagButton.mdx
+++ b/docs/pages/components/TagButton.mdx
@@ -1,0 +1,62 @@
+---
+title: TagButton
+description: Component that is an extension of the existing <Button variant="tag"> that requires those properties listed in the Props Section
+source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/TagButton/index.tsx
+---
+
+import { TagButton } from '@deque/cauldron-react'
+
+```js
+import { TagButton } from '@deque/cauldron-react'
+```
+
+## Example
+
+```jsx example
+<div style={{ display: "flex", gap: "8px" }}>
+  <TagButton label={"Label"} value={"value"} action={() => console.log("test")} icon={"pencil"} />
+
+  <TagButton label={"Label"} value={"value"} action={() => console.log("test")} icon={"pencil"} disabled={true} />
+</div>
+```
+
+
+## Props
+
+### TagButton
+
+<ComponentProps 
+  className={true} 
+  props={[
+    {
+      name: 'label',
+      required: true,
+      type: ['string', 'number', 'ReactElement', 'ReactFragment', 'ReactPortal'],
+      description: 'Label for the tag button.'
+    },
+    {
+      name: 'value',
+      required: true,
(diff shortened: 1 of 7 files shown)


packages/react/src/components/TagButton/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+interface TagButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
+  label: ContentNode;
+  value: ContentNode;
+  icon: IconType;
+  action: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  disabled?: boolean;
+}
+
+const TagButton = ({
```
scurker, line 16: This needs to follow our [component standards](https://cauldron.dequelabs.com/component-guidelines#component-standards) and is missing: 

- `React.forwardRef`

packages/react/src/components/TagButton/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+      variant="tag"
+      className={classNames('TagButton', className)}
+      onClick={action}
+      disabled={disabled}
+    >
+      <TagLabel>{label}</TagLabel>
+      {value}
+      <Icon className="TagButton__icon" type={icon} />
+    </Button>
+  );
```
scurker, line 53: I don't think we should be swapping out the component types here as that can cause issues since it should semantically still be a button in both instances. 

It would be better to always keep it as a button (abbreviated example below):

```jsx
return (
  <Button
    variant="tag"
    disabled={disabled}
    {...props}
  >
    ...
  </Button>
)
```

If the content author also sets `aria-disabled`, we would need to prevent the onclick/action callback from occurring.

packages/styles/tag-button.css
```diff
@@ -0,0 +1,8 @@
+.TagButton {
+  vertical-align: middle;
+}
```
scurker, line 3: I'm not sure this is needed.

packages/styles/tag-button.css
```diff
@@ -0,0 +1,8 @@
...
+
+.TagButton__icon svg {
+  width: 15px;
+  height: 15px;
+}
```
scurker, line 8: I'm assuming these sizes came from @dequelabs/cauldron-design? Unfortunately I don't see it listed on the original issue/uxpin but if these are the right sizes, they need to be set via css vars:

```css
:root {
  --tag-button-icon-size: ...
}
```

packages/react/src/components/TagButton/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { ContentNode } from '../../types';
+import Tag, { TagLabel } from '../Tag';
+import classNames from 'classnames';
+import Button from '../Button';
+
+interface TagButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
+  label: ContentNode;
+  value: ContentNode;
+  icon: IconType;
+  action: (e: React.MouseEvent<HTMLButtonElement>) => void;
```
scurker, line 12: What's the expected behavior if someone sets props on the component like this?

```jsx
<TagButton action={...} onClick={...} />
```

It looks like we allow both via the type, but in the implementation below we clobber `onClick` with `action`. If the desired behavior is the action is only set via that prop, we likely need to omit `onClick`:

```tsx
interface TagButtonProps extends Omit<React.HTMLAttributes<HTMLButtonElement>, 'onClick'>
```

docs/pages/components/TagButton.mdx
```diff
@@ -0,0 +1,62 @@
...
+      name: 'value',
+      required: true,
+      type: ['string', 'number', 'ReactElement', 'ReactFragment', 'ReactPortal'],
+      description: 'Value for the tag button.'
+    },
+    {
+      name: 'action',
+      required: true,
+      type: '(e: React.MouseEvent<HTMLButtonElement>) => void',
+      description: 'Label for the tag button.'
```
scurker, line 47: This should be describing the action.

docs/pages/components/TagButton.mdx
```diff
@@ -0,0 +1,62 @@
...
+      description: 'Icon to use for the TagButton'
+    },
+    {
+      name: 'disabled',
+      required: false,
+      type: 'boolean',
+      description: 'Disabled state of the element'
+    },
+  ]}
+/>
```
scurker, line 62: We should include "related components" here, and also cross-link with those related components. I think `Button` and `Tag` would suffice.

scurker (changes requested):

docs/pages/components/TagButton.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+```js
+import { TagButton } from '@deque/cauldron-react'
+```
+
+## Example
+
+```jsx example
+<div style={{ display: "flex", gap: "8px" }}>
+  <TagButton label={"Label"} value={"value"} action={() => console.log("test")} icon={"pencil"} />
```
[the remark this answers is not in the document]
orest-s: Actually, it seems that actions is the same that onClick, so probably we just need to rely on onClick props, or if we want to have "action prop", we need to have some additional checking there?

packages/react/src/components/TagButton/index.tsx
orest-s: I was thinking, that statement in the task

> The disabled state is not implicitly supported, the author should instead use Tag in absence of the disabled state.

means that "author" is author of the component, so do we need to remove using of Tag component here?

packages/styles/tag-button.css
orest-s: I have added this, because of <Tag> component. Button has this style, when we add Tag and Button in a row, it won't be on the one level height, but since we need to remove Tag component (as i understood from above messages) it won't be conflicted

packages/styles/tag-button.css
orest-s: I'm wrong, sorry i get it from Monitor, but now i see that real design even on the monitor has 10px..

docs/pages/components/TagButton.mdx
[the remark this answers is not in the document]
scurker: If we want to enforce `onClick` we would need to make it required prop and not optional. I think having `onClick` in place of `action` is acceptable.

packages/react/src/components/TagButton/index.tsx
scurker: Ah yes, sorry I forgot that requirement.

That's to mean the "author" as in content author and not component author. We don't need to support disabled here because an tag should either be a `TagButton` state or it should not be, and we already have a component that  supports the latter. So I think we should nix `disabled` entirely.

docs/pages/components/TagButton.mdx
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## Example
+
+```jsx example
+<TagButton
+  label={'Label'}
+  value={'value'}
+  onClick={() => console.log('test')}
+  icon={'pencil'}
```
scurker, line 20: Nit

```suggestion
  label="Label"
  value="value"
  onClick={() => console.log('test')}
  icon="pencil"
```

packages/styles/tag-button.css
```diff
@@ -0,0 +1,12 @@
...
+  vertical-align: middle;
+}
+
+.TagButton__icon {
+  margin-left: var(--space-half);
+}
+
+.TagButton__icon svg {
+  width: var(--tag-button-icon-size);
+  height: var(--tag-button-icon-size);
```
scurker, line 11: This variable should be local to this file and not global (as in variables.css).

Additionally, this doesn't appear to be the right size. The icon looks super tiny in the preview:

scurker (changes requested):
Just a few more small things

packages/styles/variables.css
```diff
@@ -109,6 +109,7 @@
 
   /* icon sizes */
   --icon-size: 24px;
+  --tag-button-icon-size: 15px;
```
scurker, line 112: To clarify, this is the right size but needs to be moved into the `tag-button.css` file, only global vars should exist in `variables.css`:

```css
:root {
  --tag-button-icon-size: 15px;
}
```

scurker (changes requested):
last thing!

scurker (approved):

PR #1200, opened 2023-09-21
fix(react): ensure that no results get announced for NVDA

### LINKED ISSUE #1201: Combobox Not Announcing "No Results" status (NVDA)
Auditor Issue: https://axeauditor.dequecloud.com/test-run/87ff92fc-25f1-11ee-a855-1b2298d1b894/issue/c5cc4b1c-5886-11ee-9399-9336eb8a0144?sortField=ordinal&sortDir=asc&filter%5Brule%5D=manual&filter%5Bpage_number%5D=49&row=0

Status message is not automatically announced by the screen reader.
Element description: The status message "No results found".
Element location: This applies to each combo box in the examples section.

Steps to reproduce: Type an option that is not present in the combo box. The status message appears but is not announced automatically.

scurker:
closes #1201

### DIFF
--- a/packages/react/src/components/Combobox/Combobox.tsx
+++ b/packages/react/src/components/Combobox/Combobox.tsx
@@ -65,20 +65,9 @@ const ComboboxNoResults = ({
 }: {
   children?: React.ReactNode;
 }): JSX.Element => {
-  const [liveText, setLiveText] = useState<React.ReactNode>('');
-
-  useEffect(() => {
-    // Note: NVDA will not announce live regions unless initially mounted,
-    // we work around this by arbitrarily updating the live region text
-    // on the next animation frame.
-    requestAnimationFrame(() => {
-      setLiveText(children || 'No results found.');
-    });
-  }, []);
-
   return (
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -60,10 +60,14 @@ const defaultAutoCompleteMatches = (inputValue: string, value: string) => {
...
-const ComboboxNoResults = (): JSX.Element => {
+const ComboboxNoResults = ({
+  children
+}: {
+  children?: React.ReactNode;
+}): JSX.Element => {
   return (
-    <div className="ComboboxListbox__empty" aria-live="polite">
-      No results found.
+    <div className="ComboboxListbox__empty" role="alert" aria-live="polite">
```
scurker, line 69: This was the best combination of roles that I could find that would work across the board. Unfortunately for NVDA/Jaws both prepend the announcement with "ALERT" but I could not find a combination with `aria-live`, `aria-atomic`, or `aria-relevant` that would work across the board.

thuey (approved):

PR #1204, opened 2023-09-22
docs: document design tokens

### LINKED ISSUE #1050: Document Design Tokens (styles, css variables, etc)
We need to document some of the different aspects of our design system to include all the various design tokens for color, typography, sizing, etc.

Some examples: 

- https://build.washingtonpost.com/foundations/color

UXPin example: https://app.uxpin.com/design-system/c933c74928e2195cc540/colors

orest-s:
Closes #1050


docs/components/Colors.tsx
```diff
@@ -0,0 +1,113 @@
...
+  );
+};
+
+export default function Colors({ colorGroup, colorNames }: ColorsProps) {
+  const colors: ColorProps[] = colorNames.map((color: string) => ({
+    name: color,
+    value: normalizeVarValue(color)
+  }));
+
+  console.log({ colors });
```
scurker, line 104:
```suggestion
```

scurker (commented):
Thanks for putting this together! I have some different thoughts I need to mull on, and will try to review this later today.

docs/pages/colors.mdx
```diff
@@ -0,0 +1,93 @@
...
+    "accent-danger",
+    "accent-danger-light",
+    "accent-warning-light",
+    "accent-warning",
+    "accent-caution",
+    "accent-info"
+  ]}
+/>
+
+### Gray
```
scurker, line 40:
```suggestion
## Gray
```

docs/pages/colors.mdx
```diff
@@ -0,0 +1,93 @@
...
+    "gray-20",
+    "gray-30",
+    "gray-40",
+    "gray-60",
+    "gray-80",
+    "gray-90"
+  ]}
+/>
+
+### Light
```
scurker, line 55:
```suggestion
## Light
```

docs/pages/colors.mdx
```diff
@@ -0,0 +1,93 @@
...
+  colorNames={[
+    "light-workspace-color",
+    "background-light",
+    "stroke-light",
+    "focus-light",
+    "accent-success-high-contrast",
+  ]}
+/>
+
+### Dark
```
scurker, line 68:
```suggestion
## Dark
```

docs/pages/spacing.mdx
```diff
@@ -0,0 +1,13 @@
+---
+title: Spacing
+name: Spacing
+path: /spacing
+---
+
+import Spacing from '../components/Spacing'
+
+### Spaces
```
scurker, line 9:
```suggestion
## Spaces
```

docs/pages/typography.mdx
```diff
@@ -0,0 +1,49 @@
...
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from '@deque/cauldron-react'
+
+Documentation and examples for Cauldron typography, including global settings, headings, body text, lists, and more.
+
+### Font
```
scurker, line 18:
```suggestion
## Font
```

docs/pages/typography.mdx
```diff
@@ -0,0 +1,49 @@
...
+  <TableBody>
+    <TableRow>
+      <TableCell>Base</TableCell>
+      <TableCell>Roboto', Helvetica, Arial, sans-serif</TableCell>
+      <TableCell>-</TableCell>
+    </TableRow>
+  </TableBody>
+</Table>
+
+### Font size
```
scurker, line 39:
```suggestion
## Font size
```

docs/pages/typography.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+Font size tokens are relative to the <b>--text-size-small</b>. Use font size tokens to define the size of the font.
+
+<CssParamsTable param={"--text-size"} />
+
+### Font weight
```
scurker, line 45:
```suggestion
## Font weight
```

packages/styles/button.css
```diff
@@ -10,7 +10,7 @@
   --button-background-color-secondary-active: var(--gray-30);
 
   --button-background-color-error: var(--error);
-  --button-background-color-error-disabled: #db6379;
+  --button-background-color-error-disabled: var(----accent-error-disabled);
```
scurker, line 13: Is this supposed to be this?

```suggestion
  --button-background-color-error-disabled: var(--accent-error-disabled);
```

I do appreciate syncing missing vars, but I think it might be better to split this into 2 PRs to isolate unrelated changes. A `feat` PR to implement any missing tokens/css vars and a `docs` PR to actually document the changes.

docs/components/Colors.css
```diff
@@ -0,0 +1,51 @@
...
+  --color-text-gray: var(--accent-dark);
+}
+
+.Colors {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 20px var(--space-medium);
+}
+
+.Color {
```
scurker, line 16: I'm thinking every color tile needs a border around it to better associate the color variable with the name and to help prevent "blending" with the background color when there's a non-contrasting background, something like what's displayed below: 

gray-40 border around the color tile

docs/components/Colors.css
```diff
@@ -0,0 +1,51 @@
...
+
+.Color .Color__bg {
+  height: 80%;
+  position: relative;
+  border-radius: 2px;
+}
+
+.Color .Color__bg--border {
+  border: 1px solid var(--border-color);
+}
```
scurker, line 31: If we change every color to always have a border, this won't be needed.

docs/components/Colors.css
```diff
@@ -0,0 +1,51 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  gap: 20px var(--space-medium);
+}
+
+.Color {
+  width: var(--color-width);
+  height: var(--color-height);
+  display: inline-block;
+  margin-right: 20px var(--space-medium);
```
scurker, line 20: I think we can achieve what you're doing here in more simple terms:

```css
.Colors {
  display: grid;
  grid-template-columns: repeat(4, 1fr);
  gap: var(--space-small);
}

.Color__bg {
  height: 3rem;
  position: relative;
}
```

The other advantage of removing the set fixed/widths is that the color tiles will always fill the container, and we can apply media queries to shrink down the column count when appropriate:

```css
@media(max-width: 38rem) { 
  .Colors {
    grid-template-columns: repeat(3, 1fr);
  }
}

@media(max-width: 36rem) { 
  .Colors {
    grid-template-columns: repeat(2, 1fr);
  }
}

...

```

The above code is an example, you will have to play around with the media queries to see what feels right here.

docs/components/Colors.css
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.Color .Color__bg {
+  height: 80%;
+  position: relative;
+  border-radius: 2px;
+}
```
scurker, line 27: Nit: One advantage of using BEM is that it tries to avoid nesting selectors so that there's no higher specificity getting applied. For this and other styles I think you can just use a single selector.

```suggestion
.Color__bg {
  height: 80%;
  position: relative;
  border-radius: 2px;
}
```

docs/pages/colors.mdx
```diff
@@ -0,0 +1,93 @@
+---
```
scurker, line 1: These don't really belong in the "getting started" section, but might need to be in a separate section just for our design system tokens. Most design systems have this as "Foundations" which should fit here as well.

docs/pages/spacing.mdx
```diff
@@ -0,0 +1,13 @@
...
+
+Use space tokens to define padding or margin when designing a component or layout.
+
+<Spacing />
```
scurker, line 13: I apologize that I didn't do a very good job of documenting this ticket, I had some ideas in my head that I didn't fully flesh out here as I was expecting myself to pick it up eventually.

We don't need a 1:1 mapping of what's in UXPin to what's here in cauldron but rather something more fleshed out. Here's an example of something that is a little more useful:

- https://designsystem.digital.gov/design-tokens/spacing-units/

docs/pages/zIndex.mdx
```diff
@@ -0,0 +1,9 @@
+---
```
scurker, line 1: I think we should omit this for now, we might need to review some of these to ensure they're what we want.

docs/pages/typography.mdx
```diff
@@ -0,0 +1,49 @@
...
+import {
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from '@deque/cauldron-react'
+
+Documentation and examples for Cauldron typography, including global settings, headings, body text, lists, and more.
```
scurker, line 16: ?

docs/pages/typography.mdx
```diff
@@ -0,0 +1,49 @@
...
+    <TableRow>
+      <TableHeader scope="col">Name</TableHeader>
+      <TableHeader scope="col">Value</TableHeader>
+      <TableHeader scope="col">Description</TableHeader>
+    </TableRow>
+  </TableHead>
+  <TableBody>
+    <TableRow>
+      <TableCell>Base</TableCell>
+      <TableCell>Roboto', Helvetica, Arial, sans-serif</TableCell>
```
scurker, line 33: Can we pull this value from the css var? There's a risk it will drift if we don't.

docs/pages/typography.mdx
```diff
@@ -0,0 +1,49 @@
...
+      <TableCell>-</TableCell>
+    </TableRow>
+  </TableBody>
+</Table>
+
+### Font size
+
+Font size tokens are relative to the <b>--text-size-small</b>. Use font size tokens to define the size of the font.
+
+<CssParamsTable param={"--text-size"} />
```
scurker, line 43: This is looking pretty good, although it would be super nice if we could include an example of text with the applied style.

scurker (changes requested):
Thanks for putting this together! I know there's a lot of comments/requested changes but I appreciate the work put into this and it's a really great start.

docs/pages/foundations/typography.mdx
```diff
@@ -0,0 +1,50 @@
...
+import {
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableHeader,
+  TableRow
+} from '@deque/cauldron-react';
+
+Cauldron harnesses a range of text Tokens to ensure a polished and engaging user experience.
```
scurker, line 16: Nit:

```suggestion
Cauldron harnesses a range of text tokens to ensure a polished and engaging user experience.
```

docs/pages/foundations/typography.mdx
```diff
@@ -0,0 +1,50 @@
...
+  renderExample={(name, value) => (
+    <span style={{ fontFamily: value }}>
+      This is an example of the font family
+    </span>
+  )}
+/>
+
+## Font size
+
+Font size tokens are relative to the <b>--text-size-small</b>. Use font size tokens to define the size of the font.
```
scurker, line 34: We should just be able to use raw markdown here.

```suggestion
Font size tokens are relative to the **--text-size-small**. Use font size tokens to define the size of the font.
```

scurker (commented):
This is looking really good! 2 small nits, and I noticed that I was getting the following error on the preview build:

> Error accessing stylesheet: DOMException: Failed to read the 'cssRules' property from 'CSSStyleSheet': Cannot access rules

scurker (commented):
Code LGTM, the last thing is in the docs preview it looks like the css vars for spacing/typography isn't appearing.

docs/utils/getCssVariablesStartingWith.ts
```diff
@@ -3,7 +3,7 @@ export function getCssVariablesStartingWith(prefix: string) {
 
   // Iterate through all style sheets in the document
   for (const styleSheet of Array.from(document.styleSheets).filter(
-    (s) => !s.href
+    (s) => !s.href || s.href.startsWith(window.location.origin)
```
scurker, line 6: Interesting, was it trying to get cross-origin stylesheets?

orest-s: Yeah, exactly, i added this checking because of error message, which we catch in case getting stylesheets from another origin, but it excluded our stylesheets too, so i was need to avoid excluding our origin, adding `startsWith` function check. :)
Do you have any concerns on this?

scurker: My only thought here is that it might be better to parse the href instead of using `startsWith` since I'm not sure how compatible this might be cross browser. Something like:

```js
new URL(s.href).origin === window.location.origin
```

It might need to be wrapped in a try/catch since `new URL` can thrown.

scurker (approved):
I updated the title to `docs` because we wouldn't want this to bump our cauldron version.

PR #1205, opened 2023-09-25
feat(react): remove large overlay

### LINKED ISSUE #989: Deprecate "large" variants of Overlay, LoaderOverlay
We don't appear to use the large variants of this component, so we should remove the "large" variants and default to a single variant.

orest-s:
Close #989 

BREAKING CHANGE: removed variant property from Loader/LoaderOverlay components, since it's not used anymore.


docs/pages/components/Loader.mdx
```diff
@@ -31,7 +31,7 @@ import { Loader } from '@deque/cauldron-react'
     {
       name: 'variant',
       type: ['small', 'large'],
-      description: 'Loader variant, this property will be removed in a future version.',
+      description: 'Loader variant, this property was removed.',
```
scurker, line 34: If we're deprecating the variant, we can safely remove this object.

docs/pages/components/LoaderOverlay.mdx
```diff
@@ -64,7 +64,7 @@ function LoaderOverlayExample() {
     {
       name: 'variant',
       type: ['small', 'large'],
-      description: 'Loader variant, this property will be removed in a future version.',
+      description: 'Loader variant, this property was removed',
```
scurker, line 67: If we're deprecating the variant, we can safely remove this object.

scurker (commented):
Just 2 comments.

For now, we're going to hold off on actually removing it to comply with our [deprecation guidelines](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#deprecating) but we can leave this PR in a holding pattern.

scurker:
We would also need to make this `feat` and include `BREAKING CHANGE: [description]` in the body of the commit message.

scurker (approved):

Issue #1070, opened 2023-05-09
Sync TextField with UXPin


scurker:
Some minor UX changes needed:

- Disabled border color should match UXPin
- Focus ring should be 1px smaller (currently appears to be 3px)
- Error field should have a 2px border in its resting state
- Dark theme `field-border-color` should be #74818B

Do we want to get rid of the bottom-border color difference and make it a single solid color that passes WCAG 2.2 AA? See screenshot below:

![Screenshot 2023-06-22 at 12 44 20 PM](https://github.com/dequelabs/cauldron/assets/28606186/285f11dc-3a1a-4732-9a6c-0faeae9ed912)

PR #1206, opened 2023-09-25
fix(style): synced text field styles with pattern library

orest-s:
Closes: #1070


scurker (changes requested):
A couple of things here:

The error field focus ring shifts content during focus (see below video), the footprint of the element should not be shifting content on the page:

https://github.com/dequelabs/cauldron/assets/1062039/012556e3-a49e-4a5e-97b0-b8a515d43c01

The default focus ring still appears to be too large, according to our [pattern library](https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161871161/simulate/no-panels?mode=i) it needs to be 1px smaller.

Minor nit: the commit message / pr title should be more descriptive in describing what changed. Something like "synced text field styles with pattern library". Otherwise, it will be difficult to know what actually changed when it gets placed in the changelog.

We need to also ensure that any style changes are reflected on Combobox as well. Combobox isn't a 1:1 of textfield and has to duplicate a few styles over as it cannot use the textfield styles directly.

scurker (changes requested):
Changes here look good for the textfield/combobox error focus ring changes.

It does look like we're still missing the following from the issue:
> - Focus ring should be 1px smaller (currently appears to be 3px)

The above applies to the default "blue" focus ring that gets applied, so we just need to shrink both TextField and Combobox by 1px.

This is also an existing issue, but I noticed that hover state of the combobox error is currently using the wrong focus color that would be nice to sync up here as well:

cauldron combobox with blue focus ring

packages/styles/forms.css
```diff
@@ -94,10 +95,9 @@ textarea:focus,
 [role='listbox']:focus,
 [role='spinbutton']:focus {
   outline: 0;
-  border: 1px solid var(--field-border-color-focus);
-  box-shadow: 0 0 0 2px var(--field-border-color-focus),
-    0 0 5px var(--field-border-color-focus-glow),
-    inset 0 1px 2px rgba(0, 0, 0, 0.05);
+  border: 2px solid var(--field-border-color-focus);
+  box-shadow: 0px 0px 5px var(--field-border-color-focus-glow),
+    inset 0px 1px 2px rgba(0, 0, 0, 0.05);
```
scurker, line 100: So the trick we had before with the focus ring to not displace content was a combination of:

- 1px outline
- 2px outer solid box-shadow (+ focus glow)

To shrink the solid outline while keeping the same focus footprint, we just need to reduce the box-shadow by 1px:

```suggestion
  border: 1px solid var(--field-border-color-focus);
  box-shadow: 0 0 0 1px var(--field-border-color-focus),
    0 0 5px var(--field-border-color-focus-glow),
    inset 0px 1px 2px rgba(0, 0, 0, 0.05)
```

This will need to be applied to both `TextField` and `Combobox`

scurker (changes requested):

packages/styles/combobox.css
```diff
@@ -91,10 +91,11 @@
 
 .Combobox__input:focus-within {
   border: 1px solid var(--combobox-input-border-focus-color);
-  box-shadow: 0 0 0 2px var(--combobox-input-border-focus-color),
-    0 0 5px var(--combobox-input-border-focus-glow-color),
-    inset 0 1px 2px rgba(0, 0, 0, 0.05);
+  box-shadow: 0 0 0 1px var(--combobox-input-border-focus-color),
+    0px 0px 5px var(--combobox-input-border-focus-glow-color),
```
scurker, line 95: Nit: 0 values can be unitless in css.

```suggestion
    0 0 5px var(--combobox-input-border-focus-glow-color),
```

packages/styles/combobox.css
```diff
@@ -231,13 +232,28 @@
 
 .Combobox__input--error {
   border: 1px solid var(--combobox-input-error-border-color);
+  box-shadow: 0 0 0 1px var(--combobox-input-error-border-color),
+    inset 0px 1px 2px rgba(0, 0, 0, 0.05);
```
scurker, line 236:
```suggestion
    inset 0 1px 2px rgba(0, 0, 0, 0.05);
```

packages/styles/combobox.css
```diff
@@ -231,13 +232,28 @@
...
 }
 
 .Combobox__input--error:focus-within {
   border: 1px solid var(--combobox-input-error-border-color);
-  box-shadow: 0 0 0 1px var(--combobox-input-error-focus-border-color),
-    inset 0 2px 3px 0 rgba(0, 0, 0, 0.05),
-    0 0 5px 0 var(--combobox-input-error-focus-border-glow-color);
+  box-shadow: 0 0 0 1px var(--combobox-input-error-border-color),
+    0px 0px 5px var(--field-border-color-error-focus-glow),
```
scurker, line 242:
```suggestion
    0 0 5px var(--field-border-color-error-focus-glow),
```

packages/styles/combobox.css
```diff
@@ -231,13 +232,28 @@
...
+    inset 0px 1px 2px rgba(0, 0, 0, 0.05);
+}
+
+.Combobox__input--error:focus-within:hover {
+  border: 1px solid var(--combobox-input-error-border-color);
+  box-shadow: 0 0 0 1px var(--combobox-input-error-border-color),
+    0px 0px 5px var(--field-border-color-error-focus-glow),
```
scurker, line 249:
```suggestion
    0 0 5px var(--field-border-color-error-focus-glow),
```

packages/styles/combobox.css
```diff
@@ -231,13 +232,28 @@
...
+    inset 0px 1px 2px rgba(0, 0, 0, 0.05);
+}
+
+.Combobox__input--error:hover {
+  border: 1px solid var(--combobox-input-error-border-color);
+  box-shadow: 0 0 0 1px var(--combobox-input-error-border-color),
+    inset 0px 1px 2px rgba(0, 0, 0, 0.05);
```
scurker, line 256:
```suggestion
    inset 0 1px 2px rgba(0, 0, 0, 0.05);
```

packages/styles/forms.css
```diff
@@ -95,9 +96,9 @@ textarea:focus,
 [role='spinbutton']:focus {
   outline: 0;
   border: 1px solid var(--field-border-color-focus);
-  box-shadow: 0 0 0 2px var(--field-border-color-focus),
-    0 0 5px var(--field-border-color-focus-glow),
-    inset 0 1px 2px rgba(0, 0, 0, 0.05);
+  box-shadow: 0 0 0 1px var(--field-border-color-focus),
+    0px 0px 5px var(--field-border-color-focus-glow),
```
scurker, line 100:
```suggestion
    0 0 5px var(--field-border-color-focus-glow),
```

packages/styles/forms.css
```diff
@@ -95,9 +96,9 @@ textarea:focus,
...
+    inset 0px 1px 2px rgba(0, 0, 0, 0.05);
```
scurker, line 101:
```suggestion
    inset 0 1px 2px rgba(0, 0, 0, 0.05);
```

packages/styles/forms.css
```diff
@@ -120,6 +121,9 @@ textarea:focus:hover,
 [role='listbox']:focus:hover,
 [role='spinbutton']:focus:hover {
   border: 1px solid var(--field-border-color-focus-hover);
+  box-shadow: 0 0 0 1px var(--field-border-color-focus-hover),
+    0px 0px 5px var(--field-border-color-focus-glow),
```
scurker, line 125:
```suggestion
    0 0 5px var(--field-border-color-focus-glow),
```

packages/styles/forms.css
```diff
@@ -120,6 +121,9 @@ textarea:focus:hover,
...
+    inset 0px 1px 2px rgba(0, 0, 0, 0.05);
```
scurker, line 126:
```suggestion
    inset 0 1px 2px rgba(0, 0, 0, 0.05);
```

packages/styles/forms.css
```diff
@@ -131,6 +135,8 @@ textarea.Field--has-error,
 [role='listbox'].Field--has-error,
 [role='spinbutton'].Field--has-error {
   border: 1px solid var(--field-border-color-error);
+  box-shadow: 0 0 0 1px var(--field-border-color-error),
+    inset 0px 1px 2px rgba(0, 0, 0, 0.05);
```
scurker, line 139:
```suggestion
    inset 0 1px 2px rgba(0, 0, 0, 0.05);
```

packages/styles/forms.css
```diff
@@ -163,9 +169,9 @@ textarea.Field--has-error:focus,
 [role='listbox'].Field--has-error:focus,
 [role='spinbutton'].Field--has-error:focus {
   border: 1px solid var(--field-border-color-error);
-  box-shadow: 0 0 0 1px var(--field-border-color-error-hover),
-    inset 0 2px 3px 0 rgba(0, 0, 0, 0.05),
-    0 0 5px 0 var(--field-border-color-error-focus-glow);
+  box-shadow: 0 0 0 1px var(--field-border-color-error),
+    0px 0px 5px var(--field-border-color-error-focus-glow),
```
scurker, line 173:
```suggestion
    0 0 5px var(--field-border-color-error-focus-glow),
```

scurker (commented):
One remaining nit, and this should be good to go!

scurker (dismissed):

scurker (changes requested):
This is very close, last remaining this is that the Combobox error hover ring appears to be the wrong color in dark mode:

scurker (approved):

PR #1207, opened 2023-09-25
fix(react): prevent esm modules from appearing in cjs dist

scurker:
It appears something in the [npm changes with babel/rollup](https://github.com/dequelabs/cauldron/compare/v5.7.1...v5.8.0) caused react-syntax-highlighter to be pulled in as a dependency in the production bundle instead of inlined. Using the cjs dist of react-syntax-highlighter should prevent this issue from occuring.

### DIFF
--- a/packages/react/rollup.config.js
+++ b/packages/react/rollup.config.js
@@ -8,7 +8,12 @@ export default {
   input: 'src/index.ts',
   external: [
     ...Object.keys(pkg.dependencies),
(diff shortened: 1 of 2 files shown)


thuey (approved):

scurker:
There shouldn't be any component changes here as this is just fixing a build issue, so I'm going to go ahead and merge before waiting for the a11y testing to complete.

PR #1212, opened 2023-09-26
docs: fix table of contents not scrolling when vertically overflowing

### LINKED ISSUE #1203: Documentation Table of Contents not visible during tab focus
Auditor issue: https://axeauditor.dequecloud.com/test-run/87ff92fc-25f1-11ee-a855-1b2298d1b894/issue/60646e32-5884-11ee-be02-0f9218df4b24?sortField=ordinal&sortDir=asc&filter%5Brule%5D=manual&filter%5Bpage_number%5D=49&row=0

The element is not visible when it receives focus by default.
Element description: Below links.
ComboboxGroup
Related Components

Element location: In the section "On this page"

Additional Information: When the keyboard user is navigating through the page using tab key navigation, the elements are focused but are not visible to the keyboard users.
However when the user scrolls the page using a down arrow key, the keyboard user is able to see those links, hence raised as a best practice.

## Dev Notes

This appears to be an issue in shorter viewports with regards to the Table of Contents links on the right area of documentation. We should consider making this area scrollable when the on this page links do not fit.

scurker:
closes #1203

### DIFF
--- a/docs/components/ComponentLayout.css
+++ b/docs/components/ComponentLayout.css
@@ -22,6 +22,11 @@ main > section[aria-labelledby='on-this-page'] {
   position: sticky;
   top: var(--space-medium);
   align-self: start;
+  overflow-x: auto;
+  /* account for top + bottom bars and spacing when enforcing max-height for scroll behavior */
+  max-height: calc(
+    100vh - var(--top-bar-height) - var(--top-bar-height) - var(--space-large)
+  );
 }
 
 section[aria-labelledby='on-this-page'] > .toc {


thuey (approved):
This feels a bit hacky, but I'm struggling to think of a better solution.

scurker:
> This feels a bit hacky, but I'm struggling to think of a better solution.

How so? With the way we currently manage layouts in cauldron and with this being position fixed, we have a bit of our hands tied as to what solutions we can use. 

It's not an uncommon method, [MDN uses a very similar method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) for their fixed table of contents:

mdn css rules for their sticky sidebar container with a height using a calculated offset with 100vh and position sticky

thuey:
> How so? With the way we currently manage layouts in cauldron and with this being position fixed, we have a bit of our hands tied as to what solutions we can use.

I think it's the best approach given how the layout works--I can't think of a better solution. I just wish we didn't have to add all of those subtractions to the height. If, for example, we added some other amount of padding/margin somewhere, it could throw this off and we would need to revisit

If our layout worked differently, the ToC could be a sibling of the component area, instead of a child. But that would then throw off how the scrolling works. So gain, this is probably the best solution given how things work.

PR #1227, opened 2023-10-04
docs: set aria-current on breadcrumb docs

### LINKED ISSUE #1215: Breadcrumb: Selected Aria State is Missing
Auditor Issue: https://axeauditor.dequecloud.com/test-run/87ff92fc-25f1-11ee-a855-1b2298d1b894/issue/7cf7860a-5cfe-11ee-933f-07f04e8012cd?activeTab=dt-issue&page=0&sortField=ordinal&sortDir=asc&row=3&issuesCount=&issuesCount=&filter=

## Description

State: Selected state of the element is missing or incorrect

Element description:  "Four" - aria current is missing for the current location 
Element location: In the "Basic" and "Custom Separators" examples

## Dev Notes

We should add `aria-current="true"` and toss up a note that as a best practice, the author should provide an [`aria-current`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) value where necessary.

scurker:
closes #1215

### DIFF
--- a/docs/pages/components/Breadcrumb.mdx
+++ b/docs/pages/components/Breadcrumb.mdx
@@ -23,7 +23,7 @@ import {
   <BreadcrumbLink href="/one">One</BreadcrumbLink>
   <BreadcrumbLink href="/two">Two</BreadcrumbLink>
   <BreadcrumbLink href="/three">Three</BreadcrumbLink>
-  <BreadcrumbItem>Four</BreadcrumbItem>
+  <BreadcrumbItem aria-current="true">Four</BreadcrumbItem>
 </Breadcrumb>
 ```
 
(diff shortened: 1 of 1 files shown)


thuey (approved):
Besides fixing this in the docs, would we want to do this automatically? For example, `BreadcrumbItem` could set this by default. If that is too opinionated, then the `Breadcrumb` component could tack the property on to the last child. Just a thought.

scurker:
> Besides fixing this in the docs, would we want to do this automatically? For example, BreadcrumbItem could set this by default. If that is too opinionated, then the Breadcrumb component could tack the property on to the last child. Just a thought.

It's a best practice, and it may not always be the case that the last breadcrumb item is the current one. There's also the issue where `aria-current` could be something else like `page` or `step`.

Additionally, the [ARIA authoring practices for breadcrumb](https://www.w3.org/WAI/ARIA/apg/patterns/breadcrumb/) specifically mentions:

> If the element representing the current page is not a link, aria-current is optional.

So I'm not sure we need to make assumptions here.

Issue #1197, opened 2023-09-16
Migrate from Enzyme to React Testing Library or Playwright Component Testing


anastasialanz:
# Background
In order to update Cauldron to React 18, we need to update our testing framework. Currently, [Enzyme](https://enzymejs.github.io/enzyme/) does not support [React 17](https://github.com/enzymejs/enzyme/issues/2429) or [React 18](https://github.com/enzymejs/enzyme/issues/2524).

Related resources:
https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl
https://stackoverflow.com/questions/72029189/does-enzyme-support-react-version-18

# Recommendations
### Pros
- Other teams already use React Testing Library and updating component tests in Cauldron would be familiar to them
- Testing components the way a user would use them
### Cons
- React Testing Library is tied to a React version

In this branch https://github.com/dequelabs/cauldron/tree/rfc--testing-library, I've used the matching Reat Testing Library to React version and updated the tests for the Button component as a proof of concept.

## Migration Plan

https://testing-library.com/docs/react-testing-library/migrate-from-enzyme

> To ensure a successful migration, we recommend doing it incrementally by running the two test libraries side by side in the same application, porting your Enzyme tests to React Testing Library one by one. That makes it possible to migrate even large and complex applications without disrupting other business because the work can be done collaboratively and spread out over time.

I recommend that any new components switch to using RTL in their tests and that we make separate issues for each component testing migration.

## ⚠️  Switching to [Playwright Component Testing](https://playwright.dev/docs/test-components)
@scurker has mentioned in the past about trying out Playwright Component Testing.
https://deque.slack.com/archives/C5MUFQ7UM/p1653585095625409?thread_ts=1653582814.665619&cid=C5MUFQ7UM

### Pros
- Would allow for testing in a real browser

### Cons
- ⚠️ Playwright component testing is considered experimental. The repo says, “BEWARE This package is EXPERIMENTAL and does not respect semver.”
- Since it's new, the team is not as familiar with this framework

Additional resources:
https://github.com/microsoft/playwright/tree/main/packages/playwright-ct-react
https://www.lambdatest.com/learning-hub/playwright-component-testing

scurker:
Thanks for putting this together! 

I think switching to RTL probably makes sense as our teams are likely already familiar with it, even with some of the caveats that might exist in RTL.

A couple of questions/thoughts:

[...]

PR #1229, opened 2023-10-06
test(button): switch button tests to react testing library

anastasialanz:
Relates to https://github.com/dequelabs/cauldron/issues/1197 

This PR updates the `Button` component tests to use the React Testing Library and switches the test file to use TypeScript. It also adds @types packages for enzyme, jest, and jest-axe.

Screenshot 2024-01-04 at 1 01 13 PM

## Validation Steps
- [ ] Run `yarn test` and see a pass for `src/components/Button/index.test.tsx`
- [ ] Verify that there is coverage for the `Button` component
- [ ] Verify that the new test file uses TypeScript and is in the same directory as the component folder


packages/react/src/axe.ts
anastasialanz, line 1: Since we're getting rid of the `__tests__` directory eventually, I copied this file to the root of the project.

packages/react/__tests__/setupTests.ts
anastasialanz, line 1: Should I also move this to `src`?

packages/react/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "__tests__/setupTests.ts"],
```
anastasialanz, line 3: https://github.com/testing-library/jest-dom?tab=readme-ov-file#with-typescript

scurker (approved):
Nice work! 🎉

PR #1232, opened 2023-10-10
feat(Combobox): allow Combobox to set hidden form values

### LINKED ISSUE #1084: Add Combobox Component
Requirements TBD

### LINKED ISSUE #1184: Combobox Hidden Form Values
see #1088 for initial implementation

Comboboxes typically have an input value and a label value, which are not necessarily the same. Authors may want to have a value that is different from both when the input gets submitted via a form.

To support opaque or hidden form values, a `formValue` property will be added to Combobox to support this capability.

## Via Combobox Children:

```jsx
<Combobox name="fruit" value="Apple">
  <ComboboxOption key="123" value="Apple" formValue="123">🍎 Apple</ComboboxOption>
  <ComboboxOption key="456" value="Banana" formValue="456">🍌 Banana</ComboboxOption>
  <ComboboxOption key="789" value="Cucumber" formValue="789">🥒 Cucumber</ComboboxOption>
</Combobox>
```

## Via Combobox Options:

```jsx
function ComboboxFormValues() {
  const options = [
    { key: 123, label: 'Apple', formValue: '123' },
    { key: 456, label: 'Banana', formValue: '456' },
    { key: 789, label: 'Cucumber', formValue: '789' }
  ]
  return <Combobox name="fruit" value="Apple" options={options} />
}
```

When `name` is set, a hidden input field will be rendered with the name prop that will have its value set to the `formValue` of the currently selected combobox option. When `formValue` is omitted or not set for the selected combobox option, the value of the hidden input will default to the set `value` or the inner text contents of the selected combobox option.

scurker:
closes #1184

### DIFF
--- a/packages/react/src/components/Combobox/Combobox.tsx
+++ b/packages/react/src/components/Combobox/Combobox.tsx
@@ -49,6 +49,7 @@ interface ComboboxProps
   }) => void;
   onActiveChange?: (option: ListboxOption) => void;
   renderNoResults?: (() => JSX.Element) | React.ReactElement;
+  formKey?: string;
   portal?: React.RefObject<HTMLElement> | HTMLElement;
 }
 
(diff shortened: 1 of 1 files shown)


scurker:
Proposal has been updated and the approach was pivoted, see #1264

PR #1234, opened 2023-10-10
fix(styles): prevent overflow for TwoColumnPanel left container at narrow viewports

### LINKED ISSUE #1220: TwoColumnPanel: Content lost at reflow
Auditor Issue: https://axeauditor.dequecloud.com/test-run/87ff92fc-25f1-11ee-a855-1b2298d1b894/issue/25eca8fe-5d35-11ee-a207-6752144d7996?activeTab=dt-issue&page=0&sortField=ordinal&sortDir=asc&row=3&issuesCount=&issuesCount=&filter=

## Description

Content is lost at reflow

Element description: Hide panel button is lost at reflow
Element location: Below the "Default" and "With Optional Group Headings" 

Steps to reproduce:
- Resize the viewport to 640px width by 512px height at 100% and then zoom to 200% 
- Now click on "Show panel" button
- The hide panel button is lost at reflow

scurker:
closes #1220

### DIFF
--- a/packages/styles/two-column-panel.css
+++ b/packages/styles/two-column-panel.css
@@ -261,6 +261,7 @@
   .TwoColumnPanel--hide .TwoColumnPanel__Left,
   .TwoColumnPanel--show .TwoColumnPanel__Left {
     max-width: 100%;
+    /* This will constrain the component to the max-width of its container, as well as adjusting the offset margin to the max negative value to prevent incorrect offset when the max-width is less than the column width. */
     margin-left: max(
       calc(var(--two-column-panel-left-column-width) * -1),
       calc(100% * -1)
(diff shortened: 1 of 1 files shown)


packages/styles/two-column-panel.css
```diff
@@ -260,7 +260,11 @@
 
   .TwoColumnPanel--hide .TwoColumnPanel__Left,
   .TwoColumnPanel--show .TwoColumnPanel__Left {
-    margin-left: calc(var(--two-column-panel-left-column-width) * -1);
+    max-width: 100%;
+    margin-left: max(
```
scurker, line 264: This will constrain the component to the max-width of its container, as well as adjusting the offset margin to the max negative value to prevent incorrect offset when the max-width is less than the column width.

thuey: nit:
```suggestion
    max-width: 100%;
    /* This will constrain the component to the max-width of its container, as well as adjusting the offset margin to the max negative value to prevent incorrect offset when the max-width is less than the column width. */
    margin-left: max(
```

thuey (dismissed):

PR #1236, opened 2023-10-13
feat(Modal): sync modal styles with pattern library and include alignment props

### LINKED ISSUE #990: Sync Modal with UXPin
Update the dismiss action to match the new design:

modal with close action

`<ModalContent>` and `<ModalFooter>` should allow for alignment, i.e. `align="right"`. Defaults as `align="left"`.

Update the padding to the `ModalFooter` to match `16px` all around.

scurker:
closes #990

### DIFF
--- a/packages/styles/dialog.css
+++ b/packages/styles/dialog.css
@@ -160,11 +160,11 @@
 }
 
 .cauldron--theme-dark .Dialog__header {
-  border-bottom: 1px solid #5d676f;
+  border-bottom: 1px solid var(--stroke-dark);
 }
 
 .cauldron--theme-dark .Dialog__footer {
-  border-top: 1px solid #5d676f;
+  border-top: 1px solid var(--stroke-dark);
 }
 
 .cauldron--theme-dark .Dialog__header,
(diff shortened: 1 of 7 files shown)


packages/styles/dialog.css
```diff
@@ -166,20 +177,14 @@
...
+  color: var(--accent-medium);
 }
 
 .cauldron--theme-dark .Modal--info .Dialog__header,
 .cauldron--theme-dark .Modal--info .Dialog__content,
 .cauldron--theme-dark .Modal--info .Dialog__footer {
-  background-color: var(--accent-medium);
-}
-
-.cauldron--theme-dark .Modal--info .Dialog__header {
```
dequejosie, line 181: I wonder if it is necessary to add the border-bottom for info Modal. At least the new ad modal I am looking at does not have that.

dequejosie:
Left one question. Not blocking. Changes make sense to me.

dequejosie (approved):

packages/styles/dialog.css
scurker: This is what's in the pattern library at least for the "simple" variation: https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161871161/simulate/no-panels?mode=i

dequejosie: I see. I guess we will have to remove the bottom border for multiple places in the extension.

PR #1245, opened 2023-10-17
feat(react): remove PropTypes from all react components

### LINKED ISSUE #1164: Remove PropTypes from React Components
Maintaining both PropTypes and typescript props has been troublesome and challenging to keep in sync. Given that typescript provides the functionality to verify props, we will be removing PropTypes from components and _only_ use typescript types going forward.

For teams/authors that need prop validation, the recommendation going forward will be that you use typescript.

scurker:
BREAKING CHANGE: This removes all PropTypes from react components. If an author needs to depend on specific types, the recommendation going forward is to use Typescript types to validate React component props.

closes #1164


scurker:
99% of the output here was done via the following codemod and jscodeshift commands:

## Codemod

```ts
import type {
  API,
  FileInfo
} from 'jscodeshift'

export default function (file: FileInfo, api: API) {
  const j = api.jscodeshift;
  const root = j(file.source);

  // Find assignment propTypes and remove them
  const assignmentPropTypes = root
    .find(j.AssignmentExpression, {
      left: {
        property: {
          name: 'propTypes'
        }
      }
    })
  assignmentPropTypes.remove();

  // Find static propTypes from class based components and remove them
  const staticPropTypes = root
    .find(j.ClassProperty, {
      key: {
        type: 'Identifier',
        name: 'propTypes'
      },
      static: true
    })
  staticPropTypes.remove()

  // Remove import declaration
  const importPropType = root
    .find(j.ImportDeclaration, {
      source: {
        value: 'prop-types'
      }
    })
  importPropType.remove()

  return root.toSource();
}
```

## CLI Command

```sh
npx jscodeshift -t ./remove-react-prop-types.ts --parser=tsx --extensions=tsx ./packages/react/src/**/*.tsx
```

packages/react/src/components/Address/index.tsx
```diff
@@ -13,12 +12,10 @@ export const Address = ({ children, className, ...other }: AddressProps) => (
...
 Address.displayName = 'Address';
-Address.propTypes = {
-  children: PropTypes.node.isRequired,
-  className: PropTypes.string
-};
 
-type AddressLineProps = React.HTMLAttributes<HTMLElement>;
+type AddressLineProps = React.HTMLAttributes<HTMLElement> & {
+  className?: string;
+};
```
thuey, line 18: Isn't the `className` already implied?

thuey (commented):
There are a lot of places where `className` was added, but it seems unnecessary, unless I'm missing something.

packages/react/src/components/Address/index.tsx
scurker: It is, but for some reason `react/prop-types` isn't picking it up, so it's throwing all kinds of linter errors where it was previous validating `className` via `propTypes`.

So the options are either A) ignore `className` with some linter config, or B) ensure that `className` is included as part of props.

thuey: Ah, I see. It's because `className` isn't an `HTMLAttribute` of `HTMLElement`. We could instead potentially do `React.HTMLProps` or `React.ComponentProps` if our version of React supports it. See: https://stackoverflow.com/questions/48198180/what-is-the-difference-between-react-htmlprops-and-react-htmlattributest

scurker: I'm down for that change, but I think that work should be handled separately from this since it does seems like there's some differences between `ComponentProps` and `HTMLAttributes` that we should try to document first before diving in.

thuey (approved):

packages/react/src/components/Address/index.tsx
scurker: https://github.com/dequelabs/cauldron/issues/1252

thuey (approved):

PR #1250, opened 2023-10-19
chore: include wcag22aa as part of accessibility checks

### LINKED ISSUE #1249: Run WCAG22AA checks as part of testing in CI
Now that WCAG22 is an official standard, we should include both WCAG22A and WCAG22AA as part of our accessibility testing in CI.

scurker:
closes #1249

### DIFF
--- a/e2e/accessibility.ts
+++ b/e2e/accessibility.ts
@@ -7,6 +7,7 @@ import chalk from 'chalk';
 import fs from 'fs';
 import path from 'path';
 import assert from 'assert';
+import PQueue from 'p-queue';
 import { AddressInfo } from 'net';
 
 const DIST_PATH = path.join(__dirname, '..', 'docs', 'dist');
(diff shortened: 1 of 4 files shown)


e2e/accessibility.ts
```diff
@@ -45,69 +35,106 @@ const main = async (): Promise<void> => {
...
-      console.log(title, dots, symbol);
-
-      for (const { id, help, nodes } of violations) {
-        console.log('↳', chalk.underline(chalk.magenta(id)), '━', help);
-        nodes.map(node => {
-          console.log(
-            '  ↳',
-            chalk.bgGreen(chalk.black(`${node.target}`)),
-            chalk.yellowBright(node.html)
+  await Promise.all(
```
scurker, line 62: Changes here were primarily wrapping the puppeteer browse spin-up into a queue to run in paralle.

e2e/accessibility.ts
```diff
@@ -16,23 +17,12 @@ const MAX_WIDTH = IS_CI ? 80 : process.stdout.columns - 8;
 const THEMES = ['light', 'dark'];
 const AXE_PATH = require.resolve('axe-core');
 const AXE_SOURCE = fs.readFileSync(AXE_PATH, 'utf8');
+const CONCURRENCY = parseInt(process.env.CONCURRENCY as string, 10) || 8;
```
scurker, line 20: I played around with this in CI, `8` seems like the sweet spot of concurrency as higher numbers don't reduce the time and lower numbers increase it. 

With this change, it cuts down total CI time from about 15 minutes to about 5 minutes.

Issue #1052, opened 2023-05-02
Sync Stepper with UXPin


scurker:
## Dark Mode

![stepper component with different states in dark mode](https://user-images.githubusercontent.com/1062039/235750694-107eaba4-8b82-4fc6-a3d0-759de8dc72ce.png)

## Light Mode

![stepper component with different states in light mode](https://user-images.githubusercontent.com/1062039/235750818-cd17ce2c-fa91-4e64-b8a9-00d821b76e64.png)

Main differences are:

- Text weight / size has changed
- Progress circle size has changed
- Slight gap change
- "Completed" steps have a checkmark

Refer to https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161871161/simulate/no-panels?mode=i for all of the changes

PR #1251, opened 2023-10-19
fix(style): synced stepper with pattern library

orest-s:
Closes: #1052


packages/react/src/components/Stepper/index.tsx
```diff
@@ -59,11 +60,16 @@ export const Step = (props: StepProps) => {
             association="aria-labelledby"
             aria-label={tooltipText}
           >
-            <div className="Stepper__step-indicator" />
+            <div className="Stepper__step-indicator">
+              {props.status === 'complete' && <Icon type={'check'} />}
```
scurker, line 64: Nit:

```suggestion
              {status === 'complete' && <Icon type="check" />}
```

packages/react/src/components/Stepper/index.tsx
```diff
@@ -59,11 +60,16 @@ export const Step = (props: StepProps) => {
...
+            </div>
           </TooltipTabstop>
         ) : (
           <>
-            <div className="Stepper__step-indicator" />
+            <div className="Stepper__step-indicator">
+              {props.status === 'complete' && <Icon type={'check'} />}
```
scurker, line 70: Nit:

```suggestion
              {status === 'complete' && <Icon type="check" />}
```

packages/styles/stepper.css
```diff
@@ -72,8 +75,13 @@
   box-shadow: 0 0 0 2px var(--accent-dark);
 }
 
+.Stepper__step--current .Stepper__step-indicator::before {
```
scurker, line 78: Nit: we typically use `:before` anywhere we set pseudo selectors, so this is mostly for consistency purposes.

```suggestion
.Stepper__step--current .Stepper__step-indicator:before {
```

packages/styles/stepper.css
```diff
@@ -86,12 +94,26 @@
   color: var(--white);
 }
 
+.Stepper__step--complete .Stepper__step-indicator::before {
```
scurker, line 97: Same as above.

packages/styles/stepper.css
```diff
@@ -116,3 +138,11 @@
...
   color: var(--accent-dark);
 }
+
+.cauldron--theme-dark .Stepper__step--current .Stepper__step-label {
+  color: var(--white);
+}
+
+.cauldron--theme-dark .Stepper__step--complete .Stepper__step-indicator svg {
+  color: var(--black);
+}
```
scurker, line 148: A couple of thoughts here:

This is an older component, so some of the "patterns" here are a little bit outdated. Our guidelines are [here](https://cauldron.dequelabs.com/style-guidelines#theming) if you need a refresher, but we should generally lean towards avoiding long selectors as it makes it harder to style in an app and instead lean on css variables to set the colors according to the theme according to the current scope of the theme.

Additionally, `--black` doesn't exist and `--white` is primarily there for legacy purposes so `#fff` and `#000` are fine to set as the colors since those definitions won't be changing.

scurker (changes requested):
Component looks good, just leaving a couple of nits and 1 requested change here.

packages/styles/stepper.css
```diff
@@ -1,12 +1,58 @@
-.Stepper {
-  --step-indicator-size: 2em;
+:root {
+  --step-indicator-size: 1.867em;
   --step-line-height: 0.53em;
-  --step-line-offset: 1.33em;
+  --step-line-offset: 1.53em;
```
scurker, line 4: I realize these were `em` before, but they should probably be `rem` so they don't scale relative to the inherited font size but the root font size.

I think `step-indicator-size` needs to be `1.75rem` to match the `28px` set in the uxpin. The others should be updated to match accordingly.

And FYI, line-height values are already relative, so `em` can be omitted entirely, instead specifying the number without the unit.

scurker (changes requested):
Component looks good, just one last change needed to update the `em` values.

scurker (approved):

PR #1256, opened 2023-10-20
chore: prevent preview action from running against master branch

scurker:
The action frequently fails on master, and we really only need previews against pull requests as `develop` will always have the latest branch.

### DIFF
--- a/.github/workflows/preview-cleanup.yml
+++ b/.github/workflows/preview-cleanup.yml
@@ -3,6 +3,8 @@ name: PR preview cleanup
 on:
   pull_request:
     types: [closed]
+    branches-ignore:
+      - master
 
 jobs:
   deploy:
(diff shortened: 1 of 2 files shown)


.github/workflows/preview-cleanup.yml
```diff
@@ -3,6 +3,8 @@ name: PR preview cleanup
 on:
   pull_request:
     types: [closed]
+    branches-ignore:
+      - master
```
thuey, line 7: Would it make sense to just whitelist this for develop?
```suggestion
    branches:
      - develop
```

.github/workflows/preview.yml
```diff
@@ -3,6 +3,8 @@ name: PR preview
 on:
   pull_request:
     types: [opened]
+    branches-ignore:
+      - master
```
thuey, line 7: (see above)
```suggestion
    branches:
      - develop
```

.github/workflows/preview-cleanup.yml
thuey: Makes sense

thuey (approved):

PR #1264, opened 2023-10-31
feat(Combobox): add custom formValue prop to ComboboxOption

### LINKED ISSUE #1184: Combobox Hidden Form Values

scurker:
closes #1184

### DIFF
--- a/packages/react/src/components/Combobox/ComboboxOption.tsx
+++ b/packages/react/src/components/Combobox/ComboboxOption.tsx
@@ -14,6 +14,7 @@ export type ComboboxValue = Exclude<ListboxValue, number>;
 interface ComboboxOptionProps extends React.HTMLAttributes<HTMLLIElement> {
   disabled?: boolean;
   value?: ComboboxValue;
+  formValue?: ComboboxValue;
   description?: ContentNode;
   children: string;
 }
@@ -58,13 +59,15 @@ const ComboboxOption = forwardRef<HTMLLIElement, ComboboxOptionProps>(
       id: propId,
       description,
       value: propValue,
+      formValue,
       ...props
     },
     ref
   ): JSX.Element | null => {
     const [id] = propId ? [propId] : useId(1, 'combobox-option');
     const { selected, active } = useListboxContext();
-    const { matches, setMatchingOptions } = useComboboxContext();
+    const { selectedValue, matches, setMatchingOptions, setFormValue } =
+      useComboboxContext();
     const comboboxOptionRef = useSharedRef<HTMLElement>(ref);
     const intersectionRef = useIntersectionRef<HTMLElement>(comboboxOptionRef, {
       root: null,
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Combobox/ComboboxOption.tsx
```diff
@@ -100,6 +103,19 @@ const ComboboxOption = forwardRef<HTMLLIElement, ComboboxOptionProps>(
...
+        typeof propValue !== 'undefined'
+          ? propValue
+          : comboboxOptionRef.current?.innerText;
+
+      if (selectedValue === comboboxValue) {
+        setFormValue(
+          typeof formValue === 'undefined' ? comboboxValue : formValue
+        );
+      }
+    }, [selectedValue, formValue]);
```
thuey, line 117: Should we also watch for changes to `propValue`?
```suggestion
    }, [selectedValue, formValue, propValue]);
```

Although I realize this will only cover when the prop value changes and not the innerText :/

scurker: Huh, that's an interesting question.

The setting of the form value should really only change when the selected value changes, but I'm now realizing that `formValue` may be unintended. If I select a value, then somewhere in code the value of that field changes I would think the value of the input shouldn't change but let me play around with native inputs to see what the default behavior is in react as we'd likely want to match that.

scurker: I tested using a native `select` input and updating the values via a `setTimeout` independent of of a user selection.

It does appear that native selects will update the value of the select when an option value changes, even if `<select value={...}>` is set to a no longer valid value, so we probably want to try to mimic that behavior here.

This is probably a bit of an extreme edge case as I'm trying to imagine how often this might pop up. The only scenario where it might be feasible is if the options were returned asynchronously via a fetch request and the same option element had a different value/innerText for the same key. I'm inclined to try to cover what we can here and see if `innerText` does present itself as a problem later on.

scurker: I think it's also good to keep this in mind: this hook should trigger again upon a selection change, so if a new selection is made the value should re-sync.

thuey (approved):

PR #1267, opened 2023-11-02
chore: increate max workers for jest

scurker:


scurker:
Doesn't appear to be any performance gains in ci ¯\\\_(ツ)_/¯

PR #1269, opened 2023-11-03
ci: convert circle jobs to github actions

scurker:
This change will migrate us away from using circle and using github actions instead. There was already a mix of circle/github and this just keeps everything in one spot.

It is currently expected that Circle will be failing on this PR and that the required checks will not run. Once this PR is merged, we will update the branches to require the new github actions checks and remove any integrations with circle. All of the github actions that do run will need to pass on this PR before merging.


.github/workflows/create-release-pr.yml
```diff
@@ -10,9 +10,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 18
```
thuey, line 15: mostly just curious, but why did you remove this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,76 @@
...
+          PACKAGE_VERSION="$(npm pkg get version | tr -d \")"
+          npm version "$PACKAGE_VERSION-canary.${GITHUB_SHA:0:8}" --allow-same-version --no-git-tag-version
+          npm publish --tag=next
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+      - name: Publish @deque/cauldron-react
+        run: |
+          cd packages/react
+          PACKAGE_VERSION="$(npm pkg get version | tr -d \")"
+          npm version "$PACKAGE_VERSION-canary.${GITHUB_SHA:0:8}" --allow-same-version --no-git-tag-version
```
thuey, line 43: Should this be in the `canary` job instead of the `publish` job?

scurker: Whoops, good catch! 🤦 

Looks like I swapped them around accidentally.

.github/workflows/create-release-pr.yml
scurker: Got a little too aggressive 🗡️

thuey (approved):

PR #1270, opened 2023-11-03
chore(dependabot): monthly updates for actions

anastasialanz:
Since many of the GitHub Actions versions in this repo are older, this PR adds a Dependabot config for updating dependencies for GitHub Actions similar to how it's done in `axe-core`.

https://github.com/dequelabs/axe-core/blob/develop/.github/dependabot.yml#L3-L16

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10
+    commit-message:
+      prefix: 'chore'
+    groups:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
+version: 2
+
+updates:
+  - package-ecosystem: "github-actions"
```
scurker, line 4: Don't we need `npm` here as well? Seems like this would skip updates to npm packages.

anastasialanz: I didn't know if you wanted to update packages so I started with actions. 😄

anastasialanz: I can add npm packages to this if you would like or make a separate PR for it.

scurker (approved):

Issue #358, opened 2021-10-04
Sync Alert modals with UXpin


scurker:
## Light Mode

[uxpin](https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161666306/simulate/no-panels?mode=i)

![light mode alert modal info variant](https://github.com/dequelabs/cauldron/assets/1062039/fd5cb421-e47e-4c7f-be48-92b2a9afb8e2)

![light mode alert modal warning variant](https://github.com/dequelabs/cauldron/assets/1062039/725c3ffd-0ba7-4983-966b-cb1f19c8d11c)

## Dark Mode

![dark mode alert modal info variant](https://github.com/dequelabs/cauldron/assets/1062039/e001f5b3-cd12-45f9-aca1-e8df41c3ad64)

![dark mode alert modal warning variant](https://github.com/dequelabs/cauldron/assets/1062039/e80d4bd9-4da6-4f0b-8011-56038f7b4c96)

scurker:
@bobbyomari Is this still out of line with cauldron?

bobbyomari:
There are 3 types of modals and should be styled as such: https://preview.uxpin.com/4052c8b6f05ce03e28813f28deb3a682950c2ac6#/pages/140808127

In this case, the OP example was for a destructive action so it should use the Error Modal example instead of a standard modal.

scurker:
These are specifically the alert modals (which are a variation of dialog): https://cauldron.dequelabs.com/components/Alert

scurker:
fixed by #1275

PR #1275, opened 2023-11-08
feat(react): sync alert modals with uxpin

chornonoh-vova:
Implementation of the #358


packages/react/__tests__/src/components/Alert/index.js
```diff
@@ -27,3 +27,12 @@ test('should return no axe violations', async () => {
   );
   expect(await axe(alert.html())).toHaveNoViolations();
 });
+
+test('should return no axe violations warning variant', async () => {
+  const alert = mount(
+    <Alert show={true} variant={'warning'} heading={'title'}>
```
scurker, line 33: Just a couple of nits:

```suggestion
    <Alert show={true} variant="warning" heading="title">
```

packages/styles/alert.css
```diff
@@ -1,22 +1,29 @@
 :root {
-  --alert-warning-background-color: #d93251;
-  --alert-warning-header-text-color: #fff;
+  --alert-default-color: #3c7aae;
+  --alert-warning-color: #d93251;
+
+  --alert-default-color-dark: #6cdaf2;
+  --alert-warning-color-dark: #fe6d6b;
```
scurker, line 6: This doesn't follow our theming guidelines, see: https://cauldron.dequelabs.com/style-guidelines#theming

The correct way to do this is to style the "default" values on `:root` and have an additional selector to match the more specific styles. Something like:

```css
:root {
  --alert-border-color: ...
  --alert-warning-border-color: ...
}

.cauldron-theme-dark {
  --alert-border-color: ...
  --alert-warning-border-color: ...
}
```

Additionally, these should be pulling from global [`variables.css`](https://github.com/dequelabs/cauldron/blob/develop/packages/styles/variables.css) colors, example below:

```css
--alert-border-color: var(--accent-primary)
```

Additionally, removal of existing css variables would constitute a breaking change per our [contributing guidelines](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#breaking-changes). We can use new variables that better fit the new component, but need to keep the existing ones around to accommodate a deprecation period.

scurker:
I just closed / re-opened to re-trigger the new preview check.

packages/react/src/components/Alert/index.tsx
```diff
@@ -18,13 +18,14 @@ const Alert = ({
   <Dialog
     className={classnames(
       { Alert__warning: variant === 'warning' },
+      { Alert__default: variant === 'default' },
```
scurker, line 21: We can treat the default variant as the variant without a class, so we can remove this line.

scurker (changes requested):
The actual component looks spot on, just need to make a few code tweaks to be consistent with how we piece together our other components.

packages/styles/alert.css
```diff
@@ -25,8 +32,26 @@
   display: flex;
   justify-content: center;
   background-color: transparent;
+  padding: 0 var(--space-large) var(--space-large);
```
scurker, line 35: I checked with @dequelabs/cauldron-design on this and this can inherit from `--dialog-padding`.

packages/react/__tests__/src/components/Alert/index.js
chornonoh-vova: Thank you for feedback, updated

packages/styles/alert.css
chornonoh-vova: Thank you for highlighting this, I've added previous colors back and updated colors according to your suggestions. I've separated the colors for borders and icons, even though they are identical. Not sure if that's the right way to go, but I thought it might help if these colors change in the future in any way. If that's not acceptable, I'll change it back to use one variable.

chornonoh-vova: Also, I needed to make a couple of changes to dialog styles, where I saw these theming guidelines weren't followed fully :) It was preventing my changes from taking effect in dark mode.

scurker: > Not sure if that's the right way to go, but I thought it might help if these colors change in the future in any way. If that's not acceptable, I'll change it back to use one variable.

I could go either way, but what you have is fine as it provides more granularity.

scurker (approved):

PR #1286, opened 2023-11-09
feat(icons): add step forward/back icons

qazwsxedcrfvtgb1111:
This pr adds step forward and back icons


scurker (changes requested):
I think we might have to change the name here as `-(left|right)` postfixed names have a special meaning that determines the direction icons are facing: https://github.com/dequelabs/cauldron/blob/e30d9f6a47d3798d0b2f5c14ea9e3c50b8f3430d/packages/react/src/components/Icon/index.tsx#L17-L20

We do this as a measure to cut down on the amount of icons that we have to ship.

We might want something like "step-forward" and "step-back".

scurker (approved):

Issue #1293, opened 2023-11-17
Bug: required TextField components announces required twice


JoshMK:
### Steps to reproduce

1. Open VoiceOver and tab to a TextField component with required set

### Expected Behavior

Required is only announced once.

### Actual Behavior

Required is announced twice.

### Version

5.12.0

### Anything else we should know?

<img width="1102" alt="Required being announced twice on the TextField component in VoiceOver" src="https://github.com/dequelabs/cauldron/assets/6422248/d1f52b6a-dd52-4e66-af49-92cff2177bb6">

scurker:
I'm hesitant to make this change as the _required_ text exists in the label for support of older screen readers. I haven't seen our own guidelines suggest that this as an issue, so I'm going to defer to our experts here to determine if this is necessary.

scurker:
For now, we'll leave this as is and won't fix. If the need arrises where the design changes or we want to be more comprehensive with form instructions we can update this.

PR #1294, opened 2023-11-17
fix(react): announce required once through attribute only

JoshMK:
- uses the `required` attribute to announce 'required' to screen readers while still displaying 'required' for sighted users.

required being announced on TextField component once to screenreader users while being displayed to sighted users

Closes: #1293


JoshMK:
Closing this out pending further discussion of the best way to handle cases where `requiredText` is customized by the user.

Issue #1175, opened 2023-08-22
Sync IconButton with UXPin


scurker:
There's a small number of stylistic changes for the IconButton that are needed to sync up with our updated designs for dark mode only:

![icon button in different states shown in uxpin](https://github.com/dequelabs/cauldron/assets/1062039/5f8ff410-e329-42fd-829e-ab12a64f1164)

[UXPin](https://preview.uxpin.com/520e80315786a67ba43a3ca50598e6d69782627c#/pages/161666306/simulate/no-panels?mode=i)

Related issues:
- #306 
- #969

Issue #969, opened 2023-03-21
Need "large" variant for IconButton 


scurker:
![primary icon button large](https://user-images.githubusercontent.com/1062039/226709723-32654bd3-1ec6-43a9-9365-cf647f15a0c7.png)

https://preview.uxpin.com/design-system/c933c74928e2195cc540/6764127,light-button-regular-icon-primary

This should implement all the other variant styles of buttons. We need to set the modifier to be `large` to not introduce a breaking change.

scurker:
This is done

Issue #938, opened 2023-03-09
disabled/aria-disabled buttons can have an active state


scurker:
We probably don't want aria-disabled/disabled buttons to have a different background color in their active state than their resting state. You can see this on the [Button](https://cauldron.dequelabs.com/components/Button) page by activating a disabled button and see the background color changing. This also appears to be happening for [IconButton](https://cauldron.dequelabs.com/components/IconButton) as well.

PR #1304, opened 2023-12-27
fix(style): Synced styles and added large variant of IconButton

orest-s:
Closes: #1175 and #969.
Also PR closes partly (only for icon-button): #938

### DIFF
--- a/packages/styles/icon-button.css
+++ b/packages/styles/icon-button.css
@@ -1,15 +1,4 @@
 :root {
-  --icon-button-content-color-light-disabled: #ccc;
-  --icon-button-background-color-light: #dadada;
-  --icon-button-background-color-light-active: #c2c2c2;
-  --icon-button-outline-color-light: var(--icon-button-background-color-light);
-
-  --icon-button-content-color-dark-disabled: #fff 40%;
-  --icon-button-background-color-dark: var(--accent-medium);
-  --icon-button-background-color-dark-disabled: var(--accent-medium);
-  --icon-button-background-color-dark-active: var(--accent-dark);
-  --icon-button-outline-color-dark: var(--icon-button-background-color-dark);
-
   /* Primary */
   --icon-button-primary-color: var(--button-text-color-light);
   --icon-button-background-color-primary: var(
@@ -20,7 +9,7 @@
   );
   --icon-button-outline-color-primary: var(--button-outline-color-primary);
 
-  --icon-button-large-size: 36px;
+  --icon-button-large-height: var(--button-height);
 
   /* Secondary */
   --icon-button-secondary-color: var(--button-text-color-dark);
@@ -36,6 +25,27 @@
   /* Error */
   --icon-button-background-color-error: var(--button-background-color-error);
   --icon-button-outline-color-error: var(--button-outline-color-error);
+  --icon-button-background-color-error-active: var(
+    --button-background-color-error-active
+  );
+
+  /* Deprecated variables */
+  --icon-button-content-color-light-disabled: #ccc;
+  --icon-button-background-color-light: #dadada;
+  --icon-button-background-color-light-active: #c2c2c2;
+  --icon-button-outline-color-light: var(--icon-button-background-color-light);
+
+  --icon-button-content-color-dark-disabled: #fff 40%;
+  --icon-button-background-color-dark: var(--accent-medium);
+  --icon-button-background-color-dark-disabled: var(--accent-medium);
+  --icon-button-background-color-dark-active: var(--accent-dark);
+  --icon-button-outline-color-dark: var(--icon-button-background-color-dark);
+  --icon-button-background-color-secondary-disabled: var(
+    --button-background-color-secondary-disabled
+  );
+  --icon-button-background-color-error-disabled: var(
+    --button-background-color-error-disabled
+  );
 }
 
 .IconButton {
@@ -87,8 +97,8 @@ a.IconButton {
 }
 
 .IconButton--large {
-  height: var(--icon-button-large-size);
-  width: var(--icon-button-large-size);
+  height: var(--icon-button-large-height);
+  width: var(--icon-button-large-height);
 }
 
 .IconButton--secondary[aria-disabled='true'],
@@ -154,7 +164,7 @@ a.IconButton {
 }
 
 .IconButton--error:active {
-  background-color: var(--button-background-color-error-active);
+  background-color: var(--icon-button-background-color-error-active);
 }
 
 /* Dark Theme */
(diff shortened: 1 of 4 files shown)


packages/styles/icon-button.css
```diff
@@ -1,46 +1,41 @@
...
+  --icon-button-content-color-light-disabled: #ccc;
+  --icon-button-background-color-light: #dadada;
+  --icon-button-background-color-light-active: #c2c2c2;
+  --icon-button-outline-color-light: var(--icon-button-background-color-light);
+
+  --icon-button-content-color-dark-disabled: #fff 40%;
+  --icon-button-background-color-dark: var(--accent-medium);
+  --icon-button-background-color-dark-disabled: var(--accent-medium);
+  --icon-button-background-color-dark-active: var(--accent-dark);
+  --icon-button-outline-color-dark: var(--icon-button-background-color-dark);
```
scurker, line 11: I can't find these variables used anywhere. We should [mark these as deprecated](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#styles-1), and add a ticket to remove these in the future as a breaking change.

Since these should also not be used, they should be positioned back towards the bottom of `:root`.

packages/styles/icon-button.css
```diff
@@ -1,46 +1,41 @@
...
+  --icon-button-large-size: 36px;
+
+  /* Secondary */
+  --icon-button-secondary-color: var(--button-text-color-dark);
   --icon-button-background-color-secondary: var(
     --button-background-color-secondary
   );
-  --icon-button-background-color-secondary-disabled: var(
-    --button-background-color-secondary-disabled
-  );
```
scurker, line 18: This can probably be omitted, but we can't remove it without deprecation first (see related comment).

packages/styles/icon-button.css
```diff
@@ -1,46 +1,41 @@
...
   --icon-button-background-color-secondary-active: var(
     --button-background-color-secondary-active
   );
   --icon-button-outline-color-secondary: var(--button-outline-color-secondary);
   --icon-button-border-color-secondary: var(--gray-40);
 
+  /* Error */
   --icon-button-background-color-error: var(--button-background-color-error);
-  --icon-button-background-color-error-disabled: var(
```
scurker, line 26: nding
This can probably be omitted, but we can't remove it without deprecation first (see related comment).

packages/styles/icon-button.css
```diff
@@ -1,46 +1,41 @@
...
-    --button-background-color-error-disabled
-  );
-  --icon-button-background-color-error-active: var(
```
scurker, line 29: Did you mean to remove this? This looks correct.

packages/styles/icon-button.css
```diff
@@ -152,61 +154,83 @@ a.IconButton {
 }
 
 .IconButton--error:active {
-  background-color: var(--icon-button-background-color-error-active);
+  background-color: var(--button-background-color-error-active);
```
scurker, line 157: This should not be referencing button variables directly, but icon variables:

```suggestion
  background-color: var(--icon-button-background-color-error-active);
```

packages/styles/icon-button.css
```diff
@@ -152,61 +154,83 @@ a.IconButton {
...
 
-.cauldron--theme-dark .IconButton--primary:active {
-  background-color: #e5ebec;
+.cauldron--theme-dark .IconButton--primary[aria-disabled='true'],
+.cauldron--theme-dark .IconButton--primary[disabled],
+.cauldron--theme-dark .IconButton--secondary[aria-disabled='true'],
+.cauldron--theme-dark .IconButton--secondary[disabled],
+.cauldron--theme-dark .IconButton--error[aria-disabled='true'],
+.cauldron--theme-dark .IconButton--error[disabled] {
+  color: var(--dark-workspace-color);
```
scurker, line 202: This should not referencing a global variable directly, but an aliased color. Something like `--icon-button-disabled-color`.

packages/styles/icon-button.css
```diff
@@ -152,61 +154,83 @@ a.IconButton {
...
 }
 
-.cauldron--theme-dark .IconButton--primary:active {
-  background-color: #e5ebec;
+.cauldron--theme-dark .IconButton--primary[aria-disabled='true'],
+.cauldron--theme-dark .IconButton--primary[disabled],
+.cauldron--theme-dark .IconButton--secondary[aria-disabled='true'],
+.cauldron--theme-dark .IconButton--secondary[disabled],
+.cauldron--theme-dark .IconButton--error[aria-disabled='true'],
+.cauldron--theme-dark .IconButton--error[disabled] {
```
scurker, line 201: Small nit:

```suggestion
.cauldron--theme-dark .IconButton--primary:is([aria-disabled='true'], [disabled]),
.cauldron--theme-dark .IconButton--secondary:is([aria-disabled='true'], [disabled]),
.cauldron--theme-dark .IconButton--error:is([aria-disabled='true'], [disabled]) {
```

scurker (changes requested):
It also appears that the disabled active state has a background color (see https://github.com/dequelabs/cauldron/issues/938) which we likely want to fix here. We can leave #938 open to address the standard buttons at a later time.

packages/styles/icon-button.css
```diff
@@ -1,46 +1,41 @@
...
   );
-  --icon-button-background-color-primary-disabled: var(
-    --button-background-color-primary-disabled
-  );
   --icon-button-background-color-primary-active: var(
     --button-background-color-primary-active
   );
   --icon-button-outline-color-primary: var(--button-outline-color-primary);
 
+  --icon-button-large-size: 36px;
```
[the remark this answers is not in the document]
orest-s: Does this mean the height of the button is equal to the height of the large icon button?
We want to have exactly the same size for button and for large variant of icon-button, right?

packages/styles/icon-button.css
```diff
@@ -152,61 +154,83 @@ a.IconButton {
...
+
+.cauldron--theme-dark
+  .IconButton--error:not([disabled]):not([aria-disabled='true']):hover:before {
+  box-shadow: 0px 0px 0px 1px var(--icon-button-outline-shadow-color),
+    0px 0px 0px 2px var(--icon-button-outline-color-error);
+}
+
+.cauldron--theme-dark .IconButton:focus:before {
+  box-shadow: 0px 0px 0px 2px var(--icon-button-outline-shadow-color),
+    0px 0px 0px 4px var(--button-focus-ring-color, --focus);
```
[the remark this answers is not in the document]
orest-s: Sorry here and above, forgot to check, just copied it from uxpin and didn't delete px :)

packages/styles/icon-button.css
orest-s: yeah, you are right, just a mistake.

packages/styles/icon-button.css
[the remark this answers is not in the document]
scurker: Correct.

scurker (commented):
Code LGTM! Just want to get a 👍 from @dequelabs/cauldron-design

shawnsharpDQ (approved):
Design gives 2 👍 up!

scurker (approved):

Issue #915, opened 2023-02-22
Add additional progress bar variant


scurker:
[TBD]

bobbyomari:
We've added light and dark theme styles for progress bars - we will not be using "flat" style progress bars and instead keep the version with a stroke for better accessibility.

Component in design system: https://app.uxpin.com/design-system/c933c74928e2195cc540/ui-patterns/213784

**Some other details:**
- Thick is 24px in height
- thin is 16px in height
- Author can change the color as needed, these are default styles
- Padding around inner bar is 4px
- Set a margin-bottom only of 8px (no margin-top)

**Inspect Components:**
- Light / Thick: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6770639,light-progress-bar-thick
- Light / Thin: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6770640,light-progress-bar-thin
- Dark / Thick: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6770641,dark-progress-bar-thick
- Dark / Thin: https://preview.uxpin.com/design-system/c933c74928e2195cc540/6770642,dark-progress-bar-thin

PR #1309, opened 2024-01-04
feat(react,styles): synced styles for progress bar and added thin variant

orest-s:
Closes: #915

### DIFF
--- a/docs/pages/components/ProgressBar.mdx
+++ b/docs/pages/components/ProgressBar.mdx
@@ -17,6 +17,14 @@ import { ProgressBar } from '@deque/cauldron-react'
 <ProgressBar
   aria-label="Progress %"
   progress={75}
+  thin
+/>
+```
+
+```jsx example
+<ProgressBar
+  aria-label="Progress %"
+  progress={75}
 />
 ```
 
@@ -121,5 +129,11 @@ function ProgressExample() {
       type: 'string',
       description: 'A label for the ProgressBar is required. This means you must provide either an aria-label or aria-labelledby prop.'
     },
+    {
+      name: 'thin',
+      type: 'boolean',
+      defaultValue: 'false',
+      description: 'Thinner visual style of icon button.'
+    }
   ]}
 />
--- a/docs/pages/components/ProgressBar.mdx
+++ b/docs/pages/components/ProgressBar.mdx
@@ -13,6 +13,8 @@ import { ProgressBar } from '@deque/cauldron-react'
 
 ## Examples
 
+### Thin
+
 ```jsx example
 <ProgressBar
   aria-label="Progress %"
@@ -20,6 +22,8 @@ import { ProgressBar } from '@deque/cauldron-react'
   thin
 />
 ```
+
+### Normal
 
 ```jsx example
 <ProgressBar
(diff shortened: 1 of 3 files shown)


packages/react/src/components/ProgressBar/index.tsx
```diff
@@ -6,15 +6,22 @@ type ProgressBarProps = {
...
+    ref
+  ) => {
     const { className, ...otherProps } = props;
     return (
       <div
-        className={classnames('ProgressBar', className)}
+        className={classnames(className, {
+          ProgressBar: true,
+          'ProgressBar--thin': thin
+        })}
```
scurker, line 24: Nit: `classnames` allows for an array of arguments:

```suggestion
        className={classnames(className, 'Progressbar', {
          'ProgressBar--thin': thin
        })}
```

packages/styles/progressbar.css
```diff
@@ -1,13 +1,15 @@
...
+  --progress-bar-background-color: var(--white);
+  --progress-bar-fill-color: var(--accent-primary);
+  --progress-bar-border-color: var(--gray-40);
   --progress-bar-animation-timing: 150ms;
 }
 
 .ProgressBar {
   background-color: var(--progress-bar-background-color);
-  padding: 2px;
+  padding: var(--space-half);
```
scurker, line 10: This is an instance where we don't really want spacing to scale with the global padding variable. The padding in this instance is acting like a border, which we do want to have a fixed size. Additionally, I think the padding is wrong here. The thin/normal variants have a `16px` and `24px` footprint respectively, so this appears to be `3px` from the design.

docs/pages/components/ProgressBar.mdx
```diff
@@ -13,6 +13,14 @@ import { ProgressBar } from '@deque/cauldron-react'
 
 ## Examples
 
+```jsx example
```
scurker, line 16: With multiple examples here, we probably want each example under a level 3 heading.

packages/styles/progressbar.css
```diff
@@ -18,3 +20,16 @@
   min-width: var(--space-small);
   transition: width linear var(--progress-bar-animation-timing);
 }
+
+.ProgressBar--thin .ProgressBar--fill {
+  height: var(--space-smallest);
```
scurker, line 25: Same issue as above comment. This should be a fixed size and not inheriting from `--space-smallest`.

packages/styles/progressbar.css
```diff
@@ -18,3 +20,16 @@
...
+}
+
+.cauldron--theme-dark {
+  --progress-bar-background-color: var(--accent-dark);
+  --progress-bar-fill-color: var(--accent-info-light);
+}
+
+.cauldron--theme-dark .ProgressBar {
+  border: none;
```
scurker, line 34: This gives the dark themed progress bar a different footprint. It would be better to update `--progress-bar-border-color` to match the background color for dark theme.

packages/styles/progressbar.css
```diff
@@ -18,3 +20,16 @@
...
 }
+
+.ProgressBar--thin .ProgressBar--fill {
+  height: var(--space-smallest);
+}
+
+.cauldron--theme-dark {
+  --progress-bar-background-color: var(--accent-dark);
+  --progress-bar-fill-color: var(--accent-info-light);
+}
```
scurker, line 31: This should be immediately below `:root`.

scurker (changes requested):
This is looking really good! Just a few minor changes needed.

Issue #856, opened 2023-01-18
Update accordion to match UXPin


bobbyomari:
The current accordion in Cauldron is outdated and needs to be updated to the latest iteration of accordion styles with two versions - basic and advanced. We no longer use the blue bar as a hover state and has been updated as such to be used as an active state instead (when the accordion item is open). There are also some changes to the icon used, box-shadow (only applied when accordion group is open), and spacing. 

Light theme: https://preview.uxpin.com/68df515d0023cc4b79f25d43f7663e24e363b1b9#/pages/150823418
Dark theme: https://preview.uxpin.com/4052c8b6f05ce03e28813f28deb3a682950c2ac6#/pages/150951058

scurker:
Related: #716 

Update styles and add "advanced variant of accordion". The advanced variant of the accordion would need to use `ExpandCollapsePanel` under the hood and match the styles of the accordion in order to support the action buttons on the right side of the header.

scurker:
Also related #902

scurker:
Regarding the spacing of the component, the trigger text should be aligned to the left the same as the content text should be. I think we just need to adjust the left padding of the trigger element.

Issue #869, opened 2023-01-25
PROPOSAL: Consolidate icons into a single library and map old icons to new icons


bobbyomari:
### The Problem
Today, our designs are using multiple icon libraries – this is creating inconsistencies between icons throughout our products where icons might be too thin, too thick, too small, too large, etc. We are mixing libraries of UXPin, Font Awesome, Material UI, and none of which were made to work together. 

### Proposal
Let's consolidate the icons to use Font Awesome 6 Pro so we can take advantage of their full repository of icons that are all accessible and create consistency between our iconography library. I am suggesting Font Awesome because they have a very extensive and allow for a lot of parameters to be passed such as icon rotation, icon stacking, light/solid/thin/duotone icons, and much more. Additionally, if there are unique icons that are not part of FA6, we could upload custom icons as well as they have a method for that in their [documentation](https://fontawesome.com/docs/web/add-icons/upload-icons/). For more information about Font Awesome 6 visit: https://fontawesome.com/plans

### Need to do
If we can confirm that it is OK to switch, then I can work on selecting icons that will map from old to new and include them in this ticket.

scurker:
When this gets initially implemented, we don't want to change the name yet but save that for a later breaking change. We may want to consider alias for the time being to support both names until we remove support for the old name.

scurker:
Whatever we decide on, we should document the actual guidelines we expect icons to follow. Some examples below:

- https://primer.style/foundations/icons/design-guidelines
- https://m3.material.io/styles/icons/designing-icons
- https://developer.apple.com/design/human-interface-guidelines/icons

I don't think we need to be as verbose as what's above, but namely document the grid + padding we're using as well as general style to ensure new icons are consistent with the guidelines. If we're using font awesome, we could just specify that it must originate from there and list whatever guidelines we want to use.

anastasialanz:
Font Awesome has a Setup with React page https://docs.fontawesome.com/web/use-with/react/

PR #1313, opened 2024-01-10
fix(style): basic accordion uxpin sync

orest-s:
Closes: #856


packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -84,7 +84,7 @@ const Accordion = ({
   }
 
   return (
-    <div className="Accordion" {...props}>
+    <div className={'Accordion'} {...props}>
```
scurker, line 87: Nit:

```suggestion
    <div className="Accordion" {...props}>
```

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -93,8 +93,8 @@ const Accordion = ({
         {...panelElement.props}
       >
         <PanelTrigger
-          iconCollapsed="triangle-right"
-          iconExpanded="triangle-down"
+          iconCollapsed="triangle-solid-right"
+          iconExpanded="triangle-solid-down"
```
scurker, line 97: Need to check with @dequelabs/cauldron-design on this, I'm not sure we need 2 different styles of triangle icons. It's possible that it's intended for this triangle to replace the existing one.

packages/styles/accordion.css
```diff
@@ -3,67 +3,74 @@
   --accordion-trigger-background-color-expanded: var(--white);
   --accordion-trigger-text-color: var(--gray-90);
   --accordion-trigger-text-color-hover: var(--accent-primary);
-  --accordion-trigger-box-shadow-hover: var(--accent-primary);
```
scurker, line 6: This can't be removed without it being a breaking change.

packages/styles/accordion.css
```diff
@@ -3,67 +3,74 @@
...
   --accordion-panel-border-color: var(--gray-40);
   --accordion-box-shadow-color: rgba(0, 0, 0, 0.15);
 }
 
 .cauldron--theme-dark {
   --accordion-trigger-background-color: var(--accent-medium);
-  --accordion-trigger-background-color-expanded: var(--accent-dark);
-  --accordion-trigger-text-color: var(--white);
-  --accordion-trigger-text-color-hover: var(--accent-light);
-  --accordion-trigger-box-shadow-hover: var(--accent-info);
```
scurker, line 20: This can't be removed without it being a breaking change.

packages/styles/accordion.css
```diff
@@ -3,67 +3,74 @@
...
-
 .Accordion .Icon {
   transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
   transition-duration: 0.2s;
+  color: var(--accordion-trigger-icon-color);
+}
+
+.Accordion .Icon svg {
+  height: var(--space-small);
+  width: var(--space-small);
```
scurker, line 73: Spacing variables should not be used for height/width.

packages/styles/accordion.css
```diff
@@ -3,67 +3,74 @@
...
   --accordion-trigger-border-color: var(--stroke-dark);
   --accordion-trigger-icon-color: var(--white);
   --accordion-panel-background-color: var(--accent-medium);
-  --accordion-panel-text-color: var(--white);
+  --accordion-panel-text-color: var(--accent-light);
   --accordion-panel-border-color: var(--stroke-dark);
 }
 
+.Accordion:has(.Accordion__trigger[aria-expanded='true']) {
+  filter: drop-shadow(0px 2px 2px rgba(0, 0, 0, 0.15));
```
scurker, line 29: Why `filter: drop-shadow`?

scurker (changes requested):

packages/styles/accordion.css
scurker: This should be using `box-shadow` and not `filter`. `box-shadow` is more widely supported, and the css properties aren't always a 1:1. I'm guessing this should be:

```css
box-shadow: 0 2px 2px rgba(0, 0, 0, 0.15);
```

packages/react/src/components/Accordion/Accordion.tsx
scurker: For now, keep the existing icon. There will be a bunch of work in the future to address the greater body of work where the icons don't match up with the design system (https://github.com/dequelabs/cauldron/issues/869)

packages/styles/accordion.css
scurker: These variables are in 2 different scopes, `:root: ` and `.cauldron--theme-dark`. If someone was depending on that variable existing within the `:root` scope, it would no longer have a value.

scurker (changes requested):

packages/styles/accordion.css
scurker: I realize that this variable is below, dismissed

packages/styles/accordion.css
scurker: I realize Github UI was being funky and showing this as changed when it got added below, dismissed.

scurker (approved):

PR #1316, opened 2024-01-10
ci: generate screenshots via playwright for visual component testing

scurker:
This uses Playwright to generate screenshot snapshots for various components in different states. This PR primarily focuses on interactive components and doesn't establish full visual regression coverage (for now).

- Captures screenshots for rendered elements in both light/dark mode
- Groups different component states together
- Runs tests in CI to ensure there are no visual regressions for saved snapshots

It's currently somewhat difficult to determine when a component visually changes so this should help to provide some automation to actually know when there's a visual change - intended or not intended. In the future, we may want to establish best practices and guidelines for visually testing all components, but this should provide a template to jump start from.


packages/react/src/components/TextField/index.tsx
```diff
@@ -72,6 +72,7 @@ export default class TextField extends React.Component<
       requiredText,
       multiline,
       'aria-describedby': ariaDescribedby,
+      children,
```
scurker, line 75: Textfield shouldn't have children and vite was real angry when this was not defined.

e2e/screenshots.test.tsx
thuey, line 1: nit: It seems like this file is not going to scale well as we add more components. I realize this is probably just an MVP stage, but I'm curious if you have any thoughts on how these screenshot tests can be structured as we add more components. Also curious if there is any guidance on when/what to do screenshot testing.

It might be nice if we could get these screenshot tests working for our docs site, but that might also be too much.

thuey (approved):
Nice work!

e2e/screenshots.test.tsx
scurker: > It seems like this file is not going to scale well as we add more components. I realize this is probably just an MVP stage, but I'm curious if you have any thoughts on how these screenshot tests can be structured as we add more components. 

I thought about this for a bit; I agree a huge monolithic file isn't ideal. Ideally this would go into existing unit tests but this has a different test runner (playright-ct vs jest). The other alternative that comes to mind is having a screenshot specific test file for each component, but we'd have to filter those files for the playwright tests, then ignore them for jest. I'm okay cleaning that up here if we feel that's more scalable.

> Also curious if there is any guidance on when/what to do screenshot testing.

Not yet. I want to get this in place before establishing rules/guidelines to try and iron out any kinks before we establish the set of requirements. But generally speaking, the practice would be similar to the [testing strategies](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#testing-strategies) outlined in the contributing guidelines of including all interactive states + variants.

> It might be nice if we could get these screenshot tests working for our docs site, but that might also be too much.

Can you expand what you mean by this?

thuey: > Can you expand what you mean by this?

Right now, you are doing component-based testing, which is great since it isolates the component. However, it does mean we have to write and maintain an extra set of test cases. Integrating this with unit testing somehow might be a good compromise

What I was referring to was basically doing page screenshot testing on the cauldrons docs site, since it already has example components written. Then it would just be a matter of traversing through the component pages in the cauldron docs. But the scope of the screenshot might be too large and we would want to target specific areas anyway.

scurker: > But the scope of the screenshot might be too large and we would want to target specific areas anyway.

This is exactly why we don't want to do screenshot testing of the docs site as there's a couple of flaws with that approach. We might not have all the components in the right combination of interactions, states, and variants, so writing documentation would potentially have two competing goals. Additionally, we would likely want to clip the screenshot as a small change to unrelated components could have unintended impact of snapshots on other components. This would also potentially have side effects, such as modifying text or other properties on a component for documentation would introduce an unnecessary visual change requiring an updated snapshot.

Page screenshot testing would in theory work and while it may potentially be less maintenance, it adds a lot more noise to the process that we would need to filter through. In this case, component testing gives us a lot more confidence that we're only testing visual changes directly related to changes to the component and not changes elsewhere. We can avoid having competing goals and guarantee that we're testing exactly what we want when we want.

scurker: @thuey I made a slight change to move the button screenshot tests out of the monolithic file, is that more in line with your expectations?

thuey: Yep! That approach looks good to me. And I like the idea of collocating the screenshot tests with the component

thuey (approved):

e2e/screenshots.test.tsx
scurker: I'll update the rest of the components hopefully today or later this week.

scurker:
@thuey when you get a chance, would you mind giving this another review?

thuey (approved):

Issue #1159, opened 2023-08-09
Test Migration (React Upgrade / React Testing Library)


scurker:
We need to migrate away from using Enzyme as it doesn't support newer versions of React. This means we need to refactor all of our existing tests to utilize React Testing Library.
```[tasklist]
### Tasks
- [x] Accordion
- [x] Address
- [x] Alert
- [x] Breadcrumb
- [x] Button
- [x] Checkbox
- [x] ClickOutsideListener
- [x] Code
- [x] Combobox
- [x] DescriptionList
- [x] Dialog
- [x] ExpandCollapsePanel
- [x] FieldWrap
- [x] Icon
- [x] IconButton
- [x] Line
- [x] Link
- [x] Loader
- [x] LoaderOverlay
- [x] Listbox
- [x] MenuBar
- [x] MenuItem
- [x] Modal
- [x] NavBar
- [x] NavItem
- [x] Notice
- [x] OptionsMenu
- [x] OptionsMenuItem
- [x] Pagination
- [x] Panel
- [x] Popover
- [x] Progressbar
- [x] RadioCardGroup
- [x] RadioGroup
- [x] Scrim
- [x] Select
- [x] SideBar
- [x] SkipLink
- [x] Stepper
- [x] Table
- [x] Tabs
- [x] Tag
- [x] TagButton
- [x] TextField
- [x] Toast
- [x] Tooltip
- [x] TooltipTabstop
- [x] TopBar
- [x] TwoColumnPanel
- [x] Workspace
- [x] contexts/theme
- [x] utils/aria-isolate
- [x] utils/rndid
- [x] utils/stylesheets
- [x] hooks useIntersectionRef
- [x] hooks useSharedRef
```

scurker:
All tests have been migrated! 🎉

PR #1317, opened 2024-01-18
test(react): convert Combobox tests to react-testing-library

scurker:
ref #1159


packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -219,7 +219,9 @@ const Combobox = forwardRef<HTMLInputElement, ComboboxProps>(
 
     const handleComboboxOptionClick = useCallback(() => {
       // maintain focus on the input
-      inputRef.current?.focus();
+      if (inputRef.current !== document.activeElement) {
+        inputRef.current?.focus();
+      }
```
scurker, line 224: This was previously causing a test to fail for in rtl, but we should only try to focus the element again when the element doesn't already have focus.

packages/react/src/components/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,914 @@
...
+  render(
+    <Combobox label="label">
+      <ComboboxOption>Apple</ComboboxOption>
+      <ComboboxOption>Banana</ComboboxOption>
+      <ComboboxOption>Cantaloupe</ComboboxOption>
+    </Combobox>
+  );
+
+  assertListboxIsOpen(false);
+  fireEvent.click(screen.getByRole('combobox'));
```
thuey, line 273: nit: I don't know what cauldron's standard practices are for this, but it's considered a best RTL practice to use the userEvents library to more closely simulate a real user. See: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event

packages/react/src/components/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,914 @@
...
+  render(
+    <Combobox label="label" portal={element}>
+      <ComboboxOption>Apple</ComboboxOption>
+      <ComboboxOption>Banana</ComboboxOption>
+      <ComboboxOption>Cantaloupe</ComboboxOption>
+    </Combobox>
+  );
+  expect(
+    element.contains(element.querySelector('[role="listbox"]'))
+  ).toBeTruthy();
```
thuey, line 899: You might be looking for [within](https://testing-library.com/docs/dom-testing-library/api-within/) here, which would allow you to use RTL's selectors

packages/react/src/components/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,914 @@
...
+  );
+  const combobox = screen.getByRole('combobox');
+
+  fireEvent.focus(combobox);
+  assertListboxIsOpen(true);
+  fireEvent.click(screen.getAllByRole('option')[1]);
+  fireEvent.blur(combobox);
+  assertListboxIsOpen(false);
+  fireEvent.focus(combobox);
+  expect(screen.getByRole('combobox').getAttribute('value')).toEqual('');
```
thuey, line 545: You might want to consider using `getByDisplayValue`: https://testing-library.com/docs/queries/bydisplayvalue/

packages/react/src/components/Combobox/Combobox.test.tsx
scurker: Yes, I'm aware.

I evaluated using userEvents here, but I think in this instance we do want `fireEvent` since we do individually test many of the different sequences of events. In this particular example, we need to test that the combobox does open on click regardless of focus. userEvents would first focus the element, then fire the click which isn't the exact sequence we want to test here.

packages/react/src/components/Combobox/Combobox.test.tsx
scurker: Is that actually what I want though? I want to assert that the combobox has an empty value.

`screen.getByDisplayValue('')` doesn't quite seem like the right assertion here, unless you were thinking of handling that differently somehow. I also don't necessarily want to make a negative assertion, i.e. `expect(screen.getByDisplayValue('banana')).isFalsy()`

thuey: This one probably isn't the best example. I was more thinking of other places where you do things like this:

```
expect(screen.getByRole('combobox').getAttribute('value')).toEqual('Banana');
```

scurker: I'm not following, and I'm not sure I agree that's the assertion that we want to make.

`screen.getByDisplayValue('Banana')` would fail if no input had that value, but I want a more specific assertion here since there's potentially multiple input fields with the visible + hidden field. I want a tighter assertion because I do think we want to validate that there is a `role="combobox"` field that has the matching field value. `getByDisplayValue` only solves half of that, unless we were to do something like `expect(screen.getByDisplayValue('banana').getAttribute('role')).toEqual('combobox')` which is the inverse of what we have above.

thuey: Gotcha. That makes sense. The (mild) concern I have with using "getAttribute" is that it doesn't necessarily match what the user sees, which is the advantage of using `getByDisplayValue`.

It looks like cauldron is using @testing-library/jest-dom, so an alternative would be to use [toHaveDisplayValue](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohavedisplayvalue):

```ts
expect(screen.getByRole('combobox')).toHaveDisplayValue('Banana');
```

scurker: Yup, that's a fair assessment. I'll make that change.

thuey (approved):

PR #1333, opened 2024-02-03
test: switch tag tests to react testing library

anastasialanz:

This PR switches the `Tag` tests to React Testing Library. It also corrects the paths for `setupFilesAfterEnv` and `collectCoverageFrom` in the `jest.config.ts`.

## Validation steps
- Run `yarn test`


packages/react/src/components/Tag/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+import axe from '../../axe';
+
+test('renders children', () => {
+  render(
+    <Tag>
+      <TagLabel>Label:</TagLabel> value
+    </Tag>
+  );
+  expect(screen.getByText('Label:')).toBeInTheDocument();
+});
```
anastasialanz, line 13: This test seems redundant because the ones below already verify that it renders.

packages/react/jest.config.ts
```diff
@@ -2,13 +2,13 @@ import type { Config } from 'jest';
...
   testEnvironment: 'jsdom',
-  setupFilesAfterEnv: ['<rootDir>__tests__/setupTests.ts'],
+  setupFilesAfterEnv: ['<rootDir>/__tests__/setupTests.ts'],
   testMatch: [
     '**/__tests__/src/**/*.js',
     '**/__tests__/demo/**/*.js',
     '**/src/components/**/*.test.tsx'
   ],
-  collectCoverageFrom: ['src/**/*.{ts.tsx}'],
+  collectCoverageFrom: ['<rootDir>/src/components/**/*.{tsx,js}'],
```
scurker, line 11: Whoops, typo on my part, good catch! 🤦 

This should probably be:

```suggestion
  collectCoverageFrom: ['<rootDir>/src/components/**/*.{tsx,ts}'],
```

Since there are `.ts` files that are not just index files. Are there actually any js files located under components? I couldn't find any.

scurker (commented):
Code changes LGTM, just one question on the coverage change.

packages/react/jest.config.ts
anastasialanz: @scurker I double-checked and didn't find any. Updated!

scurker (approved):

PR #1334, opened 2024-02-05
test: convert icon tests to react testing library

anastasialanz:

This PR switches the `Icon` tests to React Testing Library.

## Validation steps
- Run `yarn test` to see the tests pass


packages/react/src/components/Icon/index.test.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { createRef } from 'react';
+import { render } from '@testing-library/react';
+import Icon from './index';
```
scurker, line 3: Very very minor nit.

```suggestion
import Icon from './';
```

scurker (changes requested):
Sorry I missed this somehow! I think we just need to add a test to run the component through jest-axe, then we should be good.

anastasialanz:
> Sorry I missed this somehow! I think we just need to add a test to run the component through jest-axe, then we should be good.

Updated here https://github.com/dequelabs/cauldron/pull/1334/commits/b94d0e912a01c9f47b0aa93ea88527de6eb52870

scurker (approved):

Issue #1336, opened 2024-02-06
Update `react-testing-library`


scurker:
We're currently "stuck" on react testing library v12, because versions 13+ have removed support for React v16 and below:

> React Testing Library versions 13+ require React v18. If your project uses an older version of React, be sure to install version 12:

This means we can't take advantage of any new queries/matchers that get added to [`react-testing-library/dom`](https://github.com/testing-library/dom-testing-library) due to the following issues: 
- https://github.com/testing-library/react-testing-library/issues/1184
- https://github.com/testing-library/react-testing-library/issues/1103

If any of the above issues get resolved, it might be possible for us to install the dom library as a peer dependency to take advantage of any new features OR if we fully migrate away from enzyme in order to upgrade to react 17+.

scurker:
This depends on #1473

PR #1337, opened 2024-02-06
test: update TextField tests to use react testing library

scurker:
Upgrading to jest@12 as that's the highest version we can support while being on react 16 (see #1336). This does completely re-write the existing tests as they did not provide adequate coverage.

ref #1159


packages/react/src/components/TextField/index.tsx
```diff
@@ -13,7 +13,7 @@ export interface TextFieldProps
     value: string,
     e: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => void;
-  fieldRef: React.Ref<HTMLInputElement | HTMLTextAreaElement>;
+  fieldRef?: React.Ref<HTMLInputElement | HTMLTextAreaElement>;
```
scurker, line 16: This should already be optional as the default value is set below, but this is just updating the type to match the expected interface.

thuey (dismissed):

thuey (approved):

Issue #1339, opened 2024-02-07
Disabled checkbox components have a pointer cursor


JoshMK:
### Steps to reproduce

Hover over a Checkbox component with the mouse.

### Expected Behavior

The cursor will stay as the default one.

### Actual Behavior

The cursor becomes a pointer for a intractable element.

### Version

6.1.0

### Anything else we should know?

Additional information

PR #1340, opened 2024-02-07
fix(react): use default hover cursor when checkboxes are disabled

JoshMK:
* CSS adjustment to use the default hover cursor on non-intractable disabled checkboxes

Closes issue: #1339

### DIFF
--- a/packages/styles/forms.css
+++ b/packages/styles/forms.css
@@ -439,11 +439,13 @@ textarea.Field--has-error:focus:hover,
 .Radio__overlay--disabled.Icon--radio-checked,
 .Checkbox__overlay--disabled.Icon--checkbox-checked {
   color: var(--field-icon-checked-disabled-color);
+  cursor: default;
 }
 
 .Checkbox__overlay--disabled.Icon--checkbox-unchecked,
 .Radio__overlay--disabled.Icon--radio-unchecked {
   color: var(--field-icon-unchecked-disabled-color);
+  cursor: default;
 }
 
 .Field__label:hover ~ .Radio__overlay:not(.Radio__overlay--disabled),


packages/styles/forms.css
```diff
@@ -463,3 +463,8 @@ textarea.Field--has-error:focus:hover,
 .Checkbox__overlay:hover:not(.Checkbox__overlay--disabled) {
   border: 1px solid currentColor;
 }
+
+.Checkbox__overlay--disabled.Icon--checkbox-checked,
```
scurker, line 467: I cannot wait to get rid of these selectors so we can use `:has`.

scurker (approved):

packages/styles/forms.css
```diff
@@ -463,3 +463,8 @@ textarea.Field--has-error:focus:hover,
...
+.Checkbox__overlay--disabled.Icon--checkbox-unchecked {
```
scurker, line 468: I'm wondering if this should be added in the above rules (lines 439~447) so it covers both checkbox and radio instead of an additional rule?

JoshMK: @scurker Done.

packages/styles/forms.css
```diff
@@ -437,8 +437,11 @@ textarea.Field--has-error:focus:hover,
 }
 
 .Radio__overlay--disabled.Icon--radio-checked,
-.Checkbox__overlay--disabled.Icon--checkbox-checked {
+.Radio__overlay--disabled.Icon--radio-unchecked,
+.Checkbox__overlay--disabled.Icon--checkbox-checked,
+.Checkbox__overlay--disabled.Icon--checkbox-unchecked {
   color: var(--field-icon-checked-disabled-color);
+  cursor: default;
```
scurker, line 444: Sorry, one last request: These are split into 2 different style rules, one for checked checkboxes and one for unchecked checkbox so the change should set `cursor: default` onto both instead of adding new selectors:

```css
.Radio__overlay--disabled.Icon--radio-checked,
.Checkbox__overlay--disabled.Icon--checkbox-checked {
  color: var(--field-icon-checked-disabled-color);
  cursor: default;
}

.Checkbox__overlay--disabled.Icon--checkbox-unchecked,
.Radio__overlay--disabled.Icon--radio-unchecked {
  color: var(--field-icon-unchecked-disabled-color);
  cursor: default;
}
```

JoshMK: Good now? @scurker

scurker (changes requested):
These rules are duplicative of the lines below so you only need to include `cursor: default` once for both the unchecked and checked styles. There should only be a single set of rules for `.Checkbox__overlay--disabled.Icon--checkbox-unchecked,
.Radio__overlay--disabled.Icon--radio-unchecked`

https://github.com/dequelabs/cauldron/blob/e98d332ef14ae376482efe9416adb878fbb6f08a/packages/styles/forms.css#L445-L454

packages/styles/forms.css
scurker: Almost, see review comment

JoshMK: Oops. 😅

scurker (approved):
🎉

PR #1341, opened 2024-02-07
test: update Checkbox tests to use react testing library

scurker:
ref https://github.com/dequelabs/cauldron/issues/1159


packages/react/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+  expect(onFocus.calledOnce).toBeTruthy();
+});
+
+test('should handle blur correctly', () => {
+  const onBlur = spy();
+  const input = renderCheckbox({ onBlur, checked: true });
+  const checkboxIcon = input.parentElement!.querySelector(
+    '.Checkbox__overlay'
+  ) as HTMLElement;
+  expect(checkboxIcon).not.toHaveClass('.Checkbox__overlay--focused');
```
JoshMK, line 92: Not a huge deal, but we can use `expect(checkboxIcon).toHaveFocus()` and `expect(checkboxIcon).not.toHaveFocus()` in addition to checking against classes here. It'd be a nice to make sure we're explicitly looking for the focus attribute.

JoshMK (dismissed):
Looks amazing! 🙌  Just left a small change request on two tests.

packages/react/src/components/Checkbox/Checkbox.test.tsx
scurker: Yeah, totally fair. That's an easy change to make.

JoshMK (approved):

Issue #1006, opened 2023-04-04
Sync Select with UXPin


scurker:
Currently, this looks like it's just the disabled state, but will update this ticket once we verify with @bobbyomari

PR #1342, opened 2024-02-09
feat(styles): synced disabled and error states for select

orest-s:
Closes: #1006


scurker (changes requested):
I realize this file hasn't been touched in a while and has some old patterns, but our new guidelines specify these should be component level variables. 

example:

```css
:root {
  --select-background-color-disabled: var(--field-background-color-disabled)
}

.cauldron--theme-light {
  --select-background-color-disabled: /* whatever color is supposed to be here */
}

.Field__select--disabled select {
  background: var(--select-background-color-disabled);
}
...
```

It's okay if this takes a little more time since we're trying to align our component css patterns as well.

packages/styles/select.css
```diff
@@ -85,3 +87,11 @@
 .Field__select--wrapper.Field--has-error select:focus:hover {
   border-color: var(--field-border-color-error-hover);
 }
+
+.cauldron--theme-dark .Field__select--wrapper select:invalid:focus:hover,
+.cauldron--theme-dark
+  .Field__select--wrapper.Field--has-error
+  select:focus:hover {
+  box-shadow: 0px 0px 5px var(--accent-danger),
```
scurker, line 95: 0 values can be unitless.

```suggestion
  box-shadow: 0 0 5px var(--accent-danger),
```

packages/styles/select.css
```diff
@@ -37,51 +72,53 @@
...
 
 .Field__select--wrapper select:invalid:focus,
 .Field__select--wrapper.Field--has-error select:focus {
-  border-width: var(--space-quarter);
-  border-color: var(--field-border-color-error);
+  border-width: 2px;
+  border-color: var(--select-field-border-color-error);
   box-shadow: rgba(0, 0, 0, 0.05) 0 1px 2px 0 inset,
-    var(--field-border-color-error-focus-glow) 0 0 4px 0;
+    var(--select-field-border-color-error) 0 0 0 1px,
```
scurker, line 117: This was done after a discussion with @dequelabs/cauldron-design in which we decided a temporary solution to address an a11y issue would be to increase the size of the focus ring upon focus.

PR #1348, opened 2024-02-16
fix(react): restore missing types from lib

scurker:
This restores types that are missing from the lib.

### DIFF
--- a/packages/react/rollup.config.js
+++ b/packages/react/rollup.config.js
@@ -22,7 +22,16 @@ export default {
     chunkFileNames: '[name].js'
   },
   plugins: [
-    typescript(),
+    typescript({
+      tsconfig: './tsconfig.json',
+      exclude: [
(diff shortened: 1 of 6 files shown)


packages/react/rollup.config.js
```diff
@@ -22,7 +22,16 @@ export default {
     chunkFileNames: '[name].js'
   },
   plugins: [
-    typescript(),
+    typescript({
```
scurker, line 25: I went this route because the default `tsconfig` and `tsc` was picking up _everything_ under the src folder, including test files meaning we were packaging types such as `jest-dom` that do not need to be distributed.

There doesn't seem to be a default way to get `tsc` to ignore specific files for output while allowing it for type hints, so letting rollup do the heavy lifting to exclude what we don't want seems like the right route.

schne324 (approved):

Issue #1343, opened 2024-02-13
Add TextEllipsis Component


scurker:
There are various cases where the width of a container cannot accommodate a given length of text and wrapping text is undesirable. Adding `text-overflow: ellipsis` as a style can accommodate this need in some cases, but can introduce issues with [WCAG 1.4.12](https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html) if not handled correctly.

`TextEllipsis` will be a utility component that does not infer any specific styles on the containing text, but will apply a CSS ellipsis to the text when the containing text overflows any wrapping containers, and provide an accessible means for showing the full text content.

## Examples

```jsx
<div class="container">
  <TextEllipsis>Text to Ellipsis</TextEllipsis>
</div>
```

By default, text will be limited to 1 lines and prevent overflow with an ellipsis. When ellipsized, the content should become a tabbable element with the full text content available via a tooltip on focus.

Under the hood, when no element is specified this should render an unstyled `button` element with `aria-disabled="true"` set when the text becomes clipped and a `span` or element when not. The tooltip should also be conditionally rendered dependent upon the text becoming clipped.

### Interface

```ts
interface TextEllipsis {
  maxLines?: number
  as?: React.ElementType
  tooltipProps?: TooltipProps
  children: string
}
```

Alternatively, the text ellipsizing can allow for multiple lines using [`-webkit-line-clamp`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp):

```jsx
<div class="container">
  <TextEllipsis maxLines={2}>
    Some long text that should clip text and add ellipsis when it wraps beyond 2 lines
  </TextEllipsis>
</div>
```

If the component containing the text is already interactive, we do not want to create an interactive element inside of another interactive element. Instead, `TextEllipsis` should allow for polymorphic components to allow for the interactive element to take on the tooltip:

```jsx
<div class="container">
  <TextEllipsis as={Button}>
    Text to TextEllipsis
  </TextEllipsis>
</div>
```

> [!NOTE]  
> Ideally, the component should _not_ be introducing additional tabstops where unnecessary as this can lead to a negative user experience for keyboard users. Instead, we should explore if it's realistic to programmatically determine if we can conditionally display the tabstop + focus element _only_ when it's necessary for the text to be clipped.

PR #1354, opened 2024-02-20
feat(react): text ellipsis

scurker:
ref #1343

### DIFF
--- a/packages/react/src/components/TextEllipsis/index.tsx
+++ b/packages/react/src/components/TextEllipsis/index.tsx
@@ -1,0 +1,103 @@
+import React, { Ref, useEffect, useState } from 'react';
+import classnames from 'classnames';
+import useSharedRef from '../../utils/useSharedRef';
+import Tooltip, { type TooltipProps } from '../Tooltip';
+
+interface TextEllipsisProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: string;
+  maxLines?: number;
+  as?: React.ElementType;
+  tooltipProps?: Omit<TooltipProps, 'target' | 'association'>;
+}
+
+const TextEllipsis = React.forwardRef(
+  (
+    {
+      className,
+      children,
+      maxLines,
+      as,
+      tooltipProps,
+      ...props
+    }: TextEllipsisProps,
+    ref: Ref<HTMLDivElement>
+  ) => {
+    let Element: React.ElementType<any> = 'div';
+    const sharedRef = useSharedRef<HTMLElement>(ref);
+    const [showTooltip, setShowTooltip] = useState(false);
+
+    if (as) {
+      Element = as;
+    } else if (showTooltip) {
+      props = Object.assign(
+        {
+          role: 'button',
+          'aria-disabled': true,
+          tabIndex: 0
+        },
+        props
+      );
+    }
+
+    if (typeof maxLines === 'number') {
+      props.style = {
+        WebkitLineClamp: maxLines || 2,
+        ...props.style
+      };
+    }
+
+    useEffect(() => {
+      const listener: ResizeObserverCallback = () => {
+        requestAnimationFrame(() => {
+          const { current: overflowElement } = sharedRef;
+          if (!overflowElement) {
+            return;
+          }
+
+          setShowTooltip((overflow) => {
+            const hasOverflow =
+              typeof maxLines === 'number'
+                ? overflowElement.clientHeight < overflowElement.scrollHeight
+                : overflowElement.offsetWidth < overflowElement.scrollWidth;
+            return overflow !== hasOverflow ? hasOverflow : overflow;
+          });
+        });
+      };
+
+      const observer = new ResizeObserver(listener);
+      observer.observe(sharedRef.current);
+
+      return () => {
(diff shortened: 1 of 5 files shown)


packages/react/src/components/TextEllipsis/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+import useSharedRef from '../../utils/useSharedRef';
+import Tooltip, { type TooltipProps } from '../Tooltip';
+
+interface TextEllipsisProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: string;
+  maxLines?: number;
+  as?: React.ElementType;
+  refProp?: string;
+  tooltipProps?: Omit<TooltipProps, 'target' | 'association'>;
+}
```
thuey, line 12: Should we be using the polymorphic type util? https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/utils/polymorphic-type/index.ts

In the docs, you gave an example of specifying `as={Link}`. However, without using the polymorphic type utility, TS would complain if we try to provide properties specific to the component we are rendering `as` (e.g. `href`):
image

packages/react/src/components/TextEllipsis/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+    }: TextEllipsisProps,
+    ref: Ref<HTMLElement>
+  ) => {
+    let Element: React.ElementType<any> = 'div';
+    const sharedRef = useSharedRef<HTMLElement>(ref);
+    const [showTooltip, setShowTooltip] = useState(false);
+
+    if (as) {
+      Element = as;
+    } else if (showTooltip) {
```
thuey, line 32: So is the expectation that if you pass `as`, it _has_ to be interactive (either semantically or the caller is responsible themselves for setting the proper interactive attributes)? Might be good to call that out in the docs, since someone might pass `p` or some other non-interactive as the `as`.

packages/react/src/components/TextEllipsis/index.tsx
scurker: Looks like that polymorphic type doesn't quite pass the mustard here so there's still some work to do. 😞 



I'm inclined to leave this as-is for now, but I do agree we need to address polymorphic types and likely to address that as a separate issue. What do you think?

packages/react/src/components/TextEllipsis/index.tsx
scurker: It's already called out a bit:

> If your component is a Button, Link, or some other kind of interactive element with a `tabIndex`, you _must_ provide your component as a Polymorphic component using the `as` prop to avoid nesting interactive elements

IMO authors _should not_ be passing in a non-interactive element. We can make that indication stronger if you think it's needed, but we have no way of validating that the component that they pass in is interactive.

packages/react/src/components/TextEllipsis/index.tsx
scurker: Digging into this a little bit more, it seems the current `ForwardRefComponent` type only infers types of intrinsic elements. So `as="a"` is able to infer the type, but not `as={Link}`. As such, there's more work to be done here but I would like to defer that to another issue.

thuey: I'd be fine with deferring it, but I wonder if the `as` prop should be removed or caveats documented, since it's not fully supported.

packages/react/src/components/TextEllipsis/index.tsx
thuey: As a nit, I would suggest documenting that the expectation is that `as` should be interactive, even if we can't verify it.

packages/react/src/components/TextEllipsis/index.tsx
scurker: > I'd be fine with deferring it, but I wonder if the as prop should be removed or caveats documented

It _can not_ be removed because the absence of it would create accessibility issues when the element is contained within another interactive element. I've been looking at other instances in different libraries and there doesn't appear to be a way that I've found yet to support non-intrinsic elements.

packages/react/src/components/TextEllipsis/index.tsx
scurker: Added a clarifying note.

packages/react/src/components/TextEllipsis/index.tsx
scurker: Created https://github.com/dequelabs/cauldron/issues/1407, going to do a bit of digging to see what we need to do here. My guess is if it is possible the type may need to be updated.

thuey (approved):

packages/react/src/components/TextEllipsis/index.tsx
thuey: > It _can not_ be removed because the absence of it would create accessibility issues when the element is contained within another interactive element. I've been looking at other instances in different libraries and there doesn't appear to be a way that I've found yet to support non-intrinsic elements.

I just meant remove it until it's fully supported. But I guess `Link` is probably the worst case scenario. I imagine a lot of button usages would probably work as long as we're just using things like `onClick`

PR #1362, opened 2024-02-26
test: switch breadcrumb tests to react testing library

lsprr:
Issue #1159

This PR switches the Line tests to React Testing Library.

### Validation steps

- Run yarn test

### DIFF
--- a/packages/react/src/components/Breadcrumb/index.test.tsx
+++ b/packages/react/src/components/Breadcrumb/index.test.tsx
@@ -1,60 +1,55 @@
-import React, { createRef } from 'react';
-import { shallow } from 'enzyme';
-import {
-  default as Breadcrumb,
-  BreadcrumbLink,
-  BreadcrumbItem
-} from 'src/components/Breadcrumb';
-import axe from '../../../axe';
+import React from 'react';
+import { render } from '@testing-library/react';
+import Breadcrumb, { BreadcrumbLink, BreadcrumbItem } from './index';
+import axe from '../../axe';
 
 test('should render breadcrumbs', () => {
-  const breadcrumb = shallow(
+  const { container } = render(
     <Breadcrumb aria-label="breadcrumb">
       <BreadcrumbLink href="#">one</BreadcrumbLink>
       <BreadcrumbLink href="#">two</BreadcrumbLink>
       <BreadcrumbItem>three</BreadcrumbItem>
     </Breadcrumb>
   );
-  expect(breadcrumb.hasClass('Breadcrumb')).toBe(true);
-  expect(breadcrumb.find('li').length).toBe(3);
+  expect(container.querySelector('.Breadcrumb')).toBeInTheDocument();
+  expect(container.querySelectorAll('li').length).toBe(3);
 });
 
 test('should render separators between breadcrumbs', () => {
-  const breadcrumb = shallow(
+  const { container } = render(
     <Breadcrumb aria-label="breadcrumb">
       <BreadcrumbLink href="#">one</BreadcrumbLink>
       <BreadcrumbLink href="#">two</BreadcrumbLink>
       <BreadcrumbItem>three</BreadcrumbItem>
     </Breadcrumb>
   );
-  expect(breadcrumb.find('.Breadcrumb__Separator').length).toBe(2);
-  expect(
-    breadcrumb
-      .find(BreadcrumbItem)
-      .find('.Breadcrumb__Separator')
-      .exists()
-  ).toBeFalsy();
+  const item = container.querySelector('Breadcrumb__Item');
+  expect(container.querySelectorAll('.Breadcrumb__Separator').length).toBe(2);
+  if (item) {
+    const separator = item.querySelector('.Breadcrumb__Separator');
+    expect(separator).not.toBeInTheDocument();
+  }
 });
 
 test('should render custom separators between breadcrumbs', () => {
-  const breadcrumb = shallow(
+  const { container } = render(
     <Breadcrumb aria-label="breadcrumb" separator="💩">
       <BreadcrumbLink href="#">one</BreadcrumbLink>
       <BreadcrumbLink href="#">two</BreadcrumbLink>
       <BreadcrumbItem>three</BreadcrumbItem>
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Breadcrumb/index.test.tsx
```diff
@@ -1,60 +1,55 @@
...
-import { shallow } from 'enzyme';
-import {
-  default as Breadcrumb,
-  BreadcrumbLink,
-  BreadcrumbItem
-} from 'src/components/Breadcrumb';
-import axe from '../../../axe';
+import React from 'react';
+import { render } from '@testing-library/react';
+import Breadcrumb, { BreadcrumbLink, BreadcrumbItem } from './index';
```
scurker, line 3: Very very minor nit:

```suggestion
import Breadcrumb, { BreadcrumbLink, BreadcrumbItem } from './';
```

packages/react/src/components/Breadcrumb/index.test.tsx
```diff
@@ -1,60 +1,55 @@
...
+import axe from '../../axe';
 
 test('should render breadcrumbs', () => {
-  const breadcrumb = shallow(
+  const { container } = render(
```
scurker, line 7: When we've been updating our tests, we've been trying to follow better [test writing practices](https://testing-library.com/docs/guiding-principles) with the following guideline:

> It should be generally useful for testing the application components in the way the user would use it.

In this instance, we want this to be looking for a navigation type element, as that's what the user would see. To meet that expectation, we would write (as an example):

```jsx
render(<Breadcrumb aria-label="breadcrumb">...</BreadCrumb>
await expect(screen.getByRole('navigation', { name: 'breadcrumb' })).toBeInTheDocument()
await expect(screen.queryByRole('listitem')).toHaveLength(3)
```

As far as classes and other props, we can split those out into other tests specific to those properties as seen as an example below:

https://github.com/dequelabs/cauldron/blob/5d92f1e2b49f6a5ce814de4c996c285c79d4fab4/packages/react/src/components/Link/Link.test.tsx#L41-L59

packages/react/src/components/Breadcrumb/index.test.tsx
```diff
@@ -1,60 +1,55 @@
...
   );
-  expect(breadcrumb.hasClass('Breadcrumb')).toBe(true);
-  expect(breadcrumb.find('li').length).toBe(3);
+  expect(container.querySelector('.Breadcrumb')).toBeInTheDocument();
+  expect(container.querySelectorAll('li').length).toBe(3);
 });
 
 test('should render separators between breadcrumbs', () => {
-  const breadcrumb = shallow(
+  const { container } = render(
```
scurker, line 19: Instead of using container, this should be using:
```jsx
screen.getByRole('navigation', { name: 'breadcrumb' })
```

packages/react/src/components/Breadcrumb/index.test.tsx
```diff
@@ -1,60 +1,55 @@
...
+  expect(container.querySelectorAll('.Breadcrumb__Separator').length).toBe(2);
+  if (item) {
+    const separator = item.querySelector('.Breadcrumb__Separator');
+    expect(separator).not.toBeInTheDocument();
+  }
 });
 
 test('should render custom separators between breadcrumbs', () => {
-  const breadcrumb = shallow(
+  const { container } = render(
```
scurker, line 35: Instead of using container, this should be using:
```jsx
screen.getByRole('navigation', { name: 'breadcrumb' })
```

packages/react/src/components/Breadcrumb/index.test.tsx
```diff
@@ -1,60 +1,55 @@
...
       <BreadcrumbItem>three</BreadcrumbItem>
     </Breadcrumb>
   );
-  expect(breadcrumb.text()).toEqual('one💩two💩three');
+  expect(container.textContent).toEqual('one💩two💩three');
 });
 
 test('should return no axe violations', async () => {
-  const breadcrumb = shallow(
+  const { container } = render(
```
scurker, line 46: Instead of using container, this should be using:
```jsx
screen.getByRole('navigation', { name: 'breadcrumb' })
```

packages/react/src/components/Breadcrumb/index.test.tsx
```diff
@@ -1,60 +1,55 @@
...
     <Breadcrumb aria-label="breadcrumb" separator="💩">
       <BreadcrumbLink href="#">one</BreadcrumbLink>
       <BreadcrumbLink href="#">two</BreadcrumbLink>
       <BreadcrumbItem>three</BreadcrumbItem>
     </Breadcrumb>
   );
-  expect(breadcrumb.text()).toEqual('one💩two💩three');
+  expect(container.textContent).toEqual('one💩two💩three');
```
scurker, line 42: Based on the above suggestion...

```suggestion
  expect(screen.getByRole('navigation', { name: 'breadcrumb' })).toHaveTextContentl('one💩two💩three');
```

packages/react/src/components/Breadcrumb/index.test.tsx
```diff
@@ -1,60 +1,55 @@
...
-      .find(BreadcrumbItem)
-      .find('.Breadcrumb__Separator')
-      .exists()
-  ).toBeFalsy();
+  const item = container.querySelector('Breadcrumb__Item');
+  expect(container.querySelectorAll('.Breadcrumb__Separator').length).toBe(2);
+  if (item) {
+    const separator = item.querySelector('.Breadcrumb__Separator');
+    expect(separator).not.toBeInTheDocument();
+  }
```
scurker, line 31: I'm a little unclear why we need the logic branch here for item?

packages/react/src/components/Breadcrumb/index.test.tsx
lsprr: Hi @scurker, I did update the code to adhere to RTL best practices. Thank you for the feedback!

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

PR #1363, opened 2024-02-27
test: switch description list tests to react testing library

lsprr:
Issue #1159

This PR switches the Line tests to React Testing Library.

### Validation steps

- Run yarn test


packages/react/src/components/DescriptionList/index.test.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import {
+  DescriptionList,
+  DescriptionListItem,
+  DescriptionTerm,
+  DescriptionDetails
+} from './index';
```
scurker, line 8: Very minor nit.

```suggestion
} from './';
```

packages/react/src/components/DescriptionList/index.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+import axe from '../../axe';
+
+describe('DescriptionList components', () => {
+  test('renders a dl element', () => {
+    const { container } = render(<DescriptionList>a</DescriptionList>);
+    const dl = container.querySelector('dl');
+    expect(dl).toBeInTheDocument();
```
scurker, line 15: RTL typically recommends using a test id for instances like this:

```jsx
render(<DescriptionList data-testid="dl">a</DescriptionList>)
expect(screen.getByTestTestId('dl').tagName).toBe('DL')
```

packages/react/src/components/DescriptionList/index.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  test('renders a dt element', () => {
+    const { container } = render(<DescriptionTerm>a</DescriptionTerm>);
+    const dt = container.querySelector('dt');
+    expect(dt).toBeInTheDocument();
+  });
+
+  test('renders a dd element', () => {
+    const { container } = render(<DescriptionDetails>a</DescriptionDetails>);
```
scurker, line 25: Same as above, but I believe it should be `definition` as the role.

packages/react/src/components/DescriptionList/index.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const dd = container.querySelector('dd');
+    expect(dd).toBeInTheDocument();
+  });
+
+  test('handles uncollapsed prop', () => {
+    const { container } = render(<DescriptionList>a</DescriptionList>);
+    const dl = container.querySelector('dl');
```
scurker, line 32: See above about `data-testid`

packages/react/src/components/DescriptionList/index.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    expect(dl).not.toHaveClass('DescriptionList--collapsed');
+  });
+
+  test('handles collapsed prop', () => {
+    const { container } = render(
```
scurker, line 37: See above about `data-testid`

packages/react/src/components/DescriptionList/index.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <DescriptionList collapsed>a</DescriptionList>
+    );
+    const dl = container.querySelector('dl');
+    expect(dl).toHaveClass('DescriptionList--collapsed');
+  });
+
+  test('passes classNames through', () => {
+    render(<DescriptionList className="a">a</DescriptionList>);
+    const dl = screen.getByText('a');
+    expect(dl).toHaveClass('a');
```
scurker, line 47: We should be sure that we're getting the right component here (by its role or test id) and that it doesn't clobber the existing class name that's already on the component.

scurker (commented):
Just a few minor things to fix, thanks for your help here!

packages/react/src/components/DescriptionList/index.test.tsx
lsprr: Thank you for the snippet! I wasn't aware of the tagname property until now. It's helpful for identifying HTML tags that don't have an assigned ARIA role.

lsprr:
@scurker 

Thank you for the feedback and assistance. I greatly appreciate it!

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

PR #1367, opened 2024-02-28
test: switch loader tests to react testing library

lsprr:
Issue #1159

This PR switches the Loader tests to React Testing Library.

### Validation steps

- Run yarn test


packages/react/src/components/Loader/index.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+  render(<Loader data-testid="loader" label="hi" />);
+  expect(screen.getByTestId('loader')).not.toHaveAttribute(
+    'aria-hidden',
+    'false'
+  );
+});
+
+test('sets expected role attributes given an aria-label', () => {
+  render(<Loader data-testid="loader" label="bananas" />);
+  expect(screen.getByTestId('loader')).toHaveAttribute('role', 'alert');
```
scurker, line 31: I think we want to check for the role here:

```suggestion
  expect(screen.getByRole('alert')).toBeInTheDocument()
```

We can also check that the element has an accessible name as well:

```
  expect(screen.getByRole('alert')).toHaveAccessibleName('bananas')
```

packages/react/src/components/Loader/index.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+});
+
+test('supports ref={React.createRef()}', () => {
```
scurker, line 34: Small nit:

```suggestion
test('supports ref prop', () => {
```

scurker (changes requested):
A few small things

packages/react/src/components/Loader/index.test.tsx
lsprr: Hi, the label is within a '<span>' element with the class 'Offscreen'.

`const Loader = React.forwardRef<HTMLDivElement, LoaderProps>(
  ({ className, label, ...props }: LoaderProps, ref) => {
    if (label?.length) {
      props['role'] = 'alert';
      props.children = <Offscreen>{label}</Offscreen>;
    } else {
      props['aria-hidden'] = true;
    }
    return (
      <div ref={ref} className={classNames('Loader', className)} {...props} />
    );
  }
);
`

So, this is what I came up with to find the label. 
`
  expect(screen.getByText('bananas', { selector: '.Offscreen' })).toBeInTheDocument();
`

Please inform me if you are satisfied with this solution, or if you have any alternative suggestions.

scurker: Ah right, the loader doesn't have an accessible name but wraps offscreen text. In that instance, I think this should be:

```js
expect(screen.getByRole('alert').toHaveTextContent('bananas'))
```

packages/react/src/components/Loader/index.test.tsx
scurker: Looks like this wasn't updated? This was just to keep the naming consistent with other tests. Just checking.

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

PR #1370, opened 2024-02-28
fix(react): ensure Tooltip cleans up association id after removal

### LINKED ISSUE #1353: Bug: Tooltip does not cleanup aria association after unmount
### Steps to reproduce

To replicate this scenario, have an element with a dynamically rendered tooltip:

```jsx
<Button ref={buttonRef}>Button</Button>
{showTooltip && (
  <Tooltip target={buttonRef}>
    Tooltip Description
  </Tooltip>
)}
```

### Expected Behavior

When mounted, the tooltip will normally associate the target with the tooltip with an aria association, by default with `aria-describedby`:

```html
<button aria-describedby="tooltipid">Button</button>
```

However in cases where the tooltip is unmounted or removed from the page, the above association remains. When removed, Tooltip should clean up any relevant associations with the targeted element. If the aria association is the only id referenced, the aria association attribute should be removed. Alternatively if there are other associations, the attribute should remain with the referenced id removed.

### Actual Behavior

When an associated element with a tooltip is removed, the association remains meaning there's an aria associated pointed to an element that is no longer present on the page.

### Version

6.x

_No response_

### Relevant log output

_No response_

N/A

scurker:
closes #1353


packages/react/src/utils/idRefs.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Returns a unique set of id refs from the provided string
+ * @param ids - string of id refs
+ */
+export default function idRefs(ids: IdRefs): Set<string> {
+  if (!ids || !ids.trim()) {
+    return new Set<string>();
+  }
+  return new Set(ids.trim().split(/\s+/));
+}
```
thuey, line 12: Sets are un-ordered, so it's possible when adding or removing an ID that the order of the IDs change. Unless I'm missing something, we should change this implementation to make sure ordering is preserved, such as with an `Array`

thuey (changes requested):

packages/react/src/utils/idRefs.ts
scurker: I think you're missing something. Sets maintain their order via [insertion order](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/add):
> The insertion order corresponds to the order in which each element was inserted into the set by the [add()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/add) method successfully (that is, there wasn't an identical element already in the set when add() was called).

Adding an id that already exists should be a noop as the id already exists, and it should always be appended to the end when it doesn't exist so there's no order change there. I'm not quite following on how removing an id should change the order of the set. The below test shows that the order still is the same after removal:

https://github.com/dequelabs/cauldron/blob/1a2038d03ab58ea716ae28df3280bf5cc800d160/packages/react/src/utils/idRefs.test.ts#L28-L30

thuey: Cool! My Computer Science days about the definition of a Set were just haunting me, then. It's cool the JS maintains insertion order

PR #1372, opened 2024-02-28
feat(react,styles): add SearchField and internal TextFieldWrapper components

### LINKED ISSUE #1178: Placeholder text has default styling
We're not styling placeholder text meaning that it looks odd when set for dark mode.

## Light Mode

gray placeholder text on white background

## Dark Mode

gray placeholder text on dark background

We probably should minimally set this for both light/dark modes so that it does look correct is someone uses it.

Change light mode placeholder text to: `#757575` (this should be a new global gray color, `--gray-50`)
Change dark mode placeholder text to: `#8c9fab`

Also add `font-style: italic` to both light/dark modes.

### LINKED ISSUE #1190: Add Search Field
TBD

yhafez:
Closes #1190
Closes #1178


packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -424,10 +425,8 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
           )}
         </label>
         {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
-        <div
-          className={classnames('Combobox__input', {
-            'Combobox__input--error': hasError
-          })}
+        <TextFieldWrapper
```
yhafez, line 428: Replacing one-off wrapper implementation with reusable implementation of a styled div that applies native `input` styles on the `div` and removes them from the nested `input`.

Discussed with @scurker

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+        {hideLabel ? <Offscreen>{label}</Offscreen> : <span>{label}</span>}
+      </label>
+      <TextFieldWrapper
+        className={
+          otherProps.disabled ? 'TextFieldWrapper--disabled' : undefined
+        }
+      >
+        <Icon
+          type="magnifying-glass"
+          aria-hidden="true"
```
yhafez, line 75: Hide from screen readers since this is decorative and label is required

packages/styles/combobox.css
```diff
@@ -1,20 +1,4 @@
...
-  );
-  --combobox-input-error-border-color: var(--field-border-color-error);
-  --combobox-input-error-focus-border-color: var(
-    --field-border-color-error-hover
-  );
-  --combobox-input-error-focus-border-glow-color: var(
-    --field-border-color-focus-glow
-  );
-  --combobox-input-font-size: var(--field-font-size);
-  --combobox-input-font-color: var(--field-content-color);
```
yhafez, line 17: Styles and variables prefixed with `combobox-input` have been refactored to be implemented by `TextFieldWrapper` component

packages/styles/forms.css
```diff
@@ -12,6 +12,7 @@
   --field-border-color-error-focus-glow: rgba(217, 50, 81, 0.7);
   --field-border-color-disabled: rgba(204, 204, 204, 0.25);
   --field-background-color-disabled: var(--light-workspace-color);
+  --field-placeholder-color: var(--gray-60);
```
yhafez, line 15: Added variables for placeholder text color

packages/styles/text-field-wrapper.css
```diff
@@ -0,0 +1,127 @@
...
+.cauldron--theme-dark .TextFieldWrapper--error {
+  border: 1px solid var(--text-field-wrapper-error-border-color);
+}
+
+.TextFieldWrapper:hover {
+  border-color: var(--text-field-wrapper-border-hover-color);
+}
+
+.TextFieldWrapper input[type='text'],
+.TextFieldWrapper input[type='search'] {
```
scurker, line 73: I think we can omit `type` here since type could be a number of different things (number, password, search, etc)

```suggestion
.TextFieldWrapper input {
```

packages/react/src/components/internal/TextFieldWrapper/TextFieldWrapper.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    'banana'
+  );
+});
+
+test('should render TextFieldWrapper with onClick', async () => {
+  const onClick = spy();
+  render({ onClick });
+  await userEvent.click(screen.getByText('Children'));
+  expect(onClick.calledOnce).toBe(true);
+});
```
scurker, line 45: This seems like an odd test to add for this component.

docs/pages/components/SearchField.mdx
```diff
@@ -0,0 +1,92 @@
...
+```jsx example
+<FieldWrap>
+  <SearchField label="Hidden" hideLabel />
+</FieldWrap>
+```
+
+## Props
+
+<ComponentProps
+  className={true}
```
scurker, line 49:
```suggestion
  className={true}
  refType="HTMLInputElement"
```

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  extends Omit<InputHTMLAttributes<HTMLInputElement>, 'onChange'> {
+  label: ReactNode;
+  value?: string;
+  defaultValue?: string;
+  onChange?: (value: string, e: ChangeEvent<HTMLInputElement>) => void;
+  hideLabel?: boolean;
+  isForm?: boolean;
+}
+
+const SearchField: ForwardRefRenderFunction<
```
scurker, line 27: This is missing display name:

```js
SearchField.displayName = 'SearchField';
```

I think it might have to be set after wrapping this component with `forwardRef` instead of returning `forwardRef`.

packages/styles/combobox.css
```diff
@@ -63,42 +32,6 @@
   padding: var(--space-smallest);
```
dequejosie, line 32: Is this block still needed for other `input` in Combobox?

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -452,7 +451,7 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
             onBlur={handleBlur}
           />
           <span className="Combobox__arrow" />
-        </div>
+        </TextFieldWrapper>
```
dequejosie, line 454: Nothing related to this particular PR, I am just thinking if we will refactor `<span className="Combobox__arrow" />` into `trailingChildren` of `TextFieldWrapper`? @scurker

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+} from 'react';
+import classNames from 'classnames';
+
+import TextFieldWrapper from '../internal/TextFieldWrapper';
+import randomId from '../../utils/rndid';
+import Offscreen from '../Offscreen';
+import Icon from '../Icon';
+
+interface SearchFieldProps
+  extends Omit<InputHTMLAttributes<HTMLInputElement>, 'onChange'> {
```
dequejosie, line 18: Just curious `HTMLAttributes<HTMLInputElement>` does not work? Just looks a bit redundant.

packages/react/src/components/Combobox/Combobox.tsx
scurker: Yes, that should be how it works.

shawnsharpDQ:
Could we remove the padding on the right next to the clear button? 
Search with text and remove button visable

The clear button needs to have a 24px hit area and I think it is measuring close to 14px

scurker:
> The clear button needs to have a 24px hit area and I think it is measuring close to 14px

The x to clear the field is coming from chrome ui and is not Cauldron. The mechanism for 24px also accounts for spacing, which means we need to account for the input field to be spaced appropriately, which it is not currently:

image

packages/react/src/components/internal/TextFieldWrapper/TextFieldWrapper.test.tsx
yhafez: That's fair. I wanted to include it because the usage of `ComboBox` which this replaced passed an `onClick`, so wanted to make sure there's a test confirming that that would still work. I can remove it if you feel like it's unneccessary.

packages/react/src/components/SearchField/index.tsx
yhafez: Ya, it's necessary because we're overriding the `onChange` prop with a different type. Here's the error it produces if we don't omit it:
```
Interface 'SearchFieldProps' incorrectly extends interface 'InputHTMLAttributes<HTMLInputElement>'.
  Types of property 'onChange' are incompatible.
    Type '((value: string, e: ChangeEvent<HTMLInputElement>) => void) | undefined' is not assignable to type 'ChangeEventHandler<HTMLInputElement> | undefined'.
      Type '(value: string, e: ChangeEvent<HTMLInputElement>) => void' is not assignable to type 'ChangeEventHandler<HTMLInputElement>'.
        Target signature provides too few arguments. Expected 2 or more, but got 1.
```

yhafez: I think we want it to possibly be undefined in case it's an uncontrolled component

packages/styles/combobox.css
yhafez: Nope, removed!

yhafez:
> > The clear button needs to have a 24px hit area and I think it is measuring close to 14px
> 
> The x to clear the field is coming from chrome ui and is not Cauldron. The mechanism for 24px also accounts for spacing, which means we need to account for the input field to be spaced appropriately, which it is not currently:
> 
> 
> 
> To be clear, a target can be under 24px and still pass minimum target size as long as there are no adjacent targets within that target. Since this is part of browser chrome that we can't customize the sizing, we need to be sure that the spacing passes instead.

Hmmm, I'm not exactly sure how to handle this. Chrome renders the "x" button inside the padding of the input, so if I were to increase the padding, it wouldn't create space between the text and the input. I could add more padding to the right of the "x" button, but I'm not sure if that technically fixes things, especially since clicking on the padding doesn't trigger the clearing functionality.

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+        <Offscreen>
+          <label htmlFor={inputId} id={`${inputId}-label`}>
+            {label}
+          </label>
+        </Offscreen>
+      ) : (
+        <label
+          className={'Field__label'}
+          htmlFor={inputId}
+          id={`${inputId}-label`}
```
yhafez, line 85: Added this `aria-labelledby` and the `id` on the labels. Please confirm this is the correct implementation. Not sure if any `aria-labelledby` needs to also be applied to the `input` as well?

shawnsharpDQ:
> Hmmm, I'm not exactly sure how to handle this. Chrome renders the "x" button inside the padding of the input, so if I were to increase the padding, it wouldn't create space between the text and the input. I could add more padding to the right of the "x" button, but I'm not sure if that technically fixes things, especially since clicking on the padding doesn't trigger the clearing functionality.

Is it possible to control the target size? I still find it hard to click the x. I am ok with the size of the icon and placement, hoping we could increase the target size so it is a little easier to click.

packages/react/src/components/internal/TextFieldWrapper/TextFieldWrapper.test.tsx
scurker: As long as it's covered by the Combobox tests, we can omit it here.

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    >
+      {hideLabel ? (
+        <Offscreen>
+          <label htmlFor={inputId} id={`${inputId}-label`}>
+            {label}
+          </label>
+        </Offscreen>
+      ) : (
+        <label
+          className={'Field__label'}
```
scurker, line 83: Nit:

```suggestion
          className="Field__label"
```

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+        </label>
+      )}
+      <TextFieldWrapper
+        className={classNames({
+          'TextFieldWrapper--disabled': otherProps.disabled
+        })}
+      >
+        <Icon
+          type="magnifying-glass"
+          aria-hidden="true"
```
scurker, line 97: This is redundant since it's already hidden when there's no label, see: 
https://github.com/dequelabs/cauldron/blob/b7bf54c6c627676834f383c78bd31aaf9d6ace0e/packages/react/src/components/Icon/index.tsx#L46

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+          {...otherProps}
+          className={classNames(otherProps.className, 'Field__text-input')}
+          type="search"
+        />
+      </TextFieldWrapper>
+    </Field>
+  );
+};
+
+const SearchField = forwardRef(SearchFieldComponent);
```
scurker, line 115: Just a very minor nit to make this consistent with other components:
```suggestion
```

We typically wrap the component itself and return that element with `forwardRef` with no intermediate variables, see:
https://github.com/dequelabs/cauldron/blob/b7bf54c6c627676834f383c78bd31aaf9d6ace0e/packages/react/src/components/Button/index.tsx#L10

packages/react/src/components/internal/TextFieldWrapper/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import classNames from 'classnames';
+
+export interface TextFieldWrapperProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: React.ReactNode;
+  className?: string;
+}
+
+const TextFieldWrapper = ({
```
scurker, line 9: This component is just missing `ref`, see [component standards](https://cauldron.dequelabs.com/component-guidelines#component-standards)

docs/pages/components/SearchField.mdx
```diff
@@ -0,0 +1,105 @@
...
+  refType="HTMLInputElement"
+  props={[
+    {
+      name: 'label',
+      type: 'React.ReactNode',
```
scurker, line 66: This should be using `ContentNode`, see: 
https://github.com/dequelabs/cauldron/blob/b7bf54c6c627676834f383c78bd31aaf9d6ace0e/packages/react/src/components/Combobox/Combobox.tsx#L36

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+import classNames from 'classnames';
+import { useId } from 'react-id-generator';
+
+import TextFieldWrapper from '../internal/TextFieldWrapper';
+import Offscreen from '../Offscreen';
+import Icon from '../Icon';
+
+interface SearchFieldProps
+  extends Omit<InputHTMLAttributes<HTMLInputElement>, 'onChange'> {
+  label: ReactNode;
```
scurker, line 19: This should be using `ContentNode`, see:
https://github.com/dequelabs/cauldron/blob/b7bf54c6c627676834f383c78bd31aaf9d6ace0e/packages/react/src/components/Combobox/Combobox.tsx#L36

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const SearchFieldComponent: ForwardRefRenderFunction<
+  HTMLInputElement,
+  SearchFieldProps
+> = (
+  {
+    label,
+    defaultValue = '',
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    onChange = () => {},
```
scurker, line 35: Now that we're checking `typeof` below, this default can be omitted.

packages/styles/search-field.css
```diff
@@ -0,0 +1,125 @@
...
+    --field-border-color-error-hover
+  );
+  --search-input-error-focus-border-glow-color: var(
+    --field-border-color-focus-glow
+  );
+  --search-input-font-color: var(--field-content-color);
+  --search-icon-color: var(--accent-light);
+}
+
+.Field__wrapper .TextFieldWrapper {
```
scurker, line 36: This is a wee bit too generic, this should probably be `SearchField`:
```suggestion
.SearchField :where(.TextFieldWrapper) {
```

scurker (changes requested):
Mostly nit picks and things to bring things up to alignment with other cauldron components.

scurker (changes requested):

packages/react/src/components/SearchField/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const [id] = propId ? [propId] : useId(1, 'search-field');
+    const inputId = useRef(id).current;
+
+    const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+      const newValue = e.target.value;
+      if (typeof onChange === 'function') {
+        onChange(newValue, e);
+      }
+
+      const isControlled = typeof propValue !== 'undefined';
```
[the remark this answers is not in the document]
yhafez: Good catch!

packages/styles/index.css
```diff
@@ -42,3 +42,5 @@
 @import './notice.css';
 @import './combobox.css';
 @import './timeline.css';
+@import './search-field.css';
+@import './text-field-wrapper.css';
```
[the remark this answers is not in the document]
yhafez: That makes sense!

packages/styles/text-field-wrapper.css
```diff
@@ -0,0 +1,126 @@
...
+  background-color: var(--text-field-wrapper-background-color);
+}
+
+.cauldron--theme-dark .TextFieldWrapper {
+  border: 1px solid var(--text-field-wrapper-border-color);
+}
+
+.cauldron--theme-dark .TextFieldWrapper--error {
+  border: 1px solid var(--text-field-wrapper-error-border-color);
+}
```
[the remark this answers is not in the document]
yhafez: Good question. This was in place [when it was in the styles for `combox.css`](https://github.com/dequelabs/cauldron/pull/1372/files#diff-59a57960f4b6f12ce55d8bfe0635cdf29d12b023097c6d3cf50266b495227473L77-L83). I believe it's like this to eliminate the border-bottom that we talked about in Slack. Let me know if this is alright.

scurker (dismissed):

scurker (approved):
🎉 Nice Work!

PR #1373, opened 2024-02-29
test: switch stepper tests to react testing library

lsprr:
Issue #1159

This PR switches the Stepper tests to React Testing Library.

### Validation steps

- Run yarn test


packages/react/src/components/Stepper/index.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <Step status="current">
+      <span>Foo</span>
+    </Step>
+  );
+  expect(screen.getByRole('listitem')).toBeInTheDocument();
+  expect(screen.getByText('Foo')).toBeInTheDocument();
+  expect(screen.getByRole('listitem')).toContainElement(
+    screen.getByText('Foo')
+  );
+  expect(screen.getByText('Foo')).toHaveTextContent('Foo');
```
scurker, line 48: I think we can omit this one, it seems a little strange to get the element by its text then validate that it has that text. The `toBeInTheDocument` above should cover us.

```suggestion
```

scurker (changes requested):
One very minor nit, otherwise change LGTM!

scurker (approved):
🚨 Reviewed for Security 🚨

Issue #1086, opened 2023-05-30
Sync NavBar with uxpin


scurker:
Very minor changes needed. 

Per @bobbyomari there should be a slight left/right padding (see [uxpin](https://app.uxpin.com/design-system/c933c74928e2195cc540/ui-patterns/213525)) with `--space-small` at `< 64rem`.

Additionally in cauldron there appears to be an extra border creating a 2px line on the bottom of navbar. There should only be a 1px border.

PR #1375, opened 2024-03-01
fix(react): fix collapsed navbar styles

scurker:
ref #1086 

This is re-applying a few styles to fix some issues that were introduced with #1303. This also brings it into alignment with the recent designs including the box-shadow for the expanded mobile nav.


thuey:
Did we get clarification on whether or not it's intentional to remove the bottom border of the entire nav when not collapsed?

PR #1376, opened 2024-03-01
test: switch tab panel tests to react testing library

lsprr:
Issue #1159

This PR switches the Tab Panel tests to React Testing Library.

### Validation steps

- Run yarn test


packages/react/src/components/Tabs/tabPanel.test.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useRef } from 'react';
```
scurker, line 1: Very minor nit on the test file name, it should be `TabPanel.test.tsx`

packages/react/src/components/Tabs/tabPanel.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <TabPanel>
+          <p>a simple paragraph</p>
+          <p>a complicated paragraph</p>
+        </TabPanel>
+      </div>
+    );
+  };
+
+  render(<TabPanelwithRef />);
+  await waitFor(() => {
```
scurker, line 19: Just curious, what was the `waitFor` added for? This seems to work without it.

packages/react/src/components/Tabs/tabPanel.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    expect(screen.getAllByRole('tabpanel')).toHaveLength(1);
```
scurker, line 20:
```suggestion
    expect(screen.getAllByRole('tabpanel')).toBeInTheDocument();
```

packages/react/src/components/Tabs/tabPanel.test.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useRef } from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
+import { TabPanel } from './';
+
+test('should render tab panel with two paragraphs correctly', async () => {
+  const TabPanelwithRef = () => {
+    const ref = useRef(null);
+    return (
+      <div ref={ref}>
```
scurker, line 9: To be perfectly honest, I don't really remember why we have this wrapped in a ref, so we may be able to omit the wrapping `div` and just render `TabPanel` directly for this and the other examples.

packages/react/src/components/Tabs/tabPanel.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+test('should pass classNames through', async () => {
+  const TabPanelwithRef = () => {
+    const ref = useRef(null);
+    return <TabPanel ref={ref} className="find--me" />;
+  };
+
+  render(<TabPanelwithRef />);
+  await waitFor(() => {
+    expect(screen.getByRole('tabpanel')).toHaveClass('find--me');
```
scurker, line 32:
```suggestion
    expect(screen.getByRole('tabpanel')).toHaveClass('TabPanel', 'find--me');
```

packages/react/src/components/Tabs/tabPanel.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+  };
+
+  render(<TabPanelwithRef />);
+  await waitFor(() => {
+    expect(screen.getByRole('tabpanel')).toHaveAttribute(
+      'id',
+      'I am a panelId'
+    );
+  });
+});
```
scurker, line 49: This is an older test so it's missing a few newer tests that we started to require, including testing that `ref` prop is supported, and running the component through axe.

scurker (changes requested):

packages/react/src/components/Tabs/tabPanel.test.tsx
lsprr: Thank you for pointing that out. I forgot to remove it, and it is not necessary to include.

packages/react/src/components/Tabs/tabPanel.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const ref = React.createRef<HTMLDivElement>();
+  const TabPanelwithRef = () => {
+    return <TabPanel ref={ref}>Content</TabPanel>;
+  };
+
+  render(<TabPanelwithRef />);
+  await waitFor(() => {
+    expect(ref.current).toBeInTheDocument();
+    expect(ref.current).toHaveTextContent('Content');
+  });
```
scurker, line 36: Just a very very minor nit, I think we can just render the tab panel with the ref directly.

```tsx
render(<TabPanel ref={ref}>Content</TabPanel>)
```

And I think based on the above discussion we can also omit the `waitFor`.

scurker (changes requested):
One last change, and we should be good to go.

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

PR #1377, opened 2024-03-02
test: switch tab tests to react testing library

lsprr:
Issue #1159

This PR switches the Table tests to React Testing Library.

### Validation steps

- Run yarn test


ryomtoob (commented):
What's here looks great! I'm wondering if we should add a test for the `ref` property. Also curious about scanning with axe. Are we thinking it's not needed here because `<Tab>` will always be used as a child of `<Tabs>`? Would love to get your thoughts @scurker 😄

scurker:
> What's here looks great! I'm wondering if we should add a test for the `ref` property. Also curious about scanning with axe. Are we thinking it's not needed here because `<Tab>` will always be used as a child of `<Tabs>`? Would love to get your thoughts @scurker 😄

I agree, since we support `ref` here we should add that as a test case similarly to what we did with other components.

I do think accessibility tests might still be useful here. A `role="tab"` element even in isolation shouldn't have any accessibility violations.

packages/react/src/components/Tabs/tab.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+test('returns no axe violations', async () => {
+  const ref = React.createRef<HTMLDivElement>();
+
+  const { container } = render(
+    <Tabs aria-label={ariaLabel}>
+      <Tab target={ref} id={'I am a tabId'} role="tab">
+        <p>a simple paragraph</p>
+      </Tab>
+    </Tabs>
```
anastasialanz, line 95: The `id` of the tab gets automatically set by whatever is inside the `Tab` element. We also don't need to set `role="tab"` because that gets set automatically.

```suggestion
      <Tab target={ref}>
         Tab 1
      </Tab>
    </Tabs>
```

packages/react/src/components/Tabs/tab.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+  expect(screen.getAllByRole('tab')[0]).toBeInTheDocument();
+  expect(screen.getAllByRole('tab')[1]).not.toHaveClass('Tab--active');
+  expect(screen.getAllByRole('tab')[1]).toBeInTheDocument();
+});
+
+test('should pass id through', async () => {
+  const ref = React.createRef<HTMLDivElement>();
+
+  render(
+    <Tab target={ref} id={'I am a tabId'}>
```
anastasialanz, line 42:
```suggestion
    <Tab target={ref}>
```

packages/react/src/components/Tabs/tab.test.tsx
anastasialanz, line 95: We don't need to set `role="tab"` because that gets set automatically. Valid `id`'s must not contain any space characters.

```suggestion
    <Tabs aria-label={ariaLabel}>
      <Tab target={ref} id="I-am-a-tabId">
        <p>a simple paragraph</p>
      </Tab>
    </Tabs>
```

packages/react/src/components/Tabs/tab.test.tsx
anastasialanz, line 42:
```suggestion
    <Tab target={ref} id="I-am-a-tabId">
```

packages/react/src/components/Tabs/tab.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+      option 1
+    </Tab>
+  );
+
+  expect(screen.getByRole('tab')).toHaveAttribute('id', 'I am a tabId');
```
anastasialanz, line 47:
```suggestion
  expect(screen.getByRole('tab')).toHaveAttribute('id', 'I-am-a-tabId');
```

anastasialanz (approved):

scurker (approved):
Reviewed for security

PR #1378, opened 2024-03-02
test: switch tabs tests to react testing library

lsprr:
Issue #1159

This PR switches the Tabs tests to React Testing Library.

### Validation steps

- Run yarn test


packages/react/src/components/Tabs/Tabs.tsx
```diff
@@ -6,6 +6,7 @@ import { useId } from 'react-id-generator';
 import useDidUpdate from '../../utils/use-did-update';
 
 type TabsProps = {
+  id?: string;
```
lsprr, line 9: While transitioning the old test from Enzymes to RTL and rewiring the entire code to align with RTL best practices, I gave it a go. Discovered a couple of hiccups in Line 220 and 334 related to the ID, but I sorted it out by adding the "id" to TabsProps.

packages/react/src/components/Tabs/tabs.test.tsx
```diff
@@ -0,0 +1,411 @@
...
+        </TabPanel>
+        <TabPanel ref={ref2}>
+          <p>Panel 2</p>
+        </TabPanel>
+      </>
+    );
+  };
+
+  render(<TabsComponent />, { container: mountElement });
+  fireEvent.keyDown(screen.getByText('option 2'), { key: 'ArrowRight' });
```
ryomtoob, line 207: According to [the RTL docs](https://testing-library.com/docs/guide-events/#keydown), I believe it's best practice to focus the element first then fire the key down event on the active element. 
```suggestion
  screen.getByText('option 2').focus()
  fireEvent.keyDown(document.activeElement || document.body, { key: 'ArrowRight' });
```

packages/react/src/components/Tabs/Tabs.tsx
ryomtoob: hmm I don't see `id` listed as a prop in [the documentation for this component](https://cauldron.dequelabs.com/components/Tabs#tabs), however it looks like the old tests had a couple cases for this. @scurker is this a feature that we'd want to support and hence add to the docs? Or should we not add the prop and remove the tests for "customized id"?

packages/react/src/components/Tabs/tabs.test.tsx
```diff
@@ -0,0 +1,411 @@
...
+  fireEvent.click(screen.getAllByRole('tab')[1]);
+  await waitFor(() => {
+    expect(screen.getAllByRole('tabpanel')[0]).toHaveClass('TabPanel--hidden');
+    expect(screen.getAllByRole('tabpanel')[1]).not.toHaveClass(
+      'TabPanel--hidden'
+    );
+  });
+});
+
+test('should display correct tabpanel when pressing left, right, home, or end keys', async () => {
```
ryomtoob, line 244: This doesn't actually test the left keydown event. I'm guessing that's because it's covered in the test on line 179, maybe we should just combine them?

ryomtoob, line 244: I think this test case is missing the logic for the arrow left keydown event.

ryomtoob (commented):
I left a few comments, but overall this conversion looks good! In addition to converting the existing cases, this is an opportunity to evaluate if we should add anything to increase coverage and better follow [the CONTRIBUTING guidelines](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#testing-strategies). That being said, I think we should make sure we are running axe on all the component states. Right now, we're missing the vertical and thin variants.

lsprr:
> I left a few comments, but overall this conversion looks good! In addition to converting the existing cases, this is an opportunity to evaluate if we should add anything to increase coverage and better follow [the CONTRIBUTING guidelines](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#testing-strategies). That being said, I think we should make sure we are running axe on all the component states. Right now, we're missing the vertical and thin variants.

I have just added those variants to the test file.

packages/react/src/components/Tabs/Tabs.tsx
scurker: I think this component existed before we implemented typescript, which is probably why that initially slipped through. Changing of props would change this from `test` to `fix`, so I think it would be fine to omit those tests for now and not include the id as looking at the tests I'm scratching my head trying to figure out what we were actually testing for.

ryomtoob: Makes sense to me! @lsprr could you please remove the id prop that you added as well as the tests for "customized id"?

packages/react/src/components/Tabs/tabs.test.tsx
```diff
@@ -0,0 +1,402 @@
...
+        <TabPanel ref={ref}>
+          <p>Panel 1</p>
+        </TabPanel>
+      </>
+    );
+  };
+
+  render(<TabswithRef />);
+  await waitFor(() => {
+    expect(screen.getByRole('tablist').parentElement).toHaveClass('find--me');
```
scurker, line 89:
```suggestion
    expect(screen.getByRole('tablist').parentElement).toHaveClass('Tabs', 'find--me');
```

packages/react/src/components/Tabs/tabs.test.tsx
```diff
@@ -0,0 +1,402 @@
...
+        </Tabs>
+        <TabPanel ref={ref}>
+          <p>Panel 1</p>
+        </TabPanel>
+      </>
+    );
+  };
+
+  render(<TabswithRef />);
+  expect(screen.getByRole('tablist')).toHaveAttribute('aria-label', 'find-me');
```
scurker, line 109: In this case, checking for the accessible name would be better since that's what the user expects:

```suggestion
  expect(screen.getByRole('tablist')).toHaveAccessibleName('aria-label');
```

packages/react/src/components/Tabs/tabs.test.tsx
```diff
@@ -0,0 +1,402 @@
...
+        </Tabs>
+        <TabPanel ref={ref}>
+          <p>Panel 1</p>
+        </TabPanel>
+      </>
+    );
+  };
+
+  render(<TabswithRef />);
+  expect(screen.getByRole('tablist')).toHaveAttribute(
```
scurker, line 128: Similar to above, but we should have an id ref for `find-me` that the tablist can be labelledby.

scurker (commented):
Just a few minor things, looking good!

anastasialanz (approved):
Looks great! Thank you!

anastasialanz:
🚨 Reviewed for Security 🚨

PR #1381, opened 2024-03-05
test: switch alert tests to react testing library

lsprr:
Issue #1159

This PR switches the Alert tests to React Testing Library.

Validation steps

- Run yarn test


packages/react/src/components/Alert/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const defaults = { show: false, heading: { text: 'hi' } };
+
+test('should return null when passed a falsey "show" prop', () => {
+  render(
+    <Alert data-testid="alert" {...defaults}>
+      Test Alert
+    </Alert>
+  );
+  expect(screen.queryByTestId('alert')).toBeNull();
```
scurker, line 14: This was an older test style, so we can clean this up slightly to check that it does not render when show is false:

```suggestion
  expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
```

packages/react/src/components/Alert/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+});
+
+test('should show modal when passed a truthy "show" prop', () => {
+  render(
+    <Alert data-testid="alert" {...defaults} show>
+      Test Alert
+    </Alert>
+  );
+  expect(screen.getByTestId('alert')).toBeInTheDocument();
```
scurker, line 23:
```suggestion
  expect(screen.queryByRole('dialog')).toBeInTheDocument();
```

scurker (changes requested):
The only other things we need to check for here are some of the different props that we weren't testing for before, including:

- `ref`
- `className` (includes 'Alert' and custom classnames)

Once these few things are addressed, it should be good to go!

scurker (approved):
🚨 Reviewed for Security 🚨

Issue #1358, opened 2024-02-23
Search Field Enhancements


scurker:
Outlined are additional enhancements needed for `SearchField`: #1190 

Include a right-oriented "slot" to allow for additional input context to be provided, such as texts, buttons, icons, etc. To achieve this, `SearchField` takes an additional prop named `trailingChildren`, which will be "slotted" into the component. 

Here are a few characteristics of `trailingChildren`:

- it can be anything including strings, elements, or components. 
- it is a child of `TextFieldWrapper` within `SearchField` and is rendered after the native element `input` . 
- visually it is placed after the typing area (including the cancel button) of `TextFieldWrapper`, which is styled as an `input`

Here is a visual example of `SearchField` with the text `1 of 1` and two `IconButton`s as `trailingChildren`:

![search field showing an "X" indicator with a "1 of 1" text and navigational up/down chevron buttons](https://github.com/dequelabs/cauldron/assets/1062039/53140e0d-4f9f-447a-ab06-211105e26e2c)

Here is an usage example:

```js
<SearchField 
  trailingChildren={<MyComponent />} 
  // other props
/>
```

Here is a code snippet of a possible implementation of `trailingChildren` within the `SearchField` component: 

```js
const SearchField = ({
  trailingChildren = "I am a string"
  // other props
}) => {
  // other implementation details
  ...
       <TextFieldWrapper
          className={classNames({
            'TextFieldWrapper--disabled': otherProps.disabled
          })}
       >
          <Icon type="magnifying-glass" className="SearchField__search-icon" />
          <input
            id={inputId}
            value={value}
            onChange={handleChange}
            placeholder={placeholder}
            ref={ref}
            {...otherProps}
            className={classNames(otherProps.className, 'Field__text-input')}
            type="search"
          />
          <span>{trailingChildren}</span>
       </TextFieldWrapper>
}
```

Inside the `SearchField` component, the `trailingChildren` is placed after the native element `input` wrapped by `TextFieldWrapper`. We render it based on the type of `trailingChildren`. In the example above, since we have a string, we render it in a `span`.

JoshMK:
The proposed enhancements look good and make sense. 👍  I just have one question:

In the example implementation, we hardcode a span as the container for `trailingChildren`. Should we give users control to determine the container element? From `<span>{trailingChildren}</span>` to `{trailingChildren}` 👈 (now includes the container itself.)

[...]

PR #1385, opened 2024-03-07
feat(react): Add property to include trailing children for SearchField component

dequejosie:
closes https://github.com/dequelabs/cauldron/issues/1358

### DIFF
--- a/docs/index.css
+++ b/docs/index.css
@@ -151,6 +151,11 @@ h6:has(> .Permalink):hover .Permalink,
   margin: var(--space-large);
 }
 
+.SearchField .TextFieldWrapper .page {
+  flex-shrink: 0;
+  margin-right: var(--space-smallest);
+}
+
 @media (max-width: 64rem) {
   .Content {
     grid-template-columns: 1fr;
--- a/docs/index.css
+++ b/docs/index.css
@@ -151,11 +151,6 @@ h6:has(> .Permalink):hover .Permalink,
   margin: var(--space-large);
 }
 
-.SearchField .TextFieldWrapper .page {
-  flex-shrink: 0;
-  margin-right: var(--space-smallest);
-}
-
 @media (max-width: 64rem) {
   .Content {
     grid-template-columns: 1fr;
--- a/docs/pages/components/SearchField.mdx
+++ b/docs/pages/components/SearchField.mdx
@@ -57,11 +57,11 @@ By default, SearchField is wrapped in a `<form role="search">` to provide better
 
 ### SearchField with trailing children
 
-We include a right-oriented "slot" to allow for additional input context to be provided, such as texts, buttons, icons, etc. To achieve this, SearchField takes an additional prop named trailingChildren, which will be "slotted" into the component as a child of TextFieldWrapper after the native input.
+We include trailing children to allow for additional input context to be provided, such as texts, buttons, icons, etc. The trailing children will be placed at the right end of the SearchField.
 
 ```jsx example
 <FieldWrap>
-  <SearchField label="Search Field with trailing children" trailingChildren={<span class="page">
+  <SearchField label="Search Field with trailing children" trailingChildren={<span style={{ marginRight: '8px'}}>
     <strong>1</strong> of <strong>1</strong>
   </span>} />
 </FieldWrap>
--- a/docs/pages/components/SearchField.mdx
+++ b/docs/pages/components/SearchField.mdx
@@ -4,7 +4,7 @@ description: A form element that allows users to type in text for searching.
 source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/SearchField/index.tsx
 ---
 
-import { SearchField, FieldWrap } from '@deque/cauldron-react';
+import { SearchField, FieldWrap, IconButton } from '@deque/cauldron-react';
 
 ```js
 import { SearchField } from '@deque/cauldron-react';
@@ -55,6 +55,16 @@ By default, SearchField is wrapped in a `<form role="search">` to provide better
 </form>
 ```
 
+### SearchField with trailingChildren
+
+```jsx example
+<FieldWrap>
+  <SearchField label="Search Field with trailingChildren" trailingChildren={<span class="page">
+    <strong>1</strong> of <strong>1</strong>
+  </span>} />
+</FieldWrap>
+```
+
 ## Props
 
 <ComponentProps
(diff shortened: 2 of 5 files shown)


docs/pages/components/SearchField.mdx
```diff
@@ -96,6 +106,12 @@ By default, SearchField is wrapped in a `<form role="search">` to provide better
       description:
         'If true, the label and input will be rendered in a `<form>` tag. If false, the label and input will be rendered in a `<div>` tag.',
       defaultValue: 'true'
+    },
+    {
+      name: 'trailingChildren',
+      type: ['ReactNode, string'],
```
dequejosie, line 112: Debating between ReactNode vs ReactElement, or vs both. Not 100% sure about the difference.

docs/pages/components/SearchField.mdx
```diff
@@ -55,6 +55,16 @@ By default, SearchField is wrapped in a `<form role="search">` to provide better
 </form>
 ```
 
+### SearchField with trailingChildren
```
scurker, line 58: Nit:

```suggestion
### SearchField with Trailing Children
```

Also, can we describe a little more about the use cases here about when/why we would want to use this?

docs/pages/components/SearchField.mdx
```diff
@@ -55,6 +55,16 @@ By default, SearchField is wrapped in a `<form role="search">` to provide better
...
+
+```jsx example
+<FieldWrap>
+  <SearchField label="Search Field with trailingChildren" trailingChildren={<span class="page">
```
scurker, line 62:
```suggestion
  <SearchField label="Search Field with Trailing Children" trailingChildren={<span class="page">
```

docs/pages/components/SearchField.mdx
scurker, line 112: [`ReactElement`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L318-L325) is an element/component type where `ReactNode` covers a wider range. `ReactNode` also already includes string so we should just be able to use a single type here.

```suggestion
      type: 'ReactNode',
```

packages/react/src/components/SearchField/index.tsx
```diff
@@ -99,6 +101,12 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
             className={classNames(otherProps.className, 'Field__text-input')}
             type="search"
           />
+          {trailingChildren &&
```
scurker, line 104: Nit: We should move this up above to the body:

```jsx
if (trailingChildren === 'string') {
  trailingChildren = <span>{trailingChildren}</span>
}
```

Then we can just render `{trailingChildren}` in the component since `null` or `undefined` values won't render anything.

packages/react/src/components/SearchField/index.tsx
```diff
@@ -21,6 +21,7 @@ interface SearchFieldProps
   onChange?: (value: string, e: ChangeEvent<HTMLInputElement>) => void;
   hideLabel?: boolean;
   isForm?: boolean;
+  trailingChildren?: React.ReactNode | string;
```
scurker, line 24: `React.ReactNode` already includes string.

```suggestion
  trailingChildren?: React.ReactNode;
```

packages/react/src/components/SearchField/SearchField.test.tsx
```diff
@@ -150,6 +150,31 @@ test('should support name prop', () => {
   expect(input).toHaveDisplayValue('bananas');
 });
 
+test('SearchField should render trailingChildren as a component', async () => {
```
scurker, line 153: These tests and the other tests below don't actually validate that the component is rendered correctly since it should be a child of `TextFieldWrapper`.

What we likely want here is to check that the expected component contains the element:

```jsx
const input = render()
expect(input.parentElement.contains(screen.getByText('1 of 1'))).toBeTruthy() 
```

I would also write the tests more like the actual intended usage, so it makes the patterns more clear.

packages/react/src/components/SearchField/SearchField.test.tsx
```diff
@@ -150,6 +150,31 @@ test('should support name prop', () => {
...
+});
+
+test('SearchField should render trailingChildren as an element', async () => {
+  render({
+    trailingChildren: <div className="element">I am a component</div>
+  });
+
+  expect(screen.getByText('I am a component')).toHaveClass('element');
+});
+
```
scurker, line 177: Since this is an additional "state" of the component, we should test it with axe below.

docs/index.css
```diff
@@ -151,6 +151,11 @@ h6:has(> .Permalink):hover .Permalink,
   margin: var(--space-large);
 }
 
+.SearchField .TextFieldWrapper .page {
+  flex-shrink: 0;
+  margin-right: var(--space-smallest);
+}
+
```
scurker, line 158: We've largely moved away from using global documentation styles here, you can see [Panel](https://cauldron.dequelabs.com/components/Panel) as an example. You could include a style property in the documentation to help demonstrate that the component does not come with inherited styles.

I'm also thinking `flex-shrink` may need to be applied at the component level since we likely don't want components in the wrapper to wrap:

```css
.TextFieldWrapper > :not(input) {
  flex-shrink: 0;
}
```

scurker (changes requested):

docs/pages/components/SearchField.mdx
```diff
@@ -55,6 +55,18 @@ By default, SearchField is wrapped in a `<form role="search">` to provide better
...
+### SearchField with trailing children
+
+We include trailing children to allow for additional input context to be provided, such as texts, buttons, icons, etc. The trailing children will be placed at the right end of SearchField.
+
+```jsx example
+<FieldWrap>
+  <SearchField label="Search Field with trailing children" trailingChildren={<span style={{ marginRight: '8px'}}>
+    <strong>1</strong> of <strong>1</strong>
+  </span>} />
+</FieldWrap>
```
scurker, line 67: Nit: Formatting

```suggestion
<FieldWrap>
  <SearchField 
    label="Search Field with trailing children" 
    trailingChildren={
      <span style={{ marginRight: '8px' }}>
        <strong>1</strong> of <strong>1</strong>
      </span>
    }
  />
</FieldWrap>
```

scurker (changes requested):
One small change to make docs more readable, code LGTM!

scurker (approved):

PR #1386, opened 2024-03-08
test: switch select tests to react testing library

lsprr:
Issue #1159

This PR switches the Select tests to React Testing Library.

Validation steps

- Run yarn test


packages/react/src/components/Select/index.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+const withCustomOptions = (otherProps = {}) => {
+  const options = [
+    { key: '1', value: 'Fred' },
+    { key: '2', value: 'Bill' },
+    { key: '3', value: 'Ted' },
+    { key: '4', value: 'Bob' }
+  ];
+
+  return render(
+    <div>
```
scurker, line 21: Just curious, is the outer `<div>` needed here for any reason? I realize this was converted over from the old test but would be nice to clean it up if it's not needed.

packages/react/src/components/Select/index.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+      options={[
+        { key: '1', value: 'a' },
+        { disabled: true, key: '2', value: 'b' },
+        { key: '3', value: 'c' }
+      ]}
+    />
+  );
+
+  expect(screen.getAllByRole('option')).toHaveLength(3);
+  expect(screen.getByText('b')).toBeInTheDocument();
```
scurker, line 59: This checks that it exists in the document, but we probably want something more strict here such as there's an option role with the desired text.

packages/react/src/components/Select/index.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+  render(
+    <Select {...defaultProps} defaultValue="a">
+      <option>a</option>
+      <option disabled>b</option>
+      <option>c</option>
+    </Select>
+  );
+
+  expect(screen.getAllByRole('option')).toHaveLength(3);
+  expect(screen.getByText('b')).toBeInTheDocument();
```
scurker, line 88: Same issue as above regarding option role.

packages/react/src/components/Select/index.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  expect(screen.getByText(errorText)).toBeInTheDocument();
+  expect(screen.getByText(errorText)).toHaveAttribute(
+    'id',
+    screen.getByText(errorText).id
+  );
+  expect(screen.getByRole('combobox')).toHaveAttribute(
+    'aria-describedby',
+    screen.getByText(errorText).id
+  );
```
scurker, line 111: In this instance, we can use 
```suggestion
  expect(screen.getByRole('combobox')).toHaveAccessibleDescription(errorText);
```

packages/react/src/components/Select/index.test.tsx
```diff
@@ -0,0 +1,263 @@
...
+        options={options}
+      />
+    </div>
+  );
+};
+
+test('Should render the expected UI elements', () => {
+  withCustomOptions();
+
+  expect(screen.getByText(defaultProps.name)).toBeInTheDocument();
```
scurker, line 35: We probably want to ensure that an element with the right role exists here:

```tsx
expect(screen.getByRole('combobox')).toBeInTheDocument()
```

scurker (changes requested):
Just a few minor things, everything else LGTM!

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

PR #1400, opened 2024-03-08
feat(react): add prop to SearchField to render error messages

dequejosie:
Follow up of https://github.com/dequelabs/cauldron/issues/1190 to render error messages as in the following case.

Screenshot 2024-03-08 at 3 01 56 PM


scurker (changes requested):
@shawnsharpDQ and I discussed this, as this was a bit of a surprise to me as I'm not sure we want Search field to visually error in this way.

We discussed this on slack but we may not need any changes in Cauldron as this may be a pattern that exists per each search component.

docs/pages/components/SearchField.mdx
```diff
@@ -35,6 +35,14 @@ import { SearchField } from '@deque/cauldron-react';
...
 ```
 
+### Error
+
+```jsx example
+<FieldWrap>
+  <SearchField label="Error" error="This field has an error" />
+</FieldWrap>
+```
+
```
yhafez, line 45: I think @scurker mentioned to me in a DM that we don't need to handle error states for this component (since searches don't usually error)

yhafez: Oh jk, looks like you already commented

yhafez: @dequejosie is there a need for an error state in what you're implementing in the extension?

dequejosie:
@shawnsharpDQ and I discussed about this, we will go for a different approach.

PR #1401, opened 2024-03-08
feat: adjust TagButton icon size and add position prop

yhafez:


packages/styles/tag-button.css
```diff
@@ -1,16 +1,25 @@
...
+.Icon-left {
+  margin-right: var(--space-half);
+}
+
+.Icon-center {
+  margin-left: var(--space-quarter);
+  margin-right: var(--space-quarter);
+}
+
+.TagButton .TagButton__icon svg {
```
yhafez, line 22: This was being overridden by `.Icon svg`, so I increased its specificity

scurker (changes requested):
I think the only change that can land for sure right now is the icon size change.

Knowing what we know now, this is not likely how we would implement icon customization in this component - so that's a bit of a miss on our part. I think this is in a bit of a holding pattern and @shawnsharpDQ is aware but wanted to at least let you know we're still looking into what we actually need to do here.

PR #1403, opened 2024-03-09
test: switch nav bar tests to react testing library

lsprr:
Issue #1159

This PR switches the Nav Bar tests to React Testing Library.

Validation steps

- Run yarn test


packages/react/src/components/NavBar/navBar.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+      </NavItem>
+    </NavBar>
+  );
+
+  fireEvent.click(screen.getByRole('button'));
+  fireEvent.focusIn(screen.getByRole('listitem'));
+  expect(screen.getByRole('listitem')).toBeInTheDocument();
+});
+
+test('returns no axe violations', async () => {
```
scurker, line 155: Only thing missing is running axe against both collapsed/not-collapsed states.

scurker (changes requested):
Sorry I'm so far behind on these, but thank you again for your contribution! Just one small change needed here to test for an additional a11y state.

packages/react/src/components/NavBar/navBar.test.tsx
lsprr: No worries! I understand you're very busy. Anything for you and Deque. 😄

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

PR #1404, opened 2024-03-12
test: switch pagination tests to react testing library

lsprr:
Issue #1159

This PR switches the Pagination tests to React Testing Library.

Validation steps

- Run yarn test


packages/react/src/components/Pagination/index.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+      currentPage={1}
+      onFirstPageClick={onFirstPageClick}
+      onPreviousPageClick={onPreviousPageClick}
+    />
+  );
+
+  expect(screen.getAllByRole('button')[0]).toHaveAttribute(
+    'aria-disabled',
+    'true'
+  );
```
scurker, line 29: We should probably use the built-in jest matchers for this:

```suggestion
  expect(screen.getAllByRole('button')[0]).toBeDisabled()
```

packages/react/src/components/Pagination/index.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+  render(
+    <Pagination
+      totalItems={18}
+      currentPage={2}
+      onNextPageClick={onNextPageClick}
+      onLastPageClick={onLastPageClick}
+    />
+  );
+
+  expect(screen.getAllByRole('button')[0]).toHaveAttribute(
```
scurker, line 63: Same suggestion as above regarding disabled.

packages/react/src/components/Pagination/index.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+  expect(onNextPageClick).not.toHaveBeenCalled();
+  expect(onLastPageClick).not.toHaveBeenCalled();
+});
+
+test('should support custom status label text', () => {
+  render(
+    <Pagination totalItems={18} statusLabel={<div id="foo">hello world</div>} />
+  );
+
+  expect(screen.getByText('hello world')).toHaveAttribute('id', 'foo');
```
scurker, line 92: The one change I would make here is to do what you did below to check the log role:

```tsx
expect(screen.getByRole('log')).toHaveTextContent('Showing 35 to 51 of 500');
```

scurker (changes requested):
Just a few minor changes needed, comments below.

packages/react/src/components/Pagination/index.test.tsx
lsprr: Hi @scurker, I cannot use toBeDisabled() because the element does not have the disabled attribute. toBeDisabled() specifically checks for the presence of the disabled attribute to determine if an element is disabled. However, in our case, we're controlling the disabled state using the aria-disabled attribute instead of the disabled attribute. I can include the disabled state in the component if you'd like. This would require creating a new PR for the changes. Let me know your thoughts.

packages/react/src/components/Pagination/index.test.tsx
lsprr: See my comment as above regarding disabled.

packages/react/src/components/Pagination/index.test.tsx
lsprr: Here it is. 

`<button class="IconButton IconButton--secondary" tabindex="0" type="button" aria-disabled="true" aria-labelledby="tooltip18"><span aria-hidden="true" class="Icon Icon--chevron-double-left Icon__left"><svg overflow="visible" preserveAspectRatio="none" viewBox="0 0 24 24" height="24" width="24"><path d="M17.2 11.84a.35.35 0 0 0-.1-.24l-.52-.52a.332.332 0 0 0-.47-.01l-.01.01-4.1 4.1-4.09-4.1a.332.332 0 0 0-.47-.01l-.01.01-.52.52c-.13.13-.14.33-.01.47l.01.01 4.85 4.85c.13.13.33.14.47.01l.01-.01 4.85-4.85c.06-.06.1-.15.1-.24h.01zm0-4a.35.35 0 0 0-.1-.24l-.52-.52a.332.332 0 0 0-.47-.01l-.01.01-4.1 4.1-4.09-4.09a.333.333 0 0 0-.47-.02l-.01.01-.52.53a.32.32 0 0 0-.02.46l.01.01 4.85 4.85c.13.13.33.14.47.01l.01-.01 4.85-4.85c.06-.06.1-.15.1-.24h.02z" fill="currentColor" vector-effect="non-scaling-stroke"></path></svg></span></button>`

scurker: Yup, you're right - we can disregard this.

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

PR #1416, opened 2024-03-18
test: switch skip link tests to react testing library

lsprr:
Issue #1159

This PR switches the Skip Link tests to React Testing Library.

Validation steps

- Run yarn test


packages/react/src/components/SkipLink/index.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+    />
+  );
+};
+
+test('should focus main content when SkipLink is clicked', () => {
+  expect.assertions(2);
+
+  const target = window.document.createElement('div');
+  target.id = 'main-content';
+  window.document.body.appendChild(target);
```
scurker, line 26: Just one small change here, we should probably but this fixture into before/after each hooks to ensure the global document cleans up after itself:

```ts
let target: HTMLDivElement

afterEach(() => {
  target = window.document.createElement('div');
  target.id = 'main-content';
  window.document.body.appendChild(target);
})

beforeEach(() => {
  window.document.body.removeChild(target);
})
```

scurker (commented):
Just one small change needed.

scurker (approved):
Thank you, LGTM! 🎉

scurker:
🚨 Reviewed for Security 🚨

PR #1425, opened 2024-03-20
fix(react): prevent search field input from inheriting unintended styles from default TextField

### LINKED ISSUE #1424: Bug: Search field can expand beyond it's container
### Steps to reproduce

Have a search field in an element, then shrink the page down (example below)

### Expected Behavior

`SearchField` should responsively adapt.

### Actual Behavior

example from cauldron docs with search field overflows outside of its container

### Version

6.x

scurker:
closes #1424

### DIFF
--- a/packages/react/src/components/SearchField/index.tsx
+++ b/packages/react/src/components/SearchField/index.tsx
@@ -102,7 +102,6 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
             placeholder={placeholder}
             ref={ref}
             {...otherProps}
-            className={classNames(otherProps.className, 'Field__text-input')}
             type="search"
           />
           {trailingChildren}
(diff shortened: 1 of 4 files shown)


packages/react/src/components/SearchField/index.tsx
```diff
@@ -101,8 +102,7 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
             onChange={handleChange}
             placeholder={placeholder}
             ref={ref}
-            {...otherProps}
-            className={classNames(otherProps.className, 'Field__text-input')}
```
scurker, line 105: This was likely a mistake in the initial implementation, `TextFieldWrapper` should appropriately handle any styles associated with input and this component should not have this class.

yhafez (approved):

PR #1426, opened 2024-03-20
test: switch notice tests to react testing library

lsprr:
Issue #1159

This PR switches the Notice tests to React Testing Library.

### Validation steps

- Run yarn test


packages/react/src/components/Notice/index.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+test('should allow a ref to be forwarded', () => {
+  const ref = React.createRef<HTMLDivElement>();
+
+  render(<Notice data-testid="notice" ref={ref} title={''} />);
+
+  expect(screen.getByTestId('notice')).toBeInTheDocument();
+  expect(ref.current).toBeTruthy();
+});
+
+test('should return no axe violations', async () => {
```
scurker, line 78: Can we break this down into different types? That way we know specifically which case failed if someone broke something in the future.

scurker (changes requested):
Just one very minor thing with a11y test, otherwise looks good!

scurker (approved):
🚨 Reviewed for Security 🚨

PR #1437, opened 2024-04-01
fix: fix toast bug where screen readers dont read toast

### LINKED ISSUE #1436: Bug: Screen reader doesn't read `Toast` content if `Pagination` component is on page
### Steps to reproduce

Steps to reproduce

1) Add a `Pagination`, `Loader`, and `Toast` component to the DOM.
2) Trigger the toast to load after the loader is dismissed.

### Expected Behavior

Loading content is read, then pagination content is read, then toast content is read by screen reader.

### Actual Behavior

Loading content is read Pagination content is read, but Toast component content is not read by screen reader.

### Version

6.3.1

Chrome

### Relevant log output

_No response_

_No response_

yhafez:
Closes #1436

### DIFF
--- a/packages/react/src/components/Pagination/Pagination.tsx
+++ b/packages/react/src/components/Pagination/Pagination.tsx
@@ -79,7 +79,7 @@ const Pagination = React.forwardRef<HTMLDivElement, Props>(
           </li>
 
           <li>
-            <span role="log" aria-atomic="true">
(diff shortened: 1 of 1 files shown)


scurker (approved):
Code changes LGTM, might be good to describe in more detail as to why we're making this change since `role="log"` has an implicit role of `aria-live="polite"`.

PR #1443, opened 2024-04-09
test: switch panel tests to react testing library

lsprr:
Issue #1159

This PR switches the Panel tests to React Testing Library.

Validation steps

- Run yarn test


packages/react/src/components/Panel/index.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+test('should render a custom heading level', () => {
+  render(<Panel heading={{ text: 'Title', level: 3 }}>Content</Panel>);
+
+  expect(screen.getByRole('heading', { name: 'Title' }).tagName).toBe('H3');
+  expect(screen.getByText('Content')).toBeInTheDocument();
+});
+
+test('should render the label the panel by the heading by default', () => {
+  render(<Panel heading={'Title'}>Content</Panel>);
```
scurker, line 21: Very minor style nit. (and for all other <code>{`Title`}</code> ones.

```suggestion
  render(<Panel heading="Title">Content</Panel>);
```

packages/react/src/components/Panel/index.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+  const panel = screen.getByRole('region');
+  const heading = screen.getByRole('heading', { name: 'Title' });
+  const body = screen.getByText('Content');
+
+  const headingId = heading.getAttribute('id');
+  const ariaLabelledBy = panel.getAttribute('aria-labelledby');
+
+  expect(heading.tagName).toBe('H2');
+  expect(ariaLabelledBy).toBe(headingId);
+  expect(body).toBeInTheDocument();
```
scurker, line 32: I realize this was in the original test, but some of these things are redundant since they're tested above. I'm also scratching my head on the test name here, because I think it's saying "should label the panel by the heading by default".

What I think may need to be tested here is validating that the panel has an accessible name (`screen.queryByRole('region').hasAccessibleName(...)`)

packages/react/src/components/Panel/index.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+});
+
+test('should render the label the panel by the heading id', () => {
+  render(
```
scurker, line 36: This is similar to the above, except that it should support the custom id for the heading.

scurker (commented):
Mostly minor style nits and a little bit of test cleanup, but looking good otherwise!

packages/react/src/components/Panel/index.test.tsx
lsprr: I took a look at my work and couldn't remember why I wrote it, so I decided to revise it to make more sense.

packages/react/src/components/Panel/index.test.tsx
lsprr: I've revised it to make more sense.

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

Issue #1445, opened 2024-04-10
Handle multiple toasts


scurker:
Cauldron has an existing limitation of only allowing a single toast to be displayed, meaning multiple toasts get stacked on top of each other. There's currently an unwritten rule that only a single toast should be displayed at a time, but we may have a need in the future for allow for toasts to stack or some other mechanism for displaying multiple toasts.

## Info Needed

- How do we want to display multiple toasts?
- Is there a change in the design to make toasts more in-line with toasts from other design systems?

Dependent on designs, #1364 might be part of whatever solution we land on here as well.

PR #1446, opened 2024-04-10
docs: prevent more than one toast notification from appearing for copy actions

### LINKED ISSUE #1395: [A11y] - Hidden or empty element receives focus
### Current Behaviour 
`Keyboard focus or touch screen swiping falls on an element that is hidden or empty. `<br/><hr/>
### Expected Behaviour 
`RULE :  The reading and navigation order MUST be logical and intuitive. All focusable elements MUST have a visual focus indicator when in focus. HOW TO FIX:  Fix this issue by using ONE of the following techniques: 1.If the element should not be visible or focusable, remove the element from the DOM or hide it from all users with CSS such as display:none. 2.If the element is meant to be visible but not interactive, make it visible and remove the focus state. 3.If the element is meant to be visible and interactive, then ensure that it is both focusable and visible to all users. `<br/><hr/>
### Impact 
 <p>

- [ ] Blocker
- [x] Serious
- [ ] Critical
- [ ] Moderate
- [ ] Minor

### Issue Details

<details><summary>View Issue 1513116 </summary> 
 <p>

_**Auditor ID**_ : [1513116](https://axeauditor.dequecloud.com/issue-global-search?searchKeyword=1513116&searchOn=id) 

_**description**_ :  Example copied to clipboard dismiss button receives focus but is not visible to the keyboard-only/ visible users.

Element location: On the "Accordion" page

Steps to reproduce: 
- Without dismissing the "example copied to clipboard" status message, trigger another Copy button. 
-The same message appears below the existing message. So when navigating using a tab, the dismiss button receives focus but is not visible to the keyboard-only/ visible users.

Note: It is applicable to all the pages 

_**Page Name**_ : Project Wide 

_**source**_ : `<button type="button" class="Toast__dismiss" aria-label="Dismiss"><span aria-hidden="true" class="Icon Icon--close">` 

_**Screenshots**_ : [Keyboard focus order](https://axeauditor.dequecloud.com/api/v1/file/df9d6d7e-5d0d-11ee-816d-036a4d3bb847) 

</p>
</details>

scurker:
closes #1395 

This is primarily to address the a11y issue raised, and a more comprehensive approach should be taken with #1445 which should replace the work being done here.

### DIFF
--- a/docs/components/CopyToClipboardButton.tsx
+++ b/docs/components/CopyToClipboardButton.tsx
@@ -27,6 +27,38 @@ function copyTextToClipboard(text: string) {
   return copied;
 }
 
+// Note: eventually we want to natively handle multiple toasts, but for now we will limit the copy toast notification
+// to only display a single notification to prevent weird focus issues
+const notificationsMap = new Map<Symbol, React.ComponentProps<Toast>>();
+function CopyNotificationToast(
+  props: React.ComponentProps<Toast>
+): JSX.Element {
+  const { show, onDismiss } = props;
+  const id = useMemo(() => Symbol('toast'), []);
+  const [deferredShow, setDeferredShow] = useState(false);
+
+  useEffect(() => {
+    if (show) {
+      Array.from(notificationsMap.values()).forEach(({ onDismiss }) => {
+        // force any open toasts to dismiss themselves
+        onDismiss();
+      });
+      notificationsMap.set(id, props);
+      // toast sets show via set timeout, so this matches the behavior to avoid a race condition
+      setTimeout(() => setDeferredShow(show));
+    } else {
+      notificationsMap.delete(id);
+      setDeferredShow(false);
+    }
+
+    return () => {
+      notificationsMap.delete(id);
+    };
+  }, [show]);
+
+  return <Toast {...props} show={deferredShow} />;
+}
+
 export default function CopyToClipboardButton({
   value,
   label = 'copy to clipboard',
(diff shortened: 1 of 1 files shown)


docs/components/CopyToClipboardButton.tsx
```diff
@@ -27,6 +27,38 @@ function copyTextToClipboard(text: string) {
   return copied;
 }
 
+// Note: eventually we want to natively handle multiple toasts, but for now we will limit the copy toast notification
+// to only display a single notification to prevent weird focus issues
+const notificationsMap = new Map<Symbol, React.ComponentProps<Toast>>();
```
thuey, line 32: I think this will work, although admittedly, the pattern feels a little weird to have a component call the props of another component stored in a module-level map.

I'm curious, did you consider doing something like dispatching a custom event? I know we do something like that with tooltips. Something like:

```ts
useEffect(() => {
  if (show) {
    const event = new CustomEvent('cauldron:copy-to-clipboard-toast:hide')
    event.source = ref.current
    window.dispatch(event)
  }
}, [show])

useEffect(() => {
  const removeListener = window.addEventListener('cauldron:copy-to-clipboard-toast:hide', (e) => {
    if (e.source !== ref.current) {
      setShowToast(false)
    }
  })
  
  return () => {
    removeListener()
  }
}, [])
```

scurker: I agree it's gnarly, but I didn't want to get too down in the weeds trying to find the perfect solution. We want to actually address this scenario eventually in Cauldron itself and this solution should go away.

thuey (approved):

Issue #619, opened 2022-04-05
Toast: Add missing `className` to available props for toast component


akornmeier:
Toast component does not currently have the ability to accept new classes in a `className` prop.

scurker:
We should probably eventually define a base of what every component should take:

* `className`
* `ref` (via `forwardRef`)
* ...?

akornmeier:
Agreed. I also think these older class components should be converted to functional ones for brevity

PR #1447, opened 2024-04-11
feat(react): support use of custom classNames with Toast component

tbusillo:
closes: https://github.com/dequelabs/cauldron/issues/619

### DIFF
--- a/packages/react/__tests__/src/components/Toast/index.js
+++ b/packages/react/__tests__/src/components/Toast/index.js
@@ -246,3 +246,16 @@ test('dismiss control is not rendered when dismissible is `false`', (done) => {
     done();
   }); // wait for animation timeouts / async setState calls
 });
+
+test.only('allows for custom class names', () => {
+  const mountElement = document.createElement('div');
+  document.body.appendChild(mountElement);
+  const wrapper = mount(
+    <Toast {...defaultProps} show={true} className="custom-class" type="info">
+      {'hi'}
+    </Toast>,
+    { attachTo: mountElement }
+  );
+
+  expect(wrapper.find('.Toast').hasClass('custom-class')).toBeTruthy();
+});
--- a/packages/react/src/components/Toast/index.tsx
+++ b/packages/react/src/components/Toast/index.tsx
@@ -5,7 +5,7 @@ import AriaIsolate from '../../utils/aria-isolate';
 import { typeMap, tabIndexHandler } from './utils';
 import setRef from '../../utils/setRef';
 
-export interface ToastProps {
+export interface ToastProps extends React.HTMLAttributes<HTMLDivElement> {
   type: 'confirmation' | 'caution' | 'error' | 'action-needed' | 'info';
   onDismiss: () => void;
   dismissText?: string;
@@ -106,7 +106,8 @@ export default class Toast extends React.Component<ToastProps, ToastState> {
         'Toast',
         `Toast--${typeMap[type].className}`,
         animationClass,
-        { 'Toast--non-dismissible': !dismissible }
+        { 'Toast--non-dismissible': !dismissible },
+        otherProps.className
       )
     };
 
@@ -122,7 +123,7 @@ export default class Toast extends React.Component<ToastProps, ToastState> {
             setRef(toastRef, el);
           }}
           {...defaultProps}
-          {...otherProps}
+          className={defaultProps.className}
         >
           <div className="Toast__message">
             <Icon type={typeMap[type].icon} />


packages/react/src/components/Toast/index.tsx
```diff
@@ -122,7 +123,7 @@ export default class Toast extends React.Component<ToastProps, ToastState> {
             setRef(toastRef, el);
           }}
           {...defaultProps}
-          {...otherProps}
+          className={defaultProps.className}
```
tbusillo, line 126: I realize this is a bit unorthodox and we shouldn't really have to do it this way. 

However, in this case, much of the toast's functionality relies upon certain classNames always existing/being set and I don't think consumers of the component will expect adding a custom classname will break all existing functionality/styling (e.g., just trying to add a max-width or some other superficial change). 

```tsx
// It is equivalent to a usage like this:
<Toast show={show} type="caution" className="test" onDismiss={() => setShow(false)}>

// rendering as this
<div tabindex="-1" class="Toast Toast--warning FadeIn--flex FadeIn test">

// instead of this
<div tabindex="-1" class="test">
```

I am definitely open to suggestions on how this could be done differently, though.

packages/react/__tests__/src/components/Toast/index.js
```diff
@@ -246,3 +246,16 @@ test('dismiss control is not rendered when dismissible is `false`', (done) => {
...
+  const mountElement = document.createElement('div');
+  document.body.appendChild(mountElement);
+  const wrapper = mount(
+    <Toast {...defaultProps} show={true} className="custom-class" type="info">
+      {'hi'}
+    </Toast>,
+    { attachTo: mountElement }
+  );
+
+  expect(wrapper.find('.Toast').hasClass('custom-class')).toBeTruthy();
```
scurker, line 260: I know this is probably more than what you bargained for, but we've been [migrating to use RTL](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#unit-tests)

Additionally, this test should probably cover that the extra class is added and keeps the existing class.

packages/react/src/components/Toast/index.tsx
scurker: You're correct in your assumption that `className` should not clobber toast's class name. The cleaner way to implement this might be to destructure `className` from `props` and just spread that as part of default props without setting `className` again. This is primarily how we've implemented it elsewhere, and looks something like this:

```tsx
const { className, ...other } = props
return <div className={className('CauldronClass', classname)} {...other} /> 
```

scurker (changes requested):

packages/react/__tests__/src/components/Toast/index.js
tbusillo: No worries, more comfortable with RTL these days than enzyme to be honest. Is the expectation that the entire test suite be rewritten in RTL, or just new test cases (with others converted over time)? 

I don't have strong feelings either way, just want to clarify what's needed.

packages/react/src/components/Toast/toast.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+      <Toast type="info" show={show} data-testid="toast">
+        {testString}
+      </Toast>
+    );
+    const toast = screen.getByTestId('toast');
+    expect(toast).toHaveClass('Toast Toast--info is--hidden');
+    show = true;
+    setTimeout(() => {
+      expect(toast).toHaveClass('Toast Toast--info FadeIn--flex');
+    }, 300);
```
scurker, line 70: So the RTL way we've been doing this and I've seen recommended is to use [`rerender`](https://testing-library.com/docs/example-update-props/), example here: https://github.com/dequelabs/cauldron/blob/b25e7c3390360af6e5261af41b120b56e87012ef/packages/react/src/components/Select/index.test.tsx#L130

packages/react/src/components/Toast/toast.test.tsx
```diff
@@ -0,0 +1,288 @@
...
+  });
+});
+
+test('should render toast with dismiss button', async () => {
+  const setState = jest.fn();
+  jest
+    .spyOn(React, 'useState')
+    .mockImplementationOnce(
+      () => [undefined, setState] as [unknown, React.Dispatch<unknown>]
+    );
```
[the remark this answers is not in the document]
tbusillo: I was trying to match the existing test cases as close as possible, is the simplest answer.

It’s definitely more the enzyme way of testing though and I can remove the ones where internals are mocked (think their functioning is implicitly tested anyways).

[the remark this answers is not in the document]
scurker: We've been trying to cleanup things like this as we go along to conform the what's the present style at least.

scurker (approved):

scurker:
Thanks for putting this together @tbusillo!

scurker (changes requested):
Sorry, I just realized I forgot one thing 🤡 

We need to add this to documentation. It should be as simple as including `className={true}` on the `ComponentProps` component.

tbusillo:
> ld be as simple as including `className={true}` on the `ComponentProps`

No problem! I'll add it in after standup!

scurker (approved):

PR #1450, opened 2024-04-16
test: switch scrim tests to react testing library

lsprr:
Issue #1159

This PR switches the Scrim tests to React Testing Library.

Validation steps

- Run yarn test


packages/react/src/components/Scrim/index.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+test('should fade scrim in when initial truthy show prop is provided', async () => {
+  render(<Scrim show={true} />);
+
+  expect(screen.getAllByRole('generic')[1]).toHaveClass('Scrim--show');
+  await waitFor(() => {
+    expect(screen.getAllByRole('generic')[1]).toHaveClass(
+      'Scrim--show Scrim--fade-in'
+    );
+  });
```
thuey, line 14: @scurker What do you think about this pattern when there is not an otherwise good RTL selector to use?

Alternatives to this include:
1. Passing an optional `data-testid`
2. Using the `container` returned from `render()` to just do `container.querySelector` or `container.firstChild`
3. Using `document.querySelector`

Personally, I lean toward option 2 or 3. I realize that doing `getAllByRole` follows the RTL practice of avoiding DOM queries. However, it feels a bit dogmatic and unnatural here, since we're already checking for CSS class names anyway.

Whatever we land on, I think we should document.

thuey (commented):
Thanks for contributing!

packages/react/src/components/Scrim/index.test.tsx
scurker: I agree, the `getByRole` with generic role doesn't really buy all that much here.

I'm not going to be as dogmatic about the approach however since I would accept either `data-testid` or the `container` here since both would be acceptable means. If we were to document, I would probably specify the preference would be setting `data-testid` for a component that's accessible in the render function.

lsprr: The Scrim component only accepts the show prop and does not support custom props. This means that adding data-testid or any other custom prop to the Scrim component will not have any effect. Therefore, I will lean toward option 2.

Please let me know if this is OK with you. @thuey @scurker

scurker: Ah yes - in this instance you will need to use `container` then.

thuey (approved):

scurker:
🚨 Reviewed for Security 🚨

scurker (approved):

PR #1456, opened 2024-04-19
test: switch accordion tests to react testing library

lsprr:
Issue #1159

This PR switches the Accordion tests to React Testing Library.

### Validation steps

- Run yarn test


packages/react/src/components/Accordion/index.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+test('should pass through props', () => {
+  render(
+    <Accordion role="region">
+      <AccordionTrigger>Trigger</AccordionTrigger>
+      <AccordionContent>Content</AccordionContent>
+    </Accordion>
+  );
+
+  expect(screen.getByRole('region')).toBeInTheDocument();
```
scurker, line 37: This test is a bit different from before since it doesn't validate that the accordion component is a region.

packages/react/src/components/Accordion/index.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+    <Accordion>
+      <AccordionTrigger>Trigger</AccordionTrigger>
+      <AccordionContent>
+        <div data-testid="content">foo</div>
+      </AccordionContent>
+    </Accordion>
+  );
+
+  setTimeout(() => {
+    expect(screen.queryByTestId('content')).toBeNull();
```
scurker, line 90: Very minor nit:

```suggestion
    expect(screen.queryByTestId('content')).not.toBeInTheDocument()
```

scurker (changes requested):
Just 2 very minor things, otherwise code changes LGTM.

PR #1458, opened 2024-04-23
test: switch tag button tests to react testing library

lsprr:
Issue #1159

This PR switches the Tag Button tests to React Testing Library.

Validation steps

- Run yarn test


packages/react/src/components/TagButton/index.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+  value: string;
+  onClick: () => void;
+};
+
+const defaultProps: TagButtonProps = {
+  icon: 'pencil',
+  label: 'Label: ',
+  value: 'value',
+  onClick: () => console.log('test')
+};
```
scurker, line 19: Instead of retyping the types, we can use component props.

```suggestion
const defaultProps: React.ComponentProps<typeof TagButton> = {
  icon: 'pencil',
  label: 'Label: ',
  value: 'value',
  onClick: () => console.log('test')
};
```

packages/react/src/components/TagButton/index.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  expect(screen.getByRole('button')).toBeInTheDocument();
+  expect(screen.getByRole('button')).not.toHaveAttribute('role');
+  expect(screen.getByRole('button')).toHaveTextContent('Label: value');
+});
+
+test('should support className prop', () => {
+  renderDefaultTagButton({ className: 'foo' });
+
+  expect(screen.getByRole('button')).toHaveClass('foo');
```
scurker, line 37: We should validate that it still includes the `TagButton` class:
```suggestion
  expect(screen.getByRole('button')).toHaveClass('TagButton', 'foo');
```

scurker (changes requested):
A few small changes need, otherwise looks good!

scurker (approved):

scurker:
🚨 Reviewed for Security 🚨

PR #1464, opened 2024-04-24
test: switch options menu list tests to react testing library

lsprr:
Issue #1159

This PR switches the Options Menu List tests to React Testing Library.

Validation steps

- Run yarn test


lsprr:
Hi @scurker. 

I wanted to inform you that I discovered the `closeOnSelect` prop isn't being used in the component's logic. I'm unsure why it's included if it's not utilized. Furthermore, I strongly suggest either removing it entirely or incorporating it into the component's functionality. That's why I haven't tested `closeOnSelect` in the tests.

packages/react/src/components/OptionsMenu/OptionsMenuList.test.tsx
```diff
@@ -0,0 +1,307 @@
...
+      <li>option 2</li>
+    </OptionsMenuList>
+  );
+
+  fireEvent.click(screen.getByRole('link'));
+
+  expect(onClick).toBeCalledTimes(1);
+});
+
+test('should focus child links with keydown events', () => {
```
scurker, line 249: This test name mentions focus, but it looks like we're checking for `click` here. Is this test right?

scurker (commented):
One test stood out to me, so just wanted to check to make sure it's testing the right thing.

scurker (approved):
🚨 Reviewed for Security 🚨

PR #1465, opened 2024-04-24
feat(react): deprecate individual tooltip props for IconButton replacing with tooltipProps

### LINKED ISSUE #1457: Support controlling Tooltip's `association` prop via `<IconButton />`
In cases where the icon button needs a complex accessible name joined from several tokens in the `aria-labelledby` attribute, it can be difficult to create the "ideal" accessible name. Specifically the fact that the tooltip's ID gets added to the **end** of the token list can create some sub-optimal accNames.

## Example

```
<h2 id="heading">Some contextual heading</h2>
<p id="paragraph">additional info</p>
<IconButton
  icon="foo"
  label={i18n`Some action`}
  aria-labelledby={`heading paragraph`}
  onClick={handleInspect}
/>
```

This would create an accessible name of:

> Some contextual heading additional info Some action

Where it is sometimes ideal for "Some action" to be at the beginning of the accName

scurker:
closes #1457 
ref #1159


packages/react/src/components/IconButton/index.tsx
```diff
@@ -13,55 +13,95 @@ import {
   PolymorphicProps,
   PolymorphicComponent
 } from '../../utils/polymorphicComponent';
-import { HTMLAttributes } from 'enzyme';
 
 export interface IconButtonProps
-  extends PolymorphicProps<React.HTMLAttributes<HTMLButtonElement>, 'button'> {
+  extends PolymorphicProps<
+    React.HTMLAttributes<HTMLButtonElement | HTMLAnchorElement>,
+    'button' | 'a'
```
scurker, line 20: When writing tests, the type was a bit too restrictive in the sense that we do support both `button`, `a`, and linkish elements. We could address separately if needed but I don't think this is a breaking change as it's widening the type.

schne324 (approved):

PR #1466, opened 2024-04-24
test: switch toast's utils tests to react testing library

lsprr:
Issue #1159

This PR switches the Toast's utils tests to React Testing Library.

Validation steps

- Run yarn test


packages/react/src/components/Toast/utils.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+  {
+    /* eslint-disable jsx-a11y/no-noninteractive-tabindex */
+  }
+  const { container } = render(
+    <div>
+      <div className="toaster"></div>
+      <div className="zero" tabIndex={0}></div>
+      <div className="minus-one" tabIndex={-1}></div>
+    </div>
+  );
```
scurker, line 18: This will disable this rule for the whole file. We should just be able to disable it where needed:

```suggestion
  const { container } = render(
    <div>
      <div className="toaster"></div>
      {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
      <div className="zero" tabIndex={0}></div>
      {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
      <div className="minus-one" tabIndex={-1}></div>
    </div>
  );
```

scurker (changes requested):

scurker (approved):
🚨 Reviewed for Security 🚨

PR #1474, opened 2024-05-03
feat: add multiselect to listbox and combobox

### LINKED ISSUE #1103: Combobox Component (Multiselect)
Follow general ARIA practices guidelines: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

This should include guidelines outlined on #1088 with the addition of the following

Multiselect is configurable via a prop:

```jsx
<Combobox label="Combobox" multiselect>
  <ComboboxItem value="...">
    Item
    Longer Description
  </ComboboxItem>
</Combobox>
```

When multiple items are selected, each item selected should be displayed via a removable tag:

combobox with multiple tags displayed

When the selected items overflows the max-width of the contained area, they should naturally wrap and provide space for the input

combobox with multiple tags that wrap within the text field

## Expected Behaviors

Backspace (with no text entered) should highlight the last selected "pill". Backspace again (or <kbd>delete</kbd>) should delete that selection. When any pill has active focus, backspace or <kbd>delete</kbd> should delete that selection.

Each pill should not be a tabstop but navigated with <kbd>left</kbd> / <kbd>right</kbd> arrow keys. We should support arrow key navigation when the inner text field does not have any text.

When a selection is made to select an active selection, the combobox inner text field should be reset to blank.

When disabled, do not show the "x" for any of the pills and clicking on a pill should not remove it from the selections.

## Form Fields

A single select combobox typically displays form field values as a single hidden input, e.g. `<input type="hidden" name={name} value={value} />`. When multiple values are selected, the hidden input field should be individual fields mapping to the same name for each value, example:

```tsx
{selectedValues.forEach(value => (
  <input type="hidden" name={name} value={value} />
))}
```

chornonoh-vova:
closes #1103

### DIFF
--- a/docs/pages/components/Listbox.mdx
+++ b/docs/pages/components/Listbox.mdx
@@ -96,7 +96,6 @@ Controlled listboxes require updating the value to reflect the newly selected st
 function ControlledListboxExample() {
   const [value, setValue] = useState('Two');
   const handleSelectionChange = ({ value, previousValue }) => {
-    console.log(value, previousValue);
     setValue(value);
   };
 
@@ -137,7 +136,6 @@ Uncontrolled listboxes will automatically set `aria-selected="true"` for the sel
 function MultiselectListboxExample() {
   const [value, setValue] = useState(['Two']);
   const handleSelectionChange = ({ value, previousValue }) => {
-      console.log(value, previousValue);
       setValue(value);
   };
   return (
--- a/docs/pages/components/Listbox.mdx
+++ b/docs/pages/components/Listbox.mdx
@@ -5,7 +5,7 @@ source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/co
 ---
 
 import { useState } from 'react';
-import { Listbox, ListboxOption, ListboxGroup } from '@deque/cauldron-react';
+import { FieldWrap, Listbox, ListboxOption, ListboxGroup } from '@deque/cauldron-react';
 
 ```js
 import { 
@@ -95,20 +95,24 @@ Controlled listboxes require updating the value to reflect the newly selected st
 ```jsx example
 function ControlledListboxExample() {
   const [value, setValue] = useState('Two');
-  const handleSelect = ({ value }) => setValue(value);
+  const handleSelectionChange = ({ value, previousValue }) => {
+    console.log(value, previousValue);
+    setValue(value);
+  };
+
   return (
-    <>
+    <FieldWrap>
       <div id="listbox-controlled-example">Controlled Listbox</div>
       <Listbox 
         aria-labelledby="listbox-controlled-example" 
         value={value}
-        onSelect={handleSelect}
+        onSelectionChange={handleSelectionChange}
       >
         <ListboxOption>One</ListboxOption>
         <ListboxOption>Two</ListboxOption>
         <ListboxOption>Three</ListboxOption>
       </Listbox>
-    </>
+    </FieldWrap>
   );
 }
 ```
@@ -126,6 +130,32 @@ Uncontrolled listboxes will automatically set `aria-selected="true"` for the sel
     <ListboxOption>Three</ListboxOption>
   </Listbox>
 </>
+```
+### Multiselect
+
+```jsx example
+function MultiselectListboxExample() {
+  const [value, setValue] = useState(['Two']);
+  const handleSelectionChange = ({ value, previousValue }) => {
+      console.log(value, previousValue);
+      setValue(value);
+  };
+  return (
+    <FieldWrap>
+      <div id="listbox-multiselect-example">Multiselect Listbox</div>
+      <Listbox 
+        aria-labelledby="listbox-multiselect-example" 
+        value={value}
+        onSelectionChange={handleSelectionChange}
+        multiselect
+      >
+        <ListboxOption>One</ListboxOption>
+        <ListboxOption>Two</ListboxOption>
+        <ListboxOption>Three</ListboxOption>
+      </Listbox>
+    </FieldWrap>
+  );
+}
 ```
 
 ## Props
--- a/packages/react/scripts/buildIconTypes.js
+++ b/packages/react/scripts/buildIconTypes.js
@@ -36,7 +36,7 @@ for (const file of files) {
 
   const name = file.replace('.svg', '');
   if (ICONS_WITH_DIRECTIONS.includes(name)) {
-    iconTypes.push(...DIRECTIONS.map(d => `${name}-${d}`));
+    iconTypes.push(...DIRECTIONS.map((d) => `${name}-${d}`));
   } else {
     iconTypes.push(name);
   }
@@ -51,11 +51,11 @@ fs.writeFileSync(
 
 /** IconType represents each valid icon type. */
 export type IconType =
-${iconTypes.map(i => `  | '${i}'`).join('\n')};
+${iconTypes.map((i) => `  | '${i}'`).join('\n')};
 
 /** iconTypes holds each valid icon type. */
 export const iconTypes = [
-${iconTypes.map(i => `  '${i}'`).join(',\n')}
+${iconTypes.map((i) => `  '${i}'`).join(',\n')}
 ];
 `.trim()
 );
--- a/packages/react/src/components/Combobox/Combobox.tsx
+++ b/packages/react/src/components/Combobox/Combobox.tsx
@@ -32,20 +32,16 @@ interface ComboboxOption {
   description?: string;
 }
 
-interface ComboboxProps
-  extends Omit<
-    React.InputHTMLAttributes<HTMLInputElement>,
-    'value' | 'defaultValue'
-  > {
+type ComboboxProps = Omit<
+  React.InputHTMLAttributes<HTMLInputElement>,
+  'value' | 'defaultValue'
+> & {
   label: ContentNode;
   options?: ComboboxOption[];
   inputValue?: string;
-  value?: ComboboxValue | ComboboxValue[];
-  defaultValue?: ComboboxValue | ComboboxValue[];
   requiredText?: React.ReactNode;
   error?: React.ReactNode;
   autocomplete?: 'none' | 'manual' | 'automatic';
-  multiselect?: boolean;
   onSelectionChange?: <T extends HTMLElement = HTMLElement>({
     target,
     value,
@@ -59,7 +55,18 @@ interface ComboboxProps
   renderNoResults?: (() => JSX.Element) | React.ReactElement;
   portal?: React.RefObject<HTMLElement> | HTMLElement;
   inputRef?: React.Ref<HTMLInputElement>;
-}
+} & (
+    | {
+        value?: ComboboxValue;
+        defaultValue?: ComboboxValue;
+        multiselect?: false;
+      }
+    | {
+        value?: ComboboxValue[];
+        defaultValue?: ComboboxValue[];
+        multiselect: true;
+      }
+  );
 
 const getComboboxValue = (
   listboxValue: ListboxValue | ListboxValue[]
@@ -94,8 +101,8 @@ const ComboboxNoResults = ({
 };
 
 const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
-  (
-    {
+  (props, ref): JSX.Element => {
+    const {
       id: propId,
       className,
       label,
@@ -119,10 +126,8 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
       portal,
       inputRef: propInputRef = null,
       'aria-describedby': ariaDescribedby,
-      ...props
-    },
-    ref
-  ): JSX.Element => {
+      ...rest
+    } = props;
     const [value, setValue] = useState<ComboboxValue | ComboboxValue[]>(
       defaultValue || propValue || (multiselect ? [] : '')
     );
@@ -132,7 +137,7 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
     >(new Map());
     const [selectedValue, setSelectedValue] = useState<
       ComboboxValue | ComboboxValue[]
-    >(value || multiselect ? [] : '');
+    >(value);
     const [formValue, setFormValue] = useState<string | undefined>('');
     const [open, setOpen] = useState(false);
     const [activeDescendant, setActiveDescendant] =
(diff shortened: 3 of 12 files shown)


docs/pages/components/Listbox.mdx
```diff
@@ -127,6 +131,32 @@ Uncontrolled listboxes will automatically set `aria-selected="true"` for the sel
   </Listbox>
 </>
 ```
+### Multiselect
+
+```jsx example
+function MultiselectListboxExample() {
+  const [value, setValue] = useState(['Two']);
+  const handleSelectionChange = ({ value, previousValue }) => {
+      console.log(value, previousValue);
```
scurker, line 140: Nit:

```suggestion
```

packages/react/scripts/buildIconTypes.js
scurker, line 1: Do you know why this file changed?

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -16,34 +16,43 @@ import useSharedRef from '../../utils/useSharedRef';
...
-  }) => void;
   onActiveChange?: (option: ListboxOption) => void;
-}
+  onSelectionChange?: <T extends HTMLElement = HTMLElement>(selection: {
+    target: T | undefined;
+    previousValue: ListboxValue | ListboxValue[];
+    value: ListboxValue | ListboxValue[];
+  }) => void;
+} & (
+    | {
```
scurker, line 30: I'm not sure the discriminated union buys a whole lot here, although if we did want a discriminated union I would make this more readable by keeping the existing type and unioning the types below where needed (example):

```tsx
interfact ListboxProps  = ....
type MultiselectListBoxProps = Omit<ListboxProps, 'value', 'defaultValue', 'multiselect'> & {
  value?: ListboxValue | ListboxValue[]
  defaultValue?: ListboxValue | ListboxValue[]
  multiselect: true
}
```

Part of the reasoning behind allowing array and non-array values is that it allows the component to be more flexible when specifying the value. When referenced, we would use `ListboxProps | MultiselectListboxProps`.

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -16,34 +16,43 @@ import useSharedRef from '../../utils/useSharedRef';
...
-    value
-  }: {
-    target: T;
-    previousValue: ListboxValue;
-    value: ListboxValue;
-  }) => void;
   onActiveChange?: (option: ListboxOption) => void;
-}
+  onSelectionChange?: <T extends HTMLElement = HTMLElement>(selection: {
+    target: T | undefined;
```
scurker, line 25: Why `| undefined`?

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -65,9 +75,9 @@ const Listbox = forwardRef<HTMLElement, ListboxProps>(
     const [activeOption, setActiveOption] = useState<ListboxOption | null>(
       null
     );
-    const [selectedOption, setSelectedOption] = useState<ListboxOption | null>(
-      null
-    );
+    const [selectedOption, setSelectedOption] = useState<
```
scurker, line 78: Since we potentially want to handle array values, I would change this to be `selectedOptions` and `setSelectedOptions` because the existing naming strategy doesn't make sense for both single/multi select.

packages/react/src/components/Listbox/Listbox.tsx
```diff
@@ -179,13 +239,13 @@ const Listbox = forwardRef<HTMLElement, ListboxProps>(
...
           );
-          // istanbul ignore else
           if (firstOption) {
             setActiveOption(firstOption);
           }
-          // istanbul ignore else
         } else if (event.target === listboxRef.current) {
-          setActiveOption(selectedOption);
+          setActiveOption(
+            Array.isArray(selectedOption) ? selectedOption[0] : selectedOption
```
scurker, line 247: Ideally, this would be keying off the prop and not the value type.

```tsx
!multiselect ? selectedOption[0] : selectedOption
```

packages/react/src/components/Listbox/ListboxOption.tsx
```diff
@@ -68,14 +77,11 @@ const ListboxOption = forwardRef<HTMLElement, ListboxOptionProps>(
           return [...options, option];
         }
 
-        for (const opt of options) {
+        for (let i = 0; i < options.length; i++) {
```
scurker, line 80: Why the change here?

packages/styles/combobox.css
```diff
@@ -111,8 +111,28 @@
...
 
+.Combobox .TextFieldWrapper {
+  flex-wrap: wrap;
+}
+
 .Combobox .TextFieldWrapper input[type='text'] {
   margin-right: 1.25rem;
+  flex-grow: 1;
+  width: unset;
+  min-width: var(--input-min-width);
```
scurker, line 122: Why was `min-width` added here? This creates an accessibility issue at lower viewport widths.

packages/styles/combobox.css
```diff
@@ -111,8 +111,28 @@
...
+}
+
+.Combobox__selected-tag {
+  margin-block: var(--space-smallest);
```
scurker, line 126: I'm not sure this is correct as the vertical spacing between the selected items doesn't appear to match the designs.

Actual:
image

Proposed Design:
image

The down carat is also not positioned correctly in the top right.

packages/styles/combobox.css
```diff
@@ -111,8 +111,28 @@
...
+}
+
+.Combobox__selected-tag:first-of-type {
+  margin-left: var(--space-smallest);
+}
+
+.Combobox__selected-tag .TagButton__icon svg {
+  width: 16px;
+  height: 16px;
```
scurker, line 135: I think instead of defining the size here, we may be able to just ensure we're using `--icon-size: 16px` scoped to the right css selector.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -46,15 +47,38 @@ interface ComboboxProps
...
+    value: ComboboxValue | ComboboxValue[];
+    previousValue: ComboboxValue | ComboboxValue[];
   }) => void;
   onActiveChange?: (option: ListboxOption) => void;
   renderNoResults?: (() => JSX.Element) | React.ReactElement;
   portal?: React.RefObject<HTMLElement> | HTMLElement;
   inputRef?: React.Ref<HTMLInputElement>;
-}
+} & (
+    | {
```
scurker, line 59: Same comment as above regarding Listbox and its discriminated unions.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -30,14 +32,13 @@ interface ComboboxOption {
...
-  > {
+type ComboboxProps = Omit<
+  React.InputHTMLAttributes<HTMLInputElement>,
+  'value' | 'defaultValue'
+> & {
   label: ContentNode;
   options?: ComboboxOption[];
-  value?: ComboboxValue;
-  defaultValue?: ComboboxValue;
+  inputValue?: string;
```
scurker, line 41: I apologize because this wasn't well defined, but I'm not sure we should be changing `value` to `inputValue` here. Existing users of the component are expecting to use `value` for non-multiselect components so we need to maintain that compatibility in order to not be a breaking change.

What we may need is a mechanism for converting a value string into a multiselect value. Feel free to update the RFC as we need to reach a consensus to ensure the changes we make here are compatible and make sense.

scurker (changes requested):
There's a lot of great work here, but I'm realizing looking at your PR that we missed defining some of the more refined behaviors of how we expect users to interact with this component. I left some comments inline of some of the things I noticed, but we may need a few rounds to ensure we land on the right spot here.

packages/react/scripts/buildIconTypes.js
chornonoh-vova: This file was changed after running `yarn fmt`, actually a couple of other files in `packages/react/__test__` also changed, but I didn't include them

packages/react/src/components/Listbox/Listbox.tsx
chornonoh-vova: By doing this I'm only allowing `value: string[]` when `multiselect = true` and `value: string` when `multiselect = false`. I don't really understand why we should allow value to be both string and string[] when multiselect is true, behavior will just be broken in my opinion. But if that's required, I'll change that.

packages/react/src/components/Listbox/Listbox.tsx
chornonoh-vova: But in the single select case name `selectedOptions` doesn't make sense either 😅

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: I had no idea how to convert value into multiselect value, therefore added this prop. It's just not compatible - `onChange` emits `string` and `onSelectionChange` emits `string[]` values. One of the possible solutions - albeit radical one - would be only left `value` and `onSelectionChange` props, and remove `onChange`. Although I don't know if that's plausible for users of the component.

packages/react/src/components/Listbox/ListboxOption.tsx
chornonoh-vova: It's just easier, previously we had line `const index = options.indexOf(opt);` inside of the loop, so why not have this index already when iterating?

packages/react/src/components/Combobox/Combobox.tsx
longmatthewh: @scurker what would you think about creating a new `MultiSelectCombobox` component? I don't totally love the idea (feels a little lazy), but I also don't hate it. Thoughts?

packages/react/scripts/buildIconTypes.js
scurker: We really want to keep the changes focused to what needs to change for this component and isolate changes that are unrelated as a separate PR/issue.

packages/react/src/components/Listbox/Listbox.tsx
scurker: Part of the reasoning would be to allow for more graceful inputs since we can't necessarily enforce that everyone is using typescript. Someone using `<Combobox value="foo" multiselect />` at a minimum should not cause an error although we could potentially log a warning if this wasn't the desired interface.

packages/react/src/components/Listbox/Listbox.tsx
scurker: My thinking here was if the internal state is always an array, then it's always "selected options of 1" but I'm open to discussion here.

packages/react/src/components/Combobox/Combobox.tsx
scurker: > @scurker what would you think about creating a new MultiSelectCombobox component? I don't totally love the idea (feels a little lazy), but I also don't hate it. Thoughts?

There's a lot of overlap between the single/multiselect components, so that introduces an additional challenge of keeping those components in sync if we split them up with future updates. It's also a higher lift for authors since swapping between a single/multiselect component is potentially as small as changing the property with the current proposal.

scurker: > I had no idea how to convert value into multiselect value, therefore added this prop. It's just not compatible - onChange emits string and onSelectionChange emits string[] values.

That's where the discriminated union could potentially help, example below.

```tsx
type ComboboxProps = {
  ...
  onChange: (value: string, ...) => {} 
}

type MultiselectComboboxProps extends Omit<ComboboxProps, 'onChange'> = {
  ...
  onChange: (value: string[]) => {}
}
```

packages/react/src/components/Combobox/ComboboxOption.tsx
```diff
@@ -113,10 +123,23 @@ const ComboboxOption = forwardRef<HTMLLIElement, ComboboxOptionProps>(
...
+        if ((selectedValue as ComboboxValue[]).includes(comboboxValue)) {
+          setFormValue((prev) => {
+            const formValueToSet =
+              typeof formValue === 'undefined' ? comboboxValue : formValue;
+
+            if (!prev?.includes(formValueToSet)) {
+              return prev ? prev + ',' + formValueToSet : formValueToSet;
+            }
+          });
+        }
```
scurker, line 142: This is worth discussing, I'm not sure what the best approach is (and it needs to be addressed as part of the RFC). There's 2 distinct approaches we could take:

- Comma delimited form values
- Individual form values

It appears you've taken the first approach, but what if the value itself includes a comma? It can introduce challenges since the server will need to be aware of the specific format to parse by and be able to translate comma delimited values correctly.

Alternatively, each value itself could be a set of hidden fields that map to the same name:

```html
<input type="hidden" name={name} value={value1} />
<input type="hidden" name={name} value={value2} />
<input type="hidden" name={name} value={value3} />
```

I feel like this approach is more scalable and avoids having to directly parse the submitted values. Thoughts?

packages/react/src/components/Listbox/ListboxOption.tsx
scurker: I'm partially nitpicking here because it's not clear why the change was made and isn't conventionally how iterations have been performed elsewhere. I understand this can be subjective, but given there's not a problem that's being identified it would be better to leave this untouched unless we're fixing a specific issue.

packages/react/src/components/Combobox/ComboboxOption.tsx
chornonoh-vova: I really like this approach, thanks for suggestion

packages/styles/combobox.css
```diff
@@ -111,8 +111,24 @@
   position: relative;
 }
 
+.Combobox .TextFieldWrapper {
+  min-height: 32px;
+  flex-wrap: wrap;
+  gap: var(--space-half);
+  padding-inline: var(--space-smallest);
+  padding-block: var(--space-half);
```
scurker, line 119: Is there a reason why these styles were added? This appears to change the size of all Comboboxes. Comboboxes should match the existing size of other comboboxes (which is currently 35px).

packages/styles/combobox.css
```diff
@@ -111,8 +111,24 @@
...
+
 .Combobox .TextFieldWrapper input[type='text'] {
   margin-right: 1.25rem;
+  flex-grow: 1;
+  width: unset;
+  padding: unset;
+}
+
+.Combobox__selected-tag {
+  margin-left: 0 !important;
```
scurker, line 130: Why `!important`? This should be limited to extremely specific use cases (such as utility classes) where the desired behavior is to take on that style without respect of specificity. There are likely better alternatives that we need to consider before using `!important`.

PR #1475, opened 2024-05-08
chore: update playwright dependencies

scurker:
Playwright is currently failing all screenshots, so this is changing the build to pull the latest version when building the docker file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "test:a11y": "ts-node e2e/accessibility.ts",
     "postinstall": "if [ -z \"$CI\" ]; then yarn --cwd=packages/styles && yarn --cwd=packages/react; fi",
     "screenshots": "docker run --rm --ipc=host -v $(pwd)/playwright-ct.config.ts:/app/playwright-ct.config.ts -v $(pwd)/packages:/app/packages -v $(pwd)/e2e:/app/e2e -v $(pwd)/e2e:/app/e2e cauldron-playwright-e2e-screenshots",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "test:a11y": "ts-node e2e/accessibility.ts",
     "postinstall": "if [ -z \"$CI\" ]; then yarn --cwd=packages/styles && yarn --cwd=packages/react; fi",
     "screenshots": "docker run --rm --ipc=host -v $(pwd)/playwright-ct.config.ts:/app/playwright-ct.config.ts -v $(pwd)/packages:/app/packages -v $(pwd)/e2e:/app/e2e -v $(pwd)/e2e:/app/e2e cauldron-playwright-e2e-screenshots",
-    "screenshots:docker": "docker build . -t cauldron-playwright-e2e-screenshots -f e2e/Dockerfile",
+    "screenshots:docker": "docker build --build-arg playwright_version=$(npm view playwright version) . -t cauldron-playwright-e2e-screenshots -f e2e/Dockerfile",
```
scurker, line 25: I'm torn on whether this should be using the latest version of playwright, or what's currently in package but ideally we want to avoid failures in CI due to playwright not being up to date https://github.com/dequelabs/cauldron/actions/runs/9007968743/job/24748981345?pr=1342

thuey: IMO, we should be using the version that's declared in `package.json`

Can we address the issue of not being up to date by configuring dependabot to help us keep playwright up to date?

scurker: Yeah that's what I was initially considering. Pulling the version out of what we have installed will take a bit more time.

scurker: For now, I created https://github.com/dequelabs/cauldron/issues/1476 and this PR can address the fix short term

PR #1477, opened 2024-05-09
feat(react): image combobox

orest-s:


packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -54,6 +55,7 @@ interface ComboboxProps
   renderNoResults?: (() => JSX.Element) | React.ReactElement;
   portal?: React.RefObject<HTMLElement> | HTMLElement;
   inputRef?: React.Ref<HTMLInputElement>;
+  variant?: 'default' | 'advanced';
```
stephenmathieson, line 58: Why is this change necessary?

PR #1486, opened 2024-05-20
fix(react): make focusOnInitialRender work without setTimeout

tbusillo:
- This pull request removes the loader overlay's `focusOnInitialRender`'s dependency on setTimeout + the loader's ref as a dependency in the `useEffect`'s dependency array. 
- closes: #747 (replaces #833)

### DIFF
--- a/packages/react/src/components/LoaderOverlay/index.tsx
+++ b/packages/react/src/components/LoaderOverlay/index.tsx
@@ -42,14 +42,9 @@ const LoaderOverlay = forwardRef<HTMLDivElement, LoaderOverlayProps>(
       };
     }, [focusTrap, overlayRef.current]);
 
-    useEffect(() => {
-      if (!!focusOnInitialRender && overlayRef.current) {
-        setTimeout(() => {
-          return overlayRef.current?.focus();
-        });
-      }
-      return;
-    }, [overlayRef.current]);
+    if (!!focusOnInitialRender && overlayRef.current) {
+      overlayRef.current.focus();
+    }
 
     const Wrapper = focusTrap ? FocusTrap : React.Fragment;
     const wrapperProps = focusTrap
--- a/packages/react/src/components/LoaderOverlay/index.tsx
+++ b/packages/react/src/components/LoaderOverlay/index.tsx
@@ -42,9 +42,11 @@ const LoaderOverlay = forwardRef<HTMLDivElement, LoaderOverlayProps>(
       };
     }, [focusTrap, overlayRef.current]);
 
-    if (!!focusOnInitialRender && overlayRef.current) {
-      overlayRef.current.focus();
-    }
+    useEffect(() => {
+      if (!!focusOnInitialRender && overlayRef.current) {
+        overlayRef.current.focus();
+      }
+    }, []);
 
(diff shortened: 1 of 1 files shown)


packages/react/src/components/LoaderOverlay/index.tsx
```diff
@@ -44,12 +44,9 @@ const LoaderOverlay = forwardRef<HTMLDivElement, LoaderOverlayProps>(
 
     useEffect(() => {
       if (!!focusOnInitialRender && overlayRef.current) {
-        setTimeout(() => {
-          return overlayRef.current?.focus();
-        });
+        overlayRef.current.focus();
```
tbusillo, line 47: I tried just getting rid of the `useEffect` altogether, which functionally did seem to work. However, RTL (maybe JSDOM?) seemed to have some issues with the `activeElement` not moving from `document.body` when doing it that way. 

I visually/programmatically (see below) confirmed both with and without the `useEffect`, the `document.activeElement` is the loader overlay, so not 100% sure why the matcher is coming back the way it is. 

## Verification

1. Checkout the branch, run dev and go to http://localhost:8003/components/LoaderOverlay

2. Open up the browser console and use something similar to the following to continuously output the result of `document.activeElement` to the console.

```ts
setInterval(() => {
    console.log(document.activeElement)
}, 500)
```

3. You should see something like the below output:

browser console output with the loader overlay as the active element

4. Remove the `useEffect` from the surrounding conditional/`ref.current.focus()`, then repeat steps 2-3. 

I tried addressing this a few different ways, but the tests themselves seem to be where the issue is coming from (they're not picking up on the focus movement, for whatever reason). 

Possibly related: [jest-dom issue](https://github.com/testing-library/jest-dom/issues/53)

scurker: So I'm curious what has changed based on the following comment: https://github.com/dequelabs/cauldron/pull/562#discussion_r831497553

Your comment seems to suggest that setting focus on the next tick _was_ necessary, and it's not immediately clear if anything has changed. It does appear you've done some testing around this already, so I just wanted to circle back to make sure there wasn't something we were missing.

tbusillo: If I had to guess, it was probably a browser update more than anything we changed or React changed. 

I do remember when this was initially done there was a timing issue, I'm 99% sure because of [the 100ms setTimeout](https://github.com/dequelabs/cauldron/blob/82f43f21593fd7cb6248d228ecba10e6999cf5d2/packages/react/src/components/Scrim/index.tsx#L52-L56) when applying the animation class in the Scrim component. 

_But_, when I revisited it in the other PR (the one replaced by this one), noticed it did not seem to be an issue anymore 🤷‍♂️

scurker, line 47: I'm a bit surprised this doesn't complain, wouldn't it be possible for `.current` to be null?

tbusillo: We check for its truthiness in the condition surrounding this

scurker (approved):

PR #1497, opened 2024-06-03
test: switch menubar tests to react testing library

denysfedorov:

This PR switches the MenuBar tests to React Testing Library.

Validation steps

Run yarn test


packages/react/src/components/MenuBar/index.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+      {child}
+      {false && child}
+    </MenuBar>
+  );
+  expect(container).toBeTruthy();
+  expect(container.querySelectorAll('div').length).toBe(1);
+});
+
+test('handles left arrow', async () => {
+  const { getAllByRole } = render(
```
scurker, line 32: I would recommend using `screen.getByRole` here because that does 2 things, it validates that we're rendering the element in the document, there's only 1 type of that element, and that we should be firing events on the menubar itself.

packages/react/src/components/MenuBar/index.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  fireEvent.keyDown(menuItems[2], { key: 'ArrowLeft', keyCode: left });
+  await waitFor(() => {
+    expect(menuItems[1]).toHaveFocus();
+    expect(document.activeElement).toBe(menuItems[1]);
+  });
+});
+
+test('handles right arrow', async () => {
+  const { getAllByRole } = render(
```
scurker, line 55: Same suggestion as above.

scurker (changes requested):
Good work! The only thing that is missing (per our [guidelines](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#testing-strategies)) is testing for the alternate states (wide and thin).

packages/react/src/components/MenuBar/index.test.tsx
denysfedorov: added thin\wide state tests
using scree.getAllByRole to get needed elements

packages/react/src/components/MenuBar/index.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+test('handles left arrow', async () => {
+  render(
+    <MenuBar>
+      <TopBarItem>test-1</TopBarItem>
+      <TopBarItem>test-2</TopBarItem>
+      <TopBarItem>test-3</TopBarItem>
+    </MenuBar>
+  );
+
+  const menuItems = screen.getAllByRole('menuitem');
```
scurker, line 40: We should be firing events against menubar, not menuitems.

```suggestion
  const menuItems = screen.getAllByRole('menuitem')
  const menubar = screen.getByRole('menubar');
```

Access to both might still be useful, so you can validate that the right menuitem was activated.

packages/react/src/components/MenuBar/index.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+test('handles right arrow', async () => {
+  render(
+    <MenuBar>
+      <TopBarItem>test-1</TopBarItem>
+      <TopBarItem>test-2</TopBarItem>
+      <TopBarItem>test-3</TopBarItem>
+    </MenuBar>
+  );
+
+  const menuItems = screen.getAllByRole('menuitem');
```
scurker, line 64: Same suggestion as above.

scurker (changes requested):
Last few things, otherwise LGTM!

packages/react/src/components/MenuBar/index.test.tsx
denysfedorov: Found this issue, but my case is 100% opposite

denysfedorov: And also, found, that we do not handle `keyDownEvent` on the `menuBar` component, we pass this handler to the children. So I believe that we should test `keyDown` handler in menuItem although

If we try to trigger `onKeyDown` on `MenuBar` it has not even triggered

scurker: Regarding your first comment on focusing `menuBar`, that's expected because menu bar isn't focusable.

For the latter comment, you're right - we do handle it on the child elements, I just didn't read the code fully. 🤦

scurker (approved):

PR #1500, opened 2024-06-03
fix: set `value` rather than `label` when using `onChange` and `onSelectionChange` with `Combobox`

### LINKED ISSUE #1499: Bug: `Combobox` `onChange` and `onSelectionChange` sets `value` argument to `label` rather than `value` when options prop is passed
### Steps to reproduce

Steps to reproduce

1. Create a `Combobox`
2. Provide an array of objects with keys of label and value to the `options` prop of the combobox
3. console.log `value` in `onChange` and `onSelectionChange` events

### Expected Behavior

`value` in console.log reflects the `value` from the options array object

### Actual Behavior

`value` in console.log reflects the `label` from the options array object

### Version

6.3.0

Chrome

### Relevant log output

_No response_

Additional information

yhafez:
Closes #1499

### DIFF
--- a/packages/react/src/components/Combobox/Combobox.test.tsx
+++ b/packages/react/src/components/Combobox/Combobox.test.tsx
@@ -439,6 +439,66 @@ test('should close combobox listbox when selecting option via keypress', () => {
   assertListboxIsOpen(false);
 });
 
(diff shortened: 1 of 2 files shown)


scurker (approved):
Good catch!

PR #1501, opened 2024-06-04
fix(Dialog): switch dialog tests to react testing library and fix heading type

### LINKED ISSUE #140: The Modal heading prop should take text as a prop as well as JSX.Elements
Currently the Modal only takes JSX.Elements as a prop for the heading. The heading prop should take both text and JSX.Elements for ease of use.

scurker:
ref #1159 
closes #140


packages/react/src/components/Dialog/index.tsx
```diff
@@ -117,6 +118,7 @@ export default class Dialog extends React.Component<DialogProps, DialogState> {
     const Dialog = (
       <FocusTrap
         focusTrapOptions={{
+          allowOutsideClick: true,
```
scurker, line 121: I spent way too much time here getting stuck, but apparently we weren't testing this correctly earlier since we were just checking that the dialog handler was invoked but not the prop function. Since we already have `ClickOutsideListener` we don't need to let focus trap handle it.

abhideo-deque (approved):
LGTM!!!

scurker:
Looks like actions got stuck from yesterday, so closing/reopening to retrigger them.

packages/react/src/components/Dialog/index.tsx
```diff
@@ -18,9 +18,10 @@ export interface DialogProps extends React.HTMLAttributes<HTMLDivElement> {
   onClose?: () => void;
   forceAction?: boolean;
   heading:
+    | string
```
scurker, line 21: As a result of writing tests, I discovered this type was too restrictive so expanding the type to allow for strings to be set as well.

thuey (commented):
Approved

PR #1511, opened 2024-06-10
test: moved `theme` from `Enzyme` to `RTL`

orest-s:

This PR switches the `context.theme` tests to React Testing Library.

Validation steps

Run yarn test


packages/react/src/contexts/theme.test.tsx
```diff
@@ -63,8 +71,8 @@ test('it provides toggleTheme functionality', () => {
 });
 
 test('handles mutations', () => {
-  mountWrapper();
-  expect(observe.called).toBe(true);
+  renderProvider();
+  expect(observe).toHaveBeenCalled;
```
scurker, line 75: Don't we need to invoke this?

```suggestion
  expect(observe).toHaveBeenCalled();
```

scurker (approved):

PR #1519, opened 2024-06-11
test: moved `useSharedRef` tests from `Enzyme` to `RTL`

orest-s:

This PR switches the `useSharedRef` tests to React Testing Library.

Validation steps

Run yarn test

### DIFF
--- a/packages/react/src/utils/polymorphicComponent.test.tsx
+++ b/packages/react/src/utils/polymorphicComponent.test.tsx
@@ -143,3 +143,7 @@ const htmlAnchorRef = React.createRef<HTMLAnchorElement>();
     bananas
   </ConstrainedComponent>
 );
+
+test('polymorphicComponent', () => {
+  // avoid `yarn test` error (Your test suite must contain at least one test.)
+});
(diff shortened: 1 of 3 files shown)


packages/react/src/utils/polymorphicComponent.test.tsx
```diff
@@ -143,3 +143,7 @@ const htmlAnchorRef = React.createRef<HTMLAnchorElement>();
     bananas
   </ConstrainedComponent>
 );
+
+test('polymorphicComponent', () => {
```
scurker, line 147: I think instead of adding an empty test here, it would be better to exclude it from coverage.

```js
  collectCoverageFrom: [
    '<rootDir>/src/components/**/*.{tsx,ts}',
    '<rootDir>/src/utils/**/*.{tsx,ts}',
    '!<rootDir>/src/utils/**/polymorphicComponent.test.{tsx,ts}',
  ],
```

scurker (changes requested):
Just one small change requested, otherwise test changes LGTM.

scurker (approved):

PR #1521, opened 2024-06-12
fix(react): fix incorrect React.Fragment for ContentNode type

### LINKED ISSUE #1507: ReactFragment is deprecated
In `packages/react/src/types.ts`, VS Code is showing that `ReactFragment` is deprecated.

> @deprecated — - This type is not relevant when using React. Inline the type instead to make the intent clear.

Screenshot 2024-06-09 at 8 57 35 AM

scurker:
closes #1507 

It appears this does appear to be safe to remove and should be a non-breaking change, as I can still set `const type: ContentNode = <></>` and types are happy.


packages/react/src/types.ts
```diff
@@ -1,4 +1,4 @@
-import { ReactElement, ReactFragment, ReactPortal } from 'react';
+import type { ReactElement, ReactFragment, ReactPortal } from 'react';
```
anastasialanz, line 1: Should we remove the imported `ReactFragment` here too?

scurker: Whoops, I was testing this by commenting it out and the tests passed and I forgot to check 🤣

anastasialanz (approved):

PR #1523, opened 2024-06-17
fix: fixed `menuItem` test warning for link

denysfedorov:


denysfedorov:
> What ticket is this referencing? Why is it `fix`?

This PR connected with this comment https://github.com/dequelabs/cauldron/pull/1505#discussion_r1635338350
We do not have an additional ticket for it

PR #1592, opened 2024-07-25
feat(react): add Button tertiary variant

### LINKED ISSUE #1193: Add Tertiary Button
tertiary button styles in their resting, hover, focus, active, and disabled states

Tertiary buttons are an alternate style of button that presents a low priority action taking less visual prominence than primary or secondary buttons. The above image shows the tertiary button in its resting, hover, focus, active, and disabled state.

## Implementation

Add an additional variant of `tertiary` for the `Button` component.

### States

The hover, focus, and active states are visually identical to the states of the secondary variant. The main different is that the resting and disabled states do not have a visible background or border color for the button. The `thin` modifier should still be supported for the `tertiary` variant.

## Additional Notes

Dark mode's text is white to contrast with the background, all other state changes are applicable.

This is co-dependent on #1595 which will handle the associated text sizes changing.

scurker:
closes #1193


e2e/screenshots/button-thin-secondary-.png
scurker, line 1: There should be no visual change here, I noticed the text label was incorrect.

shawnsharpDQ (dismissed):

docs/pages/components/Button.mdx
thuey, line 1: Looks like the props documentation needs to be updated, too:

image

thuey (dismissed):

docs/pages/components/Button.mdx
scurker: Ah good catch 🤦 

One day, I'd like to automate this. One day....

thuey (approved):

PR #1594, opened 2024-07-25
fix(react): Dialog closes instead of contained tooltip on escape key

DelishusCake:
Fixes an issue where, when a tooltip is within a modal, the modal 'eats' the escape key event and closes first, preventing the tooltip from closing when it has focus.

Closes: #1597

Part of: dequelabs/comply#6311

### DIFF
--- a/packages/react/src/components/Dialog/Dialog.test.tsx
+++ b/packages/react/src/components/Dialog/Dialog.test.tsx
@@ -276,24 +276,16 @@ describe('when a tooltip is rendered within the dialog', () => {
     return { user, dialog, button, getTooltip };
   };
 
-  test('it should render the tooltip on hover', async () => {
+  test('it should close the tooltip but not the dialog on escape', async () => {
     const { user, dialog, button, getTooltip } = renderTooltipInDialog();
 
     expect(dialog).toBeVisible();
-
     expect(getTooltip()).not.toBeInTheDocument();
 
     await user.hover(button);
 
     expect(getTooltip()).toBeInTheDocument();
-  });
 
-  test('it should close the tooltip but not the dialog on escape', async () => {
-    const { user, dialog, button, getTooltip } = renderTooltipInDialog();
-
-    expect(dialog).toBeVisible();
-
-    await user.hover(button);
     await user.keyboard('{Escape}');
 
     expect(getTooltip()).not.toBeInTheDocument();
--- a/packages/react/src/components/Dialog/Dialog.test.tsx
+++ b/packages/react/src/components/Dialog/Dialog.test.tsx
@@ -1,9 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { render, screen, within, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { spy } from 'sinon';
 import Dialog, { DialogContent, DialogFooter } from './';
+import TooltipTabstop from '../TooltipTabstop';
 import axe from '../../axe';
-import { userEvent } from '@testing-library/user-event';
 
 const defaultProps: Omit<React.ComponentProps<typeof Dialog>, 'children'> = {
   show: true,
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Tooltip/index.tsx
```diff
@@ -78,13 +77,28 @@ export default function Tooltip({
     }
   );
 
-  // Show the tooltip
-  const show: EventListener = useCallback(async () => {
-    // Clear the hide timeout if there was one pending
+  const hideTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const maybeSetHideTimeout = (fn: () => void, delay: number) => {
```
scurker, line 81: What's this change for?

packages/react/src/components/Tooltip/index.tsx
```diff
@@ -130,21 +141,26 @@ export default function Tooltip({
         event.key === 'Esc' ||
         event.keyCode === 27
       ) {
+        event.preventDefault();
+        event.stopPropagation();
```
scurker, line 145: Generally speaking, `stopPropagation()` is an anti-pattern and something we should strictly avoid especially within a pattern library / design system.

[This article by Philip Walton is old](https://css-tricks.com/dangers-stopping-event-propagation/), but still relevant:

> This solution is roughly equivalent to fixing a leaky shower by turning off the water to the bathroom. It works, but it completely ignores the possibility that any other code on the page might need to know about that event.

scurker (changes requested):
Instead of making significant changes to `Tooltip`, I'm wondering if we need to be making changes to `Modal`.

We discussed this in slack, but you mentioned that `focus-trap` attaches listeners to `keydown`: https://github.com/focus-trap/focus-trap/blob/d1a6a897dd3b573a684c1c50b5f6927e1579c94a/index.js#L836

I think instead of relying on whatever behavior that focus-trap has implemented, we should instead lift that functionality and rely on behaviors that we can depend on, such as what we're already doing in tooltip: https://github.com/dequelabs/cauldron/blob/2657ce8bab2446496b82d65f1f16e13dee257b26/packages/react/src/components/Tooltip/index.tsx#L126-L151

Since the `keyup` function of tooltip happens during the capture phase, if we manually handle escape in modal and check for `event.defaultPrevented` this would be a guard against any "interactive widget" that uses escape key as an escape mechanism including comboboxes, popovers, tooltips, etc.

DelishusCake:
I've moved the changes into the `Dialog` component directly. We now control the dialog exit on the escape `keyup` event, and prevent closing if a component further down the tree called `preventDefault` on the event.

packages/react/src/components/Tooltip/index.tsx
DelishusCake: This was part of some cleanup for a dangling timeout that was causing a memory leak, but it's no longer relevant

packages/react/src/components/Dialog/index.tsx
```diff
@@ -53,24 +58,29 @@ export default class Dialog extends React.Component<DialogProps, DialogState> {
     this.close = this.close.bind(this);
     this.focusHeading = this.focusHeading.bind(this);
     this.handleClickOutside = this.handleClickOutside.bind(this);
+    this.bindKeyboardEvents = this.bindKeyboardEvents.bind(this);
     this.state = {};
   }
 
   componentDidMount() {
     if (this.props.show) {
+      this.bindKeyboardEvents('added');
```
scurker, line 67: Keeping in line with how we name these functions elsewhere (see https://github.com/dequelabs/cauldron/blob/256641b5883108529152d2a299088dcc6f7ef146/packages/react/src/components/ClickOutsideListener/index.tsx#L42) I would use `attachEventListeners` and `removeEventListeners` to make it a little more declarative. 

This is also in line with `attachIsolator`.

packages/react/src/components/Dialog/Dialog.test.tsx
```diff
@@ -233,6 +234,65 @@ test('should set alignment on dialog footer', () => {
   expect(footer).toHaveClass('text--align-right');
 });
 
+test('it should call "onClose" when escape is pressed', async () => {
```
scurker, line 237: This is great! This is coverage that we probably should've had before but didn't.

scurker (changes requested):
Just a very nitpicking comment on consistency with naming, but otherwise the approach looks good!

packages/react/src/components/Dialog/index.tsx
```diff
@@ -187,6 +195,28 @@ export default class Dialog extends React.Component<DialogProps, DialogState> {
     }
     this.state.isolator?.activate();
   }
+
+  bindKeyboardEvents(event: DialogEvent) {
+    const handleEscape = (keyboardEvent: KeyboardEvent) => {
+      if (!keyboardEvent.defaultPrevented && isEscape(keyboardEvent)) {
```
scurker, line 201: I think one thing we're missing either here or when we bind events is checking for `forceAction`. "Esc" appears to still calling `.close()` and returning focus to the original trigger without actually closing the modal when `forceAction` is set to true.

You can see this behavior in the deploy preview on the `Alert` component.

DelishusCake: Shoot, I forgot about `forceAction`. I've made some changes to fix that, and added an additional test case to ensure it functions as expected in both states.

DelishusCake:
Screenshot tests are currently failing, marking this as `In Progress` until the PR to get Playwright updated gets merged and I can rebase off that

packages/react/src/components/Dialog/index.tsx
```diff
@@ -119,8 +127,6 @@ export default class Dialog extends React.Component<DialogProps, DialogState> {
       <FocusTrap
         focusTrapOptions={{
           allowOutsideClick: true,
-          onDeactivate: this.close,
-          escapeDeactivates: !forceAction,
```
scurker, line 123: Looking at [`focus-trap` options](https://github.com/focus-trap/focus-trap#createoptions) it looks like we're not correctly handling this yet. `escapeDeactivates` defaults to `true`, so we likely want to always set this to `false`.

packages/react/src/components/Dialog/Dialog.test.tsx
```diff
@@ -233,6 +234,83 @@ test('should set alignment on dialog footer', () => {
...
+          onClose={() => setShow(false)}
+        >
+          <TooltipTabstop tooltip="Hello Tooltip">show tooltip</TooltipTabstop>
+        </Dialog>
+      );
+    };
+
+    const user = userEvent.setup();
+
+    const doc = render(<CustomTooltipDialog />);
```
scurker, line 287: Nit: Instead of using the return value from `render`, we typically use `screen`, example:

```tsx
render(<Component />)
screen.getByRole(...)
```

scurker (changes requested):
Last couple of things...

scurker (approved):
🎉

Issue #1582, opened 2024-07-17
Design Sync: Breadcrumb


scurker:
![screenshot of updated breadcrumb from figma designs](https://github.com/user-attachments/assets/25874e80-c55b-4c2c-a8fb-69ebb90d8bb9)

To sync with the design, the following changes are needed to breadcrumb:

- `font-weight` for non-current items is set to `--font-weight-medium` (500)
- Remove `font-weight` from `.Breadcrumb__Link`

PR #1600, opened 2024-07-30
fix(styles): synced breadcrumb with design

orest-s:
Closes: [#1582](https://github.com/dequelabs/cauldron/issues/1582)

### DIFF
--- a/packages/styles/breadcrumb.css
+++ b/packages/styles/breadcrumb.css
@@ -21,13 +21,17 @@
 }
 
 .Breadcrumb__Link {
-  font-weight: 400;
+  font-weight: inherit;
   padding: 0;
   display: inline;
 }
 
 .Breadcrumb__Item {
-  font-weight: 500;
+  font-weight: var(--font-weight-medium);
   color: var(--link-text-color);
   display: inline;
+}
+
+.Breadcrumb__Item:not(:last-child) {
+  font-weight: var(--font-weight-semi-bold);
 }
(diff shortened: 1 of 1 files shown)


packages/styles/breadcrumb.css
```diff
@@ -21,13 +21,17 @@
...
 
 .Breadcrumb__Item {
-  font-weight: 500;
+  font-weight: var(--font-weight-medium);
   color: var(--link-text-color);
   display: inline;
 }
+
+.Breadcrumb__Item:not(:last-child) {
+  font-weight: var(--font-weight-semi-bold);
```
scurker, line 36: 2 things.

This is the wrong font-weight and reads a little bit weird with the above rule. The "standard" font-weight for all of the breadcrumb items should be `--font-weight-medium` which is correct on line 30 above.

The _last_ item in the breadcrumb has the normal font-weight, this can be rewritten as:

```css
.Breadcrumb__Item:last-child {
}
```

packages/styles/breadcrumb.css
```diff
@@ -21,13 +21,17 @@
 }
 
 .Breadcrumb__Link {
-  font-weight: 400;
+  font-weight: inherit;
```
scurker, line 24: This can be removed entirely, this should be covered by the font-weight properties below.

```suggestion
```

scurker (changes requested):

scurker (approved):

PR #1603, opened 2024-07-31
feat(react): add CopyButton component

### LINKED ISSUE #970: Add "Copy" button to cauldron
Copy button is a component that accepts an optional Cauldron Button variant (defaulting to `tertiary`) to render an in interactive element to copy a specified string of text.

## Interface

```tsx
interface CopyButtonProps extends React.ComponentProps<typeof Button> {
  value: string
  children?: ContentNode
  notificationLabel?: ContentNode
  hideVisibleLabel?: boolean
  tooltipPlacement?: React.ComponentProps<typeof Tooltip>['placement']
  variant?: Extract<typeof Button, 'primary' | 'secondary' | 'tertiary'>
  onCopy?: (text: string) => void
}
```

- `variant` should default to the Button variant, `tertiary`
- `children` should default to the string, _Copy_
- `notificationLabel` should default to the string, _Copied_
- `tooltipPlacement` optional prop to position the tooltip

## Implementation

By default, the Button will have a visible label that will be displayed with children. When `hideVisibleLabel` is set, `children` should be wrapped in an `Offscreen` component with a tooltip displayed instead.

When the button is clicked, the text should be copied to the clipboard with a visible notification and offscreen aria-live notification using the `notificationLabel` text.

The existing copy icon should be updated to match the one from the design.

### Notification

When copied, a temporary notification should be visible to indicate a successful copy. This includes the following sequence:

- Display a tooltip with the value of `notificationLabel`
- Icon changes to icon `check-solid`
- An `aria-live="polite"` announcement should occur with the value of `notificationLabel`

After 2s, all of the original content can revert back to its original state.

> [!NOTE]
> The tooltip that appears for when the label is visible or hidden via `hideVisibleLabel` should always have an `aria-association` of `none`. The button should always have an explicit accessible name that doesn't change, so it's unnecessary to associate the accessible name or accessible description with the tooltip.

## Usage

This component is essentially a thin polymorphic wrapper around `<Button>` to add additional copy functionality.

```jsx
<CopyButton
  value="Text to Copy"
  onCopy={...}
>
  Copy to Clipboard
</CopyButton>
```

scurker:
closes #970


packages/react/src/components/CopyButton/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+  ) => {
+    const [copied, setCopied] = useState(false);
+    const copyButtonRef = useSharedRef(ref);
+    const handleClick = useCallback(() => {
+      copyTextToClipboard(value);
+      setCopied(true);
+      if (typeof onCopy === 'function') {
+        onCopy(value);
+      }
+    }, [value, onCopy]);
```
thuey, line 52: We have a pattern in some of our applications where we perform an asynchronous action as part of the copy. For example, if the copy button also updates the sharing visibility of some data, we typically wait until that update happens before confirming to the user that the content has been copied.

I'm not saying that the pattern is ideal, but I'm also trying to think of what that pattern would look like if we tried to incorporate this component into those flows.

packages/react/src/components/CopyButton/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+    // when we are visually displaying the notification label
+    const showTooltip =
+      hideVisibleLabel && !copied ? undefined : copied ? true : false;
+
+    return (
+      <>
+        <Button
+          className={classnames({
+            'Button--condensed': hideVisibleLabel
+          })}
```
thuey, line 74: Could we merge `props.className` into this?

packages/react/src/components/CopyButton/index.tsx
scurker: That's not something that we have an established pattern for in Cauldron/products and may benefit from its own proposal. This implementation is intended to provide the copy mechanism for static synchronous values.

thuey (approved):

Issue #1577, opened 2024-07-16
Design Sync: Notice


scurker:
![screenshot of notice component from figma](https://github.com/user-attachments/assets/8bb61b98-813b-482b-bd70-385870be8c6d)
![screenshot of condensed notice component](https://github.com/user-attachments/assets/6e7b613b-a4fe-4aac-bd71-85fb54889dfd)

To sync with the design, the following changes are needed to notice:

- `variant` property added, defaulting to `default`
- padding changed to `--space-large` for default variant for all borders
- layout is changed for the default variant, a 2-column grid is used with a gap between the icon and title/content of `var(--space-small)`
- a `4px` border radius is added
- `font-size` is removed

A new variant is added, `condensed` with the following changes:

- padding is set to `var(--space-small)` for all borders
- items are displayed within a single column

denysfedorov:
@scurker Hey, Jason!
I have a few questions about this:
1. Should we make this component look like exactly as on Figma or we need to add these properties only?
2. Should we change\modify other parts of `Notice`?

Why I am asking:
For `condensed` in the task description (and figma) we have `Icon` above the `Information` text;
For `condensed` we have approximately `max-width: 50%` for the container (it will look totally different if we only add:
```
padding is set to var(--space-small) for all borders
items are displayed within a single column
```

Also, if we have `max-width` for `variant="condensed"` -> may you send me styles from Figma? (I do not have an access to dev mode)

Thank you in advance!

denysfedorov:
That's how it looks like if we would add changes that are described in the ticket only:


![Image](https://github.com/user-attachments/assets/a2405a0a-e019-4f10-b1c5-4d4d2400ee2b)

denysfedorov:
So, if I will say only from design: seems like we need 1-column layout for both `default` and `condensed` variants, but `condensed` should have thinner width

scurker:
There's no implied width to the notice container. "Condensed" implies the visual layout of how the elements are stacked, and it will be up to the consumer to determine at what width this property gets applied to the component.

> That's how it looks like if we would add changes that are described in the ticket only:
> [...]

That's not quite right. For the default variant, the icon should be in its own column while in the condensed variant all of the elements should be "stacked". You may need to restructure the current component in order to match the designs.

PR #1605, opened 2024-08-01
fix: synced `Notice` component with design

denysfedorov:
Closes: https://github.com/dequelabs/cauldron/issues/1577

### DIFF
--- a/docs/pages/components/Notice.mdx
+++ b/docs/pages/components/Notice.mdx
@@ -112,6 +112,48 @@ function DismissableNotice() {
 }
 ```
 
+## Variants
+
+The Notice component now includes a variant prop that allows for two display styles: default and condensed.
+
+### Default variant
+
+The default variant displays the notice with a larger padding and a two-column layout.
+
+```jsx example
+<Notice variant="default" type="info" title="Information">
+  Forem ipsum dolor sit amet, consectetur adipiscing elit. Nunc vulputate libero
+  et velit interdum, ac aliquet odio mattis.
+  <ul>
+    <li>List item</li>
+    <li>List item</li>
+    <li>List item</li>
+    <li>List item</li>
+    <li>List item</li>
+  </ul>
+  <Button variant="link">Dismiss</Button>
+</Notice>
+```
+
+### Condensed variant
+
+The condensed variant of the Notice component uses smaller padding and a single-column layout.
+
+```jsx example
+<Notice variant="condensed" type="info" title="Information">
+  Forem ipsum dolor sit amet, consectetur adipiscing elit. Nunc vulputate libero
+  et velit interdum, ac aliquet odio mattis.
+  <ul>
+    <li>List item</li>
+    <li>List item</li>
+    <li>List item</li>
+    <li>List item</li>
+    <li>List item</li>
+  </ul>
+  <Button variant="link">Dismiss</Button>
+</Notice>
+```
+
 ## Props
 
 <ComponentProps
@@ -142,6 +184,13 @@ function DismissableNotice() {
       type: ['IconType', 'string'],
       description:
         'Icon name - by default automatically set to either "info-circle" or "caution" Cauldron icons'
+    },
+    {
+      name: 'variant',
+      type: ['default', 'condensed'],
+      defaultValue: 'default',
+      description:
+        'Controls the layout and padding of the Notice component. "default" uses a larger padding and a two-column layout, while "condensed" uses smaller padding and a single-column layout.'
     }
   ]}
   refType="HTMLDivElement"
--- a/docs/pages/components/Notice.mdx
+++ b/docs/pages/components/Notice.mdx
@@ -19,9 +19,9 @@ or anywhere else in the DOM without needing to be absolutely/statically position
 
 ## Examples
 
-There are two variants of the `Notice` component: `info` and `caution`. The `info` variant is used to display general
-information, while the `caution` variant is used to display a warning. Depending on the variant, the background color
-and icon will change (can be overwritten).
+The Notice component supports three types: info, caution, and danger. The info type is used to display general
+information, while the caution type is used to display a warning, and the danger type is used to display critical warnings. 
+Depending on the type, the background color and icon will change (can be overwritten).
 
 ### Info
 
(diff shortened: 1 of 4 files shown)


packages/styles/notice.css
```diff
@@ -33,6 +33,22 @@
   font-size: var(--text-size-smaller);
 }
 
+.Notice--default {
```
scurker, line 36: `.Notice` _is_ the default, so I don't think we need an extra class here to define it as a variant.

Additionally, there's a border radius that needs to be included here.

packages/styles/notice.css
```diff
@@ -33,6 +33,22 @@
...
+  display: grid;
+  padding: var(--space-large);
+  grid-template-columns: auto 1fr;
+  gap: var(--space-small);
+  align-items: start;
+}
+
+.Notice--condensed {
```
scurker, line 44: We shouldn't duplicate properties here that are defined by all variants. If we change `.Notice--default` to be applied across all notices, we can just have properties that are specific to condensed. Primarily the number of columns and reduced padding.

packages/react/src/components/Notice/index.tsx
```diff
@@ -14,27 +14,38 @@ export interface NoticeProps
...
+      ...otherProps
+    }: NoticeProps,
     ref
   ) => {
     return (
       <div
         className={classNames('Notice', {
-          [`Notice--${type}`]: type
+          [`Notice--${type}`]: type,
+          [`Notice--${variant}`]: variant
```
scurker, line 37: This should be statically typed for the variant so we only get classes for non-default variants.

docs/pages/components/Notice.mdx
```diff
@@ -112,6 +112,48 @@ function DismissableNotice() {
 }
 ```
 
+## Variants
+
+The Notice component now includes a variant prop that allows for two display styles: default and condensed.
+
```
scurker, line 118: I think some of the existing documentation needs to be re-worked a bit. We should update to specify that the info/caution/danger properties are types and not variants.

docs/pages/components/Notice.mdx
```diff
@@ -112,6 +112,48 @@ function DismissableNotice() {
 }
 ```
 
+## Variants
+
+The Notice component now includes a variant prop that allows for two display styles: default and condensed.
```
scurker, line 117: Nit: We can just specify what the component is.

```suggestion
The Notice component includes a variant prop that allows for two display styles: default and condensed.
```

docs/pages/components/Notice.mdx
```diff
@@ -112,6 +112,48 @@ function DismissableNotice() {
...
+
+### Default variant
```
scurker, line 119:
```suggestion
### Default
```

docs/pages/components/Notice.mdx
```diff
@@ -112,6 +112,48 @@ function DismissableNotice() {
...
+    <li>List item</li>
+    <li>List item</li>
+    <li>List item</li>
+    <li>List item</li>
+  </ul>
+  <Button variant="link">Dismiss</Button>
+</Notice>
+```
+
+### Condensed variant
```
scurker, line 138:
```suggestion
### Condensed
```

packages/styles/notice.css
scurker, line 1: It's not called out in the ticket as I didn't initially realize it, but the `--notice-icon-size` is set to `1.2em`. This should be set to `1rem` here.

packages/react/src/components/Notice/index.tsx
```diff
@@ -14,27 +14,38 @@ export interface NoticeProps
...
         ref={ref}
         {...otherProps}
       >
-        <div className="Notice__title">
+        <div className="Notice__icon">
           <Icon type={icon || (iconTypeMap[type] as IconType)} />
-          {title}
         </div>
-        {children && <div className="Notice__content">{children}</div>}
+        <div className="Notice__content-wrapper">
```
scurker, line 45: Since moving to a grid here, we should use something different from a wrapper. Each element can be at the top level:

```jsx
<Icon />
<div className="Notice__title">...</div>
<div className="Notice__content">...</div>
```

From there, we can define the two-column grid and conditionally place the content in the correct grid column:

```css
.Notice {
  display: grid;
  grid-template-columns: auto 1fr;
  gap: var(--space-small)
}

.Notice__content:not(.Notice--condensed) {
  grid-column: 2;
}
```

For the condensed variant, everything will just be a single column so we could even unset the grid properties.

scurker (changes requested):
This is a good start! I left some inline comments but noticed a few other things. It looks like we need some additional CSS to handle lists as the margins of the list aren't offset correctly:

image


image

One other thing I missed is that we need to remove `font-size` so it inherits from the default font size of the body as well.

denysfedorov:
@scurker Fixed review changes, fixed mdx file, fixed lists view (margin\padding)

denysfedorov:
image
image

packages/styles/notice.css
```diff
@@ -80,7 +98,16 @@
   color: var(--notice-link-hover-color);
 }
 
-.Notice .Notice__content > p:first-of-type {
+.Notice .Notice__content-wrapper > p:first-of-type {
```
scurker, line 101: I think we can remove this class now.

```suggestion
.Notice .Notice__content-wrapper > p:first-of-type {
```

packages/styles/notice.css
```diff
@@ -45,7 +63,7 @@
   color: var(--notice-title-text-color);
 }
 
-.Notice .Notice__title + .Notice__content {
+.Notice .Notice__content-wrapper {
```
scurker, line 66: I think we can remove this class now.

packages/styles/notice.css
```diff
@@ -8,7 +8,9 @@
   --notice-background-color: var(--notice-info-color);
   --notice-border-color: var(--accent-dark);
   --notice-link-hover-color: var(--accent-medium);
-  --notice-icon-size: 1.2em;
+  --notice-icon-size: 1rem;
+  --notice-border-thickness: 1px;
+  --notice-border-radius: 4px;
```
scurker, line 13: I don't think we want these to be configurable, so these can just be set via properties of the component.

```suggestion
```

scurker (commented):
There's still a few remaining `font-size` properties that need to be removed. Looks like it's getting applied to the title as well as links/buttons.

scurker (approved):

Issue #1590, opened 2024-07-25
Add Badge Component


scurker:
Badge is a container component intended to present content with an optional leading icon with a text label.

## Design Specs

![default design](https://github.com/user-attachments/assets/e654c618-f783-4b35-af11-ffa0066c2c25)

![impact badge anatomy](https://github.com/user-attachments/assets/658b3d88-ae4e-44ea-87ce-f24a96dfb480)

The `small` badge variant should have a minimum height of `24px` using `--target-size-minimum` for the small size. The icon should have roughly `1px` spacing on the horizontal axis, with the text having a right spacing of `--space-smaller`.

The `default` size should match the height of buttons, which is currently `36px`.

Badge should have customizable background/border colors from `--badge-background-color` and `--badge-border-color` with the default values set to `transparent` and `--gray-30` respectively.

## Interface

```ts
interface Badge extends React.HTMLAttributes<HTMLElement> {
  size: 'small' | 'default'
  label: ContentNode
}

interface Badge extends React.HTMLAttributes<HTMLElement> {
  label: ContentNode
}
```
## Example:

```jsx
<Badge>
  <BadgeLabel>Status</Badge>
  Value
</Badge>
```

If you need access to the designs, please reach out to a member of the Cauldron team.

PR #1610, opened 2024-08-02
feat(react): added badge component

orest-s:
Closes: #1590

### DIFF
--- a/docs/pages/components/Badge.mdx
+++ b/docs/pages/components/Badge.mdx
@@ -60,10 +60,10 @@ import { Badge, BadgeLabel, Icon } from '@deque/cauldron-react';
       type: ['default', 'small'],
       defaultValue: 'default',
       description: 'Visual style of the tag.'
-    },
+    }
   ]}
 />
 
 ### BadgeLabel
 
-<ComponentProps children={true} className={true} />
+<ComponentProps children={true} className={true} refType="HTMLDivElement" />
--- a/docs/pages/components/Badge.mdx
+++ b/docs/pages/components/Badge.mdx
@@ -1,0 +1,68 @@
+---
+title: Badge
+description: Badge is a container component intended to present content with an optional leading icon with a text label.
+source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/Badge/index.tsx
+---
+
+import { Badge, BadgeLabel } from '@deque/cauldron-react';
+
+```js
+import { Badge, BadgeLabel } from '@deque/cauldron-react';
+```
+
+## Example
+
+### Default
+
+```jsx example
+<Badge>Value</Badge>
+```
+
+### Badge with Label
+
+```jsx example
+<Badge>
+  <BadgeLabel>Label:</BadgeLabel>
+  value
+</Badge>
+```
+
+### Small size
+
+```jsx example
+<Badge size="small">
+  <BadgeLabel>Label:</BadgeLabel>
+  value
+</Badge>
+```
+
+### Badge with icon
+
+```jsx example
+<Badge>
+  <Icon type="chevron-double-up" />
+  <BadgeLabel>Label:</BadgeLabel>
+  value
+</Badge>
+```
+
+## Props
+
+### Badge
+
+<ComponentProps
+  children={true}
+  className={true}
+  props={[
+    {
+      name: 'size',
+      type: ['default', 'small'],
+      defaultValue: 'default',
+      description: 'Visual style of the tag.'
+    },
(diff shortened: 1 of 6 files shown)


docs/pages/components/Badge.mdx
scurker, line 1: You need to import `Icon` here.

docs/pages/components/Badge.mdx
```diff
@@ -0,0 +1,68 @@
...
+</Badge>
+```
+
+## Props
+
+### Badge
+
+<ComponentProps
+  children={true}
+  className={true}
```
scurker, line 55: We should also specify `refType` here.

packages/react/src/components/Badge/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+test('passes classNames through', () => {
+  render(
+    <Badge className="baz">
+      <BadgeLabel className="jazz">Badge:</BadgeLabel>
+      hi
+    </Badge>
+  );
+  expect(screen.getByText('hi')).toHaveClass('baz');
+  expect(screen.getByText('Badge:')).toHaveClass('jazz');
```
scurker, line 23: This should also include testing for the `.Badge` and `.Badge__Label` classes to ensure they're not being overridden by `className`.

packages/styles/badge.css
```diff
@@ -0,0 +1,44 @@
+:root {
+  --badge-text-color: var(--gray-90);
+  --badge-background-color: transparent;
+  --badge-border-color: var(--gray-30);
+
+  --badge-height: var(--button-height);
+  --badge-small-height: 1.5rem;
+
+  --badge-icon-size: 20px;
```
scurker, line 9: This should be inheriting from the default `--icon-size` property.

```suggestion
  --badge-icon-size: var(--icon-size);
```

packages/react/src/components/Badge/index.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <Badge data-foo="true">
+      <BadgeLabel data-bar="yes">hi</BadgeLabel>
+      bye
+    </Badge>
+  );
+  expect(screen.getByText('bye')).toHaveAttribute('data-foo', 'true');
+  expect(screen.getByText('hi')).toHaveAttribute('data-bar', 'yes');
+});
+
+test('should return no axe violations', async () => {
```
scurker, line 37: Per our guidelines we need to include all variants as part of testing, including the "small" variant, see https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#testing-strategies

packages/react/src/components/Badge/index.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import classNames from 'classnames';
+import Icon from '../Icon';
+
+interface BadgeProps extends React.HTMLAttributes<HTMLElement> {
```
scurker, line 4:
```suggestion
interface BadgeProps extends React.HTMLAttributes<HTMLDivElement> {
```

packages/react/src/components/Badge/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  children: React.ReactNode;
+  size?: 'default' | 'small';
+}
+
+interface BadgeLabelProps extends React.HTMLAttributes<HTMLElement> {
```
scurker, line 9:
```suggestion
interface BadgeLabelProps extends React.HTMLAttributes<HTMLDivElement> {
```

scurker (changes requested):

packages/react/src/components/Badge/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface BadgeProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: React.ReactNode;
+  size?: 'default' | 'small';
+}
+
+interface BadgeLabelProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: React.ReactNode;
+}
+
+export const BadgeLabel = ({
```
scurker, line 13: This is missing `React.forwardRef`

packages/react/src/components/Badge/index.test.tsx
scurker: These guidelines also include accessibility testing, so a test for `size="small"` is needed as well.

scurker (changes requested):
Last 2 things. Looking good!

scurker (approved):

Issue #1141, opened 2023-08-01
Add 4-scale `ImpactBadge` Component


scurker:
`ImpactBadge` is an extension of the [`Badge` component](https://github.com/dequelabs/cauldron/issues/1590) with more defined requirements.

## Design Specs

![impact badge with default and small sizes](https://github.com/user-attachments/assets/bae00297-4988-472f-9fa7-6e20549982af)

![impact badge anatomy](https://github.com/user-attachments/assets/658b3d88-ae4e-44ea-87ce-f24a96dfb480)

`ImpactBadge` is a primitive component to display various impact levels with an associated icon/background color:

- Critical
  - Icon: `chevron-double-up`
  - Background Color: `#fea7a6` (new global variable `--accent-error-active`, mapped from `--issue-critical`)
  - Border Color: `#db6379` (currently `--accent-error-disabled`)
- Serious
  - Icon: `chevron-up`
  - Background Color: `--accent-caution` (mapped from `--issue-serious`)
  - Border Color: `--accent-warning` 
- Moderate
  - Icon: `chevron-down`
  - Background Color: `#f0c4f8` (mapped from `--issue-moderate`)
  - Border Color: `#eb94ff` (doesn't exist in Cauldron, use as one-off)
- Minor
  - Icon: `chevron-double-down`
  - Background Color: `--gray-20` (mapped from `--issue-minor`)
  - Border Color: `--gray-30` 

The badge itself should have a minimum height of `24px` using `--target-size-minimum` for the small size. The icon should have roughly `1px` spacing on the horizontal axis, with the text having a right spacing of `--space-smaller`.

The `default` size should match the height of buttons, which is currently `36px`.

The text color of the labels should always be set to `--gray-90`.

## Interface

```ts
interface ImpactBadge extends Badge {
  type: 'critical' | 'serious' | 'moderate' | 'minor'
  label?: ContentNode
}
```

> [!NOTE]
> Cauldron should have a default label for each impact type in order to provide an accessible name by default, but still allow content authors to provide an alternative label for translation purposes.

## Example:

```jsx
<ImpactBadge type="critical" />
```

If you need access to the designs, please reach out to a member of the Cauldron team.

scurker:
Investigate different sizing (e.g. smaller icons potentially) as an alternate variant.

shawnsharpDQ:
4 sizes (XL, L, M, S). Labels are not required. Can be paired with a label, link, or left standalone. 
 
[...]

PR #1612, opened 2024-08-05
feat(react): `ImpactBadge` component

orest-s:
Closes: #1141


docs/pages/components/ImpactBadge.mdx
```diff
@@ -0,0 +1,64 @@
...
+import { ImpactBadge } from '@deque/cauldron-react';
+```
+
+## Example
+
+### Critical
+
+```jsx example
+<ImpactBadge type="critical" />
+<ImpactBadge type="critical" size="small" style={{ marginLeft: 'var(--space-half)' }} />
```
scurker, line 19: We don't typically have styles injected in the examples (particularly because people may copy/paste these examples).

My recommendation here would be to split out the sizes separately.

docs/pages/components/ImpactBadge.mdx
```diff
@@ -0,0 +1,64 @@
...
+```
+
+## Props
+
+### ImpactBadge
+
+<ComponentProps
+  className={true}
+  refType="HTMLDivElement"
+  props={[
```
scurker, line 50: This is missing `label`.

packages/react/src/components/ImpactBadge/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  ({ type, label, className, ...other }, ref) => {
+    return (
+      <Badge
+        className={classNames(`ImpactBadge`, `ImpactBadge--${type}`, className)}
+        ref={ref}
+        {...other}
+      >
+        <Icon type={iconByType[type]} />
+        <BadgeLabel>{label || 'Impact:'}</BadgeLabel>
+        {typeValues[type]}
```
scurker, line 40: This isn't quite right. We need to be able to pass in the label/value collectively for translations elsewhere - the way this is formatted cannot be translated correctly.

This should look something like:

```tsx
<Badge {...props}>
  <Icon {...} />
  {label || (
    <>
      <BadgeLabel>Impact:</BadgeLabel>
      {typeValues[type]}
    </>
  )}
</Badge>
```

packages/react/src/components/ImpactBadge/index.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { forwardRef } from 'react';
+
+import Badge, { BadgeLabel, BadgeProps } from '../Badge';
```
scurker, line 3: Minor nit:

Instead of using the type export, we should use `React.ComponentProps<typeof BadgeLabel>` here. This gives us greater flexibility in not having to be concerned with having to export the prop type.

packages/react/src/components/ImpactBadge/index.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+  expect(screen.getByText(/critical/i)).toHaveAttribute('data-foo', 'true');
+});
+
+test('should render small `ImpactBadge`', () => {
+  render(<ImpactBadge type="critical" size="small" />);
+  const SmallBadge = screen.getByText(/critical/i);
+  expect(SmallBadge).toHaveClass('ImpactBadge', 'Badge--small');
+});
+
+test('should render custom lable', () => {
```
scurker, line 95:
```suggestion
test('should render custom label', () => {
```

packages/react/src/components/ImpactBadge/index.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+  render(<ImpactBadge type="critical" label="custom label" />);
+  const CustomLabel = screen.getByText(/custom label/i);
+  expect(CustomLabel).toBeInTheDocument();
+});
+
+test('should return no axe violations', async () => {
```
scurker, line 101: Be sure we're checking all types/variants here.

packages/styles/impact-badge.css
```diff
@@ -0,0 +1,39 @@
...
+  --impact-badge-serious-background-color: var(--accent-caution);
+  --impact-badge-serious-border-color: var(--accent-warning);
+
+  --impact-badge-moderate-background-color: var(--issue-moderate);
+  --impact-badge-moderate-border-color: #eb94ff;
+
+  --impact-badge-minor-background-color: var(--issue-minor);
+  --impact-badge-minor-border-color: var(--gray-30);
+
+  --impact-badge-text-color: var(--gray-90);
```
scurker, line 14: This is almost there.

All of the background colors should be the appropriate `--issue-[type]` colors. This is done so that if someone uses the issue colors elsewhere, they're aligned with the same colors as the badge.

If the existing color doesn't match what's in the requirements, that just means we need to update the `-issue-[type]` color to be the appropriate one from the issue.

scurker (changes requested):
Inline comments of some things we need to change, but other wise, looking good!

scurker (approved):

Issue #479, opened 2021-12-17
Add indeterminate support to `<Checkbox />`


schne324:
The indeterminate state is the state of a checkbox where the checked value of the checkbox cannot be determined.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#attr-indeterminate

## Design Specs

<details>
  <summary>Note: The designs originated from UXPin and the links are no longer accessible but the concepts are still applicable. Please note the proposed design does not match the current size of checkboxes and this is intentional.</summary>

### Dark mode
![image](https://github.com/dequelabs/cauldron/assets/28606186/da93ec65-98da-45b3-86c7-ae92e64ea4e4)

### Light mode
![image](https://github.com/dequelabs/cauldron/assets/28606186/1e939ea6-b9d7-4460-9ced-d80d1f9f4e5a)

</details>

No new icons are needed for this checkbox, the indeterminate state can be rendered purely with CSS, approximate styles below:

```css
.Checkbox input[type='checkbox']:indeterminate ~ .Checkbox__overlay:before {
  content: "";
  display: block;
  position: absolute;
  height: calc(var(--icon-size) - 8px);
  width: calc(var(--icon-size) - 8px);
  background: currentColor;
  transform: translate(4px, 4px);
}

.Checkbox input[type='checkbox']:indeterminate ~ .Checkbox__overlay:after {
  content: "";
  display: block;
  position: absolute;
  height: 3px;
  width: calc(var(--icon-size) / 2);
  background-color: var(--workspace-background-color);
  transform: translate(6px, calc(var(--icon-size) / 2 * -1 - 1.5px));
}
```

## Technical Details

For more context, this [github comment thread for react](https://github.com/facebook/react/issues/1798) provides specific details about implementation challenges with indeterminate checkboxes. The HTML spec does not allow for indeterminate properties to be set via an attribute, so it must be set via javascript.

The `checked` and `indeterminate` properties are incompatible, so we need to define the default behavior of what happens to the checkbox when it becomes `indeterminate` in either its checked or unchecked state.

schne324:
Grab the svg (and add it as a new `<Icon />` type) from the [uxpin](https://preview.uxpin.com/5a35f510bfb55e3e77f9dbd54a49203736113855#/pages/144822188/specification/sitemap)

scurker:
Closing, I'm not aware of anywhere that we currently need this and we can create a new issue if needed later.

PR #1621, opened 2024-08-07
feat(react): add indeterminate support to checkbox

chornonoh-vova:
closes: #479


packages/react/src/components/Checkbox/index.tsx
```diff
@@ -36,23 +36,35 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
...
     useEffect(() => {
       setIsChecked(checked);
     }, [checked]);
 
     const refProp = ref || checkboxRef;
-    if (typeof refProp === 'function') {
-      refProp(checkRef.current);
-    }
+
+    const refCallback = (elem: HTMLInputElement | null): void => {
```
scurker, line 55: Is there a reason this ref callback function was added?

If you're trying to access the ref, `useEffect` may be what you want here.

packages/react/src/components/Checkbox/index.tsx
```diff
@@ -92,6 +110,9 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
                 onBlur(e);
               }
             }}
+            aria-checked={
```
scurker, line 113: I'm not sure this is necessary. 

Looking at the accessibility tree, a checkbox with the value of `indeterminate` already has the right properties:

chrome accessibility tree showing a value of checked: mixed

packages/styles/forms.css
```diff
@@ -20,6 +20,11 @@
   --field-icon-inactive-color: var(--gray-90);
   --field-icon-active-color: rgba(60, 122, 174, 0.25);
   --field-icon-error-active-color: rgba(217, 50, 81, 0.25);
+  --field-icon-indeterminate-color: var(--accent-primary);
+  --field-icon-indeterminate-disabled-color: var(
+    --accent-primary-dsiabled,
+    #78a6d8
+  );
```
scurker, line 27: I don't think new colors are necessary. This should be using the existing `--field-icon-xxx` colors for the appropriate states.

packages/react/src/components/Icon/icons/checkbox-indeterminate.svg
scurker, line 1: A new icon is not necessary, please take a look at an example of how this should be styled in the component proposal.

packages/react/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -46,6 +46,29 @@ test('should render disabled checked checkbox', () => {
...
 });
 
+test('should render indeterminate checkbox', () => {
+  const ref = createRef<HTMLInputElement>();
+  render(<Checkbox id="id" label="checkbox" ref={ref} indeterminate />);
+  expect(ref.current).toBeInstanceOf(HTMLInputElement);
+  expect(ref.current?.indeterminate).toBeTruthy();
+  expect(ref.current).toEqual(
+    screen.queryByRole('checkbox', { name: 'checkbox' })
+  );
```
scurker, line 56: This isn't quite how we should be doing the assertions here. We should be limiting to just what should be tested for this specific test, i.e. that an indeterminate checkbox was rendered. The way this test is written we're testing that the ref value has an indeterminate value but not that a indeterminate value was rendered on the page. Please take a look at how some of the other tests are written here to try to eliminate down to only the assertions that are necessary.

Additionally, jest-dom already has a matcher we can use, [`toBePartiallyChecked()`](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tobepartiallychecked)

packages/react/src/components/Checkbox/Checkbox.test.tsx
scurker, line 1: We need to include the `indeterminate` state as part of the accessibility checks as well.

scurker (changes requested):
This is a great start! 

As an overview for testing, please take a look at the [testing strategies](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#testing-strategies) guidelines as there's a few things we're missing.

Additionally, it's not documented (yet) but we're capturing screenshots for all interactive elements. We should include the `indeterminate` checkbox as an additional state to capture here: https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/Checkbox/screenshots.e2e.tsx

scurker:
One additional thing that I'm realizing is we're not matching what native checkboxes do. 

A checkbox can have an `indeterminate` state and be `checked` / `unchecked`. When clicked, the indeterminate state is removed and the `checked` value is flipped from it's currently value so we will likely want to match that behavior as our current checkbox component isn't "controlled".

packages/react/src/components/Checkbox/index.tsx
chornonoh-vova: Thanks, removed

packages/react/src/components/Checkbox/index.tsx
chornonoh-vova: Thanks, removed

packages/styles/forms.css
chornonoh-vova: Thanks, removed

packages/react/src/components/Icon/icons/checkbox-indeterminate.svg
chornonoh-vova: Thanks, implemented without icon

packages/react/src/components/Checkbox/Checkbox.test.tsx
chornonoh-vova: Thanks, updated tests

packages/react/src/components/Checkbox/Checkbox.test.tsx
chornonoh-vova: Thanks, updated a11y tests

scurker (approved):
At a high level, this looks good - thanks for making changes. I will give it a final pass over tomorrow morning.

packages/react/src/components/Checkbox/index.tsx
```diff
@@ -1,15 +1,14 @@
...
-  Ref,
   useState,
   useEffect,
   useRef,
   useMemo
 } from 'react';
 import classNames from 'classnames';
 import nextId from 'react-id-generator';
-import Icon from '../Icon';
+import Icon, { IconType } from '../Icon';
```
scurker, line 11:
```suggestion
import Icon, { type IconType } from '../Icon';
```

scurker (commented):
Very very minor nit, otherwise looks good!

shawnsharpDQ:
Is there anything we can do for the visibility of the disabled indeterminate state? 
image
compared to the disabled checked 
image

shawnsharpDQ (dismissed):

scurker (approved):

PR #1646, opened 2024-08-22
feat(react,styles): add Drawer component

### LINKED ISSUE #1145: Drawer Component
The drawer component is a utility component that provide slide-in/slide-out behavior to a section of content anchored to a cardinal direction on the page. An [existing example](https://github.com/dequelabs/cauldron/blob/develop/docs/components/Drawer.tsx) can be found in [cauldron documentation](https://cauldron.dequelabs.com) by collapsing the viewport until the hamburger menu appears, and can be toggled by activating that button.

## Props

```ts
interface DrawerProps {
  children: React.ReactNode
  position?: 'left' | 'right' | 'bottom' | 'top' // defaults to 'left'
  open?: boolean
  variant?: 'slide-in' | 'slide-over' // defaults to 'slide-in'
  onOpen?: () => null
}
```

## Usage

```jsx
<Drawer
  open={true}
  onOpen={...}
>
  Content to display in drawer
</Drawer>
```

The drawer component by default is closed. When closed, the drawer component should be marked as `aria-hidden`. Additionally, any interactive elements within the drawer component _must_ be not be focusable to ensure they are not keyboard accessible when closed. If the open state gets dismissed from _opened_ to _closed_, the focus should be returned to the element that initially trigged the closed state.

When the component is opened, `aria-hidden` should be removed from the drawer component and any interactive elements made interactive by allowing them to be focused. Additionally, focus should be set to the drawer component OR the first interactive element. For the `slide-over` variant, keyboard focus should remain in the drawer component until closed via a keyboard trap.

For the drawer width, there should be no default set width or height but should be accessible via a css var:

```css
.Drawer {
  width: var(--drawer-width, auto);
  max-width: 100vw;
  max-height: 100vh;
}
```

> [!NOTE]  
> Width or height variables should be set respective to the drawer orientation. Position `left` or `right` should respect the width variable, while position `top` or `bottom` should respect the height variable.

scurker:
closes #1145


packages/react/src/components/Drawer/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+    return createPortal(
+      <>
+        <ClickOutsideListener
+          onClickOutside={handleClose}
+          mouseEvent={open ? undefined : false}
+          touchEvent={open ? undefined : false}
+          target={drawerRef}
+        >
+          <FocusTrap
```
thuey, line 137: I don't know what the right answer is, but when focus is trapped, the drawer seems to act like a modal. Similar to modals, should we make the content outside the drawer hidden to screen reader users when focus is trapped?

packages/react/src/utils/useEscapeKey.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+      callback,
+      capture: true
+    })
+  );
+
+  expect(addEventListener.mock.lastCall?.[2]).toEqual(true);
+});
+
+// TODO: fix this test
+it.skip('should check for default prevented', async () => {
```
thuey, line 141: Did you mean to leave this as a TODO?

packages/react/src/components/Drawer/index.tsx
thuey: On a related note, when focus is not trapped, I wonder how valid it is to have the scrim visible. It seems like you should only have the scrim visible when focus is trapped. Or, tabbing outside of the drawer could collapse the drawer, which is another pattern I have seen

packages/react/src/utils/useEscapeKey.test.tsx
scurker: Nope 🤦

packages/react/src/components/Drawer/index.tsx
scurker: > should we make the content outside the drawer hidden to screen reader users when focus is trapped

Yes, good catch. Oversight on my part I will address this.

scurker: > On a related note, when focus is not trapped, I wonder how valid it is to have the scrim visible. It seems like you should only have the scrim visible when focus is trapped. Or, tabbing outside of the drawer could collapse the drawer, which is another pattern I have seen

There's a lot of nuance to this where I'm not 100% sure what the behavior should be. 

The [apg dialog pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) calls this out as a best practice:

> Inert content outside an active dialog is typically visually obscured or dimmed so it is difficult to discern, and in some implementations, attempts to interact with the inert content cause the dialog to close.

I was trying to have a declarative nature for both focus trap and the scrim, but maybe that's a mistake. @schne324 do you have any thoughts on this?

schne324: I agree with the APG best practice here. Does UX have a strong opinion to NOT scrim stuff out when the drawer is open?

scurker: @schne324 There will be future implementations of the drawer that are not focus traps (e.g. it's a slide out drawer that does not cover content) that are more like non-modal dialogs OR push content out of the way so it needs to be configurable.

So it sounds like where we may want to land is that focus trap / scrim should not be handled independently but controlled via a single prop. Does that make sense?

schne324: ah yes, I agree. Focus trap and scrim should be tied at the hip.

scurker: Updated this to capture the drawer as behavior `modal` or `non-modal`.

thuey (approved):

PR #1660, opened 2024-09-06
chore(workflows): use node 22 in github actions

anastasialanz:
We were getting deprecation warnings for not using node 20.

Screenshot 2024-09-05 at 9 30 05 PM

https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/


scurker (dismissed):

scurker:
Wow playwright can be annoying sometimes

anastasialanz:
> Also non-blocking, but node 20 will be going into maintenance mode soonish. I don't see any reason we couldn't just jump to 22 as it's the current LTS version.

I will update to 22 after the Playwright PR is in. 😄

anastasialanz:
@scurker I updated to Node 22!

scurker (approved):

PR #1668, opened 2024-09-10
fix(react): Make `RadioItem` public

stephenmathieson:
An application I'm working on requires using the `RadioItem` type directly, but the type is not exported from the package's entrypoint. This patch makes `RadioItem` part of the public API.

Previously, to get access to `RadioItem`, users would have to:

```ts
import {RadioItem} from "@deque/cauldron-react/lib/components/RadioGroup"
```

This is brittle (will break if we rename files/etc.) and doesn't look nearly as nice.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -48,7 +48,7 @@ export {
   SelectOption,
   SelectProps
 } from './components/Select';
-export { default as RadioGroup } from './components/RadioGroup';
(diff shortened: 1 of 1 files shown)


scurker (commented):
We typically have not been exporting the types from components, and don't really think we should do one offs without thinking about how this impacts our public api. Do you mind writing an [RFC](https://github.com/dequelabs/cauldron/blob/develop/rfcs.md) proposing why we should make this change?

Issue #1672, opened 2024-09-11
Design Sync: Link


scurker:
The following changes need to be made to the `.Link` styles:

- remove `font-weight` 
- remove all `p .Link` styles
- remove `display: inline-block`

For links that include leading or trailing icons:

```css
.Link:has(.Icon) {
  display: inline-flex;
  align-items: center;
  gap: var(--space-half);
  flex-wrap: wrap;
}
```

And remove from `a` styles from `base.css`:

- `font-weight`

Link typography should generally be inheriting styles from css cascading.

The end result should be: All components that appear visually as Links should have a font-weight of 400.

ezirlingerDQ:
Please update link style definition to 16px (1rem), 150% line height, weight - regular. This is the same style definition as body default.

PR #1675, opened 2024-09-16
feat(styles): synced Link component

orest-s:
Closes: [#1672](https://github.com/dequelabs/cauldron/issues/1672)


e2e/screenshots/dark--link-variant-button-.png
scurker, line 1: It looks like there's some cascading styles here that we may need to apply to the button variant since the font weight is no longer inherited via `.Link`. The end result should be there's no visual changes to `<Link variant="Button">`

packages/styles/link.css
```diff
@@ -12,11 +12,10 @@
 .Link {
   text-decoration: underline;
   color: var(--link-text-color);
-  font-weight: var(--font-weight-medium);
-  display: inline-block;
   padding: var(--space-quarter);
   background: transparent;
   vertical-align: baseline;
+  line-height: var(--line-height-body);
```
scurker, line 18: What's the line height change here?

scurker (changes requested):
Looks like we need to make one small tweak for the Button variant and see comment about line height.

packages/styles/link.css
orest-s: I took it from this [comment](https://github.com/dequelabs/cauldron/issues/1672#issuecomment-2344593401)

scurker: See the description:

> Link typography should generally be inheriting styles from css cascading.

If body has the correct line height, it will cascade down to the link. We don't need to specify the property again, Eric was just clarifying what the general expectation was.

packages/react/src/components/Link/index.css
```diff
@@ -1,3 +1,9 @@
+a.Button--primary,
+a.Button--secondary,
+a.Button-thin {
```
scurker, line 3:
```suggestion
a.Button--thin {
```

e2e/screenshots/button-variant-link-.png
scurker, line 1: It looks like this is still unchanged, we're missing an additional style to ensure that `<Button variant="link">` has the right styles.

scurker (changes requested):

packages/react/src/components/Link/index.css
```diff
@@ -1,3 +1,9 @@
+a.Button--primary,
+a.Button--secondary,
+a.Button--thin {
+  font-weight: var(--font-weight-normal);
+}
+
```
scurker, line 6: I think I may have a better understanding of the confusion on zoom earlier. Anything with the visual appearance of a Link should have a normal font weight. Anything with the visual appearance of the button should have the standard button font weight. Does that help make what's needed here clear?

Based on this, I don't think we need these additional styles.

```suggestion
```

scurker: That's what this comment (https://github.com/dequelabs/cauldron/pull/1675#discussion_r1761532088) was originally referring to. Sorry for the confusion.

scurker: Sorry for the back and forth, but this isn't quite still right. 

All components that have a visual appearance of a Button have a font weight of 500:

image

This is the `<Link variant="Button">`:

image

This link button variant has a font weight of 400, which does not match the expectation of what a visual Button should look like. We want to make sure that the visual links have the correct font weight of 400 while visual buttons have the correct font weight of 500.

I believe the change needed here is:

```css
a.Button--primary,
a.Button--secondary,
a.Button--thin {
}
```

The end result should be the only visual changes we should see in screenshots are for components that are visually styled as links.

scurker (approved):
🎉

PR #1680, opened 2024-09-17
feat(react): support react 17

### LINKED ISSUE #830: Bug: OptionsMenu broken in React 17+
### Steps to reproduce

The `OutsideClickHandler` click event is delegated to `document`. In React 16 and below React delegated all events to `document`. In React 17 this changed, and events are now delegated to the root node. What appears to be happening is when you click the `OptionsMenuTrigger` button associated with `OptionsMenu`, the event attached to the button opens the menu, but then the click event attached to the `document `by the `OutsideClickHandler` which gets rendered is triggered and the menu is immediately closed. If I add a click event to my trigger element and `preventDefault` on the event the menu works as expected. Based on the `defaultPrevented` check in `OutsideClickHandler`.

https://codesandbox.io/s/optionsmenu-issue-wgq1uh?file=/src/App.js

### Expected Behavior

When using the `OptionsMenu` component in a React 17+ app, clicking the button associated with `OptionsMenu` should open the menu without having to `preventDefault` on the click event.

### Actual Behavior

When using the `OptionsMenu` component in a React 17+ app, clicking the button associated with `OptionsMenu` ONLY opens the menu when the trigger button prevents the default behavior on the event using `preventDefault`.

### Version

4.7.0

_No response_

### Relevant log output

_No response_

Additional information

### LINKED ISSUE #1087: Bug: `Modal` broken in React 17+
### Steps to reproduce

Assuming you're using React 17.0.2 in your project, when clicking on the modal toggle button to open the modal, the `ClickOutsideListener` calls `handleClickOutside` and then the `close` method in the `Dialog` component. This prevents the modal from opening.

You can switch the React and React DOM version to 16.14.0 to see it working previously.
https://codesandbox.io/s/modal-issue-vxwiht?file=/src/App.js

Related comment: https://github.com/dequelabs/cauldron/issues/631#issuecomment-1267196016

### Expected Behavior

When using the `Modal` component in React 17+, the modal should open when triggered.

### Actual Behavior

When using the `Modal` component in React 17+, the modal does not open.

### Version

5.3.0

_No response_

### Relevant log output

_No response_

Additional information

### LINKED ISSUE #1473: Update to `React 17`
Once #1159 is complete, we should upgrade as a half-step to React 17. This should include both `react` and `react-dom` for development dependencies in the following packages:

- `./` (root)
- `packages/react`

No changes needed for `peerDependencies` until React 18 (see #631)

Also see https://legacy.reactjs.org/blog/2020/08/10/react-v17-rc.html for a reference of changes that may be needed. Running the newly updated tests should be sufficient to see if any specific issues are caught as the result of the upgrade.

### LINKED ISSUE #1574: Bug: Unable to toggle Popover in React 17
### Steps to reproduce

Have React 17 installed/running with the latest version of Cauldron. Have a Popover with a triggered button:

```jsx
function Example() {
  const [show, setShow] = useState(false);
  const onTogglePopover = () => setShow(!show);
  const buttonRef = useRef();

  return (
    <>
      <Button 
        onClick={onTogglePopover} 
        variant="secondary" 
        buttonRef={buttonRef}
      >
        Prompt!
      </Button>
      <Popover
        variant="prompt"
        target={buttonRef} 
        placement="auto" 
        show={show} 
        onClose={onTogglePopover}
        infoText="Are you sure you want to exit this wizard? All changes will be lost."
      />
    </>
  )
}
```

### Expected Behavior

The popover should appear when the button is clicked.

### Actual Behavior

The popover does not appear.

This is likely related to #1087 and #830. All of these components use `ClickOutsideListener` where the listener is activated preventing the actions from occurring.

anastasialanz:
closes #1473 
closes #1574 
closes #1087 
closes #830 

This PR does a few things:
- For the `OptionsMenu` and `Popover` component, when a user opens them, clicking on the trigger buttons will now close them instead of triggering the click outside `close()` method
- Converts the `OptionsMenu` components to function components
- Updates the `TopBar` docs so that the `OptionsMenu` now has a trigger to reference

## Verification Steps
- [ ] Go to the `OptionsMenu` component
  - [ ] Open the `OptionsMenu`
  - [ ] Clicking on the trigger for the `OptionsMenu` will close the `OptionsMenu`
  - [ ] Verify that the click outside handler works
- [ ] Go to the `Popover` component
  - [ ] Open the `Popover`
  - [ ] Clicking on the trigger for the `Popover` will close the `Popover`
  - [ ] Verify that the click outside handler works
- [ ] Go to the `TopBar` with menu component
  - [ ] Open the `OptionsMenu` inside the `TopBar`
  - [ ] Clicking on the trigger for the `OptionsMenu` will close the `OptionsMenu`
  - [ ] Verify that the click outside handler works
- [ ] Go to the Modal component
  - [ ] Verify that the Modal component opens and closes with the trigger


packages/react/src/components/Combobox/Combobox.test.tsx
```diff
@@ -545,6 +548,9 @@ test('should not prevent default with enter keypress and closed listbox', () =>
     // rtl doesn't let us mock preventDefault
     // see: https://github.com/testing-library/react-testing-library/issues/572
     event.preventDefault = preventDefault;
+    if (type === 'focus') {
```
anastasialanz, line 551: From Jason:

> The changes here I suspect are related to https://github.com/testing-library/user-event/issues/592, but we're not using user-event but are calling fireEvent due to specific limitations with testing library.

packages/react/src/components/Popover/index.test.tsx
```diff
@@ -169,12 +148,8 @@ test('should call onClose on escape keypress', async () => {
...
   const onClose = jest.fn();
   const user = userEvent.setup();
-  render(<Wrapper tooltipProps={{ onClose }} />, {
-    container: mountNode as HTMLElement
-  });
-  const outsideButton = await screen.findByText(/Click Me!/i);
-  expect(outsideButton).toBeTruthy();
-  await user.click(outsideButton);
+  render(<Wrapper tooltipProps={{ onClose }} />);
+  await user.click(document.body);
```
anastasialanz, line 152: We don't need to add any other elements to test clicking outside. We can click on the `body` element instead.

packages/react/src/components/Popover/index.test.tsx
```diff
@@ -5,39 +5,20 @@ import axe from '../../axe';
...
-  document.body.innerHTML = '';
-  wrapperNode = null;
-  mountNode = null;
-});
-
-const Wrapper = ({
-  buttonProps = {},
-  tooltipProps = {}
-}: {
+interface PopoverProps {
```
anastasialanz, line 8: Since these types were repeated a few times.

packages/react/src/components/Popover/index.test.tsx
```diff
@@ -197,13 +172,23 @@ test('onClose should be called, when close button in prompt popover is clicked',
   const handleClose = jest.fn();
   render(<WrapperPrompt tooltipProps={{ onClose: handleClose }} />);
   fireEvent.click(screen.getByText('Close'));
-  await waitFor(() => expect(handleClose).toHaveBeenCalled());
+  await waitFor(() => expect(handleClose).toHaveBeenCalledTimes(1));
+});
+
+test('close opened popover when clicking on the popover trigger button', async () => {
```
anastasialanz, line 178: This test verifies that the content is not in the document once the popover button is clicked. It's failing so we know it's a bug.

packages/react/src/components/Popover/index.test.tsx
```diff
@@ -277,7 +276,7 @@ test('should have no axe violations for prompt variant', async () => {
   expect(await axe(baseElement)).toHaveNoViolations();
 });
 
-test('aria-labelleddby should exist for variant="custom"', async () => {
+test('aria-labelledby should exist for variant="custom"', async () => {
```
anastasialanz, line 279: Typo!

packages/react/src/components/Popover/index.test.tsx
```diff
@@ -286,7 +285,7 @@ test('aria-labelleddby should exist for variant="custom"', async () => {
   expect(ariaLabelledById).toBeTruthy();
 });
 
-test('aria-labelleddby should not exist if aria-label provided for variant="prompt"', async () => {
+test('aria-labelledby should not exist if aria-label provided for variant="prompt"', async () => {
```
anastasialanz, line 288: Same typo as above.

packages/react/src/components/ClickOutsideListener/index.tsx
anastasialanz, line 1: These changes are recommended here https://legacy.reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues

scurker: We'll want to be sure we call this out in the release, might be good to tag folks in the slack channel. I don't think anyone should be relying on the capture vs bubble phase but you never know...

packages/react/src/components/OptionsMenu/OptionsMenuList.tsx
```diff
@@ -9,58 +9,42 @@ const [up, down, tab, enter, space, esc] = [38, 40, 9, 13, 32, 27];
...
+
+  useEffect(() => {
+    if (show && itemRefs.current.length) {
       // handles opens
-      this.itemRefs[0]?.focus();
-      this.setState({ itemIndex: 0 });
-    } else if (prevState.itemIndex !== itemIndex) {
-      // handle up/down arrows
-      this.itemRefs[itemIndex]?.focus();
+      itemRefs.current[0]?.focus();
```
scurker, line 33: This is a prime example of why #1703 would be useful

packages/react/src/components/OptionsMenu/OptionsMenuList.tsx
```diff
@@ -9,58 +9,42 @@ const [up, down, tab, enter, space, esc] = [38, 40, 9, 13, 32, 27];
 export interface OptionsMenuListProps
   extends Omit<OptionsMenuProps, 'trigger'> {
   className?: string;
+  triggerRef: React.RefObject<HTMLButtonElement>;
 }
```
scurker, line 13: Unfortunately, this appears to be a breaking change. Does this change still need to be made even when we're checking the event in the capture phase?

scurker (changes requested):
Everything looks pretty good, but I do think we need to address the breaking change here.

packages/react/src/components/OptionsMenu/OptionsMenuList.tsx
anastasialanz: This change is needed because the trigger button is outside of the `ClickOutsideListener` component which cause the menu to not be able to close after it was opened.

scurker: I pulled down your PR, and I have a few thoughts about this.

One way we could handle this in a non-breaking manner, is to check for `event.defaultPrevented` here: 

https://github.com/dequelabs/cauldron/blob/9df13e6645fce1903a742900c9b0db304a334546/packages/react/src/components/OptionsMenu/OptionsMenu.tsx#L48-L50

Within the click outside callback we could then set `event.preventDefault` if the menu list is currently shown. This does come with the downside of it would prevent immediate clicks on elements outside of the menu while the list is open so not the best solution, but also non-breaking.

Aside from that, we need to consider these components could also be constructed piecemeal so we need to accommodate for that scenario as well. Instead of explicitly trying to set the ref on a component (and not all components correctly implement the `ref` property) we might want to consider capturing the element that triggered the open state:

```ts
const triggerRef = useRef<HTMLElement>()
const toggleMenu = (event) => {
  if (!show) {
    triggerRef.current = document.activeElement
  } else {
    triggerRef.current = null
  }
  setShow(!show);
};
```

OptionsMenuList could _optionally_ accept the ref (to be non breaking):

```ts
export interface OptionsMenuListProps
  extends Omit<OptionsMenuProps, 'trigger'> {
  className?: string;
  triggerRef?: React.RefObject<HTMLElement>
}

...

const handleClickOutside = (e: MouseEvent | TouchEvent) => {
  if (triggerRef && triggerRef.current) {
    // use existing checks
  }
}
```

IMO it makes it slightly more robust without there needing to be a breaking change. Thoughts?

scurker: In the long run, we're on the path to create a brand new component that will replace this one and deprecate usage of this component. It has a lot odd patterns that do not support some of the mechanisms we're trying to achieve and any changes we would make would need to be breaking. A brand new component to move away from some of these patterns without breaking immediate usage is the path of least resistance.

anastasialanz: > I pulled down your PR, and I have a few thoughts about this.
> 
> One way we could handle this in a non-breaking manner, is to check for `event.defaultPrevented` here:
> 
> https://github.com/dequelabs/cauldron/blob/9df13e6645fce1903a742900c9b0db304a334546/packages/react/src/components/OptionsMenu/OptionsMenu.tsx#L48-L50
> 
> Within the click outside callback we could then set `event.preventDefault` if the menu list is currently shown. This does come with the downside of it would prevent immediate clicks on elements outside of the menu while the list is open so not the best solution, but also non-breaking.

I tried out what you said above and see that it works for the `OptionsMenu` component but it doesn't fix the issue where the `OptionsMenuList` is within the `TopBarMenu` in the `TopBar` example.

> Aside from that, we need to consider these components could also be constructed piecemeal so we need to accommodate for that scenario as well. Instead of explicitly trying to set the ref on a component (and not all components correctly implement the `ref` property) we might want to consider capturing the element that triggered the open state:
> 
> ```ts
> const triggerRef = useRef<HTMLElement>()
> const toggleMenu = (event) => {
>   if (!show) {
>     triggerRef.current = document.activeElement
>   } else {
>     triggerRef.current = null
>   }
>   setShow(!show);
> };
> ```
> 
> OptionsMenuList could _optionally_ accept the ref (to be non breaking):
> 
> ```ts
> export interface OptionsMenuListProps
>   extends Omit<OptionsMenuProps, 'trigger'> {
>   className?: string;
>   triggerRef?: React.RefObject<HTMLElement>
> }
> 
> ...
> 
> const handleClickOutside = (e: MouseEvent | TouchEvent) => {
>   if (triggerRef && triggerRef.current) {
>     // use existing checks
>   }
> }
> ```
> 
> IMO it makes it slightly more robust without there needing to be a breaking change. Thoughts?

If you look at my changes for the `TopBar` docs, I set the `menuItemRef` and then pass that to the `OptionsMenuList` which would accept it optionally. Even if the triggerRef prop is optional, the OptionsMenuList still wouldn't open unless users set up the `menuItemRef`.

https://github.com/dequelabs/cauldron/blob/555e12f0f032b584fca04d427a54446c79c12409/docs/pages/components/TopBar.mdx?plain=1#L65-L85

anastasialanz: Updated here to not use a ref prop https://github.com/dequelabs/cauldron/pull/1680/commits/aa0642d9087648fb1c4636e354a19187067aec62

docs/pages/components/TopBar.mdx
```diff
@@ -4,6 +4,7 @@ description: A container component that typically contains information about the
 source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/TopBar/TopBar.tsx
 ---
 
+import { useRef } from 'react'
```
scurker, line 7: It looks like this can safely be removed.

scurker (commented):
Code changes LGTM!

I did a check with everything in the deploy preview and it looks like we've resolved all of the issues! There does appear to be 2 tests failing which we need to determine why they're failing before this can be merged, but otherwise great work!

docs/pages/components/TopBar.mdx
anastasialanz: Removed here https://github.com/dequelabs/cauldron/pull/1680/commits/0b33e195335162011cbdafa79eb49c8d477b2435

anastasialanz:
> Code changes LGTM!
> 
> I did a check with everything in the deploy preview and it looks like we've resolved all of the issues! There does appear to be 2 tests failing which we need to determine why they're failing before this can be merged, but otherwise great work!

I updated the onClose test. The new code in the `useEffect` for `OptionsMenuList` was returning early and not closing because the triggerElement was the `body` element which contains the `target`. I needed to modify the test to add a trigger, focus on it, and then rerender with the `OptionsMenuList` open which mimics what you would see when used within another component.

scurker (approved):
🎈 🎉 🎆

PR #1686, opened 2024-09-19
fix(react): update focus-trap-react to remove usage of findDomNode

### LINKED ISSUE #1685: Bug: `Modal` Component is broken in React 18+
### Steps to reproduce

Assuming you're using React 18+ in your project, when clicking on the modal toggle button to open the modal, the `ClickOutsideListener` calls `handleClickOutside` and then the `close` method in the `Dialog` component. This prevents the modal from opening. Also, you can see this error in the console: 
image

You can switch the React and React DOM version to 16.14.0 to see it working previously.
https://codesandbox.io/p/sandbox/modal-issue-vxwiht

Similar to: https://github.com/dequelabs/cauldron/issues/1087

### Expected Behavior

When using the `Modal` component in React 18+, the modal should open when triggered and we should not see an error message;

### Actual Behavior

When using the `Modal` component in React 18+, the modal does not open.

### Version

latest

_No response_

### Relevant log output

_No response_

_No response_

scurker:
closes #1685

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -25,7 +25,7 @@
   "dependencies": {
     "@popperjs/core": "^2.5.4",
     "classnames": "^2.2.6",
-    "focus-trap-react": "^10.2.3",
+    "focus-trap-react": "^9",
     "focusable": "^2.3.0",
     "keyname": "^0.1.0",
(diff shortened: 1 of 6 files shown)


scurker:
May be addressed by #1680

PR #1687, opened 2024-09-19
feat(ClickOutsideListener): refactor and allow for ref targets

scurker:
This is a slight refactor of `ClickOutsideListener` to help address some future work that will likely be needed to address React 17/18 upgrade issues, see:

- #1685 
- #1087 
- #1574


anastasialanz (approved):

anastasialanz:
I double check anything that used ClickOutsideListener and it all seems to work the same.

Issue #1679, opened 2024-09-17
Design Sync: TextField


scurker:
## Designs

![](https://github.com/user-attachments/assets/62dc9e20-5e7f-4b5e-b3f8-6cea8fb3d987)

## Changes

- Add border radius on all "textfield" components, including:
  - TextField
  - Combobox
  - SearchField
  - Select
- 1 border color for everything (`--gray-60`), no dark border underline

PR #1696, opened 2024-09-30
feat(styles): synced TextField component with designs

orest-s:
Closes: [#1679](https://github.com/dequelabs/cauldron/issues/1679)


packages/styles/forms.css
```diff
@@ -11,6 +11,7 @@
   --field-border-color-error-hover: #c42f41;
   --field-border-color-error-focus-glow: rgba(217, 50, 81, 0.7);
   --field-border-color-disabled: rgba(204, 204, 204, 0.25);
+  --field-border-radius: var(--space-half);
```
scurker, line 14: This is inappropriate usage of `--space-half`. Spacing variables are intended to be used for spacing things, like padding and margins. We can just outright set border radius to 4px.

```suggestion
  --field-border-radius: 4px;
```

packages/styles/text-field-wrapper.css
```diff
@@ -2,9 +2,7 @@
   --text-field-wrapper-border-color: var(--field-border-color);
   --text-field-wrapper-background-color: var(--field-background-color);
   --text-field-wrapper-border-focus-color: var(--field-border-color-focus);
-  --text-field-wrapper-border-hover-color: var(
-    --field-border-color-focus-hover
-  );
```
scurker, line 7: Just curious, why just this change for light mode and not dark mode?

scurker (changes requested):
Minor things.

Also design review needed from @dequelabs/cauldron-design

packages/styles/text-field-wrapper.css
orest-s: Actually, i didn't know if it applies on dark mode as well, since screenshot in the description related only to light mode, should it be applied on dark as well?

shawnsharpDQ (approved):
@orest-s no issues found.
@scurker I wanted to mention that we do not have a `hover focus state` for dark theme. Is that required and should be added? It looks like has never existed before.

packages/styles/text-field-wrapper.css
scurker: Cauldron has to handle both light/dark modes, I'm assuming this was just a small oversight in the designs.

@shawnsharpDQ ideally we have somewhat consistent behaviors for both light and dark modes. Since we're making changes to inputs here do you have any recommendations on what we want to do to align here? It looks like dark mode has the focus ring color, is that the behavior we want to stick with?

shawnsharpDQ: I would recommend to use `accent-light` for the hover color in dark theme. This is the same hover color we are using for secondary buttons, checkboxes, and radios.

packages/styles/button.css
```diff
@@ -126,7 +126,7 @@ button.Link {
 .Button--tertiary {
   background-color: var(--button-background-color-secondary);
   color: var(--button-text-color-dark);
-  border: 1px solid var(--field-border-color);
+  border: 1px solid var(--gray-40);
```
scurker, line 129: This should be using the scoped variable `--field-border-color`. If it needs to be different for light/dark modes the variable will need to be scoped to `:root` or `.cauldron--theme-dark` respectively.

scurker (changes requested):
Not sure why some of the screenshots have changed, but everything generally looks good aside from the one blocking comment.

scurker (approved):

PR #1730, opened 2024-10-18
feat(react): remove focus-trap-react replacing with internally managed focus traps

### LINKED ISSUE #1703: Replace `react-focus-trap` with internally managed Focus Management
We've been encountering a number of issues with `react-focus-trap` over the years. There's some limiting factors to `react-focus-trap`, such as:

- [`focus-trap-options` only being read upon initial renders](https://github.com/focus-trap/focus-trap-react?tab=readme-ov-file#focustrapoptions)
- A bit finicky with rendering non-ref `children`, can cause issues when used in combination with `ClickOutsideListener`
- `containerElements` requiring a list of elements and not immediately working with refs
- Can create actual focus traps in some rare scenarios (does not handle web components correctly, does not handle nameless radio groups)

Having our own mechanism of trapping/managing focus that better adapts to the paradigms that we use would help to reduce this friction, as well as provide a mechanism for this to be used for future behaviors outside of Cauldron.

As an inspiration, [Adobe Spectrum has a `FocusScope` and `useFocusManager` hook](https://react-spectrum.adobe.com/react-aria/FocusScope.html) that would be much better aligned with how we write our components today. This would provide much greater flexibility than what `react-focus-trap` currently offers today.

## Concepts

At a high level the necessary behaviors could be contained within individual hooks to encapsulate containing focus plus any necessary configuration options to initialize and configure focus traps:

```ts
type useFocusTrapProps<
  TargetElement extends Element = Element,
  FocusElement extends HTMLElement = HTMLElement
> = (
  target: ElementOrRef<TargetElement>,
  options: {
    /**
     * When set to false, deactivates the focus trap. This can be necessary if
     * a component needs to conditionally manage focus traps.
     */
    disabled?: boolean;
    /**
     * When the trap is activated, an optional custom element or ref
     * can be provided to override the default initial focus element behavior.
     */
    initialFocusElement?: ElementOrRef<FocusElement>;
    /**
     * When set to true and the trap is deactivated, this will return focus
     * back to the original active element or the return focus element if
     * provided.
     */
    returnFocus?: boolean;
    /**
     * When the trap is deactivated, an optional custom element or ref
     * can be provided to override the default active element behavior.
     */
    returnFocusElement?: ElementOrRef<FocusElement>;
  }
) => void
```

This could replace the focus trap in the following components:

- `Dialog`
- `Drawer`
- `Popover`
- `TwoColumnPanel`

The advantage of utilizing a hook vs a component is removing any complexities that from "focus scope management" having to figure out what the containing child should be and trying to align on a single responsibility of _just_ focus management. This hook should not be concerned with clicking outside, whether or not things outside of the scope are `aria-hidden`, but purely focused on just containing focus within the element.

scurker:
closes #1703 

Any existing components that were utilizing `focus-trap-react` should now be using the internal `useFocusTrap` hook. Some small refactors had to take place to accommodate the new hook, but otherwise the functionality should be identical.


packages/react/src/components/Dialog/index.tsx
```diff
@@ -28,196 +28,159 @@ export interface DialogProps extends React.HTMLAttributes<HTMLDivElement> {
...
-      },
-      done
-    );
-  }
-
-  render() {
-    const {
-      dialogRef,
-      forceAction,
+const Dialog = forwardRef<HTMLDivElement, DialogProps>(
```
scurker, line 34: Most of the primary work here was to convert Dialog to a functional component instead of class one to use the focus trap hook.

packages/react/src/utils/useFocusTrap.ts
```diff
@@ -0,0 +1,245 @@
...
+  trapGuard.setAttribute('aria-hidden', 'true');
+  return trapGuard;
+}
+
+function createFocusTrap(
+  targetElement: Element,
+  initialFocusElement: HTMLElement | null
+): FocusTrap {
+  const startGuard = createTrapGuard();
+  const endGuard = createTrapGuard();
```
scurker, line 73: Part of the methodology I evaluated to switching to use trap guards included:

- Not wanting to keep track of tabbable elements (newly added or removed) within the focus trap
- Relying on native browser behavior for forward/reverse focus tab navigation
- Not wanting to account for tabIndex order when tracking elements

Given the many nuances to tracking tabbable elements within the trap, using focus guards appeared to be a much more sustainable and maintainable approach.

docs/pages/components/Scrim.mdx
```diff
@@ -13,7 +13,7 @@ import { Scrim } from '@deque/cauldron-react'
 A scrim is a component that is a semi-transparent overlay over the viewport to provide emphasis, focus, or draw attention to certain information. This is most commonly used as a backdrop for dialogs, modals, and other elements that need primary focus on the page.
 
 <Note>
-  It is recommended that a focus trap is used to contain keyboard focus within the element that has primary focus. [Alert](./Alert) and [Modal](./Modal) both use [react-focus-trap](https://www.npmjs.com/package/focus-trap-react) to manage focus. If `Scrim` is being used with other components, a focus trap may be needed to prevent certain keyboard accessibility issues.
+  It is recommended that a focus trap is used to contain keyboard focus within the element that has primary focus. If `Scrim` is being used with other components, a focus trap may be needed to prevent certain keyboard accessibility issues.
```
schne324, line 16: nit

```suggestion
  It is recommended that a focus trap is used to contain keyboard focus within the element that has primary focus. If `Scrim` is being used with other components, a focus trap (among other AT considerations) may be needed to prevent certain keyboard accessibility issues.
```

packages/react/src/components/Dialog/index.tsx
```diff
@@ -28,196 +28,159 @@ export interface DialogProps extends React.HTMLAttributes<HTMLDivElement> {
...
+      if (show && !forceAction) {
+        handleClose();
+      }
+    }, [show, forceAction, handleClose]);
+
+    const focusHeading = useCallback(() => {
+      if (headingRef.current) {
+        headingRef.current.focus();
+      }
+      isolatorRef.current?.activate();
```
schne324, line 72: just curious - why optional chaining here but not above (`headingRef.current?.focus();`)

packages/react/src/components/Dialog/index.tsx
```diff
@@ -28,196 +28,159 @@ export interface DialogProps extends React.HTMLAttributes<HTMLDivElement> {
...
+        }
+      },
+      [handleClose]
+    );
+
+    useEffect(() => {
+      if (!show || !dialogRef.current) return;
+
+      isolatorRef.current = new AriaIsolate(dialogRef.current);
+      setTimeout(focusHeading);
```
schne324, line 88: nit - this use of `setTimeout` could probably benefit from a comment (I assume this is to wait for the dialog to be rendered, if so there may be a way to avoid setTimeout use)

packages/react/src/components/Drawer/Drawer.test.tsx
```diff
@@ -305,7 +305,7 @@ test('should not trap focus when behavior is non-modal', async () => {
     'aria-hidden',
     'true'
   );
-  await user.keyboard('{Tab}');
+  document.body.focus();
   expect(document.body).toHaveFocus();
```
schne324, line 309: This reads kinda funny:
- focus the body
- assert that the body is focused

Is the expect necessary? Just to confirm that body.focus() actually worked?

schne324 (dismissed):
Looks good to me. I left a couple of nits. I'm also wondering if we should document anything about not using iframes within things that need focus traps as it could mess with the focus trap code (if, for example, the first tab stop in a dialog is within an iframe)

schne324 (dismissed):

packages/react/src/components/Dialog/index.tsx
scurker: To be honest - I'm not sure since that's how the code was written before but I'm down to change this.

packages/react/src/components/Dialog/index.tsx
scurker: I agree, but I wish I remembered why as it's been like this since it's initial implementation: https://github.com/dequelabs/cauldron/blob/eb5def134fcae7ec53e00aa6a47d2837c6af2783/packages/react/src/components/Dialog/index.tsx#L65-L66

I'm taking an educated guess that it's queuing the focus to be on the next tick to prevent focus conflicts. We can add that as a comment if you feel that's useful?

packages/react/src/components/Drawer/Drawer.test.tsx
scurker: Yeah I guess the assertion is a bit redundant since we should assume that the body has focus after we call `body.focus()`.

scurker:
> Looks good to me. I left a couple of nits. I'm also wondering if we should document anything about not using iframes within things that need focus traps as it could mess with the focus trap code (if, for example, the first tab stop in a dialog is within an iframe)

I'm not following 100%.

I just tested an alert with an iframe inside of it and it works as expected (since we focus the header by default) and focus works exactly as you would expect. Is there a particular edge case scenario you would be looking to cover?

schne324:
> I'm not following 100%.
> 
> I just tested an alert with an iframe inside of it and it works as expected (since we focus the header by default) and focus works exactly as you would expect. Is there a particular edge case scenario you would be looking to cover?

I was making an assumption that funky things could happen because we have this code:

```ts
    const focusableElements = Array.from(
      targetElement?.querySelectorAll(focusable) || []
    ) as HTMLElement[];
```

and any focusable elements that are within an iframe would not be captured in that `focusableElements` array.

schne324 (approved):

scurker:
> I was making an assumption that funky things could happen because we have this code:
> 
> [...]
> 
> and any focusable elements that are within an iframe would not be captured in that `focusableElements` array.
> 
> [...]
> 
> the first and last elements in focusable could be, in reality, within an iframe but in this array they are not captured at all

Ah right. With that code there, the focusable elements would only really be applicable for the first or last element (to determine where the _next_ focus needs to go). In my test case I still had surrounding elements within the iframe and it works as expected. So this would be a bit of an edge case where the last or first focusable element was within an iframe.

I feel that's an acceptable outcome for us, but I'm not sure how we want to take note of it. The majoritive use case would be working within a single document context.

schne324 (approved):

Issue #631, opened 2022-04-21
Upgrade to React v18


stephenmathieson:
Once #1159 and #1473 is complete, we should upgrade to the latest version of React 18. This should include both `react` and `react-dom` for development dependencies in the following packages:

- `./` (root)
- `packages/react`

Peer dependencies can be changed to the following:

```js
"peerDependencies": {
  "react": ">=16.6",
  "react-dom": ">=16.6"
}
```

Also see https://react.dev/blog/2022/03/08/react-18-upgrade-guide for a reference of changes that may be needed. Running the newly updated tests should help catch any lingering issues, but a full breadth testing of Cauldron documentation should be done to identify any possible issues with each component.

scurker:
There's a possibility that we could unofficially support React 17/18, I'm not seeing any breaking changes for 17. We could temporarily use this [third party adapter](https://www.npmjs.com/package/@wojtekmaj/enzyme-adapter-react-17) to allow cauldron to use React 17. There may not be anything preventing someone from using cauldron in 18, we would just need to validate within the apps.

scurker:
It looks like there's some breaking changes with [v18](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html):

* [updates to type definitions](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-typescript-definitions)
* [event automatic batching](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#automatic-batching)
* changes to `useEffect` timings

stephenmathieson:
The updates to the types are what's preventing us from upgrading axe.deque.com to React 18. We should get those sorted out ASAP.

abbeyperini-deque:
Breaking change for react-dom 17: modal toggle is called twice on open in `4.7.0-canary.a816b304` (fix from Pr #698) and four times with `4.7.0-canary.87de5e62`

scurker:
[...]

PR #1731, opened 2024-10-18
feat: upgrade to react 18

anastasialanz:
https://github.com/dequelabs/cauldron/issues/631
https://github.com/dequelabs/cauldron/issues/1685

This PR updates Cauldron to use React 18.
https://react.dev/blog/2022/03/08/react-18-upgrade-guide

- Updates deprecated `JSX.Element` type to `React.JSX.Element`
- Switches some focus tests to `user.tab`
- Fixes `Warning: componentWillMount has been renamed` when viewing the docs site
- Fixes `ReactDOM.render is no longer supported in React 18` when viewing the docs site
- Fixes `ReactDOM.render is no longer supported in React 18` when running `yarn test`
- Fixes `ReferenceError: TextEncoder is not defined` error in `src/components/Select/index.test.tsx`
- createPortal types issue that came up https://github.com/DefinitelyTyped/DefinitelyTyped/issues/66841

## Verification steps
Verify each component
- [x] Accordion
- [x] Address
- [x] Alert
- [x] Badge
-  🚫  BottomSheet - `childElementRef.current.contains is not a function` appears when closing the bottom sheet. Issue for this https://github.com/dequelabs/cauldron/issues/1747 since it currently exists.
- [x] Breadcrumb
- [x] Button
- [x] Checkbox
- [x] ClickOutsideListener - it seems slow to toggle the alert
- [x] Code
- [x] Combobox - Showing `An element or ref was provided to useIntersectionRef that was not an HTMLElement.` on the docs page. I'll make a new issue for as well since it also previously existed.
- [x] CopyButton
- [x] DescriptionList
- [x] Drawer
- [x] ExpandCollapsePanel
- [x] FieldWrap
- [x] Icon
- [x] IconButton
- [x] ImpactBadge
- [x] Layout
- [x] Line
- [x] Link
- [x] Listbox
- [x] Loader
- [x] LoaderOverlay
- [x] Modal
- [x] NavBar
- [x] Notice
- [x] Offscreen
- [x] OptionsMenu
- [x] Pagination
- [x] Panel
- [x] Popover
- [x] ProgressBar
- [x] RadioCardGroup
- [x] RadioGroup
- [x] Scrim
- [x] SearchField
- [x] Select
- [x] SkipLink
- [x] Stepper
- [x] Table - docs error for Each child in a list should have a unique "key" prop.
- [x] Tabs
- [x] TagButton
- [x] TextEllipsis
- [x] TextField
- [x] Toast
- [x] Timeline
- [x] Tooltip
- [x] TooltipTabstop
- [x] TopBar
- [x] TwoColumnPanel


docs/mdx-components.tsx
```diff
@@ -52,7 +52,7 @@ function Link({
 }
 
 interface MDXComponentProps {
-  [key: string]: (props: unknown) => JSX.Element;
+  [key: string]: (props: unknown) => React.JSX.Element;
```
scurker, line 55: Ok wow, didn't realize the global namespace was deprecated and I thought it was part of the global typescript interfaces (discovered the deprecation here):  https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f5b0d6b088a3b1ff5b74dc33ef044c81c478aeae/types/react/index.d.ts#L4302-L4305 . We should ensure this is backwards compatible with 16/18.

packages/react/src/setupTests.ts
```diff
@@ -2,6 +2,9 @@
 import '@testing-library/jest-dom';
 import 'jest-axe/extend-expect';
 import { configureAxe } from 'jest-axe';
+import { TextEncoder } from 'util';
+
+global.TextEncoder = TextEncoder;
```
scurker, line 7: Do you know why this was needed?

packages/react/package.json
```diff
@@ -23,14 +23,14 @@
...
     "classnames": "^2.2.6",
-    "focus-trap-react": "^10.2.3",
+    "focus-trap-react": "^10.3.0",
     "focusable": "^2.3.0",
     "keyname": "^0.1.0",
     "react-id-generator": "^3.0.1",
-    "react-popper": "^2.2.4",
-    "react-syntax-highlighter": "^15.5.0",
+    "react-popper": "^2.3.0",
+    "react-syntax-highlighter": "^15.6.1",
```
scurker, line 33: There's quite a few dependency updates here, were they all necessary to support 18?

scurker (commented):
Nice work! I don't see anything major blocking code wise, just a few questions.

I'll take a deep dive on Monday to see if this isn't something we can't get in next week.

packages/react/package.json
anastasialanz: I don't believe they were all necessary but I did update a few to see if it would fix a test but realized they weren't the  problem.

docs/mdx-components.tsx
anastasialanz: Looks like they added support for react 16 and 17.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L3136-L3139
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v17/index.d.ts#L3136-L3139

packages/react/package.json
scurker: Generally it's a good practice to try to avoid dependency and feature changes in the same PR since it might be hard to track down bugs later, unless they're necessary to address specific issues with a PR. My preference would be to revert and implement separately if we can.

scurker (commented):
Code wise, everything LGTM. Let's resolve these last few comments and go from there.

packages/react/package.json
anastasialanz: Updated here https://github.com/dequelabs/cauldron/pull/1731/commits/afd4f6ad92477b275b1628f375f4b5bf9ed61049

packages/react/src/setupTests.ts
anastasialanz: I can fix it here!

anastasialanz: Updated here https://github.com/dequelabs/cauldron/pull/1731/commits/9521cbc351cb3c60bd1ff1037078c0b76be7a129

scurker (approved):

PR #1748, opened 2024-11-06
feat(react,styles): add grid layout for table component

### LINKED ISSUE #1744: Table Layout and Alignment
One of the challenging aspects of Cauldron is it is difficult for Cauldron to make content decisions about tables because as a design system it is unaware of how components are displayed and the contents data within a table. However, Cauldron can provide various utilities and properties to consolidate on common alignments and sizing within tables to reduce the amount of one-off styles that are needed to implement a matching design.

# Layout Properties

Since the contents of a table aren't known up front, it's necessary for the content author to provide the layout for a table and set alignment for each column of data.

To account for layouts of each column, the table should accept a new optional `columns` property that accepts an array of columns along with a `layout="grid"` property?:

```ts
interface Table {
  ...
  layout: 'grid'
  columns?: Array<Column>
}
```

## Column Layouts

To provide for more consistent table layouts and alignments, a table should be able to specify how each column should be displayed:

```ts
interface Column {
  align: 'start' | 'center' | 'end'
  width?: 'auto' | 'min-content' | 'max-content' | `${number}` | `${number}%` | `${number}fr`
}
```

Each matching cell or header within the table should inherit the column alignment from the respective index of the column definition.

If a cell has a `colspan` property with a value greater than 1, the display value of that cell will need to be set to `grid-column: span [number]` and use the leading column index alignment value.

Tables normally have a `display: table` to as the display value of the table. To accommodate the above styles, the display value of the change will need to be changed to `display: grid` when `layout` is set to `grid` to allow for the layout to be customized. The column definitions can be passed down via a `TableProvider` to sub-components to allow for each cell or column to access it's associated layout.

### Sizing

Each of the column sizing properties has the following behaviors:

- `auto`: Column will fill the amount of available space respective to other columns
- `min-content`: Column will shrink down to the size of its smallest column 
- `max-content`: Column will grow up to the size of its largest column
- `${number}`: Applies a fixed pixel width to the column's cells
- `${number}%`: Applies a fixed percentage width to the column's cells
- `${number}fr`: Applies a fractional unit width to the column's cells

## Cell Layouts

An individual cell can override the default column alignment:

```ts
interface TableCell {
  ...
  align?: 'start' | 'center' | 'end'
}
```

If a header cell a `colspan` property with a value greater than 1, the display value of that cell will need to be set to `grid-column: span [number]`.

If a cell has a `rowspan` property with a value greater than 1, the display value of that cell will need to be set to `grid-row: span [number]`.

scurker:
closes #1744

### DIFF
--- a/packages/react/src/components/Table/Table.tsx
+++ b/packages/react/src/components/Table/Table.tsx
@@ -1,22 +1,98 @@
-import React, { TableHTMLAttributes } from 'react';
+import React, { useState, useMemo } from 'react';
 import classNames from 'classnames';
+import { TableProvider } from './TableContext';
 
-interface TableProps extends TableHTMLAttributes<HTMLTableElement> {
+export type Column = {
+  align: ColumnAlignment;
+  width?: ColumnWidth;
+  minWidth?: `${number}` | `${number}%`;
+  maxWidth?: `${number}` | `${number}%`;
+};
+export type ColumnAlignment = 'start' | 'center' | 'end';
+export type ColumnWidth =
+  | 'auto'
+  | 'min-content'
+  | 'max-content'
+  | `${number}`
+  | `${number}%`
+  | `${number}fr`;
+export type RowAlignment = 'start' | 'center';
+
+interface TableProps extends React.TableHTMLAttributes<HTMLTableElement> {
+  layout: never;
+  columns: never;
   variant?: 'border';
 }
 
-const Table = ({ children, className, variant, ...other }: TableProps) => (
-  <table
-    className={classNames(
-      'Table',
-      variant === 'border' && 'Table--border',
-      className
-    )}
-    {...other}
-  >
-    {children}
-  </table>
-);
+interface TableGridProps extends Omit<TableProps, 'layout' | 'columns'> {
+  layout: 'grid';
+  columns?: Array<Column> | number;
+}
+
+const Table = ({
+  children,
+  className,
+  variant,
+  layout,
+  columns: columnsProp = [],
+  style,
+  ...other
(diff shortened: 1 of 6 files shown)


packages/react/src/components/Table/Table.tsx
```diff
@@ -1,21 +1,108 @@
...
+      }
+
+      return columns
+        .map(({ width }) => {
+          if (!width) {
+            return 'auto';
+          } else {
+            return width;
+          }
+        })
```
schne324, line 74: Since it appears that width is always a string (if set) we don't need to account for the pesky falsey `0` (number), can't we simplify this? 

```suggestion
        .map(({ width }) => width || 'auto')
```

schne324: (this is a nitpick)

schne324 (approved):
Looks good to me. The only thing that stands out now that I'm familiar with the table props is that we might want to add guidance to our documentation for the `sortAscendingAnnouncement` and `sortDescendingAnnouncement ` props that the defaults don't include the header text and that it would probably be better to provide one that includes the given header's text ("id sorted ascending" is better than "sorted ascending" for example) 🤷

scurker:
> Looks good to me. The only thing that stands out now that I'm familiar with the table props is that we might want to add guidance to our documentation for the `sortAscendingAnnouncement` and `sortDescendingAnnouncement ` props that the defaults don't include the header text and that it would probably be better to provide one that includes the given header's text ("id sorted ascending" is better than "sorted ascending" for example) 🤷

Good callout. I can create a separate ticket to track this.

PR #1764, opened 2024-12-13
fix(react): remove defaultProps from sidebar, use default parameters instead

scurker:
Fixing the following error: 

> console.error
>      Warning: SideBarItem: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
>          at SideBarItem (/home/runner/work/cauldron/cauldron/packages/react/src/components/SideBar/SideBarItem.tsx:12:8)
>          at ul
>          at nav
>          at ClickOutsideListener > (/home/runner/work/cauldron/cauldron/packages/react/src/components/ClickOutsideListener/index.tsx:17:5)
          at SideBar (/home/runner/work/cauldron/cauldron/packages/react/src/components/SideBar/SideBar.tsx:33:5)

### DIFF
--- a/packages/react/src/components/SideBar/SideBarItem.tsx
+++ b/packages/react/src/components/SideBar/SideBarItem.tsx
@@ -9,7 +9,7 @@ export interface SideBarItemProps extends React.HTMLAttributes<HTMLLIElement> {
 
 const SideBarItem: React.ComponentType<
   React.PropsWithChildren<SideBarItemProps>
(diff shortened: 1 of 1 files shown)


stephenmathieson (approved):
We should _always_ be doing this. IMO it's worth updating the contributing/style guide to say not to use React's `defaultProps`.

scurker:
> We should _always_ be doing this. IMO it's worth updating the contributing/style guide to say not to use React's `defaultProps`.

Yup. I think it was a holdover from some old code and doesn't exist anywhere else.

PR #1766, opened 2024-12-13
docs: Do not use `defaultProps`

stephenmathieson:
See https://github.com/dequelabs/cauldron/pull/1764#issuecomment-2542433467


scurker:
I'm not sure this is needed. The instance this was in was an outlier from an older component and it was caught via linting so I think we're okay with not adding this to the list of requirements unless it becomes a problem.

Issue #1783, opened 2025-01-03
Add button badge variant


dequejenn:
![figma designs](https://github.com/user-attachments/assets/88a1b02e-41f8-4706-a091-67cd30e66c72)

Add badge as an additional variant to the [button component](https://cauldron.dequelabs.com/components/Button) 

Figma https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-Library?node-id=810-3773

Similar example (implementing tag as button variant):
https://github.com/dequelabs/cauldron/commit/9a1593aecc8f24bdc5c5f515d9726b42a9c241c5

This would just be implementing the interactive states, not any additional things (such as the close icon).

stephenmathieson:
From Eric:

![Button font sizes](https://github.com/user-attachments/assets/d5bbfadd-b607-41e2-8946-e08275768d7c)

anastasialanz:
Should we also add "Badge" as a related link on the button page?

PR #1795, opened 2025-02-03
feat(button): Add "badge" variant

stephenmathieson:
This patch adds a "badge" variant to our `<Button>` component.

Closes https://github.com/dequelabs/cauldron/issues/1783

### DIFF
--- a/docs/pages/components/Button.mdx
+++ b/docs/pages/components/Button.mdx
@@ -4,7 +4,7 @@ description: An interactive component that performs a programmable action when a
 source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/Button/index.tsx
 ---
 
-import { Icon, Button, Code } from '@deque/cauldron-react'
+import { Icon, Button, Code, BadgeLabel } from '@deque/cauldron-react'
 
 ```js
 import { Button } from '@deque/cauldron-react'
@@ -76,6 +76,30 @@ When you need a semantic button with the visual appearance of a link.
 
 ```jsx example
 <Button variant="tag">Tag</Button>
+```
+
+### Badge
+
+```jsx example
+<Button variant="badge">Badge</Button>
+```
+
+### Badge with Label
+
+```jsx example
+<Button variant="badge">
+  <BadgeLabel>Label:</BadgeLabel>
+  value
+</Button>
+```
+
+### Disabled Badge with Label
+
+```jsx example
+<Button variant="badge" disabled>
+  <BadgeLabel>Label:</BadgeLabel>
+  value
+</Button>
 ```
 
 ### Additional Button States
--- a/packages/styles/button.css
+++ b/packages/styles/button.css
@@ -52,6 +52,20 @@
   position: relative;
 }
 
+.Button--badge {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  padding: 0 var(--space-small) 0 var(--space-smaller);
+  background-color: var(--badge-background-color);
+  color: var(--gray-90);
+  border: 1px solid var(--badge-border-color);
+  border-radius: var(--badge-height);
+  font-size: var(--text-size-body);
+  font-weight: var(--font-weight-normal);
+  min-height: var(--badge-height);
+}
+
 button.Link {
   cursor: pointer;
   font-size: inherit;
@@ -71,7 +85,8 @@ button.Link {
 .Button--tertiary:before,
 .Button--clear:before,
 .Button--error:before,
-.Button--tag:before {
+.Button--tag:before,
+.Button--badge:before {
   content: '';
   position: absolute;
   top: -2px;
@@ -82,7 +97,8 @@ button.Link {
   pointer-events: none;
 }
 
-.Button--tag:before {
+.Button--tag:before,
+.Button--badge:before {
   border-radius: var(--button-height);
 }
 
@@ -99,7 +115,8 @@ button.Link {
   box-shadow: 0 0 0 1px var(--button-outline-color-error);
 }
 
-.Button--tag:not([disabled]):not([aria-disabled='true']):hover:before {
+.Button--tag:not([disabled]):not([aria-disabled='true']):hover:before,
+.Button--badge:not([disabled]):not([aria-disabled='true']):hover:before {
   box-shadow: 0 0 0 1px var(--button-outline-color-secondary);
 }
 
@@ -167,7 +184,8 @@ button.Link {
   background-color: var(--button-background-color-error-active);
 }
 
-.Button--tag:is([disabled], [aria-disabled='true']) {
+.Button--tag:is([disabled], [aria-disabled='true']),
+.Button--badge:is([disabled], [aria-disabled='true']) {
   color: var(--disabled);
 }
 
--- a/packages/react/src/components/Button/index.test.tsx
+++ b/packages/react/src/components/Button/index.test.tsx
@@ -3,6 +3,7 @@ import { render, screen } from '@testing-library/react';
 import Button from '../../../src/components/Button';
 import Icon from '../../../src/components/Icon';
 import axe from '../../axe';
+import { BadgeLabel } from '../Badge';
 
 test('should render primary button', () => {
   render(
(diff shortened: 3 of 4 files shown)


docs/pages/components/Button.mdx
```diff
@@ -78,6 +78,30 @@ When you need a semantic button with the visual appearance of a link.
 <Button variant="tag">Tag</Button>
 ```
 
+### Badge
```
anastasialanz, line 81: Is there a description we have for when to use this pattern?

docs/pages/components/Button.mdx
anastasialanz, line 1: I think we need to update the props section below and add "badge".

docs/pages/components/Button.mdx
```diff
@@ -78,6 +78,30 @@ When you need a semantic button with the visual appearance of a link.
...
+
+```jsx example
+<Button variant="badge">Badge</Button>
```
anastasialanz, line 84: I'm not seeing a hover state when I hover over a badge.

https://github.com/user-attachments/assets/5d04aa75-64bc-4088-bd5f-129296aa5750

docs/pages/components/Button.mdx
stephenmathieson: Not that I'm aware of. Perhaps either @scurker or @ezirlingerDQ have a suggestion as to what we could put here?

docs/pages/components/Button.mdx
stephenmathieson: Interesting, thanks for pointing that out. I'll look into this!

packages/styles/button.css
anastasialanz, line 1: One more thing. I noticed that we also need the active state style.

anastasialanz: Screenshot 2025-02-04 at 8 04 10 AM
https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-Library?node-id=810-3773&p=f&m=dev

packages/styles/button.css
```diff
@@ -48,10 +48,29 @@
...
+  border: 1px solid var(--badge-border-color);
+  border-radius: var(--badge-height);
+  font-size: var(--text-size-body);
+  font-weight: var(--font-weight-normal);
+  min-height: var(--badge-height);
+}
+
+.Button--badge:active {
+  background-color: var(--button-background-color-secondary);
+}
```
anastasialanz, line 72: I think this needs to match the other `:active` button styles since you can click on a disabled button and it will show this active style.

```
.Button--badge:not([disabled]):not([aria-disabled='true']):active {
  background: var(--button-background-color-primary-active);
}
```

stephenmathieson: > We also need the `theme-dark` styles for the active state.

We [override `--button-background-color-secondary` for dark mode](https://github.com/dequelabs/cauldron/blob/button-badge-variant/packages/styles/button.css#L257), so this should support dark mode automatically, right?

anastasialanz: Yes, but you need to target the `.cauldron--theme-dark` class and the active class together here:
https://github.com/dequelabs/cauldron/blob/06f874efeb45d313395a33e278ab4d9314888857/packages/styles/button.css#L306-L313

anastasialanz: Looking more closely, I'm noticing some inconsistencies in the CSS.

The Figma design has the background color as "accent-secondary" `#F2F2F2` in light mode.

Light mode for `--button-background-color-secondary: var(--gray-20);`. `--gray-20` is `#f2f2f2`
Dark mode has this var as `--button-background-color-secondary: var(--accent-medium)` `--accent-medium` is `#283640`

I also noticed that there's a var for `--button-background-color-secondary-active` which is set to `var(--gray-30)`. I think this is the var we should be using but it doesn't match the Figma.

If you were to test this out by switching to dark mode, you'll see that there isn't a color change when triggering the active state. The badge default uses `--badge-background-color` which uses `--accent-medium` `#283640` in dark mode. The active state uses `--button-background-color-secondary` which also uses `--acccent-medium` `#283640`.

anastasialanz: The badge button styles are lighter than the button styles. I think we could add badge-only vars in `button.css` since the current active secondary button var colors are different than what is in the Figma file for the badge. Any thoughts @scurker? I'm also not sure what the dark mode styles would be for this badge. Since Cauldron has a concept of dark mode, how is that being communicated in the Figma files?

https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-Library?node-id=76-5310&m=dev

The active secondary button background color is `#E0E0E0` or `--gray-30`.
https://github.com/dequelabs/cauldron/blob/518bee9a3c5820d04ad6dcf434c70e135939128c/packages/styles/button.css#L10

Screenshot 2025-02-04 at 3 17 56 PM


The active badge button background color is `#F2F2F2` or `--gray-20`.

Screenshot 2025-02-04 at 3 25 59 PM

docs/pages/components/Button.mdx
scurker: We don't have strict guidelines for descriptive documentation right now. Minimally, the pattern needs to have a documented example but not everything always has a description. Since the description for "tag" is also omitted, I'm not 100% sure yet what we would want to put here.

scurker (dismissed):
Code wise looks good, just a couple of small things:

We need to add screenshot testing here for all of the states (similar to what we do for tag) https://github.com/dequelabs/cauldron/blob/518bee9a3c5820d04ad6dcf434c70e135939128c/packages/react/src/components/Button/screenshots.e2e.tsx#L415

We should also include a link to badge on the related components in the markdown documentation.

packages/styles/button.css
scurker: > Should these badge state color vars live in badge.css instead? Any thoughts @scurker?

If the state is specific to the interactive element, I think living in `button.css` would be the most fitting space for it. Otherwise there's not a hard and fast rule for where variables go aside from global vs component level ones.

> We also need the theme-dark styles for the active state.

This is a bit of a trick question 🙃 

The way this was designed, both light and dark modes should be exactly identical for all states. The only thing to check would be to make sure the right pressed state is present in dark mode which might mean the variable would need to be duplicated for cascading purposes - even if it's exactly the same.

anastasialanz: This makes sense to me. The `Badge` component itself does have dark mode colors and the `--badge-background-color` from `badge.css` is being used here. If the badge button was designed to be identical for light and dark then it looks like we need to create new vars for the badge button colors and not use what is for the `badge` component. You mentioned that the focus styles are slightly different for light and dark mode.

We can add new vars for the active badge background color:

```
:root {
  --button-background-color-badge: #FFF;
  --button-text-color-badge: var(--gray-90);
  --button-background-color-badge-active: var(--gray-20);
}
```


```
.Button--badge:not([disabled]):not([aria-disabled='true']):active {
  background-color: var(--button-background-color-badge-active);
}
```

packages/styles/button.css
```diff
@@ -275,7 +298,8 @@ button.Link {
 .cauldron--theme-dark .Button--primary:is([disabled], [aria-disabled='true']),
 .cauldron--theme-dark .Button--secondary:is([disabled], [aria-disabled='true']),
 .cauldron--theme-dark .Button--error:is([disabled], [aria-disabled='true']),
-.cauldron--theme-dark .Button--tag:is([disabled], [aria-disabled='true']) {
+.cauldron--theme-dark .Button--tag:is([disabled], [aria-disabled='true']),
+.cauldron--theme-dark .Button--badge:is([disabled], [aria-disabled='true']) {
```
anastasialanz, line 302: Looks like we can remove this since the badge button should stay white in dark mode.

packages/styles/button.css
scurker: > This makes sense to me. The Badge component itself does have dark mode colors and the --badge-background-color from badge.css is being used here.

🤦 You're right. At one point I thought we and the design team and discussed the badge would be the same for both light/dark modes but it does look like there is a dark mode implementation.

packages/styles/button.css
scurker: Sorry about the earlier comment! My suggestion was based on a misunderstanding and outdated knowledge. So we don't need to remove the styles here.

packages/styles/button.css
anastasialanz: So the [Badge component](https://cauldron.dequelabs.com/components/Badge) has dark mode styles, the [Impact Badge](https://cauldron.dequelabs.com/components/ImpactBadge) stays the same for light and dark mode, and the Badge button component follows similar dark mode styles as the [Badge component](https://cauldron.dequelabs.com/components/Badge)?

scurker: Yup! The only new missing piece would be the "active" state background color, which it appears isn't implemented for `<Button variant="tag">` in dark mode so I'm not 100% sure if it's needed here or not.

stephenmathieson: I _think_ I have updated the styles appropriately. Resolving this thread for now, but please reopen if I missed something.

anastasialanz (approved):

packages/styles/button.css
anastasialanz: Looks good to me.

shawnsharpDQ:
Noticing Active State does not match the screenshot. 
active state from designs

anastasialanz:
@shawnsharpDQ I'm not sure the screenshot is recreating a user hovering over a button and pressing on it to show both the outline and the background color change.

shawnsharpDQ:
@anastasialanz thanks for clarifying

PR #1796, opened 2025-02-06
style: Run Prettier

stephenmathieson:
Diff generated via `npx prettier --write .`


scurker (changes requested):
Can we skip `docs/pages/*` for now and `vpats`? I don't want to muck around with the documentation formatting too much and vpats are automatically generated and should be ignored. Otherwise, looks good.

PR #1799, opened 2025-02-14
chore: add action to close stale prs

scurker:
We have a lot of PRs that tend to become "stale" because people don't follow up on them. I'd like to try and keep us focused on the things that we're actively working on (since Cauldron has limited allocations).


.github/workflows/stale-prs.yaml
```diff
@@ -0,0 +1,21 @@
+name: 'Close stale PRs'
+on:
+  schedule:
+    # run every day at 1:15 AM UTC
+    - cron: '15 1 * * *'
```
anastasialanz, line 5: That's a very specific time.

scurker: The time doesn't matter too much aside from it running daily. If we didn't run daily, it would mean the action might skip days and not being "accurate" regarding when something is marked as stale and closing the pr after the appropriate time.

Github has their examples as running daily: https://github.com/actions/stale?tab=readme-ov-file#usage

anastasialanz: Ok! I asked because I was worried about our workflow usage with a daily running action.

scurker: The workflow usage should be pretty short. As an example a different stale action took about ~6s to run. https://github.com/dequelabs/cauldron/actions/runs/13337119481/job/37254677830

anastasialanz (approved):

Issue #1807, opened 2025-02-20
Update disabled style for the secondary button


anastasialanz:
The background color for the secondary button in the disabled state should be `#F2F2F2`.

**Current:**

<img width="698" alt="Image" src="https://github.com/user-attachments/assets/4af94687-23e4-42d4-a66f-d3f17bb00b57" />

**Designs:**

<img width="804" alt="Image" src="https://github.com/user-attachments/assets/e6f14a9d-9054-493e-9eb1-283ab694840b" />

PR #1808, opened 2025-02-20
fix(styles): secondary disabled button background color

anastasialanz:
Fixes https://github.com/dequelabs/cauldron/issues/1807

The secondary button disabled background color should use `--gray-20`.

## Summary by CodeRabbit

- **Style**
  - Updated the background color for disabled secondary buttons to a lighter shade for improved visual clarity.

- **Tests**
  - Improved the reliability of button component tests by refining how the active state is set during testing.


scurker (changes requested):
Screenshots should be updated (I'm surprised this didn't fail). Do you mind taking a look?

anastasialanz:
> Screenshots should be updated (I'm surprised this didn't fail). Do you mind taking a look?

I haven't fully dug into why they pass with `yarn screenshots` but they do fail using the VS Code Playright extension.

scurker:
Ok weird, this shouldn't be being picked up because it's in draft. I need to take a look at the action.

scurker:
https://github.com/dequelabs/cauldron/pull/1887

scurker:
@anastasialanz I think I might know what the root cause is as I encountered this elsewhere. Screenshot diffs is account for pixel changes but not necessarily color changes under a specific threshold. If you're trying to generate a screenshot locally that is below that threshold, it might not generate the screenshot.

If you try deleting the local screenshot and re-running `yarn screenshots --update-snapshots` does that update the snapshot?

anastasialanz:
> @anastasialanz I think I might know what the root cause is as I encountered this elsewhere. Screenshot diffs is account for pixel changes but not necessarily color changes under a specific threshold. If you're trying to generate a screenshot locally that is below that threshold, it might not generate the screenshot.
> 
> If you try deleting the local screenshot and re-running `yarn screenshots --update-snapshots` does that update the snapshot?

I tried this and new screenshots were generated, but it's still not the right background color.

shawnsharpDQ (approved):

scurker (approved):

PR #1816, opened 2025-02-28
feat(react, styles): combobox component multiselect

chornonoh-vova:
Fixes: #1103 

Adds multi select to the combobox. Mode of the combobox (single or multi select) is controlled by `multiselect` prop. When `multiselect` prop if `false` or `undefined`, `value` and `defaultValue` must be `string`, when the `multiselect` prop is `true` `value` and `defaultValue` must be `string[]`. Additionally, in the multi select mode, `inputValue` and `defaultInputValue` props of type `string` are added to allow control of the text input.


packages/react/src/utils/closestElement.ts
```diff
@@ -0,0 +1,59 @@
...
+    };
+  });
+}
+
+type Dir = 'up' | 'down';
+
+/**
+ * Perform search for the closest element in the array of elements relative to the currently selected element.
+ * Searches for the elements in the direction `dir`.
+ * If the element couldn't be found, returns currenly selected element.
```
anastasialanz, line 30:
```suggestion
 * If the element couldn't be found, returns currently selected element.
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -76,7 +116,39 @@ const ComboboxNoResults = ({
   );
 };
 
-const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
+function nextToFocus(
```
scurker, line 119: I'm... a little unclear what this is for. I'm going to leave some other comments related to this but maybe you could provide some additional clarification. It looks like this is trying to find some closest element based on positioning and adjacency and it's not clear why.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -76,7 +116,39 @@ const ComboboxNoResults = ({
...
+  fallback: HTMLElement,
+  focusedIdx: number,
+  direction: number
+): HTMLElement | null {
+  const elems = elements.concat(fallback);
+  if (direction === -1) {
+    return null;
+  }
+
+  const style = getComputedStyle(elems[focusedIdx]);
```
scurker, line 130: Generally speaking, we want to try to avoid calling things that may result in [layout or reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) unless it's absolutely necessary. It might not be a problem here since it's only being invoked on `keydown` but it's not immediately clear why we need to know anything about layout/positioning.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -100,24 +175,36 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
-    const [selectedValue, setSelectedValue] = useState<string>(value || '');
-    const [formValue, setFormValue] = useState<string | undefined>('');
+    const [inputValue, setInputValue] = useState(() => {
+      const value = defaultValue || propValue;
+      const inputVal = defaultInputValue || propInputValue;
+      return (Array.isArray(value) ? inputVal : value) || '';
+    });
+    const [selectedValues, setSelectedValues] = useState(() => {
+      const value = defaultValue || propValue;
+      if (!value) return [];
```
scurker, line 193: Nit:

```suggestion
      if (!value) { 
        return [];
      }
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -235,16 +335,33 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
+            );
+            if (idx === -1) {
+              newSelectedValues.push(activeValue);
+            }
+            setSelectedValues(newSelectedValues);
+            (onSelectionChange as OnMultiSelectionChange)?.({
+              target: activeDescendant.element,
+              value: newSelectedValues,
+              previousValue: selectedValues
+            });
```
scurker, line 363: Nit: We should avoid abbreviated names when possible. Also providing a descriptive value helps for readability since we know what index it is for.

```suggestion
            const activeValueIndex = selectedValues.indexOf(activeValue);
            const newSelectedValues = selectedValues.filter(
              (v) => v !== activeValue
            );
            if (activeValueIndex === -1) {
              newSelectedValues.push(activeValue);
            }
            setSelectedValues(newSelectedValues);
            (onSelectionChange as OnMultiSelectionChange)?.({
              target: activeDescendant.element,
              value: newSelectedValues,
              previousValue: selectedValues
            });
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -376,26 +565,31 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
     );
 
-    const noMatchingOptions = !!value?.length && !matchingOptions.size && (
+    const noMatchingOptions = !!inputValue && !matchingOptions.size && (
       <NoMatchingOptions />
     );
 
     const comboboxListbox = (
+      // eslint-disable-next-line
+      // @ts-ignore
```
scurker, line 574: Nit: If needed, this is the better way to indicate we're expecting an issue rather than just outright ignoring it.

```suggestion
      // @ts-expect-error
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -434,16 +636,54 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
+          className={classnames({
+            'Combobox__TextFieldWrapper--multiselect': multiselect,
+            'TextFieldWrapper--error': hasError
+          })}
           // We're handling click here to open the listbox when the wrapping element is clicked,
           // there's already keyboard handlers to open the listbox on the input element
           onClick={handleInputClick}
         >
+          {multiselect &&
+            selectedValues.map((value, idx) => {
```
scurker, line 648: Nit: Same suggestion as above regarding index...

```suggestion
            selectedValues.map((value, index) => {
```

scurker (changes requested):
I haven't been able to do a full in-depth review but wanted to outline a couple of quick things I noticed along with some general observations:

When a value is selected, the input value of the field is still present per the following requirement:
> When a selection is made to select an active selection, the combobox inner text field should be reset to blank.

I'm also not able to easily navigate between the pills via arrow keys. It doesn't look like I can navigate (via left/right arrow keys) when there's nothing in the input field and I can only navigate through when I press <kbd>backspace</kbd>. The interaction pattern we're expecting should be somewhat similar to this: https://atlassian.design/components/select/examples#multi-select

This is shaping up pretty good though, thank you for putting time into this! This has been a much wanted component for a while.

packages/react/src/components/Combobox/ComboboxContext.tsx
```diff
@@ -28,49 +28,49 @@ type ComboboxProvider = {
 const ComboboxContext = createContext<ComboboxContext>({
```
scurker, line 28: Would we not also want `multiselect` inside of the context? See https://github.com/dequelabs/cauldron/blob/158b965d91a283df2647c43d5f52ef89da6297d4/packages/react/src/components/Listbox/ListboxContext.tsx#L14

packages/react/src/components/Combobox/Combobox.tsx
scurker: I'm guessing the requirement of "arrow keys" wasn't detailed enough. The intention was for there to be able to navigate via left/right arrow key navigation based on the selection order, not having to do anything with dom positioning if that helps clarify.

packages/styles/combobox.css
```diff
@@ -254,3 +254,24 @@
...
 }
+
+.Combobox__TextFieldWrapper--multiselect {
+  flex-wrap: wrap;
+  padding-right: var(--space-large);
+}
+
+.Combobox__TextFieldWrapper--multiselect input {
+  width: unset;
+}
```
scurker, line 265: I would probably recommend setting `width: auto` instead.

```suggestion
.Combobox__TextFieldWrapper--multiselect input {
  width: auto;
}
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -434,16 +636,54 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
         </label>
         {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
         <TextFieldWrapper
-          className={classnames({ 'TextFieldWrapper--error': hasError })}
+          className={classnames({
+            'Combobox__TextFieldWrapper--multiselect': multiselect,
```
scurker, line 640: The naming convention on this class... is a little odd. It implies that TextFieldWrapper has a bem modifier of multiselect, which really belongs to Combobox.

My recommendation here would be instead to add the modifier to combobox and have a style cascade. Example:

```tsx
<div
  id={id}
  className={classnames('Combobox', className, { 'Combobox--multiselect': multiselect })}
  ref={comboboxRef}
>
  ...
</div>
```

Then in your styles you would just target the selector cascade:

```css
.Combobox--multiselect .TextFieldWrapper {
  ...
}
```

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: Tese calculations were done to support the following interactions with Up/Down errors when focused on the pills

https://github.com/user-attachments/assets/422b311c-8d44-41b5-b185-f52c073894cb

If it's not required, I'll remove it

packages/react/src/components/Combobox/ComboboxContext.tsx
chornonoh-vova: No, it's not needed

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: Updated

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: Updated

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: Updated

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: Updated

packages/styles/combobox.css
chornonoh-vova: Updated

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: Updated

packages/react/src/components/Combobox/Combobox.tsx
scurker: Yup, as we discussed over slack we don't need to account for visual positioning, just left/right order based on insertion order when navigating through the pills.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -413,10 +612,22 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
+          className
+        )}
         ref={comboboxRef}
       >
-        {name && <input type="hidden" name={name} value={formValue} />}
+        {name &&
+          formValues.map((formValue) => (
+            <input
+              type="hidden"
+              key={formValue}
```
scurker, line 626: While not likely, it's possible a form value could be duplicated. We should try to make sure this key is always unique regardless of the form value.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -439,11 +650,46 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
+                <TagButton
+                  label=""
+                  value={value || ''}
+                  icon="close"
+                  {...commonProps}
+                />
+              ) : (
+                <Button disabled={disabled} {...commonProps}>
+                  {value}
+                </Button>
```
scurker, line 685: These 2 components do not have similar styles. You can hold off on changes here until I confirm with @dequelabs/cauldron-design how we want these pills to appear.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -439,11 +650,46 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
+                onKeyDown: handlePillKeyDown
+              };
+
+              return !disabled ? (
+                <TagButton
+                  label=""
+                  value={value || ''}
+                  icon="close"
+                  {...commonProps}
+                />
```
scurker, line 681: One thing that isn't accounted for in the issue description, is what the accessible name for this control should be. We need some way to be able to pass in the accessible name for these pills, so they read something along the lines of "remove {label}" for users of assistive technology.

Otherwise, right now this would just read as "{label}, button" and not really providing a descriptive enough context for AT users.

packages/react/src/utils/closestElement.ts
scurker, line 1: Based on the conversation we had in slack, I don't think this is needed.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -280,10 +401,17 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
 
         setOpen(true);
 
-        if (!open && arrowKeypress && selectedValue && isAutoComplete) {
+        if (
+          inputValue === '' &&
+          (backspaceKeypress || deleteKeypress || event.key === 'ArrowLeft')
+        ) {
+          pillsRef.current[pillsRef.current.length - 1]?.focus();
```
scurker, line 408: If we're using an array ref, we need to guard the ref by either:

- Ensuring we can't get an out of range value
- Only invoking `focus` when there's actually selected items

It looks like it might be possible to go negative here.

scurker (changes requested):
A couple of issues I'm still seeing:

The text is still present in the Combobox once I select a value:



Somehow, I was able to get into a scenario where there were multiples of the same value selected:



When I have a list of selected options, type in an additional option (to search/limit what's available) and select an option, it clears out all of the previous values.

When I have a pill "focused" it's stuck in a focus trap that I can't escape out of unless I navigate via right arrow back to the input field.

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: even if it will be negative, result of 

```
const array = [];
console.log(array[-1]);
```

image

Will be undefined

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: It is not possible to have TagButton without icon, that's why I'we added a conditional here to render button instead

chornonoh-vova: Added a screenshot on how it would look like

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -3,24 +3,38 @@ import React, {
...
+  'Escape',
+  'Home',
+  'End',
+  'Backspace',
+  'Delete'
+];
+
+const ArrowKeys = ['ArrowDown', 'ArrowUp'];
+
+const PillKeys = ['Backspace', 'Delete', 'ArrowLeft', 'ArrowRight'];
```
scurker, line 37: Why duplicate values here? We should include the string values as part of the "event keys" and if you want to have a collection of event keys, just include the destructured variables names as part of that collection.

```tsx
const PillKeys = [Backspace, Delete, ArrowLeft, ArrowRight]
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -3,24 +3,38 @@ import React, {
...
+const [Enter, Escape, Home, End, Backspace, Delete] = [
+  'Enter',
+  'Escape',
+  'Home',
+  'End',
+  'Backspace',
+  'Delete'
+];
+
+const ArrowKeys = ['ArrowDown', 'ArrowUp'];
```
scurker, line 35: How are arrow down/up being used? It's not immediately clear.

packages/react/src/components/Combobox/Combobox.tsx
scurker: IMO this abstraction is pretty unclear still. Why is direction `2` and `3`? I have no idea what those numbers mean out of context. What elements are being passed in? Why do we need a fallback?

Take a look at how directional keys have been implemented in listbox for example: https://github.com/dequelabs/cauldron/blob/158b965d91a283df2647c43d5f52ef89da6297d4/packages/react/src/components/Listbox/Listbox.tsx#L215-L232

I'm not saying you can't abstract this logic out into its own function, but it should be fairly easy to read the code and have an understanding of what it does. In this instance, my recommendation would be ensure we're using naming that is self documenting, for example `left` and `right` as the directions. Secondly, since this function is only used in one location right now I'm not sure having this logic be abstracted is useful. It would be immediately more readable if we knew inline this was logic for pill arrow key functionality.

packages/react/src/components/Combobox/Combobox.tsx
scurker: I still think it would be good to avoid this logic block entirely if we knew there weren't any pills present.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -319,42 +447,115 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
-        >
-      >[0]) => {
-        const stringValue =
-          listboxValue?.toString() || /* istanbul ignore next */ '';
-
-        // istanbul ignore else
-        if (!isControlled) {
-          setValue(stringValue);
-        }
+    const handleRemove = useCallback(
```
scurker, line 456: Nit: Handle remove of what? We could benefit from a more specific name, something like:

```suggestion
    const handleRemovePill = useCallback(
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -319,42 +447,115 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
+        const backspaceKeypress = event.key === Backspace;
+
+        const focusedIndex = pillsRef.current.findIndex(
+          (p) => document.activeElement === p
+        );
+
+        const focusedPill = pillsRef.current[focusedIndex];
+
+        if (deleteKeypress || backspaceKeypress) {
+          if (disabled) return;
```
scurker, line 490: Nit: we generally don't use one line if statements.

```suggestion
          if (disabled) {
            return;
          }
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -319,42 +447,115 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
+          const nextIndex = focusedIndex + 1;
+          if (nextIndex >= pillsRef.current.length) {
+            inputRef.current?.focus();
+          } else {
+            pillsRef.current[nextIndex].focus();
+          }
+          handleRemove(focusedPill, focusedPill.innerText);
```
scurker, line 497: I think I know why you put `innerText` here, but is there some way we could not rely on the inner text of the element and use the value instead?

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -439,11 +652,36 @@ const Combobox = forwardRef<HTMLDivElement, ComboboxProps>(
...
+
+              const handleClick = () =>
+                handleRemove(pillsRef.current[index], value);
+
+              return (
+                <ComboboxPill
+                  ref={refCallback}
+                  key={value}
+                  value={value}
+                  removeValueAriaLabel={removeValueAriaLabel}
```
scurker, line 673: A couple of things with this property. We don't usually call things `xxxAriaLabel`. Secondly, the way this is written is assuming [subject-verb-object](https://en.wikipedia.org/wiki/Subject%E2%80%93verb%E2%80%93object_word_order) word order by concatenating the label with the string. This might make it impossible to translate correctly in some languages.

My recommendation here would be to set an optional property `removeOptionLabel` on `ComboboxOption` that would allow for the user to override the default, e.g. we would concatenate `remove {label}` if this property wasn't set, otherwise we'd use the value from that property. This would allow for things to be translated correctly in product land.

e2e/screenshots/combobox-multiselect-.png
scurker, line 1: I'm trying to make sure we have updated designs in the ticket (since the original designs are from a long time ago). But in the meantime, the height of the input fields do not match the height of the requirements:
screenshot of multiselect combobox with smaller pills

scurker (changes requested):
Please don't be discouraged by the amount of feedback, I think this is starting to shape up really nicely and we're getting very close. Comboboxes are pretty complex and I think you've done a great job at getting a lot of the basics right.

But just a couple of more feedback things:

I'm able to select items when a multiselect combobox is disabled. This isn't the correct behavior.

Additionally, when I type in "pea" and select "peach", "pea" remains in the input field. This value should be cleared out once an optiona is selected.

packages/react/src/components/Combobox/Combobox.tsx
chornonoh-vova: Arrow up/down keys are used in a listbox (to navigate between items)

docs/pages/components/Combobox.mdx
```diff
@@ -181,23 +185,27 @@ To set the initial value for uncontrolled Comboboxes, use `defaultValue`.
...
+  while `onSelectionChange` is only called when a selection changes from the
+  listbox.
 </Note>
 
 ```jsx example
 function ComboboxControlledExample() {
   const [value, setValue] = useState('Peach');
-  const handleChange = ({ target }) => {
-    setValue(target.value);
+  const handleChange = (value) => {
```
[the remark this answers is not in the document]
chornonoh-vova: No, this is a mistake that I accidentally did, thanks for catching it

shawnsharpDQ:
I am still able to interact with the caret icon to open the menu, but not close, when the component is disabled.
Otherwise, everything else is good to go.

shawnsharpDQ (dismissed):
Looks good!

docs/pages/components/Combobox.mdx
```diff
@@ -277,118 +297,254 @@ When autocomplete is set to "automatic" the listbox will provide a filtered list
...
+    <ComboboxOption value="Banana">🍌 Banana</ComboboxOption>
+    <ComboboxOption value="Cucumber">🥒 Cucumber</ComboboxOption>
+    <ComboboxOption value="Orange">🍊 Orange</ComboboxOption>
+    <ComboboxOption value="Peach">🍑 Peach</ComboboxOption>
+    <ComboboxOption value="Pear">🍐 Pear</ComboboxOption>
+  </Combobox>
+</FieldWrap>
+```
+
+### Multiselect disabled
```
scurker, line 319: We should have consistent casing of our headings

```suggestion
### Multiselect Disabled
```

scurker (changes requested):
Last small thing to address, great work here!

scurker (approved):

scurker:
Nice work @chornonoh-vova! We've needed this for a long time and this will help solve some long standing ux issues

PR #1826, opened 2025-03-11
fix(styles)!: remove unused sort button styles

### LINKED ISSUE #1224: Remove deprecated Table css
After two months, we should remove the styles that are commented as deprecated from this PR https://github.com/dequelabs/cauldron/pull/793.

anastasialanz:
Closes #1224 

BREAKING CHANGE: The TableHeader sort ascending and descending button text styles were overriding the sort button text with the same hex value for light or dark mode. "Descending" was spelled wrong in the class name, and the styles weren't being applied as well.

This was originally deprecated in https://github.com/dequelabs/cauldron/pull/793 1+ years ago. 🥇


scurker:
We can leave this in staging for now until we're ready to introduce other breaking changes. I don't think we're in any rush to get this out the door.

anastasialanz:
> We can leave this in staging for now until we're ready to introduce other breaking changes. I don't think we're in any rush to get this out the door.

That sounds great to me.

Issue #1812, opened 2025-02-24
Create Header Components


anastasialanz:
## Description
We need to create two new header components, _PageHeader_ and _SectionHeader_.

## `PageHeader`

Figma: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations--Cauldron--Library?node-id=4428-73&m=dev&focus-id=4428-774

<img width="839" alt="Image" src="https://github.com/user-attachments/assets/0dda49f0-fd6d-419b-b276-2c66d7ea3050" />

### Implementation

- Always uses an `h1`
- When a `string` value is passed into heading, always renders an `h1`
- Optional overline

```jsx
type PageHeaderProps {
  heading: ReactNode;
  overline?: ReactNode;
  description?: ReactNode;
  children?: ReactNode;
}
```

```jsx
<PageHeader
  heading="Page Header"
  description="An optional description"
  overline="An optional overline"
>
  <Button variant="primary">Primary</Button>
  <Button variant="secondary">Secondary</Button>
  <ActionMenu>
     ...
  </ActionMenu>
</PageHeader>
```

## `SectionHeader`

<img width="1234" alt="Image" src="https://github.com/user-attachments/assets/8c2d4bc5-d657-42fc-b3a5-1ae6c1e0bd18" />

### Implementation

- Never uses an `h1`
- Can be an `h2` - `h6`
- When a `string` value is passed into heading, always renders an `h2`

```
type SectionHeaderProps {
  heading: ReactNode;
  description?: ReactNode;
  children?: ReactNode;
}
```

```
<SectionHeader
  heading="Section Heading"
  description="An optional description"
>
  <Button variant="primary">Primary</Button>
  <Button variant="secondary">Secondary</Button>
  <ActionMenu>
     ...
  </ActionMenu>
</SectionHeader>
```

```
<SectionHeader
  heading={<h3>Section Heading</h3>}
  description="An optional description"
>
  <Button variant="primary">Primary</Button>
  <Button variant="secondary">Secondary</Button>
  <ActionMenu>
     ...
  </ActionMenu>
</SectionHeader>
```

The dark mode styles can be found in the Figma.

> [!NOTE]  
> The horizontal rule in the screenshot and Figma are not part of the component.

## Container Queries
- Large: >= 80rem
`@container (min-width: 80rem) {}`
- Medium: 79.9375rem to 48rem
`@container (min-width: 48rem) {}`
- Small: <= 79.9375rem

[...]

PR #1829, opened 2025-03-12
chore: add new component issue template

scurker:
Trying to add a little more standardization to new Cauldron components in the things we want to be considered. Adding this baseline issue template will at least get us a good starting point any time we need to add a new component.


.github/ISSUE_TEMPLATE/new_component.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Variants (optional)
+
+<!-- Does the component require different variants implementations? If so, explain: -->
+
+### Usage Examples
+
+<!-- Provide usage examples of the component and/or sub-components using different props or children: -->
+
+### Referencing Implementations
```
anastasialanz, line 58: I think this could be optional.

anastasialanz (commented):
🏆

.github/ISSUE_TEMPLATE/new_component.md
scurker: I'm not sure it should be. We shouldn't "optionally" link to different implementations, I think that should be a requirement. If there's none that exist, we can call that out.

anastasialanz: In that case, should there be an implementation link referenced in https://github.com/dequelabs/cauldron/issues/1812?

scurker: If there's some that we can find, ideally yeah. Although that issue was created before this template was implemented so...

PR #1832, opened 2025-03-12
feat(styles): provide consistent drop shadow styles across all components

### LINKED ISSUE #1774: Update Drop Shadow Styles
Shadows and fills can be used to show elevation. Our initial approach is to limit the number of shadows and provide the following semantic meaning:

Overlay = Any element that goes over content. 
Raised = Any element that rests directly on the background / base or a sunken fill. 

We should add 2 new global level variables for each of the new drop shadow styles:

```css
--drop-shadow-overlay: 0 8px 8px -4px rgba(0, 0, 0, 0.04), 0 20px 24px -4px rgba(51, 51, 51, 0.08);
--drop-shadow-raised: 0 2px 4px -2px rgba(51, 51, 51, 0.08), 0 4px 8px -2px rgba(0, 0, 0, 0.04);
```

## Overlay

All of the below components should inherit from `--drop-shadow-overlay`:

- [x] Combobox
- [x] OptionsMenu (+ future ActionMenu work)
- [x] Dialogs/Alert/Modals
- [x] LoaderOverlay
- [x] BottomSheet
- [x] Popover
- [x] Tooltip

## Raised

All of the below components should inherit from `--drop-shadow-raised`:

- [x] Accordion
- [x] Panel 
- [x] DescriptionList
- [x] Table
- [x] TwoColumnPanel

scurker:
closes #1774


shawnsharpDQ:
Would the accordion have different plans? Maybe we all missed that one when scoping the ticket. Otherwise looks good

scurker:
> Would the accordion have different plans? Maybe we all missed that one when scoping the ticket. Otherwise looks good

Oh good catch. I missed accordion, I had to go through and add a few others that we missed. Will update.

shawnsharpDQ (approved):

anastasialanz (approved):

PR #1845, opened 2025-03-18
feat(react): allow tooltip to be hidden for text ellipsis

### LINKED ISSUE #1842: TextEllipsis should be able to be shown without a Tooltip
[TextEllipses](https://cauldron.dequelabs.com/components/TextEllipsis) currently mandates that a Tooltip is shown and the ellipsized text becomes keyboard accessible when the text no longer fits within the established container. However there is an exception to [WCAG 1.4.12 Text Spacing](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html) in that there is an additional exemption to appearing on focus or activation:

> [...] where the ellipsis is part of a section of content which includes a link, the truncated text is revealed on the linked page

By default, we still want to have the truncated text appear by default for this component via activation on the page without any other property changes. But given that there are scenarios where we employ a link to content where the text can be revealed we want to allow for the tooltip to be hidden.

## Implementation

```ts
interface TextEllipsis {
  hideTooltip?: boolean
}
```

When this property is set to `true`, no tooltip should be shown and the text should be ellipsized with no additional keyboard access provided.

We should also provide additional documentation guidance as our goal is to provide accessible components _by default_ when possible. Setting `hideTooltip` to `false` will _require_ additional mechanisms leading towards an outside solution to address this specific accessibility problem.

> **Important:** This prop should only be used in specific circumstances that comply with additional WCAG guidelines. Setting `hideTooltip={true}` should **only** be used when the full non-truncated text is available via a linked page. Incorrect usage of this property will result in an accessibility issue

scurker:
closes #1842

### DIFF
--- a/docs/pages/components/TextEllipsis.mdx
+++ b/docs/pages/components/TextEllipsis.mdx
@@ -50,6 +50,21 @@ If your component is a Button, Link, or some other kind of interactive element w
 </TextEllipsis>
 ```
 
+### Ellipsis Only
+
+Ellipsis can be applied to truncated text without showing the full text upon an additional interaction on the page, however this should _only_ be used if you know what you are doing and generally should be avoided.
+
+```jsx example
+<TextEllipsis hideTooltip>
+  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+</TextEllipsis>
+```
+
+<Note variant="warning">
+This prop should only be used in specific circumstances that comply with additional WCAG guidelines. Setting `hideTooltip={true}` should only be used when the full non-truncated text is available via a linked page. Incorrect usage of this property will result in an accessibility issue
+</Note>
+
+
 ## Props
 
 <ComponentProps
@@ -74,6 +89,11 @@ If your component is a Button, Link, or some other kind of interactive element w
       name: 'tooltipProps',
       type: 'object',
       description: 'Props to pass and configure the displayed tooltip.'
+    },
+    {
+      name: 'hideTooltip',
+      type: 'boolean',
+      description: 'Prevent TextEllipsis from showing a tooltip when the text is ellipsized.'
     }
   ]}
 />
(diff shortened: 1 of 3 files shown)


packages/react/src/components/TextEllipsis/index.tsx
```diff
@@ -7,14 +7,24 @@ import type {
   PolymorphicComponent
 } from '../../utils/polymorphicComponent';
 
-interface TextEllipsisProps
+interface TextEllipsisBaseProps
   extends PolymorphicProps<React.HTMLAttributes<HTMLElement>> {
   children: string;
   maxLines?: number;
   refProp?: string;
+  hideTooltip?: boolean;
```
scurker, line 15: I debated what to call this prop. I considered using `fullTextAvailableOnLink`, but that's not really consistent with Cauldron's naming patterns. This is also a pattern that can lead to accessibility issues if not implemented correctly so I'm torn on the naming here.

docs/pages/components/TextEllipsis.mdx
```diff
@@ -50,6 +50,21 @@ If your component is a Button, Link, or some other kind of interactive element w
 </TextEllipsis>
 ```
 
+### Ellipsis Only
+
+Ellipsis can be applied to truncated text without showing the full text upon an additional interaction on the page, however this should _only_ be used if you know what you are doing and generally should be avoided.
```
anastasialanz, line 55:
```suggestion
Ellipsis can be applied to truncated text without showing the full text upon an additional interaction on the page, however, this should _only_ be used if you know what you are doing and generally should be avoided.
```

packages/react/src/components/TextEllipsis/index.tsx
anastasialanz: What if we added another comment here on this prop?

```
/**
   * This prop may cause accessibility issues by hiding important content from users.
   * Only use this prop if you have an alternative way to make the full text accessible.
   */
```

packages/react/src/components/TextEllipsis/index.tsx
```diff
@@ -52,19 +64,22 @@ const TextEllipsis = React.forwardRef(
     }
 
     useEffect(() => {
+      if (hideTooltip) {
```
anastasialanz, line 67: Nice early return!

packages/react/src/components/TextEllipsis/index.tsx
scurker: Ah, that's a good idea!

anastasialanz (approved):

PR #1883, opened 2025-04-17
feat(react,styles): add ActionMenu, ActionList components

### LINKED ISSUE #1716: Create ActionMenu Component
# ActionMenu

## Overview

_ActionMenu_ is a pattern component to contain a list of options displayed inside of a menu overlay.

Related:
- #1717 
- #1718 
- #1740 

### Problem Statement

The existing [OptionsMenu](https://cauldron.dequelabs.com/components/OptionsMenu) has been the existing component we have used to implement like functionality. However, its API is limited in that it does not allow for menu selections, doesn't allow for grouping of menu items, and has a somewhat outdated API that doesn't necessarily adhere to some of the more modern techniques we have been using to create components. Additionally, we want to avoid making significant modifications (design or functionality) that might result in breaking changes.

To allow for all of the different capabilities we want to support, the best path forward is to create a new greenfield component that gives us flexibility to design the api to meet some of our more common practices and eventually deprecate `OptionsMenu` once we reach a level of stability for this component.

### Visual Design

Figma Link: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-(Cauldron)-Library?node-id=4840-1786&p=f&m=dev

Given this is an overlay element, the drop shadow of the overlay should use the `--drop-shadow-overlay` css variable.

### Interface / Props

`ActionMenu` should be composed with [`AnchoredOverlay`](https://github.com/dequelabs/cauldron/issues/1715) and share specific properties with that implementation, excluding `open`, `focusTrapOptions`,  and `target` since those properties will be managed internally via `ActionMenu`:

```ts
type ActionMenuTriggerRenderFunction = (
  {
    onClick: (event: React.MouseEvent<HTMLElement>) => void;
    onKeyDown: (event: React.KeyboardEvent<HTMLElement>) => void;
    aria-expanded: boolean;
  },
  open: boolean
) => React.ReactNode

interface ActionMenu extends Pick<
  AnchoredOverlay<HTMLDivElement>, 'placement'
> {

  /** A render prop to display the trigger */
  trigger?: React.ReactElement | ActionMenuTriggerRenderFunction

  /** A callback function that is called when any action list item is selected. */
  closeOnAction?: boolean // defaults to true

}
```

### Accessibility

The action menu will need to control the expanded state of the button as well as the display of the menu. When the menu is opened from triggering the button keyboard navigation should occur within the menu, then focus is returned back to the triggering element that originally opened the menu.

```tsx
<button id="button-id" aria-expanded={expanded}>Open Menu</button>
<ul role="menu" aria-labelledby="button-id">
  ...
</ul>
```

### Responsiveness

The menu should have a default anchor position of `bottom-start`. If the menu expands beyond the edge of the viewport, it should not adjust its position to fit.

Additionally, the ActionMenu does not have a default max/min height/width but will naturally expand or contract based on the content that is inside of it. If a default width is necessary, `--action-menu-width` can set the `max-width` of the component and `--action-menu-height` can be set the `max-height` of the component. The menu should never horizontally scroll if there's horizontal overflow, but may vertically scroll if there's vertical overflow.

### Interactive States

The button trigger to expand or close the menu needs to have `aria-expanded` set accordingly when the menu is expanded.

### Usage Examples

```tsx
<ActionMenu trigger={<Button variant="secondary">Trigger Menu</Button>}>
  <ActionList>
    <ActionListGroup label="Group Label">
      <ActionListItem>...</ActionListItem>
      <ActionListItem>...</ActionListItem>
      <ActionListItem>...</ActionListItem>
    </ActionListGroup>
  </ActionList>
</ActionMenu>
```

The action menu component requires at least 2 children, the first being an interactive component that can be pressed (such as `Button` or `IconButton`) and an `ActionMenuList` that contains the items to display within the menu.

- [ARIA Authoring Practices Menu Button Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/)
- [GitHub Primer `ActionMenu`](https://primer.style/product/components/action-menu/)
- [Adobe Spectrum `ActionMenu`](https://react-spectrum.adobe.com/react-spectrum/ActionMenu.html)

### LINKED ISSUE #1717: Create ActionList Component
# ActionList

## Overview

_ActionList_ is a primitive component that provides a list of actionable items within a list. It is intended to be a replacement for the existing [OptionsMenuList](https://cauldron.dequelabs.com/components/OptionsMenuList) component as a standalone component.

Related:
- #1716 
- #1718 
- #1740 

### Problem Statement

The existing [OptionsMenu](https://cauldron.dequelabs.com/components/OptionsMenu) has been the existing component we have used to implement like functionality. However, its API is limited in that it does not allow for menu selections, doesn't allow for grouping of menu items, and has a somewhat outdated API that doesn't necessarily adhere to some of the more modern techniques we have been using to create components. Additionally, we want to avoid making significant modifications (design or functionality) that might result in breaking changes.

To allow for all of the different capabilities we want to support, the best path forward is to create a new greenfield component that gives us flexibility to design the api to meet some of our more common practices and eventually deprecate OptionsMenu once we reach a level of stability for this component.

### Visual Design

Figma Link: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-(Cauldron)-Library?node-id=4840-1786&p=f&m=dev

> [!Note]
> The visual design here should just be focused on the list and its behaviors. The overlay and outer container will be handled by #1716.

### Interface / Props

The action list is a container element with no direct associated styles to primarily contain _ActionListItem_, _ActionListLinkItem_, _ActionListSeparator_ or _ActionListGroup_ components.

```tsx
type onActionEvent =     
  React.MouseEvent<HTMLLiElement | HTMLAnchorElement> |
  React.KeyboardEvent<HTMLLiElement | HTMLAnchorElement>

type onActionCallbackFunction = (
  key: string, 
  event: onActionEvent
) => void

interface ActionList extends React.HTMLAttributes<HTMLUListElement> {

  children: Array<ActionListItem | ActionListLinkItem | ActionListGroup>

  /** Limits the amount of selections that can be made within an action list */
  selectionType?: 'single' | 'multiple'

  /** A callback function that is called when an action list item is selected. */
  onAction?: onActionCallbackFunction

}
```

#### Context

The action list state can be handled via context. This context should be passed down to descendants via an `ActionListContext` that includes the current selection type:

```tsx
type ActionListContext = {
  selectionType?: 'single' | 'multiple'
  onAction: onActionEvent
}
```

### Accessibility

We should be able to leverage the existing [`Listbox`](https://cauldron.dequelabs.com/components/Listbox) component to manage keyboard behaviors for <kbd>↑</kbd>/<kbd>↓</kbd> arrow key navigation. The action menu should not support cyclical navigation so if this component is used, navigation should be set to `bound`.

When displayed within an `ActionMenu` this component should have an explicit role of `menu` with an `aria-labelledby` referencing the trigger button:

```tsx
<button id="button-id" aria-expanded={expanded}>Open Menu</button>
<ul role="menu" aria-labelledby="button-id">
  ...
</ul>
```

### Responsiveness

There are no responsive considerations that need to be made here.

### Usage Examples

```tsx
<ActionList>
  <ActionListGroup label="Group Label">
    <ActionListItem>...</ActionListItem>
    <ActionListItem>...</ActionListItem>
    <ActionListItem>...</ActionListItem>
  </ActionListGroup>
</ActionList>
```

- [ARIA Authoring Practices Menu Button Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/)
- [GitHub Primer `ActionMenu`](https://primer.style/product/components/action-menu/)
- [Adobe Spectrum `ActionMenu`](https://react-spectrum.adobe.com/react-spectrum/ActionMenu.html)

### LINKED ISSUE #1718: Create ActionListItem Component
# ActionListItem

## Overview

_ActionListItem_ is a list item to be displayed inside of an `ActionList` or `ActionListGroup` inside of an `ActionMenu`.

Related:
- #1717 
- #1716  
- #1740 

> [!NOTE]
> No specialized documentation page is needed for any of the components referenced here. All component documentation will need to be contained within the documentation for `ActionMenu`.

### Problem Statement

The existing [OptionsMenuItem](https://cauldron.dequelabs.com/components/OptionsMenu) has been the existing component we have used to implement like functionality. However it doesn't support grouping of items or selection states. Additionally, we want to avoid making significant modifications (design or functionality) that might result in breaking changes.

### Visual Design

Figma Link: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-(Cauldron)-Library?node-id=4840-1786&p=f&m=dev

### Interface / Props

There are 3 different types of components that can be displayed as a list item:

- `ActionListItem`: an action item that performs an action (opening a model, performs some function)
- `ActionListLinkItem`: an action item that navigates to a url
- `ActionListSeparator`: a visual separator between action items

```ts
type ActionListItemBaseProps = {

  /** 
   * A unique key to identify the action when triggered, when not provided 
   * will use the child text content as the key. 
   */
  actionKey?: string

  /** Provides an additional description for the list item. */
  description?: React.ReactNode

  /** Indicates if the current action list item is selected. */
  selected?: boolean

  /** When an action list item is disabled, it cannot be selected or activated. */
  disabled?: boolean

  /** A callback function that is called when an action list item is selected. */
  onAction: (event: React.MouseEvent<HTMLLIElement> | React.KeyboardEvent<HTMLLIElement>) => void

}

type ActionListItemProps = ActionListItemBaseProps & React.HTMLAttributes<HTMLLIElement>

type ActionListLinkItemProps = {

  /** URL to link to for this action list item */
  href: string

} & Omit<ActionListItemBaseProps, 'selected'> & React.AnchorHTMLAttributes<HTMLAnchorElement>

type ActionListSeparatorProps = React.HTMLAttributes<HTMLLIElement>
```

### Accessibility

This is an implementation of an [aria menu pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions-active-descendant/) implementing the various roles and states necessary for menu items with selection(s).

The role of the action list item is dependent on how it it rendered within an `ActionList` or `ActionListGroup` component:

- If the selection type of the closest `ActionList` or `ActionListGroup` is not set, use `role="menuitem"`
- If the selection type of the closest `ActionList` or `ActionListGroup` is set to `single`, use `role="menuitemradio"`
- If the selection type of the closest `ActionList` of `ActionListGroup` is set to `multiple`, use `role="menuitemcheckbox"`

Only `ActionListItem` components can be selected. Each `ActionListItem` should have its tab index value set to `-1` and focus is managed virtually.

```tsx
<li role="menuitem" tabIndex={-1}>...</li>
```

Each action list item should always have an id set for aria associations, either via props or using `react-id-generator`:

```ts
const [id] = propId ? [propId] : useId(1, 'action-list-item')
```

`ActionListLinkItem` components should focus the internal `anchor` element instead of the `li` with the following markup:

```jsx
<li role="none">
  <a role="menuitem" tabIndex={-1} href="...">...</a>
</li>
```

When `disabled` is set to true, `aria-disabled` should be set to `true` and any click or keypress events for the disabled item should be prevented.

When `selected` is set to true, `aria-checked` should be set if the role of the item is `menuitemradio` or `menuitemcheckbox`.

Most of the `ActionListItem` behaviors should be able to be implemented as an extension of [ListboxOption](https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/Listbox/ListboxOption.tsx) with overriding roles and styles.

### Responsiveness

List items should never overflow the ActionList or ActionMenu item that it is in. Defensive CSS should be added to ensure that words break if the ActionList or ActionMenu container is constrained.

### Interactive States

This component should be interactive by supporting the following interactive states:

- Hover
- Focus
- Selected
- Disabled

This component should only contain a single tabstop, which focus managed virtually via arrow keys (<kbd>↑</kbd> or <kbd>↓</kbd>) that navigates between navigable items.

### Usage Examples

The `description` property should be displayed immediately below any children content of the `ActionListItem`. This can utilize a flex layout.

#### ActionListItem

```tsx
<ActionListItem>One</ActionListItem>
<ActionListItem selected>Two</ActionListItem>
<ActionListItem disabled>Three</ActionListItem>
```

#### ActionListLinkItem

This component is an extension of the `ActionListItem` component but with an anchor child. The `onAction` event should be captured as an event on the child and not the `li` element.

```tsx
<ActionListItem href="https://acme.biz">Link to thing</ActionListItem>
```

#### ActionListSeparator

This is a display component with a `role` of `separator`. This component should only accept `HTMLLiElement` properties and a ref with no children.

```tsx
<ActionListItem>One</ActionListItem>
<ActionListItem>Two</ActionListItem>
<ActionListSeparator />
<ActionListItem>Three</ActionListItem>
```

- [ARIA Authoring Practices Menu Button Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/)
- [GitHub Primer `ActionMenu`](https://primer.style/product/components/action-menu/)
- [Adobe Spectrum `ActionMenu`](https://react-spectrum.adobe.com/react-spectrum/ActionMenu.html)

### LINKED ISSUE #1740: Create ActionListGroup component
# ActionListGroup

## Overview

_ActionListGroup_ is a primitive component that presents a list of `ActionListItem`s as a semantic group with a label.

Related:
- #1716
- #1717
- #1718 

> [!NOTE]
> No specialized documentation page is needed for any of the components referenced here. All component documentation will need to be contained within the documentation for `ActionMneu`.

### Problem Statement

The existing [OptionsMenu](https://cauldron.dequelabs.com/components/OptionsMenu) has been the existing component we have used to implement similar functionality. However, its API is limited in that it does not directly allow for logical grouping of items or selections.

This grouping control is part of a new ActionMenu design to allow for different logical groups of items within an action menu and/or different selection types.

### Visual Design

Figma Link: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-(Cauldron)-Library?node-id=4840-1786&p=f&m=dev

### Interface / Props

```ts
type ActionListGroupProps = {

  /** Label for the group of action list items. */
  label: React.ReactNode;

  /** Limits the amount of selections that can be made within an action list */
  selectionType?: 'single' | 'multiple'

  children: Array<ActionListItem | ActionListLinkItem | ActionListSeparatorProps>

} & React.HTMLAttributes<HTMLLIElement>
```

### Accessibility

The action list is a container element with no to primarily contain ActionListItem, ActionListLinkItem, ActionListSeparator or ActionListGroup components within a group.

This component should render as an as `li` element with a role of `none` with a `ul` element immediately nested with a role of `group` with the label rendered as `role="presentation"`: 

```tsx
<li role="none">
  <div role="presentation" id={groupLabel}>{label}</div>
  <ul role="group" aria-labelledby={groupLabel}>
    ...
  </ul>
</li>
```

### Responsiveness

Group labels should never overflow the ActionList or ActionMenu item that it is in. Defensive CSS should be added to ensure that words break if the ActionList or ActionMenu container is constrained.

#### Logical Grouping

```tsx
<ActionMenu>
  <ActionMenuList>
    <ActionMenuGroup label="Group Label">
      <ActionMenuItem>...</ActionMenuItem>
      <ActionMenuItem>...</ActionMenuItem>
      <ActionMenuItem>...</ActionMenuItem>
    </ActionMenuGroup>
  </ActionMenuList>
</ActionMenu>
```

#### Selection Groupings

```tsx
<ActionMenu>
  <ActionMenuList>
    <ActionMenuGroup label="Multi-select" selectionType="multiple">
      <ActionMenuItem>...</ActionMenuItem>
      <ActionMenuItem>...</ActionMenuItem>
      <ActionMenuItem>...</ActionMenuItem>
    </ActionMenuGroup>
    <ActionMenuGroup label="Single-select" selectionType="single">
      <ActionMenuItem>...</ActionMenuItem>
      <ActionMenuItem>...</ActionMenuItem>
      <ActionMenuItem>...</ActionMenuItem>
    </ActionMenuGroup>
  </ActionMenuList>
</ActionMenu>
```

- [ARIA Authoring Practices Menu Button Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/)
- [GitHub Primer `ActionMenu`](https://primer.style/product/components/action-menu/)
- [Adobe Spectrum `ActionMenu`](https://react-spectrum.adobe.com/react-spectrum/ActionMenu.html)

scurker:
closes #1716 
closes #1717 
closes #1718 
closes #1740

This is an implementation of an [actions menu pattern using aria-activedescendant](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions-active-descendant/), so one will need to familiarize themselves with all of the requirements there.

Additionally this is closing multiple issues with detailed requirements so be sure to check all of the above linked tickets to understand the code changes here. The size of the PR is large, but important context might be lost if you are not reviewing the code within context of an `ActionMenu`.


packages/react/src/components/ActionList/ActionList.tsx
```diff
@@ -0,0 +1,104 @@
...
+      <Listbox
+        ref={ref}
+        /* Listbox comes with an explicit role of "listbox", but we want to either
+         * use the role from props, or default to the intrinsic role */
+        // eslint-disable-next-line jsx-a11y/aria-role
+        role={undefined}
+        // aria-multiselectable is valid for listbox roles, but not list or menu roles
+        aria-multiselectable={
+          actionListContext.role === 'listbox' ? undefined : null
+        }
```
Zidious, line 79: We aren't setting the value of [aria-multiselectable](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-multiselectable) here. Isn't this the same as omitting the property entirely?

packages/react/src/components/ActionList/ActionListContext.tsx
```diff
@@ -0,0 +1,54 @@
...
+  onAction: () => null
+});
+
+function ActionListProvider({
+  role,
+  selectionType,
+  onAction,
+  children
+}: ActionListProviderProps): React.JSX.Element {
+  const { Provider } = ActionListContext as React.Context<ActionListContext>;
```
Zidious, line 37: Does this need to be cast, the type is defined above `createContext<ActionListContext>`

packages/react/src/components/ActionList/ActionListGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+} from './ActionListContext';
+import { ListboxGroup } from '../Listbox';
+import { ContentNode } from '../../types';
+
+interface ActionListGroupProps extends React.HTMLAttributes<HTMLLIElement> {
+  /** Provides a label for the group of action items */
+  label: ContentNode;
+
+  /** Limits the amount of selections that can be made within an action group */
+  selectionType?: ActionListSelectionType | null;
```
Zidious, line 18:
```suggestion
  /** Limits the amount of selections that can be made within an action group */
  selectionType?: ActionListSelectionType
```

packages/react/src/components/ActionList/ActionListGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+          actionListContext.onAction(key, event);
+        }
+      },
+      [onAction, actionListContext.onAction]
+    );
+
+    const listItemRole: React.AriaRole | undefined = [
+      'menu',
+      'listbox'
+    ].includes(actionListContext.role as string)
```
Zidious, line 47: `role` is nullable - are we sure this is a safe cast?

packages/react/src/components/ActionList/ActionListLinkItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+  ) => {
+    const { role: contextRole } = useActionListContext();
+    const listItemRole = useMemo(() => {
+      if (contextRole === 'menu') {
+        return 'menuitem';
+      }
+
+      // istanbul ignore next
+      if (contextRole && process.env.NODE_ENV !== 'production') {
+        console.warn(
```
Zidious, line 35: If we know this will cause accessibility issues do we want to throw an error instead?

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,174 @@
...
+          setOpen(!open);
+          setFocusStrategy('first');
+        }
+      },
+      [open]
+    );
+
+    const handleTriggerKeyDown = useCallback(
+      (event: React.KeyboardEvent<HTMLButtonElement>) => {
+        // istanbul ignore else
```
Zidious, line 74: nit - Not for this PR but it feels this rule is kinda useless if we ignore it most of the time

packages/react/src/components/ActionList/ActionListSeparator.tsx
```diff
@@ -0,0 +1,30 @@
...
+type ActionListSeparatorProps = React.HTMLAttributes<HTMLLIElement>;
+
+const ActionListSeparator = forwardRef<HTMLLIElement, ActionListSeparatorProps>(
+  ({ className, ...props }, ref) => {
+    const { role: contextRole } = useActionListContext();
+    // list and listbox roles only support listitem or option roles respectively
+    // see https://github.com/w3c/aria/issues/1889
+    const separatorRole = ['list', 'listbox'].includes(contextRole as string)
+      ? 'presentation'
+      : 'separator';
```
Zidious, line 14: To avoid casting - 

```suggestion
    const separatorRole = contextRole && ['list', 'listbox'].includes(contextRole)
      ? 'presentation'
      : 'separator';
```

Zidious (changes requested):
Here's what I have so far -

packages/react/src/components/ActionList/ActionList.tsx
```diff
@@ -0,0 +1,104 @@
...
+  /** A callback function that is called when an action list item is selected. */
+  onAction?: onActionCallbackFunction;
+}
+
+const ActionList = forwardRef<HTMLUListElement, ActionListProps>(
+  ({ selectionType = null, onAction, className, children, ...props }, ref) => {
+    const actionListContext = useActionListContext();
+    const activeElement = useRef<
+      HTMLLIElement | HTMLAnchorElement
+    >() as MutableRefObject<HTMLLIElement | HTMLAnchorElement>;
```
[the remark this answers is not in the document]
scurker: Yes, unfortunately. `useRef` doesn't return a mutable ref type for some reason. Not sure why React has it typed that way.

packages/react/src/components/ActionList/ActionList.tsx
scurker: Yes, that's on purpose. It's to override the value of `aria-multiselectable` here: https://github.com/dequelabs/cauldron/blob/1266d64e7d0ea8dca2af3b00e1a6b20dc18b90f4/packages/react/src/components/Listbox/Listbox.tsx#L287

When spread, `undefined` won't override the default value from listbox, but `null` will override the value since this is not a valid property when in a non-listbox role.

packages/react/src/components/ActionList/ActionListGroup.tsx
scurker: I don't think it matters really? `[...].includes(null)` would still be a valid check.

packages/react/src/components/ActionList/ActionListLinkItem.tsx
scurker: No. It _may_ cause an an accessibility issue but we don't know the surrounding context. We shouldn't prevent the app from rendering because there might or might not be a problem.

packages/react/src/components/ActionMenu/ActionMenu.tsx
scurker: I don't think this is a rule, but specifically for coverage. I don't consider checking the else condition in this case significant for coverage unless you do.

packages/react/src/components/ActionList/ActionListSeparator.tsx
scurker: In this case, I don't think this really saves us a whole lot. `[string].includes(null)` should absolutely be a valid check.  If we're worried about the cast, we could throw a `@ts-ignore` but I think that's less desirable.

packages/react/src/components/ActionList/ActionListGroup.tsx
scurker: In this case we want to allow `null` or `undefined`. The problem is if you pass `undefined` into a nested context, it won't override a parent context but `null` will.

packages/react/src/components/ActionList/ActionList.tsx
```diff
@@ -0,0 +1,104 @@
...
+  type onActionEvent,
+  ActionListProvider
+} from './ActionListContext';
+import { useActionListContext } from './ActionListContext';
+
+interface ActionListProps extends React.HTMLAttributes<HTMLUListElement> {
+  children: React.ReactNode;
+
+  /** Limits the amount of selections that can be made within an action list */
+  selectionType?: ActionListSelectionType | null;
```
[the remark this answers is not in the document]
scurker: See response to other comment, we have to allow `null` otherwise we won't be able to override nested contexts.

packages/react/src/components/ActionList/ActionListContext.tsx
scurker: I don't remember why this was done, but looks like this is safe to remove.

packages/react/src/components/ActionList/ActionList.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const actionListContext = useActionListContext();
+    const activeElement = useRef<
+      HTMLLIElement | HTMLAnchorElement
+    >() as MutableRefObject<HTMLLIElement | HTMLAnchorElement>;
+
+    const handleActiveChange = useCallback((value: ListboxOption) => {
+      activeElement.current = value?.element as
+        | HTMLLIElement
+        | HTMLAnchorElement;
+    }, []);
```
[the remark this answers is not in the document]
scurker: This is because the callback value returns `HTMLElement` and not `HTMLLIElement` or `HTMLAnchorElement`.

packages/react/src/components/ActionList/ActionList.tsx
Zidious: Could we add a comment it's not obvious from the code that that's the reason why?

packages/react/src/components/ActionList/ActionList.tsx
```diff
@@ -0,0 +1,104 @@
+import React, {
```
Zidious, line 1: I noticed Cauldron doesn't have coverage checks setup - not for this PR but we should add that

https://github.com/dequelabs/cauldron/issues/1918

Zidious (changes requested):
LGTM modulo the one comment inline - 

I've give final 👀 to other frontend folks

packages/react/src/components/ActionList/ActionList.tsx
Zidious: Ignore - https://github.com/dequelabs/cauldron/issues/1918#issuecomment-2856816680

Bracciata (dismissed):
This looks good to me. I ran testing for accessibility and it appears to match its goals for keyboard and mouse users.

shawnsharpDQ:
- I am seeing a .25rem spacing between action items in the design.
- I am wonder if the focus state should be on the opened action menu instead of the first item in the action menu? It looks selected when opened. cc @ezirlingerDQ for clarification on that interaction if he had already looked into this.

scurker:
> * I am wonder if the focus state should be on the opened action menu instead of the first item in the action menu? It looks selected when opened. cc @ezirlingerDQ for clarification on that interaction if he had already looked into this.

No, we should not be focusing the menu. This is following the APG example where the expected pattern is to focus on the first item in the menu. 

> When a menu opens, or when a menubar receives focus, keyboard focus is placed on the first item.

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,174 @@
...
+        </ClickOutsideListener>
+        <AnchoredOverlay
+          ref={actionMenuRef}
+          role="presentation"
+          className={classnames('ActionMenu', className)}
+          open={open}
+          onOpenChange={setOpen}
+          target={triggerRef}
+          placement={placement}
+          offset={4}
```
anastasialanz, line 154: I think we want to set this to 8px.

This PR:
Screenshot 2025-05-08 at 9 00 05 PM

Figma:
https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations--Cauldron--Library?node-id=4840-1786&p=f&m=dev
Screenshot 2025-05-08 at 9 01 35 PM

docs/pages/components/ActionMenu.mdx
```diff
@@ -0,0 +1,424 @@
...
+      type: 'boolean',
+      description: 'Controls whether the action menu is open. Only necessary for controlling action menu state.'
+    }
+  ]}
+/>
+
+## Related Components
+
+- [ActionList](./ActionList)
+- [AnchoredOverlay](./AnchoredOverlay)
```
anastasialanz, line 424: Should we add `ActionMenu` to the "Related Components" section of `AnchoredOverlay`? I think we're listing components that also use `AnchoredOverlay` there like `Tooltip` and `Popover`.

docs/pages/components/ActionMenu.mdx
```diff
@@ -0,0 +1,424 @@
...
+### Basic Usage
+
+A simple action menu with a button trigger:
+
+```jsx example
+<ActionMenu
+  trigger={
+    <Button variant="secondary">
+      Actions
+      <Icon type="chevron-down" />
```
anastasialanz, line 44: What do you think about the basic example also showing the chevron changing direction when opened? I'm not seeing that in Figma, but I assume that's what we want if someone were to use it. The chevron changing direction is also in the example here. https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions-active-descendant/

docs/pages/components/ActionMenu.mdx
```diff
@@ -0,0 +1,424 @@
...
+  return (
+    <ActionMenu
+      trigger={
+        <Button variant="secondary">
+          Setup
+          <Icon type="chevron-down" />
+        </Button>
+      }
+    >
+      <ActionList selectionType="multiple">
```
anastasialanz, line 342: Under the "Experiments" header in Figma, it says 
Screenshot 2025-05-08 at 9 25 35 PM


When I was playing around with this example, I realized that the top list was a single select and the bottom list was a multiple select. Is this something that we would add later?

packages/react/src/components/ActionList/ActionListGroup.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  const listItem = screen.getByTestId('listitem');
+  const group = within(listItem).getByRole('group');
+
+  expect(listItem).toHaveRole('none');
+  expect(group).toBeInTheDocument();
+  expect(group).toHaveAccessibleName('Group Name');
+});
+
+test('should trigger onAction when a action list item is clicked within the group', async () => {
```
anastasialanz, line 64:
```suggestion
test('should trigger onAction when an action list item is clicked within the group', async () => {
```

e2e/screenshots/actionmenu-selections.png
anastasialanz, line 1: For the bottom list item, it looks like we need 8px between "Icons selected description" and the trailing icon.

Screenshot 2025-05-08 at 9 39 07 PM

anastasialanz (changes requested):
This is awesome work! I'm leaving these comments for now and will pick this back up tomorrow.

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -0,0 +1,549 @@
...
+  render(<ActionMenu {...defaultProps} />);
+
+  const triggerButton = screen.getByRole('button', { name: 'Trigger' });
+
+  triggerButton.focus();
+  await user.keyboard('{ArrowDown}');
+
+  await waitFor(() => {
+    expect(screen.queryByRole('menu')).toBeVisible();
+  });
```
anastasialanz, line 154: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find

We can use `findByRole` here instead and Kent says the error message will be better.

```suggestion
expect(await screen.findByRole('menu')).toBeVisible();
```

e2e/screenshots/actionmenu-mixed.png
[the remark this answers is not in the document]
scurker: The screenshots do work a little differently with absolute positioning, example below:

https://github.com/dequelabs/cauldron/blob/95d0265aee54f3c26da766c96b1076f968a304f1/packages/react/src/components/ActionMenu/screenshots.e2e.tsx#L16-L22

I haven't found a better way to get playwright to capture the full screenshot for absolutely positioned elements, so the screenshot is approximated. If we could determine how to capture the screenshot of these positioned elements without adding overriding styles, that would be ideal.

docs/pages/components/ActionMenu.mdx
scurker: I'm not exactly sure what you're asking. The experiment was supposed to be an "add later" thing I think for group descriptions since we opted not to handle that right now.

docs/pages/components/ActionMenu.mdx
scurker: I'm not sure, I don't have strong opinions here. Related components is typically components you would want to use together (or components that need referencing). How strongly do you feel about this? I'm probably good either way.

packages/react/src/components/ActionMenu/ActionMenu.tsx
scurker: I need to check with @ezirlingerDQ on this, I think figma is offset wrong because this is assuming that the hover state will be present and 8px is a very large gap between the trigger and the action menu.

e2e/screenshots/actionmenu-selections.png
scurker: This might be an instance where Figma didn't get updated. I talked to @ezirlingerDQ and we tweaked this a couple of weeks ago because the icon scale in Cauldron doesn't match what's in Figma, so the offset is different to account for the different iconography.

packages/styles/action-list.css
```diff
@@ -0,0 +1,195 @@
...
+  --actionlist-item-hover-background-color: var(--gray-20);
+  --actionlist-item-description-color: var(--gray-60);
+  --actionlist-item-icon-color: var(--gray-60);
+  --actionlist-item-disabled-text-color: var(--gray-40);
+}
+
+.cauldron--theme-dark {
+  --actionlist-background-color: var(--accent-medium);
+  --actionlist-text-color: #fff;
+  --actionlist-item-description-color: var(--accent-light);
```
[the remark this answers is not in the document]
scurker: I updated this

packages/react/src/components/ActionMenu/ActionMenu.tsx
scurker: Confirmed this should be 4px. Eric will update soon.

ezirlingerDQ (dismissed):
Design approved

packages/react/src/components/ActionMenu/ActionMenu.tsx
anastasialanz, line 1: Reading the docs under "Keyboard Support", it says for _Down Arrow_ , "If focus is on the last menu item, moves focus to the first menu item."

When I navigate past the last element, it does not go back to the first item.

scurker: I think we discussed this during a design review but it didn't get documented as to why we choose the "bound" behavior and not "cycle" behavior. Part of the rationale for this is we also implement <kbd>home</kbd> / <kbd>end</kbd> behavior to navigate to the first/last items and this matches the behavior of Combobox. 

This is also a one-line change if we learn something different so I don't think this should necessarily prevent us from shipping. We will need keep this in mind for future work so that we don't forget to document decisions like this.

anastasialanz: We chatted about this and the decision was made to set `navigation` to `bound` in `ActionList`.

anastasialanz: GitHub doesn't refresh when there are new comments or else I wouldn't have left that one after yours. 😆

packages/react/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * A unique key to identify the action when triggered, when not provided
+   * will use the child text content as the key.
+   */
+  actionKey?: string;
+
+  /** Displays a leading icon for the action item. */
+  leadingIcon?: IconType;
+
+  /** Displays a trailing icon for the action item.. */
```
anastasialanz, line 26:
```suggestion
  /** Displays a trailing icon for the action item. */
```

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,174 @@
...
+          style={{ display: !open ? 'none' : undefined, ...style }}
+          {...props}
+        >
+          {React.cloneElement(actionMenuList, {
```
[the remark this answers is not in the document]
scurker: In this instance, its unavoidable without a complete api change. We do use this pattern semi-regularly when we wrap an element with another element, so I'm not sure what the alternative is. In this case we do want to stamp the child with specific properties and not be dependent on the author to pass them in.

It's not an uncommon pattern in other design systems ([see primer](https://github.com/primer/react/blob/b148d1a8ef75ddfc48abe4a4e0f53008db696f31/packages/react/src/ActionMenu/ActionMenu.tsx#L116)) when dealing with complex component structures. I think a cautionary warning is probably useful for most cases so I don't think it's a _never_ use, but use cautiously and when deemed appropriate.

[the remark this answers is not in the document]
anastasialanz: We're all out here using "legacy" APIs.

packages/react/src/components/ActionList/ActionListItem.tsx
```diff
@@ -0,0 +1,184 @@
...
+          break;
+        default:
+          listItemRole = 'menuitem';
+          break;
+      }
+    } else if (contextRole === 'listbox') {
+      listItemRole = 'option';
+    }
+
+    // Keep the currently active item on screen when navigate via up/down arrow key navigation
```
anastasialanz, line 117:
```suggestion
    // Keep the currently active item on screen when navigating via up/down arrow key navigation
```

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,174 @@
...
+import type { onActionEvent } from '../ActionList/ActionListContext';
+import type Listbox from '../Listbox';
+import AnchoredOverlay from '../AnchoredOverlay';
+import ClickOutsideListener from '../ClickOutsideListener';
+
+const [ArrowDown, ArrowUp] = ['ArrowDown', 'ArrowUp'];
+
+type ActionMenuTriggerProps = Pick<
+  React.HTMLAttributes<HTMLButtonElement>,
+  'onClick' | 'onKeyUpCapture' | 'onKeyDown' | 'aria-expanded' | 'aria-haspopup'
```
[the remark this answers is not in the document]
scurker: Ah, good catch! I think I originally had capture events early on but this one got missed.

anastasialanz (approved):
I went through the APG docs and checked to see if the aria attributes and keyboard functionality match and it all looks accurate to me. 👏🏻

PR #1888, opened 2025-04-24
docs: add note about screenshot generation

scurker:
Just adding an additional note for future context.


maksym-shynkarenko (approved):

scurker:
> I had to remove the Docker `cauldron-playwright-e2e-screenshot` image and run `yarn screenshots:docker` again for this to work.

Interesting! I was able to get it working without removing it, but maybe we just keep improving this as we learn more.

denysfedorov (approved):

PR #1900, opened 2025-05-01
fix(styles): update complete indicator to display step count for safari

denysfedorov:
Closes: https://github.com/dequelabs/walnut/issues/11181

When we marked a step as "completed" by changing its content to empty (content: ''), Safari would get confused and reset the counter, causing the next step to show "1" again instead of continuing the sequence. This didn't happen in Chrome or Firefox, which handle counters more consistently.
The fix is pretty straightforward - instead of removing the counter completely in completed steps, we just make it invisible

Styling improvements:

* [`packages/styles/stepper.css`](diffhunk://#diff-d5aaa6b4c3d7bc0a682d8154b4e8e98183573173ded9f1fae909ca7b37730bc3L142-R144): Updated the `.Stepper__step--complete-indicator:before` pseudo-element to include a `counter(stepper)`. Added `color: transparent` and `position: absolute` to ensure the counter does not interfere visually while still being accessible to screen readers.


scurker (changes requested):
It looks like this is replicating the counter as well as adding additional styles on top of it.

Can you provide a reproducible test case? I'm not 100% convinced this is a Cauldron issue yet because when I view the documentation in Cauldron via Safari, the step counter does appear to be correct for completed steps: 

image

denysfedorov:
> It looks like this is replicating the counter as well as adding additional styles on top of it.
> 
> Can you provide a reproducible test case? I'm not 100% convinced this is a Cauldron issue yet because when I view the documentation in Cauldron via Safari, the step counter does appear to be correct for completed steps:
> 
> image

I am not sure that I can create fully-isolated component to use it as a test case.
It is not a Cauldron bug, it's how Safari/Chrome works with counters.

https://github.com/mdn/browser-compat-data/issues/23070
https://stackoverflow.com/questions/54822993/css-counter-reset-issue-in-safari
https://github.com/WebKit/WebKit/commit/f8d357e1814f503c9df5fdc766855eb805bf5b67

So, in our case, during initial render everything works fine, but when Safari redraws counter, it resets again, when Chrome keep counter in context

So, basically it's not a bug-fix it's a trick to make sure that counter not reset in Safari. What do you think, how can we handle this? Is it OK to have such fix? @scurker

denysfedorov:
Also found a bug on webKit bugzilla https://bugs.webkit.org/show_bug.cgi?id=260436 where we can see different behaviour for safari and chrome
image

scurker:
@denysfedorov this still does not appear to be a Safari issue at least with not how we've implemented it in Cauldron, see below screenshot from Safari: 

documentation page showing step counter in safari for stepper

Based on what I'm seeing in documentation it appears both counters appear to be working correctly, which is why I'm asking for a reproducible example. It appears we don't have a clear grasp on exactly what's causing the problem and as such I don't know what other side effects may be present as a result of this change. Once we can identify that, I think we can move forward.

denysfedorov:
```css
.Stepper {
  display: flex;
  padding: var(--space-medium);
  align-items: flex-start;
  color: var(--stepper-color);
  background-color: var(--stepper-background-color);
  counter-reset: list-item 1;
  list-style-type: none;
  margin: 0 auto;
}

.Stepper__step-indicator:before {
  content: counter(list-item);
}

.Stepper__step--complete-indicator:before {
  content: '';
}
```
This is second variant that we discussed. Since [<ol> comes with its internal](https://developer.mozilla.org/en-US/docs/Web/CSS/counter-reset#built-in_list-item_counter) we can override it's behaviour to start from 1 (by default it is starting from 0) and then just show it inside `:before`
@scurker

packages/react/src/components/Stepper/index.tsx
```diff
@@ -103,12 +103,11 @@ export const Step = (props: StepProps) => {
...
-  children: React.ReactNode;
-  className?: string;
-}
-
-const Stepper = ({ children, className, ...other }: StepperProps) => (
+const Stepper = ({
+  children,
+  className,
+  ...other
+}: React.HTMLAttributes<HTMLOListElement>) => (
```
scurker, line 110: We should keep the interface for stepper props, we can just add `extends React.HTMLAttributes<HTMLOListElement>` to the props.

scurker (changes requested):

packages/styles/stepper.css
```diff
@@ -97,8 +96,7 @@
 
 .Stepper__step-indicator:before {
   font-size: var(--text-size-small);
-  counter-increment: stepper;
-  content: counter(stepper);
+  content: counter(list-item);
```
scurker, line 99: Nice work and good catch! This appears to address the issue in Safari.

packages/react/src/components/Stepper/index.tsx
denysfedorov: Just pushed changes

scurker (approved):

PR #1905, opened 2025-05-01
feat(style): update panel radius and padding

### LINKED ISSUE #1814: Panel component style updates
## Description

The Panel component needs the following style updates:

### Panel

- Update the border radius to `16px`
- Padding for the entire component is `--space-large`

### PanelContent

Figma: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-(Cauldron)-Library?node-id=4123-1216&m=dev

scurker:
closes #1814


packages/react/src/components/Combobox/screenshots.e2e.tsx
scurker, line 1: Screenshots got updated because "FieldWrap" was using `.Panel` classes. I think in this instance, we shouldn't have screenshots getting changed as a result of a different component changing so also addressing the screenshot tests to only focus on the underlying components.

shawnsharpDQ (approved):

packages/styles/panel.css
anastasialanz, line 1: Will the padding and negative margin on the `PanelHeader` and `PanelContent` be done separately so that the border no longer extends the full panel width?

anastasialanz (approved):
:shipit:

PR #1913, opened 2025-05-05
feat(react,styles): add EmptyState component

### LINKED ISSUE #1444: Add EmptyState component
# EmptyState

## Overview

The `EmptyState` component is a primitive component to wrap contextual information when there is no items or data to show. There may be primary or secondary actions to guide the user to perform some sort of desired action. This should primarily be a container component to allow content authors to pass content in.

### Problem Statement

This pattern can be created with existing components, but we want to be able to standardize on a common pattern for displaying something that should have an empty state with optional primary or secondary call to actions.

### Visual Design

Figma: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations--Cauldron--Library?node-id=4815-19&p=f&m=dev

### Interface / Props

```ts
interface EmptyStateProps extends React.HTMLAttributes<HTMLDivElement>> {

  /** Heading text to be displayed for the empty state */
  heading: React.ReactNode
 
  /** Description for the empty state */
  description: React.ReactNode
 
  /** Slot for primary actions */
  primaryActions?: React.ReactNode

  /** Slot for secondary actions */
  secondaryActions?: React.ReactNode

}
```

When `heading` is passed in as a string, it should default to `h2`. Otherwise, it should just render the component.

### Accessibility

The heading level needs to be customizable, defaulting to using `h2`.

### Responsiveness

No specialized responsive concerns need to be addressed.

### Usage Examples

```jsx
<EmptyState
  heading="Not Found"
  description={<p>Nothing Found Here</p>}
  secondaryAction={<a href="#">Click here to learn more</a>}
/>
```

- https://component.gallery/components/empty-state/

scurker:
closes #1444


packages/react/src/components/EmptyState/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <div ref={ref} className={classnames('EmptyState', className)} {...props}>
+        {heading}
+        <p>{description}</p>
+        {primaryActions && (
+          <div className="EmptyState__primary">{primaryActions}</div>
+        )}
+        {secondaryActions && (
+          <div className="EmptyState__secondary">{secondaryActions}</div>
+        )}
+      </div>
```
denysfedorov, line 44: I am wondering if we should use `PageHeader` inside `EmptyState`?

scurker: In this case it's 2 distinct patterns

Zidious (dismissed):
LGTM - does this need sign off from @shawnsharpDQ and folks?

shawnsharpDQ:
@scurker Description text should be .25rem from the header and should be gray-90.

denysfedorov (approved):

Issue #1920, opened 2025-05-08
Document CSS Variable naming convention


anastasialanz:
Update our contributing docs about creating CSS variables to use the `[componentname]-[property]-[value]-[state]` pattern.

PR #1921, opened 2025-05-09
fix(styles): use consistent variable naming for empty state

scurker:
@anastasialanz pointed out this was different.


anastasialanz (approved):
Related issue https://github.com/dequelabs/cauldron/issues/1920

Issue #1917, opened 2025-05-06
Support ActionsMenu as a replacement for TopBarMenu


scurker:
We need to be able to utilize the ActionMenu within a MenuBar as a replacement for OptionsMenu and have the correct set of aria properties as well as positioning.

This is updating any of the existing components and documentation to ensure that the ActionMenu works correctly.

Some initial analysis:

We need to be able to render the action menu within the menu item for the correct structure:

```html
<ul role="menubar">
  <li role="menuitem">
    Menu Item Name
    <!-- action menu renders/functional here -->
  </li>
</ul>
```

With this implementation, the `menuitem` needs to function as the menu trigger while containing the menu itself.

This will need to ensure:

- Focus still returns to the correct location when the menu closed
- Click outside handlers still work as expected
- Left/right arrow key navigation still works correctly between any other sibling menu items

Part of the patterns we're seeing here is that TopBar has not largely been touched in 4-5 years and does not meld well with some of the more recent component patterns we have started to follow. The work to be done as part of this should only be a stop-gap measure to integrate ActionsMenu and render it with the correct semantics.

dequejenn:
Work Effort: High
Claude estimation: 5

Component-library work in Cauldron: make ActionsMenu render correctly inside a MenuBar menuitem as a replacement for TopBarMenu, with correct ARIA, positioning, focus-return on close, working click-outside handlers, and left/right arrow navigation between sibling menu items, plus updates to existing components and docs. Complex ARIA/focus/keyboard integration against an old, divergent TopBar (and in a shared library outside the team's primary set), so a 5.

PR #1924, opened 2025-05-13
fix(react): prevent onClickOutside from being called incorrectly when using target

### LINKED ISSUE #1925: Bug: ClickOutsideListener invokes event when clicking inside target
### Steps to reproduce

Use the `target` attribute of `ClickOutsideListener`:

```js
<ClickOutsideListener
  onClickOutside={onClickOutside}
  mouseEvent="click"
  target={target}
>
  <div>Inside!</div>
</ClickOutsideListener>
```

When clicking inside of the targeted element, the `onClickOutside` listener is invoked.

### Expected Behavior

If using `target`, we should not be invoking `onClickOutside` since `target` should take priority over any element children of `ClickOutsideListener`. 

### Version

v6.16.0

scurker:
closes #1925 

I discovered this bug while investigating #1917.


denysfedorov (approved):
Good catch!

anastasialanz (approved):

PR #1927, opened 2025-05-13
fix(style): prevent platform tooltip from appearing on ellipsized text in Safari

### LINKED ISSUE #1926: Bug: Safari shows native tooltip for text overflow elements (TextEllipsis)
### Steps to reproduce

When text is ellipsized in Safari, there appears to be a default platform behavior where a native tooltip is shown: https://bugs.webkit.org/show_bug.cgi?id=114304

There's no clear way to disable this, but this does conflict with our tooltip.

### Expected Behavior

Since we're already showing a synthetic version of the tooltip, we don't need the platform tooltip.

It does appear we can "work around" this behavior by adding the following styles to TextEllipsis:

```css
.TextEllipsis:after {
  content: '';
  display: block;
}
```

### Version

v6.16.0

scurker:
closes #1926

### DIFF
--- a/packages/styles/text-ellipsis.css
+++ b/packages/styles/text-ellipsis.css
@@ -5,6 +5,9 @@
   overflow: hidden;
 }
 
+/* Try and prevent Safari from showing a platform
+ * tooltip when visible text is truncated
+ * see: https://github.com/dequelabs/cauldron/issues/1926
 .TextEllipsis:after {
   content: '';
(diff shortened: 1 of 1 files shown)


maksym-shynkarenko (approved):
LGTM. Approved from my side, but it needs to be reviewed by Cauldrom team owner

denysfedorov (approved):

PR #1935, opened 2025-05-14
feat(react): add support to render AnchoredOverlay with a portal

scurker:
ref #1917 

This work is needed to allow authors to set where the action menu is rendered. Refactoring `AnchoredOverlay` allows us to do this consistently across ActionMenu, Tooltip, and Popover. This should not be a breaking change as we're re-working existing components to utilize the shared portal functionality via `AnchoredOverlay`.


docs/pages/components/ActionMenu.mdx
```diff
@@ -416,6 +416,11 @@ function MixedSelectionExample() {
       name: 'open',
       type: 'boolean',
       description: 'Controls whether the action menu is open. Only necessary for controlling action menu state.'
+    },
+    {
+      name: 'portal',
+      type: ['React.Ref', 'HTMLElement'],
+      description: 'The parent element to place the ActionMenu in.'
```
Zidious, line 423: Feel free to ignore - 

Is it worth linking to React documentation on [Portals](https://react.dev/reference/react-dom/createPortal) at all here?

docs/pages/components/AnchoredOverlay.mdx
```diff
@@ -188,6 +188,11 @@ function AnchoredOverlayWithOffsetExample() {
       type: 'object',
       description: 'When `focusTrap` is true, optional arguments to configure the focus trap.'
     },
+    {
+      name: 'portal',
+      type: ['React.Ref', 'HTMLElement'],
+      description: 'The parent element to place the AnchoredOverlay in.'
+    },
```
Zidious, line 195: Does this not need to include the type returned from [`createPortal`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/91176960584b69402f0e7c8745436b5d59caeebd/types/react/v18/index.d.ts#L378-L380) in this case a `ReactElement`?

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -417,6 +418,17 @@ test('should support ref prop', () => {
   expect(ref.current).toEqual(screen.getByTestId('actionmenu'));
 });
 
+test('should support portal prop', async () => {
```
Zidious, line 421: We should add a test here to ensure when the `ActionMenu` is placed in some parent element that getting the parent of `ActionMenu` is the element we expect

Zidious (dismissed):
Just some questions/minor things inline -

docs/pages/components/ActionMenu.mdx
scurker: We don't have an easy way to link in to documentation with the way this is built right now.

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
scurker: This already does:

```ts
const actionMenuInPortal = await findByTestId(portal, 'actionmenu');
```

`findByTestId` passes in `portal` to check that the action menu is present.

packages/react/src/components/AnchoredOverlay/index.tsx
```diff
@@ -128,11 +133,24 @@ const AnchoredOverlay = forwardRef(
...
+    }
+
+    return portal && typeof portal !== 'undefined'
+      ? (createPortal(
+          AnchoredOverlayComponent,
+          (portal && 'current' in portal ? portal.current : portal) ||
+            // eslint-disable-next-line ssr-friendly/no-dom-globals-in-react-fc
+            document?.body
+        ) as React.ReactPortal)
+      : AnchoredOverlayComponent;
```
anastasialanz, line 153: It is easier to read this code if we set a const for the portal element like we do for the dialog component but that may be my opinion.
```
if (portal && typeof portal !== 'undefined') {
      const portalElement = 'current' in portal ? portal.current : portal;
      return createPortal(
        AnchoredOverlayComponent,
         // eslint-disable-next-line ssr-friendly/no-dom-globals-in-react-fc
        portalElement || document?.body
      ) as React.ReactPortal;
    }
    
    return AnchoredOverlayComponent;
```

scurker: This does work slightly differently than dialog since the portal is optional, but yeah we could return early here. Let me make that change.

scurker: Good callout, this actually allowed for us to remove the cast `as React.Portal`.

anastasialanz: I meant that making a `const` like `const portalElement = 'current' in portal ? portal.current : portal;` would be easier to read.

anastasialanz (approved):
Looks good to me!

PR #1940, opened 2025-05-19
ci: automatically add SLA labels to VPAT issues

thuey:
Adding this action to help us identify VPAT issues that are at risk of breaching our SLAs. Policy document can be provided via DM upon request


thuey:
Cauldron already takes care of this: https://github.com/dequelabs/cauldron/blob/develop/.github/workflows/vpat-sla.yaml

PR #1946, opened 2025-05-22
feat(react): add onOverflowChange callback to TextEllipsis

### LINKED ISSUE #1945: Add overflow callback to TextEllipsis
The TextEllipsis component currently doesn't provide a way for authors to know when text is actually overflowing and being truncated. This creates issues for advanced accessibility use cases where different behavior is needed based on the overflow state.

For example, you might want to conditionally apply a tooltip only when text is actually truncated, or set different ARIA attributes when content is being cut off. Currently, authors have to implement their own overflow detection logic alongside the component, which duplicates the internal overflow calculations that TextEllipsis is already performing.

We should be able to add an optional `onOverflowChange` callback that gets called when the state of the overflow has changed.

scurker:
closes #1945


Bracciata (approved):
Tested by adding a console log to the mdx 
image

PR #1971, opened 2025-06-10
fix(styles): fix specificity with icon behaviors inside of links

### LINKED ISSUE #1970: Bug: ActionListLinkItem with leadingIcon or trainingIcon doesn't fill width of containing ActionMenu
### Steps to reproduce

```tsx
<ActionMenu
  trigger={
    <Button variant="secondary">
      Actions with Links
      <Icon type="chevron-down" />
    </Button>
  }
>
  <ActionList>
    <ActionListLinkItem href="#first">No Icon</ActionListLinkItem>
    <ActionListLinkItem href="#second" leadingIcon="check">Leading icon</ActionListLinkItem>
    <ActionListLinkItem href="#third" trailingIcon="check">Trailing icon</ActionListLinkItem>
  </ActionList>
</ActionMenu>
```

### Expected Behavior

All of the `<ActionListLinkItem>`s in the above example should fill the width of the containing menu, like the "No icon" case does. Instead,  when hovered/selected, the ones that contain icons only fill up to the edge of the `<a>` content.

The underlying issue is that the `action-list.css` `.ActionListItem` style that is meant to apply `display: grid` to the items in question is being overridden by the more-specific `.Link:has(.Icon)` style in `link.css` that applies `display: inline-flex`, which prevents the `.ActionListItem:has(> .ActionListItem__leadingIcon)` style from having any effect. A possible fix might be something like:

```css
/* Overrides the usual .Link:has(.Icon) styling */
.ActionListItem.Link:has(.Icon) {
  display: grid;
}
```

### Version

v6.18.2

Chrome

### Relevant log output

```shell

```

scurker:
closes #1970

The specificity of the default styles was overriding the layout set within the Action List. Reducing the specificity will allow for the action list to override the display layout where needed.

### DIFF
--- a/packages/styles/link.css
+++ b/packages/styles/link.css
@@ -27,7 +27,7 @@
   outline-offset: 0;
 }
 
-.Link:has(.Icon) {
+.Link:where(:has(.Icon)) {
   display: inline-flex;
   align-items: center;
   gap: var(--space-half);


Bracciata (approved):
Makes sense to me!

anastasialanz (approved):

PR #1976, opened 2025-06-11
fix(styles): apply overflow hidden for DescriptionList to contain overflow items

### LINKED ISSUE #1975: Bug: TextEllipsis is not contained properly within Description Details
### Steps to reproduce

Using `TextEllipsis` inside of a `DescriptionDetails` component, the text is not getting truncated properly (see screenshot below example):

```jsx
<DescriptionList>
  <DescriptionListItem>
    <DescriptionTerm>Description</DescriptionTerm>
    <DescriptionDetails>
      <TextEllipsis>Some really long text that should be ellipsized</TextEllipsis>
    </DescriptionDetails>
  </DescriptionListItem>
</DescriptionList>
```

### Expected Behavior

When using `<TextEllipsis>` inside of a description details, it should be truncated correctly, and not expand outside of its container.

### Version

v6.18.3

Chrome

### Relevant log output

```shell

```

scurker:
closes #1975

### DIFF
--- a/packages/styles/description-list.css
+++ b/packages/styles/description-list.css
@@ -86,6 +86,7 @@
   background-color: var(--data-list-value-background-color);
   font-weight: var(--font-weight-normal);
   word-break: break-word;
+  overflow: hidden;
 }
 
 .DescriptionList__term,


packages/styles/description-list.css
```diff
@@ -86,6 +86,7 @@
   background-color: var(--data-list-value-background-color);
   font-weight: var(--font-weight-normal);
   word-break: break-word;
+  overflow: hidden;
```
scurker, line 89: Because `overflow` wasn't set here, child content just expanded into infinity. Normally we wouldn't want to add overflow, but since we do break word here by default this should be relatively safe to allow for the scenario where `TextEllipsis` is used.

Zidious (approved):

Issue #1977, opened 2025-06-12
Bug: Combobox when navigating items with the keyboard focus moves out of the visible portion of items


chornonoh-vova:
### Steps to reproduce

1. Create a Combobox with a lot of items
2. Navigate with a keyboard between items
3. Observe warnings in the console

![Image](https://github.com/user-attachments/assets/9cc3ca50-4e62-477a-abda-358895291de7)

In ComboboxOption `useIntersectionRef` is called with a ref that is returned from `useSharedRef`. But because it's a ref - it's a stable reference. So, basically, this `useEffect` is only triggered once on the initial render, when the ref might not be available, and never gets called again when the ref is assigned by React. I've tried proving my idea by changing the compiled cauldron file directly in node_modules, and changing an array in the `useEffect` from `[element]`  to `[element.current]`, and it worked.

![Image](https://github.com/user-attachments/assets/b8bbab09-a501-4f73-a9ba-ef9a663971bf)

![Image](https://github.com/user-attachments/assets/3ede21e6-a83d-4c52-8381-410dcb7972d7)

![Image](https://github.com/user-attachments/assets/20ee99c0-6cba-40f8-a1e7-2bb60475ce17)

### Expected Behavior

Focus should correctly move to the next item, and the container should scroll along.

### Version

v6.17.1

### Relevant log output

```shell

```

PR #1986, opened 2025-06-18
fix: update useIntersectionRef to observer refs correctly

chornonoh-vova:
Closes: #1977


packages/react/src/utils/useIntersectionRef.ts
```diff
@@ -12,7 +12,7 @@ import { useRef, useEffect } from 'react';
  * return <span ref={elementRef}>...</span>
  */
 export default function useIntersectionRef<T extends HTMLElement>(
-  element: T | MutableRefObject<T>,
```
scurker, line 15: All of our behavior hooks follow a similar pattern of where they:

- Accept an element
- Or accept a ref

See:
- https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/utils/useFocusTrap.ts
- https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/utils/useEscapeKey.ts
- https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/utils/resolveElement.ts

(Note: there's some additional work we need to do to actually align on the interface, but the underlying principal is there; and unfortunately not well documented right now)

I'm not sure we have a convincing argument yet for removing the element interface to this hook.

packages/react/src/utils/useIntersectionRef.ts
```diff
@@ -30,12 +30,9 @@ export default function useIntersectionRef<T extends HTMLElement>(
         return;
       }
 
-      if (
```
scurker, line 33: I think the missing piece from this function is we might not be resolving the ref correctly prior to passing it into the use effect. We might need to instead resolve the element prior to passing it into the dependency array, that way we're always dealing with either an `HTMLElement` or `null`.

scurker (changes requested):
See comments

packages/react/src/utils/useIntersectionRef.ts
chornonoh-vova: Updated it to accept element or ref

packages/react/src/utils/useIntersectionRef.ts
chornonoh-vova: Added element resolving

scurker (commented):
I apologize for circling back around to this so much later, but is there any way we can have a test that validates we're fixing the issue?

chornonoh-vova:
@scurker Tested it in the product where I'm using it. It's using React 17.

Issue #1656, opened 2024-09-03
Select Description Text


scurker:
## Overview

`Select` should accept a new optional `description` property that is displayed immediately below the label text. When `description` is present, the associated input field should have `aria-describedby` with an idref pointing to the associated description.

## Related Components

- https://github.com/dequelabs/cauldron/issues/1641
- https://github.com/dequelabs/cauldron/issues/1658
- https://github.com/dequelabs/cauldron/issues/1657

### Visual Design

Figma: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-(Cauldron)-Library?node-id=50-1992&t=UIec9KMPGKQia7fz-4

![](https://github.com/user-attachments/assets/0b4458a4-ac61-4afb-a485-36e11731c571)

When the "label text" overflows with the "required" label, the label should wrap onto a new line without interfering with the placement or sizing of required.

> [!NOTE]
> This should match the description text that is current present on `RadioGroup`

### Interface / Props

```tsx
interface SelectProps {
  // ...other props
  description?: React.ReactNode
}
```

### Accessibility

> [!NOTE]
> When both `description` and `error` are set, the `aria-describedby` property should reference the description idref first, followed by the error idref.

## Additional Changes

With this change, the placement of the error message should also be visually located immediately above the select field with the addition of a `caution` icon preceding the error text.

The _Required_ text should not have any font styles applied and have italics removed.

Amandeque:
Team Drums Estimation: 3 Drum Points

Keerthi-Penukonda:
@DanylAksonov , could you please provide test notes for this ticket.

PR #1987, opened 2025-06-23
feat(react, styles): added optional description property for Select

DanylAksonov:
Closes: #1656

### DIFF
--- a/packages/styles/forms.css
+++ b/packages/styles/forms.css
@@ -18,7 +18,6 @@
   --field-label-text-color: var(--gray-90);
   --field-label-error-text-color: var(--error);
   --field-label-description-text-color: var(--gray-60);
-  --field-description-line-height: 1.5;
   --field-icon-inactive-color: var(--gray-90);
   --field-icon-active-color: rgba(60, 122, 174, 0.25);
   --field-icon-error-active-color: rgba(217, 50, 81, 0.25);
@@ -28,7 +27,6 @@
   --field-icon-focus-color: var(--focus-light);
   --field-error-text-color: var(--error);
   --field-error-border-color: var(--error);
-  --field-error-line-height: 1.5;
   --field-listbox-selected-background-color: var(--gray-50);
   --field-listbox-selected-text-color: #fff;
   --field-font-size: var(--text-size-small);
@@ -216,7 +214,7 @@ textarea.Field--has-error:focus:hover,
   font-size: var(--text-size-small);
   align-items: center;
   margin-bottom: var(--space-smallest);
-  line-height: var(--field-error-line-height);
+  line-height: var(--line-height-body);
 }
 
 .Checkbox__wrap .Error {
@@ -300,7 +298,7 @@ textarea.Field--has-error:focus:hover,
   font-size: var(--text-size-small);
   color: var(--field-label-description-text-color);
   margin-bottom: var(--space-smallest);
-  line-height: var(--field-description-line-height);
+  line-height: var(--line-height-body);
 }
 
 .Field__label + .Checkbox__wrap {


packages/react/src/components/Select/index.tsx
```diff
@@ -112,6 +128,17 @@ const Select = React.forwardRef(
             </span>
           )}
         </label>
+        {description && (
+          <div id={descriptionId} className="Field__description">
+            {description}
+          </div>
+        )}
+        {error && (
+          <div id={errorId} className="Error">
```
scurker: This should be using the existing `Field__error` class that was added for `FieldGroup` https://github.com/dequelabs/cauldron/pull/1873

packages/styles/forms.css
```diff
@@ -335,6 +334,12 @@ textarea.Field--has-error:focus:hover,
   color: var(--field-label-error-text-color);
 }
 
+.Field__select .Error {
```
scurker: See other comment.

e2e/screenshots/dark--select-error-.png
scurker, line 1: Since there's some additional variability to how the component is displayed we would want to have screenshots for the following variations of the component:

- Select with description (+ all states)
- Select with error & description (+ all states)

packages/react/src/components/Select/index.tsx
```diff
@@ -89,8 +93,20 @@ const Select = React.forwardRef(
...
+    }
+
     if (error) {
-      dynamicProps['aria-describedby'] = addIdRef(ariaDescribedby, errorId);
+      const previousAriaDescribedby = dynamicProps['aria-describedby'];
+
+      dynamicProps['aria-describedby'] = addIdRef(
+        previousAriaDescribedby || ariaDescribedby,
+        errorId
+      );
```
scurker: It took me a second to wrap my head around what was happening here. I wonder if it would read better if we did the following:

```tsx
// First, set to the initial value
dynamicProps['aria-describedby'] = ariaDescribedBy

// Then additional additions can be simplified:
if (description) {
  dynamicProps['aria-describedby'] = addIdRef(
    dynamicProps['aria-describedby'],
    descriptionId
  );
} 

if (error) {
  dynamicProps['aria-describedby'] = addIdRef(
    dynamicProps['aria-describedby'],
    errorId
  );
} 
```

Thoughts?

scurker (changes requested):

packages/react/src/components/Select/index.test.tsx
scurker, line 1: This is missing an no axe violations check when its rendered with a description.

packages/react/src/components/Select/index.tsx
DanylAksonov: Good solution, I'll apply it. Thanks.

scurker (dismissed):
Looking good! Just needs design approval.

ezirlingerDQ:
Synced with @scurker. We requested the line height of the description text be 150% and to double check the font-weight.

packages/styles/forms.css
```diff
@@ -27,6 +28,7 @@
   --field-icon-focus-color: var(--focus-light);
   --field-error-text-color: var(--error);
   --field-error-border-color: var(--error);
+  --field-error-line-height: 1.5;
```
scurker: Please read our style guidelines for variables: https://cauldron.dequelabs.com/style-guidelines#variables

We have a system level variable that should be used here: https://github.com/dequelabs/cauldron/blob/e8787b009cd9999f21bb5d2886ccd951f16c5afa/packages/styles/typography.css#L39

In this case, I'm not 100% we need a component level variable to specifically set this so we should just be able to set `line-height: var(--line-height-body);` to get the right value.

scurker (changes requested):
This is close! We just need the right token inheritance.

packages/styles/forms.css
DanylAksonov: I saw that we already have the `--line-height-body` with the required value, and I was hesitant about introducing new variables. It seemed to me that there might be cases when we need changes specifically in the form, so I introduced new variables at the form level for this, but if this is unlikely, then I agree with using the `--line-height-body` variable.

scurker (approved):
LGTM, thanks!

Issue #1641, opened 2024-08-20
Input Description Text


scurker:
## Overview

`TextField` should accept a new optional `description` property that is displayed immediately below the label text. When `description` is present, the associated input field should have `aria-describedby` with an idref pointing to the associated description.

## Related Components

- https://github.com/dequelabs/cauldron/issues/1656
- https://github.com/dequelabs/cauldron/issues/1658
- https://github.com/dequelabs/cauldron/issues/1657

### Visual Design

Figma: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-(Cauldron)-Library?node-id=50-1992&t=UIec9KMPGKQia7fz-4

![textfield design](https://github.com/user-attachments/assets/62405b8c-c667-4ab8-aa8a-e4b93fdaf6c7)
![textfield multiline design](https://github.com/user-attachments/assets/3ce2f484-9511-4fc5-b4c0-64e820cf92ae)

When the "label text" overflows with the "required" label, the label should wrap onto a new line without interfering with the placement or sizing of required.

> [!NOTE]
> This should match the description text that is current present on `RadioGroup`

### Interface / Props

```tsx
interface TextFieldProps {
  // ...other props
  description?: React.ReactNode
}
```

### Accessibility

> [!NOTE]
> When both `description` and `error` are set, the `aria-describedby` property should reference the description idref first, followed by the error idref.

## Additional Changes

With this change, the placement of the error message should also be visually located immediately above the input field with the addition of a `caution` icon preceding the error text.

The _Required_ text should not have any font styles applied and have italics removed.

shawnsharpDQ:
![Description and Error Updates](https://github.com/user-attachments/assets/3d918083-9c2e-468c-a316-ceb8a8c40582)

scurker:
Each component should be tackled separately, with a `FieldGroup` component that accepts a label, description, and error state.

amberfrontend:
I have a fix for this, but I'm not sure I have permission to push up a PR.

scurker:
Hey @amberfrontend 👋 

I believe you should have permissions to create a PR, but it might need to be from a fork. If that doesn't work let me know and I can double check the contributor settings.

scurker:
[...]

PR #1988, opened 2025-06-23
feat(react, styles): added optional description property for TextField

DanylAksonov:
Closes: #1641


e2e/screenshots/dark--textfield-error-.png
scurker, line 1: Since there's some additional variability to how the component is displayed we would want to have screenshots for the following variations of the component:

- TextField with description (+ all states)
- TextField with error & description (+ all states)

packages/react/src/components/TextField/index.tsx
```diff
@@ -80,10 +85,14 @@ export default class TextField extends React.Component<
     // textarea and input props being incompatible
     // we should probably fix this
     const Field: any = multiline ? 'textarea' : 'input';
+    const ariaDescribedbyWithDescription = description
```
scurker: Just a very small nit. The naming of this implies it comes with a description, which isn't always the case since the description is conditional. I think we can just make this more generic.
```suggestion
    const ariaDescribedby = description
```

packages/react/src/components/TextField/index.tsx
```diff
@@ -102,6 +111,17 @@ export default class TextField extends React.Component<
             </span>
           )}
         </label>
+        {description && (
+          <div className="Field__description" id={this.descriptionId}>
+            {description}
+          </div>
+        )}
+        {error && (
+          <div className="Error Field__error" id={this.errorId}>
```
scurker: We should just be able to maintain the `Field__error` class here:

```suggestion
          <div className="Field__error" id={this.errorId}>
```

scurker (dismissed):
A few minor things

packages/react/src/components/TextField/index.tsx
DanylAksonov: The variable with such name already exists, but I share your view, so I renamed the target variable

scurker (commented):
Nice work! This just needs @dequelabs/design-team approval.

ezirlingerDQ:
Synced with @scurker. We requested the line height of the description text be 150% and to double check the font-weight.

packages/styles/forms.css
```diff
@@ -18,6 +18,7 @@
   --field-label-text-color: var(--gray-90);
   --field-label-error-text-color: var(--error);
   --field-label-description-text-color: var(--gray-60);
+  --field-label-description-line-height: 1.5;
```
scurker: Please read our style guidelines for variables: https://cauldron.dequelabs.com/style-guidelines#variables

We have a system level variable that should be used here: https://github.com/dequelabs/cauldron/blob/e8787b009cd9999f21bb5d2886ccd951f16c5afa/packages/styles/typography.css#L39

In this case, I'm not 100% we need a component level variable to specifically set this so we should just be able to set `line-height: var(--line-height-body);` to get the right value.

scurker (changes requested):
This is close! We just need the right token inheritence.

scurker (approved):

Issue #1934, opened 2025-05-14
MenuBar Refactor


scurker:
# MenuBar

## Overview

`MenuBar` at present does not follow correct aria patterns, nor does it adhere to modern component guidelines we've documented in Cauldron.

### Problem Statement

There are current implementation gaps with respect to [menubar patterns](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/) including:

- Does not manage `tabIndex` correctly (each menu item has a tab index)
- Does not support menuitem separators (unsure if needed)
- Does not specify how the menu should be labeled

For current Cauldron guidelines, there's number of things that the current pattern doesn't adhere to including:

- Full ref support (supports ref of menu items via `menuItemRef`)
- Composition pattern not cleanly documented
- Class component instead of functional component
- Composition of menu items requires a very exact implementation
- Does not integrate cleanly with new replacement patterns (e.g. ActionMenu vs OptionMenu)

The above list isn't a comprehensive list of all of the issues but the primary high level issues. This component needs a detailed re-evaluation to fit within existing Cauldron component patterns.

### Visual Design

N/A

### Interface / Props

[TBD]

### Accessibility

[TBD]

### Responsiveness

[TBD]

### Interactive States (optional)

[TBD]

### Variants (optional)

[TBD]

### Usage Examples

[TBD]

### Referencing Implementations

- ARIA Authoring Practices Guide: https://www.w3.org/WAI/ARIA/apg/patterns/menubar/
- (need to find additional implementations)

rheisler-deque:
Things to fix found while working on #2368 
- spread the props onto the `ul`, or don't extend `React.HTMLAttributes<HTMLUListElement>`, [Copilot comment about ul props](https://github.com/dequelabs/cauldron/pull/2368#discussion_r3226086879)
- Also refactor the tests, e.g. use `userEvent` instead of `fireEvent`

PR #1995, opened 2025-06-27
[WIP] fix: improve keyboard focus behavior of ActionListLinkItems in TopBar

dbjorge:
Closes: #1993

### DIFF
--- a/docs/pages/components/TopBar.mdx
+++ b/docs/pages/components/TopBar.mdx
@@ -15,7 +15,8 @@ import {
   Icon,
   ActionMenu,
   ActionList,
-  ActionListItem
+  ActionListItem,
+  ActionListLinkItem
 } from '@deque/cauldron-react'
 
 ```jsx
@@ -98,6 +99,53 @@ function TopBarWithMenuExample() {
             <ActionListItem>Menu Item 1</ActionListItem>
             <ActionListItem>Menu Item 2</ActionListItem>
             <ActionListItem>Menu Item 3</ActionListItem>
+          </ActionList>
+        </ActionMenu>
+      </MenuBar>
+    </TopBar>
+  )
+}
+```
+
+### With Menu of Links
+
+This can also be combined with the use of `ActionListLinkItem`s:
+
+```jsx example
+function TopBarWithMenuLinksExample() {
+  const actionMenuRef = useRef()
+  return (
+    <TopBar>
+      <MenuBar>
+        <TopBarItem>
+          <a className="MenuItem__logo" href="#" tabIndex={-1}>
+            <img src="/public/fonts/logo.svg" alt="" />
+            Cauldron
+          </a>
+        </TopBarItem>
+        <TopBarItem>
+          <a href="#" tabIndex={-1}>
+            Item 1
+          </a>
+        </TopBarItem>
+        <ActionMenu
+          tabIndex={-1}
+          trigger={
+            ({ ref, ...props }) => {
+              return (
+                <TopBarItem menuItemRef={ref} tabIndex={0} {...props}>
+                  Menu <Icon type="chevron-down" />
+                  <div ref={actionMenuRef} />
+                </TopBarItem>
+              )
+            }
+          }
+          portal={actionMenuRef}
+        >
+          <ActionList>
+            <ActionListLinkItem href="#target-1">Menu Link 1</ActionListLinkItem>
+            <ActionListLinkItem href="#target-2">Menu Link 2</ActionListLinkItem>
+            <ActionListLinkItem href="#target-3">Menu Link 3</ActionListLinkItem>
           </ActionList>
         </ActionMenu>
       </MenuBar>


packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -31,6 +33,12 @@ type ActionMenuProps = {
   trigger: React.ReactElement | ActionMenuTriggerFunction;
   /** Render the action menu in a different location in the dom. */
   portal?: React.RefObject<HTMLElement> | HTMLElement;
+  /**
+   * If true, the menu will will be passed as a children prop to the trigger function.
+   * If false, the menu will render as a sibling of the trigger.
+   * Only supported if trigger is a function *and* portal is undefined.
+   */
+  renderInTrigger?: boolean;
```
dbjorge, line 43: This prop feels a bit janky to me; @scurker , if you have a better proposal for the name, or if you'd prefer to infer this based on whether the menu is within a top bar somehow, I'd be happy to update accordingly.

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -134,6 +133,69 @@ const ActionMenu = forwardRef<HTMLElement, ActionMenuProps>(
...
+        id: triggerId,
+        onClick: handleTriggerClick,
+        onKeyDown: handleTriggerKeyDown,
+        'aria-expanded': open,
+        'aria-haspopup': 'menu',
+        // This is specifically for the case where the ActionMenu is rendering within
+        // a TopBarItem *and* contains ActionListLinkItems; it prevents the default
+        // TopBarItem auto-click behavior that would otherwise result in menu items
+        // being clicked when the user toggles the menu's trigger. See #1993.
+        autoClickLink: false
```
scurker: I think I understand why this is needed, but I would like to consider us moving this to being part of a documentation change as opposed to a functional one. Does it only work if this property is set here?

A couple of different reasons for this:
- This is cross-dependent on props from another component
- This might generate development errors/noise on depending on how the trigger is passed in since this implementation is only specific to `TopBarItem`

I do see that we are rendering `TopBarItem` inside of the action menu for the top bar; basically is the behavior we are trying to addressed fixed if we just set `autoClickLink: false` inside of the product switcher component?

My pragmatic take is also wanting to try to keep the implementation of the ActionMenu as "clean" as possible and avoiding things that might lead to a breaking change in the near future. Generally speaking, we want to try to lean over composability of components vs applying specific behaviors inside of them. I don't think this is a good pattern to be in the component itself, but if this also works by adding the prop to the `TopBarItem` that gives us a cleaner path forward.

packages/react/src/components/ActionMenu/ActionMenu.tsx
scurker: This seems necessary, so do not take my comments below as saying that we should remove this for the problem we're trying to solve right now.

I do not think this should be a part of our "public" api, and want to make it clear, this should only remain as a short-term fix and work-around until we can address https://github.com/dequelabs/cauldron/issues/1934 or further investigate the focus problem that we are encountering. I don't think it's necessary to add this as a documented prop for the ActionMenu component, but we would want to include this in TopBar documentation as you've already done above.

Can we include in the tsdoc about specifically when this should be used?

scurker (dismissed):

packages/react/src/components/ActionMenu/ActionMenu.tsx
scurker: Yeah I agree (see other comments on this PR). If we can easily infer that its inside of a MenuBar, that would be much cleaner.

scurker:
We will also need to update the implementation here...
https://github.com/dequelabs/cauldron/blob/8f96513cb1e131c5cc9a108ea1781192329a2813/docs/index.js#L161

...with whatever outcome we land on.

packages/react/src/components/ActionMenu/ActionMenu.tsx
dbjorge: You're right that it works if you force the consumer to set it rather than doing it here. I put it here thinking that I wanted to avoid gotchas for the consumer, but I'm ok with leaving it for a doc+consumer thing if you prefer. Updated.

packages/react/src/components/ActionMenu/ActionMenu.tsx
dbjorge: Agreed on intentionally leaving it out of the ActionMenu docs. Updated the tsdoc.

dbjorge: I gave this some thought and experimenting but didn't come up with anything that felt significantly better - TopBar/MenuBar don't provide any sort of context for us to look for already and it felt wrong to add it just as a marker. It felt like the better of bad options to expose the prop; I don't think it's *totally* unreasonable for ActionMenu to allow the consumer control over this decision.

scurker: I think it's fine for an immediate fix, but I would like this to go through our full design/implementation process to figure out the right approach.

docs/index.js
```diff
@@ -160,24 +159,24 @@ const App = () => {
           </TopBarItem>
           <ActionMenu
             tabIndex={-1}
-            trigger={({ ref, ...props }) => {
+            trigger={({ ref, children, ...props }) => {
```
scurker, line 163: Don't we need to add `renderInTrigger` here?

scurker (commented):
Just one question, otherwise I think we're good to move forward.

docs/index.js
dbjorge: Yep , good catch

scurker (approved):

Issue #1996, opened 2025-06-30
Popover component placement works only for auto


maneeshparashar:
## Description

The Popover component has a positioning issue where only the 'auto' placement option works as expected. When using any other placement values (top, bottom, left, right), the popover incorrectly appears in the top-left corner of the viewport instead of the intended position relative to the trigger element.
<img width="720" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNjZMQ2c9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7519acb2e1a1687c315e6831a2f8adc180b6b5c9/image.png" alt="image.png" />

## Steps to Reproduce

Setup Popover component with a trigger element
Set the placement prop to any value other than 'auto' (e.g., 'bottom', 'top', 'right', 'left')
Click the trigger to show the popover
Observe that the popover appears in the top-left corner instead of the intended position

## Expected Behavior

Popover should appear in the specified position relative to its trigger element

## Actual Behavior

Only 'auto' placement works correctly
All other placement options cause the popover to appear at coordinates (0,0) in the top-left corner

PR #1997, opened 2025-06-30
fix: render AnchoredOverlay always to get correct floating

maneeshparashar:
Ref: #1996

### DIFF
--- a/docs/pages/components/Popover.mdx
+++ b/docs/pages/components/Popover.mdx
@@ -5,7 +5,8 @@ source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/co
 ---
 
 import { useRef, useState } from 'react'
-import { Popover, Button } from '@deque/cauldron-react'
+import { Select, Popover, Button } from '@deque/cauldron-react'
+import { placements } from './AnchoredOverlay.mdx'
 
 ```js
 import { Popover, Button } from '@deque/cauldron-react'
@@ -55,12 +56,19 @@ For custom purposes, you can use the "custom" variant, which is the default.
 
 ```jsx example
 function CustomPopoverExample() {
+  const [placement, setPlacement] = useState('top');
   const [show, setShow] = useState(false);
   const onTogglePopover = () => setShow(!show);
   const buttonRef = useRef();
+  const handlePlacementChange = ({ target }) => setPlacement(target.value);
 
   return (
     <>
+      <Select
+        label="Placement"
+        options={placements.map(placement => ({ value: placement }))}
+        onChange={handlePlacementChange}
+      />
       <Button 
         onClick={onTogglePopover} 
         variant="secondary" 
@@ -70,14 +78,14 @@ function CustomPopoverExample() {
       </Button>
       <Popover
         target={buttonRef} 
-        placement="auto" 
+        placement={placement} 
         show={show} 
         onClose={onTogglePopover}
         aria-labelledby={"popover-title"}
       >
         <div>
           <h2 id="popover-title">Popover Title</h2>
-          <p>Popover content</p>
+          <p>Popover with placement {placement}</p>
           <Button>Custom button</Button>
           <Button>Custom button</Button>
           <p>Popover content</p>
--- a/packages/react/src/components/Popover/index.tsx
+++ b/packages/react/src/components/Popover/index.tsx
@@ -1,4 +1,11 @@
-import React, { useState, useEffect, ReactNode, forwardRef, Ref } from 'react';
+import React, {
+  useState,
+  useEffect,
+  ReactNode,
+  forwardRef,
+  Ref,
+  useCallback
+} from 'react';
 import { useId } from 'react-id-generator';
 import { Cauldron } from '../../types';
 import classnames from 'classnames';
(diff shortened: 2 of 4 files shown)


scurker (commented):
Code wise, LGTM. Just one question.

packages/react/src/components/Popover/index.tsx
```diff
@@ -185,12 +195,20 @@ const Popover = forwardRef<HTMLDivElement, PopoverProps>(
 
     useFocusTrap(popoverRef, { disabled: !show, returnFocus: true });
 
-    if (!show || !isBrowser()) {
+    if (!isBrowser()) {
```
scurker: Just curious, if `show` is false is there a reason we to mount the component? What were you trying to address here?

docs/pages/components/Popover.mdx
```diff
@@ -5,7 +5,24 @@ source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/co
 ---
 
 import { useRef, useState } from 'react'
-import { Popover, Button } from '@deque/cauldron-react'
+import { Select, Popover, Button } from '@deque/cauldron-react'
+export const placements = [
```
scurker: Nit: Since these placements are inherited from [Anchored overlay](https://github.com/dequelabs/cauldron/blob/develop/docs/pages/components/AnchoredOverlay.mdx), we might be able to export them from its doc page and pull them in here?

```tsx
import { placements } from './AnchoredOverlay.mdx'
```

Not a huge deal if this doesn't work, but just a suggestion if it does.

packages/react/src/components/Popover/index.tsx
maneeshparashar: The positioning issues were caused by conditionally mounting the popover component. When the popover wasn't in the DOM, useFloating couldn't perform accurate positioning calculations, resulting in incorrect placement and visual jumps. The solution is to keep the popover element always mounted in the DOM so useFloating can consistently access it for proper positioning calculations.

docs/pages/components/Popover.mdx
maneeshparashar: Done

packages/react/src/components/Popover/index.tsx
scurker: This seems like we might have a buggy implementation here in this component that's not fully addressed yet. We have a similar pattern in Tooltip using `AnchoredOverlay` that is able to successfully hide the element and render the correct position when it's conditionally shown:

https://github.com/dequelabs/cauldron/blob/a0e15b51bbc83dc99ad8a9b3ea60f96ae488624b/packages/react/src/components/Tooltip/index.tsx#L165

Having the content of the `AnchoredOverlay` being conditionally shown is confusing because the component itself also accepts a show prop, so we shouldn't need to specify that property for both. I can try to take a look at this today to see if there's something deeper that needs to be addressed but I do think we need to attempt to get at the root of this problem, particularly because we have been able to solve this elsewhere.

scurker: I spent a little bit of time on this myself to see if I could diagnose the underlying issue. I was able to uncover a deeper problem with `AnchoredOverlay`, which I have a fix up for: https://github.com/dequelabs/cauldron/pull/2041

I did test your documentation changes with the placement, and the placement values appear to be correct with no flickery when using the changes from the above PR.

packages/styles/popover.css
```diff
@@ -51,27 +51,66 @@
   background-color: var(--popover-border-color);
 }
 
-/* Adjust position to try to center the arrow in the tooltip's border */
-[class*='Popover--top'] .Popover__popoverArrow {
+/* Adjust position to try to set the arrow in the tooltip's border */
```
scurker: When the element shifts in placement, I think the desired outcome would be to have the placement be relative to the offset value and not have these values hard-coded in css.

A similar change was made recently for tooltip:
https://github.com/dequelabs/cauldron/blob/add098fecfb0563172db4aab79745e7da7885c10/packages/react/src/components/Tooltip/index.tsx#L149-L161

maneeshparashar: @scurker 

The tooltip implementation also include CSS, but it's incomplete.
image


 It's missing CSS definitions for placement variants like top-start, top-end, right-start, etc., which is causing arrow positioning issues. Please see the attached snapshot for reference.
image

scurker: Good catch, we need to bring this back to the design team to determine how the arrows should be positioned using these start/end positions, for both the tooltips and the popover. For now I think we should just addressing the positioning fix of the popovers and not include additional styles. I should hopefully be able to have an answer after Tuesday next week during our next sync.

scurker: It would be helpful however if you could raise a separate issue regrading the arrow position.

maneeshparashar: @scurker Ticket for arrow position issue
https://github.com/dequelabs/cauldron/issues/2047

scurker: For now, we need to omit the arrow changes since we haven't made that determination yet in design how we want to tackle it. This issue should just account for the positioning of the overlay itself.

maneeshparashar: @scurker done

packages/react/src/components/Popover/index.tsx
scurker: Now that you've updated code, please double check that these below changes are now necessary. If the issue no longer exhibits itself and we're able to render conditionally based on show, I don't think the below lines will be needed.

maneeshparashar: Yes, now with your PR merged above lines are not needed.

scurker (approved):

PR #2019, opened 2025-07-10
feat(react): remove icon from ImpactBadge

### LINKED ISSUE #1978: Remove Icons from Impact Badge
## Problem

During some recent usability sessions, there was some confusion around the iconography and its purpose and this has been consistent feedback.

Image

## Solution

Remove the icons. This should not be a breaking change.

scurker:
closes #1978

### DIFF
--- a/packages/react/src/components/ImpactBadge/index.tsx
+++ b/packages/react/src/components/ImpactBadge/index.tsx
@@ -1,9 +1,6 @@
 import React, { forwardRef } from 'react';
-
 import Badge, { BadgeLabel } from '../Badge';
-
 import { ContentNode } from '../../types';
-import Icon, { IconType } from '../Icon';
 import classNames from 'classnames';
 
 type ImpactType = 'critical' | 'serious' | 'moderate' | 'minor';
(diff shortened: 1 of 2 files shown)


packages/react/src/components/ImpactBadge/index.tsx
```diff
@@ -36,7 +26,6 @@ const ImpactBadge = forwardRef<HTMLDivElement, ImpactBadgeProps>(
         ref={ref}
         {...other}
       >
-        <Icon type={iconByType[type]} />
```
Zidious, line 39: With the removal of the Icon in the badge do we need to adjust the CSS to remove the extra padding? 
image

It's not quite centered in the badge container.

Zidious (approved):
LGTM - just as minor comment about CSS feel free to resolve if not an issue.

packages/react/src/components/ImpactBadge/index.tsx
scurker: That's coming from Badge: https://github.com/dequelabs/cauldron/blob/4d3cd93fc5161192e4a65afd007f19d29d0c2afd/packages/styles/badge.css#L25

Removing the icon shouldn't change the inherited padding.

Zidious: Should that centered? Or is that intentional?

scurker: I believe that's intentional, but that's not a change that I've been made aware of from design yet.

Zidious: Understood - thanks for explaining that!

Bracciata (approved):

denysfedorov (approved):
LGTM!

Issue #1991, opened 2025-06-24
Bug: Mis-aligned Headers with Sortable Table Grids


scurker:
### Steps to reproduce

Using the [`layout="grid"`](https://cauldron.dequelabs.com/components/Table#grid-layout) table with sortable headers, the header column is misaligned when not using the default alignment (`start`):

```tsx
<Table 
  layout="grid" 
  columns={[
    { width: 'max-content', align: 'start' }, 
    { width: 'max-content', align: 'start' }, 
    { width: 'auto', maxWidth: '250', align: 'end' }
  ]}>
  <TableHead>
    <TableRow>
      <TableHeader scope="col">First Name</TableHeader>
      <TableHeader scope="col">Last Name</TableHeader>
      <TableHeader scope="col" sortDirection="ascending" onSort={() => {}}>Email</TableHeader>
    </TableRow>
  </TableHead>
  <TableBody>
   ...
  </TableBody>
</Table>
```

![](https://github.com/user-attachments/assets/70b2384d-fb55-48c5-85e9-54bf010e583b)

### Expected Behavior

When using the grid layout, the sort column should have the same alignment as the column itself.

### Version

v6.x

PR #2021, opened 2025-07-15
fix: correct table header alignment for sortable columns

Bracciata:
When the column was sortable, the header would always be left-aligned regardless of the align property. The remaining cells in the column functioned as expected. This fix corrects this bug by applying the alignment properties (using flex) to the button that is added when the column is sortable.

Closes: #1991


packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -69,6 +69,12 @@ const TableHeader = forwardRef<HTMLTableHeaderCellElement, TableHeaderProps>(
             onClick={onSort}
             className="TableHeader__sort-button"
             type="button"
+            style={{
```
scurker: Generally speaking, we want to avoid inline styles.

For the grid layout, it's a little unavoidable due to the dynamic nature of the columns, and those styles are already set here: https://github.com/dequelabs/cauldron/blob/abc96db6f05e266457ccca23a2b0f83fd879f1ee/packages/react/src/components/Table/TableHeader.tsx#L64

Rather than inlining new styles, we should take advantage of CSS inheritance to achieve the same results. We should be able to do this by removing `display: flex` and setting `text-align: inherit` here: https://github.com/dequelabs/cauldron/blob/abc96db6f05e266457ccca23a2b0f83fd879f1ee/packages/styles/table.css#L78

scurker (changes requested):

packages/react/src/components/Table/TableHeader.tsx
Bracciata: Sadly that does not fix it 
image

scurker: Your screenshot is correct. The column is left aligned, so the header should be left aligned. The alignment should match the alignment for that particular column.

Bracciata: You are correct. Please excuse that oversight, however, this still does not work when align end is set on the column

image

scurker: This was the change I made in the above linked `table.css` file. 

```diff
  font-size: var(--text-size-small);
  font-weight: var(--font-weight-medium);
  color: var(--table-header-text-color);
- display: flex;
+ text-align: inherit;
  align-items: center;
  justify-content: flex-start;
  padding: var(--table-space) var(--space-large);
}
```

I made a slight modification to the docs page to display a sortable column within the grid and this is what I'm seeing:

cauldron docs page, table

Are you doing something different?

Bracciata: I had a different property still applied locally. Thank you for patience!

scurker (approved):

PR #2041, opened 2025-07-29
fix(react): use floating-ui ref for anchored overlay to help prevent overlay flicker

scurker:


packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { createRef } from 'react';
 import {
+  act,
```
scurker, line 3: All of the "act" calls was to address a new issue where we are now calling setState, and we weren't properly awaiting for the component to complete rendering.

PR #2042, opened 2025-07-31
feat: add optional description to combobox

Bracciata:
Closes: #1657 

This adds an optional attribute for a description below the Label of a combobox. It also moves the error to above the input and below the description. Finally, it adds a caution icon before the error message.


packages/styles/forms.css
```diff
@@ -239,6 +239,15 @@ textarea.Field--has-error:focus:hover,
   margin-bottom: 0;
 }
 
+.Field.is--flex-column {
```
scurker: Why are these additional styles necessary? We haven't had to implement this for any of the other description changes for the other form fields. If there's something we missed we should include it for all of the form fields.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -640,11 +644,19 @@ const Combobox = forwardRef<
     );
 
     const errorId = `${id}-error`;
+    const descriptionId = `${id}-description`;
     const inputProps = {
       ...props,
-      'aria-describedby': error
-        ? addIdRef(ariaDescribedby, errorId)
-        : ariaDescribedby
+      'aria-describedby': (() => {
```
scurker: Is there a reason this needs to be a self-invoking function and we can't just have the value set via a variable above?

packages/react/src/components/Combobox/Combobox.test.tsx
```diff
@@ -268,6 +268,31 @@ test('should render combobox with error', () => {
...
 });
 
+test('should render combobox with both description and error', () => {
+  const errorId = 'combo-error';
+  const descriptionId = 'combo-description';
+  render(
+    <Combobox
+      label="label"
+      id="combo"
+      aria-describedby="other-id"
```
scurker: I think we need a couple of different test cases:

- The field gets associated with the description when no `aria-describedby` is set
- The field gets associated with the description when `aria-describedby` is set

Both of these things should work in conjunction when there's an error, and we should also test these states via axe-core as part of our [testing strategies](https://github.com/dequelabs/cauldron/blob/develop/CONTRIBUTING.md#testing-strategies).

It's not immediate clear why we need to set `id` here. For testing the description, I think the id is a bit of an implementation detail, we could just use the [`toHaveAccessibleDescription`](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohaveaccessibledescription) matcher to validate that the field has the right description.

scurker (changes requested):
Thanks for taking a look at this! This is definitely a good start, a few things to clean up but nothing big. We'll also need to ensure we have [updated screenshots](https://github.com/dequelabs/cauldron/blob/develop/e2e/readme.md) for this with the description. If you're unsure where to get started with that, let me know and we can pair!

packages/react/src/components/Combobox/Combobox.tsx
Bracciata: Nope. Good call out. I did do it as ifs so it is not a double ternary.

packages/styles/forms.css
Bracciata: It is not needed. it was a miss from a wrapping error earlier.

shawnsharpDQ:
Looking at the screenshots, the error and description do not look to have the correct spacing from the field

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -676,6 +686,24 @@ const Combobox = forwardRef<
             </span>
           )}
         </label>
+        <div>
```
scurker: I don't think you need the extra wrapping div, the other implementations do not have it.

```suggestion
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -676,6 +686,24 @@ const Combobox = forwardRef<
...
+          {description && (
+            <span
+              className={classNames('Field__labelDescription', {
+                'Field__labelDescription--disabled': disabled
+              })}
```
scurker: This should just be able to be `Field__description` based on https://github.com/dequelabs/cauldron/blob/d985ed72b148d58a6031ddb52dd5aed5ecbbc426/packages/react/src/components/TextField/index.tsx#L114-L118 and rendered as a `div` not a `span`.

```suggestion
              className="Field__description"
```

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -676,6 +686,24 @@ const Combobox = forwardRef<
...
+              id={descriptionId}
+            >
+              {description}
+            </span>
+          )}
+          {hasError && (
+            <div className="Error" id={errorId}>
```
scurker: This needs to be updated based on: https://github.com/dequelabs/cauldron/blob/d985ed72b148d58a6031ddb52dd5aed5ecbbc426/packages/react/src/components/TextField/index.tsx#L120-L123

```suggestion
            <div className="Field__error" id={errorId}>
```

packages/react/src/components/Combobox/Combobox.test.tsx
```diff
@@ -1638,3 +1664,223 @@ test('should have no axe violations with no matching results', async () => {
...
+
+test('should render combobox with description', () => {
+  render(
+    <Combobox label="label" description="description">
+      <ComboboxOption>Apple</ComboboxOption>
+      <ComboboxOption>Banana</ComboboxOption>
+      <ComboboxOption>Cantaloupe</ComboboxOption>
+    </Combobox>
+  );
+  expect(screen.getByText('description')).toBeInTheDocument();
```
scurker, line 1676: This just validates that the text is in the document, but not necessarily that it's associated with the field. We should have a test to check for accessible description as well, similar to this test: https://github.com/dequelabs/cauldron/blob/d985ed72b148d58a6031ddb52dd5aed5ecbbc426/packages/react/src/components/TextField/TextField.test.tsx#L66-L73

scurker (changes requested):
Just a few minor things. I think addressing the class name fixes in the component should address the differences that are being shown in the screenshot.

packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -21,6 +21,8 @@ import { addIdRef } from '../../utils/idRefs';
 import TextFieldWrapper from '../internal/TextFieldWrapper';
 import { ListboxValue } from '../Listbox/ListboxOption';
 import ComboboxPill from './ComboboxPill';
+import Icon from '../Icon';
+import classNames from 'classnames';
```
scurker: This appears to be a duplicate import (and not used)

```suggestion
```

scurker (changes requested):
One minor thing. We also need to get a final @dequelabs/design-team signoff and we're good to go.

shawnsharpDQ (approved):

PR #2074, opened 2025-08-26
feat(react,styles): add danger and danger-secondary Button variants

### LINKED ISSUE #1819: Add Button Danger Variant
# Button (danger variants)

## Overview

A new "danger" set of variants is needed for existing `Button` and `IconButton` components in Cauldron. These danger variants would provide clear visual feedback for destructive or potentially harmful actions.

### Problem Statement

Currently, Cauldron uses the "error" variant for destructive actions but "error" does not do a good job at describing the purpose of dangerous actions. We want variant names to more closely align with their intended usage and adding an additional variant of "danger" leads us towards that goal. There are also instances where we want to call out dangerous actions that we want a visual distinction from regular buttons but do not want to draw strong attention to visually.

### Visual Design

The existing "error" variant and new "danger" variant should be visually identical and should share common styles. The additional new variant should be `danger-secondary` to provide a distinct variant that is visually less heavy.

Figma: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations-(Cauldron)-Library?node-id=17-3762&p=f&m=dev

### Interface / Props

The only changes needed for both `Button`/`IconButton` would be to add `danger` and `danger-secondary` as additional variants:

```tsx
type ButtonProps {
  ...existingProps,
  variant?: 
    'primary' |
    'secondary' |
    'tertiary' |
    // @deprecated
    'error' |
    'link' |
    'tag' |
    'badge' |
    'danger' |
    'danger-secondary'
}

type IconButtonProps {
  ...existingProps,
  variant?: 
    'primary' |
    'secondary' |
    'tertiary' |
    // @deprecated
    'error' |
    'link' |
    'tag' |
    'badge' |
    'danger' |
    'danger-secondary'
}
```

### Accessibility

No additional aria changes are needed to implement these new variants.

### Responsiveness

N/A

### Interactive States

Each interactive state for resting, hover, focus, active, and disabled must be implement for each of the new variants.

### Variants 

The new variants should be supported for both `Button` and `IconButton`, including the `thin` modifier.

#### Danger

```tsx
<Button variant="danger">I'm very dangerous!</Button>
```

#### Danger Secondary

```tsx
<Button variant="danger-secondary">I'm somewhat dangerous!</Button>
```

- [Github Primer Button](https://primer.style/product/components/button/#style-variants)
- [Adobe Spectrum Button](https://spectrum.adobe.com/page/button/#Negative-variant)

scurker:
closes #1819


docs/pages/components/Button.mdx
```diff
@@ -36,12 +36,18 @@ Tertiary buttons are an alternate style of button with less visual prominence th
 <Button variant="tertiary">Tertiary</Button>
 ```
 
-### Error
+### Danger
```
scurker, line 39: Note: This is not removing error, but removing it from documentation so that its not used going forward. We don't have a good way to deprecate props, but removing from documentation should help to show it's not being used and we can communicate out to other teams that "Error" is now deprecated.

Tracked by https://github.com/dequelabs/cauldron/issues/2075

Bracciata (approved):
Looks good to me!

PR #2082, opened 2025-09-02
chore: use machine user for sync and release-pr actions

### LINKED ISSUE #1710: Fix Actions Running with Github Bot
In [release](https://github.com/dequelabs/cauldron/pull/1707)/[merge](https://github.com/dequelabs/cauldron/pull/1709) PRs github actions do not run with the github actions bot unless you close/reopen the PR manually. We should fix this so that actions run as expected without having to perform weird work arounds.

Unfortunately we either need to [use a machine account or use a PAT](https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs) as official workarounds.

scurker:
closes #1710

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -1,6 +1,11 @@
 name: Create Release PR
 
-on: [workflow_dispatch]
+on:
+  workflow_dispatch:
+  # TODO: remove, just for testing
+  push:
+    branches:
+      - fix-actions-machine-user
 
 jobs:
   create_release:
@@ -18,4 +23,4 @@ jobs:
         run: yarn run release
         env:
           COMMIT_BODY_HEADER: ":robot: This PR is created automatically ([see releasing](../blob/develop/RELEASING.md)) \n\n"
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.MACHINE_USER_PAT }}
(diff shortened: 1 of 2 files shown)


Bracciata (commented):
Is it possible to add why we are making this change to the description as it will not immediately be clear to newcomers (including myself)

scurker:
> Is it possible to add why we are making this change to the description as it will not immediately be clear to newcomers (including myself)

https://github.com/dequelabs/cauldron/issues/1710 has the details why this is necessary

.github/workflows/create-release-pr.yml
```diff
@@ -18,4 +19,4 @@ jobs:
         run: yarn run release
         env:
           COMMIT_BODY_HEADER: ":robot: This PR is created automatically ([see releasing](../blob/develop/RELEASING.md)) \n\n"
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+          GITHUB_TOKEN: ${{ secrets.MACHINE_USER_PAT }}
```
anastasialanz, line 22: What user is this for? I see "machine" user. Is this for a certain email that you made to act as a machine user or an existing one we use?

scurker: This one https://github.com/axe-core

I don't think it's named great, but we are using it for other actions and I created a test pr here https://github.com/dequelabs/cauldron/pull/2083

anastasialanz: Should we add a comment for what user it's for, or do you think it's fine as is?

scurker: A comment where? I'm not sure a comment is appropriate because the user could change in the future.

PR #2088, opened 2025-09-03
feat(react): add mnemonic typeahead for ActionList

### LINKED ISSUE #1909: Add ActionMenu Mnemonics
One usability/accessibility enhancement we could add would be to include letter key mnemonics to improve the usability of the menu. This would be a user could type the first letter of the menu item, and the first menu item matching that letter would become focused. 

Primer has an example of this: https://primer.style/product/components/action-menu/accessibility/#menu-mnemonics

scurker:
closes #1909

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "terser-webpack-plugin": "^5.3.10",
     "ts-node": "^10.9.2",
     "typescript": "~5.9.2",
-    "webpack": "^5.101.3",
+    "webpack": "^5.102.1",
     "webpack-cli": "^6.0.1",
     "webpack-dev-server": "^5.2.2"
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2642,6 +2642,11 @@ bare-stream@^2.6.4:
   dependencies:
     streamx "^2.21.0"
 
+baseline-browser-mapping@^2.8.9:
+  version "2.8.18"
+  resolved "https://registry.yarnpkg.com/baseline-browser-mapping/-/baseline-browser-mapping-2.8.18.tgz#b44b18cadddfa037ee8440dafaba4a329dfb327c"
+  integrity sha512-UYmTpOBwgPScZpS4A+YbapwWuBwasxvO/2IOHArSsAhL/+ZdmATBXTex3t+l2hXwLVYK382ibr/nKoY9GKe86w==
+
 basic-ftp@^5.0.2:
   version "5.0.5"
   resolved "https://registry.yarnpkg.com/basic-ftp/-/basic-ftp-5.0.5.tgz#14a474f5fffecca1f4f406f1c26b18f800225ac0"
@@ -2730,14 +2735,15 @@ braces@^3.0.3, braces@~3.0.2:
   dependencies:
     fill-range "^7.1.1"
 
-browserslist@^4.0.0, browserslist@^4.12.0, browserslist@^4.23.3, browserslist@^4.24.0, browserslist@^4.25.1:
-  version "4.25.1"
-  resolved "https://registry.yarnpkg.com/browserslist/-/browserslist-4.25.1.tgz#ba9e8e6f298a1d86f829c9b975e07948967bb111"
-  integrity sha512-KGj0KoOMXLpSNkkEI6Z6mShmQy0bc1I+T7K9N81k4WWMrfz+6fQ6es80B/YLAeRoKvjYE1YSHHOW1qe9xIVzHw==
+browserslist@^4.0.0, browserslist@^4.12.0, browserslist@^4.23.3, browserslist@^4.24.0, browserslist@^4.25.1, browserslist@^4.26.3:
+  version "4.26.3"
+  resolved "https://registry.yarnpkg.com/browserslist/-/browserslist-4.26.3.tgz#40fbfe2d1cd420281ce5b1caa8840049c79afb56"
+  integrity sha512-lAUU+02RFBuCKQPj/P6NgjlbCnLBMp4UtgTx7vNHd3XSIJF87s9a5rA3aH2yw3GS9DqZAUbOtZdCCiZeVRqt0w==
   dependencies:
-    caniuse-lite "^1.0.30001726"
-    electron-to-chromium "^1.5.173"
-    node-releases "^2.0.19"
+    baseline-browser-mapping "^2.8.9"
+    caniuse-lite "^1.0.30001746"
+    electron-to-chromium "^1.5.227"
+    node-releases "^2.0.21"
     update-browserslist-db "^1.1.3"
 
 buffer-crc32@~0.2.3:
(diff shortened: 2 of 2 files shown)


packages/react/src/components/ActionList/ActionList.tsx
```diff
@@ -2,10 +2,11 @@ import React, {
   type MutableRefObject,
   forwardRef,
   useCallback,
-  useRef
+  useRef,
+  useState
 } from 'react';
 import classnames from 'classnames';
-import type { ListboxOption } from '../Listbox/ListboxContext';
+import { type ListboxOption } from '../Listbox/ListboxContext';
```
anastasialanz, line 9: What's the purpose of switching this to `{ type ListboxOption }` vs `type { ListboxOption }`?

docs/pages/components/Listbox.mdx
```diff
@@ -189,6 +189,11 @@ Listboxes can also support multiple selection of listbox options.
     type: 'boolean',
     description: 'Whether or not to allow for focusing of disabled listbox options.'
   },
+  {
+    name: 'activeOption',
+    type: 'Object',
+    description: 'Used to control the active option within the list box. Used for advanced behaviors.'
```
anastasialanz:
```suggestion
    description: 'Used to control the active option within the listbox. Used for advanced behaviors.'
```

packages/react/src/components/ActionList/ActionList.tsx
```diff
@@ -47,13 +52,30 @@ const ActionList = forwardRef<HTMLUListElement, ActionListProps>(
       [onAction]
     );
 
+    const containerRef = useMnemonics<HTMLUListElement>({
```
anastasialanz, line 55: Should we update the Cauldron docs to indicate that this component includes mnemonic type-ahead?

packages/react/src/utils/useMnemonics.ts
```diff
@@ -0,0 +1,127 @@
...
+      const container = containerRef.current;
+      if (!container) {
+        return;
+      }
+
+      // Prevent default behavior and stop propagation for mnemonic keys
+      event.preventDefault();
+      event.stopPropagation();
+
+      const elements = Array.from(
```
anastasialanz, line 115: Do you think it would be worth caching the elements instead of calling `querySelectorAll` anytime a key is pressed?

packages/react/src/utils/useMnemonics.ts
```diff
@@ -0,0 +1,127 @@
...
+};
+
+type useMnemonicsResults<T extends HTMLElement> = RefObject<T>;
+
+/**
+ * Get an element's accessible name by its aria-label or text content
+ */
+function getAccessibleLabel(element: Element): string {
+  return (
+    element.getAttribute('aria-label') || element.textContent?.trim() || ''
```
anastasialanz, line 40: Could `aria-labelledby` be used here too?

packages/react/src/utils/useMnemonics.ts
```diff
@@ -0,0 +1,127 @@
...
+   */
+  onMatch: (element: HTMLElement) => void;
+
+  /**
+   * Whether mnemonic navigation is enabled. Defaults to true.
+   */
+  enabled?: boolean;
+};
+
+type useMnemonicsResults<T extends HTMLElement> = RefObject<T>;
```
anastasialanz, line 29: `useMnemonics` returns `containerRef`, which is a `MutableRefObject`. Should this be updated here?

anastasialanz (commented):
Looks pretty good! I was playing around with many items of the same letter and keyboard navigation and I noticed one thing that I wasn't sure if it seemed right.

If I modify the ActionList - Multiple Selection on the docs site to have many "F" items, and I use a combination of the arrow keys + pressing "F", I get different results, and it jumps around instead of going to the next "F" item.

packages/react/src/components/ActionList/ActionList.tsx
scurker: I can't remember, it might've been a typescript/prettier thing but they're equivalent.

packages/react/src/utils/useMnemonics.ts
scurker: We have to consider the list of elements could be "live" so I don't think caching is beneficial here.

packages/react/src/utils/useMnemonics.ts
scurker: I considered that initially, but `aria-labelledby` adds a lot of complexity since you would need to do lookups then it could also be `aria-labelledby` and just goes deep down a rabbit hole. It adds a lot of potential complexity that I'm not sure we need (yet) but I can add a comment as to why it's not included.

packages/react/src/utils/useMnemonics.ts
scurker: In this case, we don't want the return value to be mutable but it needs to be mutable within the hook itself. Does that make sense?

packages/react/src/components/ActionList/ActionList.tsx
scurker: It doesn't effect how the component is used so I'm not sure how useful that is yet.

I do think it would be useful (longterm) to eventually have this kind of documentation.

packages/react/src/utils/useMnemonics.ts
anastasialanz: I was thinking we could check if the container DOM was modified and update the cache.

packages/react/src/components/ActionList/ActionList.tsx
anastasialanz: Had to look it up. `{ type ListboxOption }` supports mixed imports.

packages/react/src/utils/useMnemonics.ts
anastasialanz: I think I was confused because it gets cast as a `MutableRefObject` in the `ActionList`.

packages/react/src/utils/useMnemonics.ts
scurker: I still am not sure caching is beneficial. There's overhead to adding dom mutation observers and `querySelectorAll` should be sufficient enough _especially_ since we're just querying on the container and not the full dom. I think we should consider caching if and only when we determine it's a performance problem.

packages/react/src/utils/useMnemonics.ts
```diff
@@ -0,0 +1,155 @@
...
+        return;
+      }
+
+      const currentActiveElement = getActiveElement(containerRef.current);
+      let nextActiveElement: HTMLElement | null = null;
+
+      if (currentActiveElement) {
+        nextActiveElement = matchingElements.find(
+          (element) =>
+            // Find the next matching element the is _after_ the current active element
```
anastasialanz:
```suggestion
            // Find the next matching element that is _after_ the current active element
```

packages/react/src/utils/useMnemonics.ts
```diff
@@ -0,0 +1,155 @@
...
+function getActiveElement(root: HTMLElement): HTMLElement | null {
+  let activeElement: HTMLElement | null;
+
+  if (
+    document.activeElement === root &&
+    root.hasAttribute('aria-activedescendant')
+  ) {
+    activeElement = document.getElementById(
+      // Validating attribute above with "hasAttribute"
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
anastasialanz, line 56: I don't think we need this anymore after the refactoring.

scurker: We don't need the non-null assertion? I think this is still needed because the root could have elements managed via active descendant.

anastasialanz: I ran `typecheck` without `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` and it didn't error.

scurker: Locally when I remove that line typescript still errors:

typescript error with removed lines

...otherwise it provides a warning 

typescript warning with no comment

anastasialanz: It works on my machine now. 😆

PR #2130, opened 2025-09-29
docs: fix broken relative links

### LINKED ISSUE #2129: Bug: Relative doc links are broken
### Steps to reproduce

1. Go to any Cauldron doc page that uses a relative link (eg, any of the "Related Components" links at the bottom of the [FieldGroup](https://cauldron.dequelabs.com/components/FieldGroup) docs)
2. Click a link
3. Observe that it is broken (goes to `https://cauldron.dequelabs.com/components/FieldGroup/TextField` instead of the expected https://cauldron.dequelabs.com/components/TextField` - note the extra `/FieldGroup/` path segment).

### Expected Behavior

The links shouldn't be broken. The links are authored in the syntax you'd expect to work (eg, `[TextField](./TextField)`), so this should ideally be fixed systematically rather than by updating all the individual links to use `../` instead of `./`.

### Version

v6.22.2

Chrome

### Relevant log output

```shell

```

dbjorge:
closes #2129

### DIFF
--- a/docs/mdx-components.tsx
+++ b/docs/mdx-components.tsx
@@ -43,11 +43,33 @@ function Link({
   href,
   ...props
 }: React.AnchorHTMLAttributes<HTMLAnchorElement>) {
+  if (!href) {
+    return <CauldronLink {...props} />;
+  }
+
   const external = href.indexOf('://') > 0 || href.indexOf('//') === 0;
-  return external ? (
-    <CauldronLink href={href} {...props} />
-  ) : (
-    <RouterLink to={href} {...props} />
+  if (external) {
+    return <CauldronLink href={href} {...props} />;
+  }
+
+  // This works around an inconsistency between normal <a href> behavior
+  // vs react-router's Link behavior. react-router resolves relative paths
+  // relative to the *current* location, not the *parent* of the current
+  // location. This fix means that if /components/Foo.mdx has
+  // a link like [Bar](./Bar), it will be resolved as "/components/Bar" rather
+  // than "/components/Foo/Bar".
+  const relative = !href.startsWith('/'); // Could be "bar" or "./bar", as opposed to "/bar"
+  const fixedHref = relative ? '../' + href : href;
+
+  return (
+    <RouterLink
+      to={fixedHref}
+      // relative="path" makes react-router resolve ".." in relative paths by skipping
+      // path components, as opposed to its normal behavior of skipping <Route>s (which
+      // might use a multi-segment path like "/components/Foo").
+      relative="path"
+      {...props}
+    />
   );
 }
 
--- a/docs/mdx-components.tsx
+++ b/docs/mdx-components.tsx
@@ -55,10 +55,10 @@ function Link({
   // This works around an inconsistency between normal <a href> behavior
   // vs react-router's Link behavior. react-router resolves relative paths
   // relative to the *current* location, not the *parent* of the current
-  // location. This fix means that if /components/Foo.mdx has
-  // a link like [Bar](./Bar), it will be resolved as "/components/Bar" rather
-  // than "/components/Foo/Bar".
-  const relative = !href.startsWith('/'); // Could be "bar" or "./bar", as opposed to "/bar"
+  // location. This fix means that if /components/Foo.mdx has a link like
+  // `[Bar](./Bar)`, it will be resolved as "/components/Bar" rather than
+  // "/components/Foo/Bar".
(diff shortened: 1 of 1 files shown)


docs/mdx-components.tsx
```diff
@@ -43,11 +43,33 @@ function Link({
...
+  }
+
+  // This works around an inconsistency between normal <a href> behavior
+  // vs react-router's Link behavior. react-router resolves relative paths
+  // relative to the *current* location, not the *parent* of the current
+  // location. This fix means that if /components/Foo.mdx has a link like
+  // `[Bar](./Bar)`, it will be resolved as "/components/Bar" rather than
+  // "/components/Foo/Bar".
+  const relative = !href.startsWith('/'); // "bar" or "./bar", as opposed to "/bar"
+  const fixedHref = relative ? '../' + href : href;
```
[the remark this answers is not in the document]
dbjorge: This is correct and desirable.

docs/mdx-components.tsx
```diff
@@ -43,11 +43,33 @@ function Link({
...
+    return <CauldronLink href={href} {...props} />;
+  }
+
+  // This works around an inconsistency between normal <a href> behavior
+  // vs react-router's Link behavior. react-router resolves relative paths
+  // relative to the *current* location, not the *parent* of the current
+  // location. This fix means that if /components/Foo.mdx has a link like
+  // `[Bar](./Bar)`, it will be resolved as "/components/Bar" rather than
+  // "/components/Foo/Bar".
+  const relative = !href.startsWith('/'); // "bar" or "./bar", as opposed to "/bar"
```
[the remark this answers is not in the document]
dbjorge: That's a great catch, way to go :robot:

docs/mdx-components.tsx
```diff
@@ -43,11 +43,36 @@ function Link({
...
+
+  // This works around an inconsistency between normal <a href> behavior
+  // vs react-router's Link behavior. react-router resolves relative paths
+  // relative to the *current* location, not the *parent* of the current
+  // location. This fix means that if /components/Foo.mdx has a link like
+  // `[Bar](./Bar)`, it will be resolved as "/components/Bar" rather than
+  // "/components/Foo/Bar".
+  //
+  // Should match "bar", "./bar", and "../bar"
+  // Should not match "/bar", "#bar", "?bar=baz", or "mailto:bar".
```
scurker, line 63: I traced this back to https://github.com/dequelabs/cauldron/commit/c02b28718c5f4e1ebfd08542231249a5284e733e and I'm a little torn on what the right fix should be. I'm not exactly sure what the exact behavior we were relying on before but I would be more inclined to implement this properly if we're treating these links as relative to the path.

I tried to dig through the react-router changelog and nothing stuck out to me as being the root cause of this change, but my guess is that we were relying on some (potentially a bug) behavior where the route was relative to the parent and not the current path. Going back to my initial inclination, I think my ideal might be to actually replace all of the incorrect paths in code from `./*` to `../*` to match what the expected behavior is now in React Router v6/v7.

Thoughts?

scurker: It does look like we're primarily using the relative linking to link to other component files: https://github.com/search?q=repo%3Adequelabs%2Fcauldron+%2F%5D%5C%28%5B%5Ehttps%3F%5D%2F+path%3A*.mdx&type=code

dbjorge: I think from the perspective of a typical contributor doing a typical doc update, the fact that the markdown syntax for `[a link](target)` happens to be implemented using a `react-router` `<Link to>` is going to be an implementation detail they aren't likely to be thinking about, even if they understand the behavior difference it implies. I think it's vastly, vastly more likely that a typical contributor making a doc change is going to assume "the markdown syntax for `[a link](./target)` is going to work similarly to how it does for markdown files in GitHub or a standard `<a href="./target">` tag: it'll look for a sibling file named `target`".

Based on that, I think it's a lot more intuitive for the markdown link behavior to work like relative links in GitHub markdown files (and regular `<a href>`) than it is for them to work like react-router's implementation, so my preference would be to keep the way they're currently specified. That said: it's not a hill I want to spend a lot of energy fighting on; if you feel strongly about it I'm willing to disagree and commit.

scurker: Ah, I hadn't considered from a markdown perspective. I still suspect this may raise problems in the long run since we won't be able to adequately use nested relative links but that can be a problem that we solve later and this is a sufficient solution for now.

PR #2135, opened 2025-10-08
fix(styles): use correct heading sizes for page/section headings

### LINKED ISSUE #2136: Implement heading typography for page/section headers
Our page/section headers do not currently specify the desired heading sizes for the appropriate headings and is dependent on either the browser defaults or the css cascade, which ever is applied. We should adopt the new typographic scale for headings so the correct heading size is applied when using these components.

scurker:
closes #2136

### DIFF
--- a/packages/styles/index.css
+++ b/packages/styles/index.css
@@ -50,8 +50,8 @@
 @import './impact-badge.css';
 @import './drawer.css';
 @import './bottom-sheet.css';
-@import './page-header.css';
 @import './section-header.css';
+@import './page-header.css';
 @import './empty-state.css';
 @import './action-menu.css';
 @import './action-list.css';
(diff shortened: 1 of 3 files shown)


packages/styles/index.css
```diff
@@ -50,8 +50,8 @@
 @import './impact-badge.css';
 @import './drawer.css';
 @import './bottom-sheet.css';
-@import './page-header.css';
 @import './section-header.css';
+@import './page-header.css';
```
scurker, line 54: I had to move the order around so that the priority would still apply correctly.

I tried to implement this using [`@layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) but we would not be able to apply it correctly in other consuming products since un-layered rules (or overrides) would still get applied as a priority.

Issue #2146, opened 2025-10-16
Bug: AnchoredOverlay shifts large overlays off-screen at top of viewport


scurker:
### Steps to reproduce

1. Create an AnchoredOverlay with content taller than the viewport
1. Position the target element in the lower third of the page
1. Set placement to bottom or use auto placement
1. Open the overlay
1. Observe the overlay shifting upward beyond the visible viewport

### Expected Behavior

Large anchored overlays should:

- Remain visible and accessible within the viewport
- Potentially apply a limiter middleware or boundary constraints
- Ensure critical content (especially the top of the overlay) remains visible

We should consider either:

Applying constraints so that the overlay doesn't overflow the top of the screen (preventing scrolling "up").

OR 

Allowing the anchored overlay shift middleware to be configured or disabled.

### Version

N/A

PR #2147, opened 2025-10-16
fix(react): prevent AnchoredOverlays from overflowing top boundaries

scurker:
closes https://github.com/dequelabs/cauldron/issues/2146

Testing this is tricky because it requires simulating real viewport and positioning calculations that are hard to replicate in unit tests. The behavior depends on the exact dimensions of the target element, the overlay content, and the viewport size all working together. I've mocked detectOverflow to verify that the component still renders. I've done manual validation of the issue to ensure we're rendering the overlay correctly when it exceeds the top of the root document and flips to an alternative placement to keep content visible.

It's hard to represent, but I tested this by forcing an overflow of an anchored element that would exceed the top of the document by zooming in 400% if it were flipped to the top:

image

After implementing, it's always shifting down to the bottom.

### DIFF
--- a/packages/react/src/components/AnchoredOverlay/index.tsx
+++ b/packages/react/src/components/AnchoredOverlay/index.tsx
@@ -72,6 +72,8 @@ const preventTopOverflowMiddleware: Middleware = {
     if (overflow?.top >= 0) {
       return {
         reset: {
+          // Replace the initial placement axis with 'bottom' while preserving alignment (start/end)
+          // Examples: 'top' -> 'bottom', 'top-start' -> 'bottom-start'
           placement: (state.placement.replace(
             /\w+?(-(start|end))?$/i,
             'bottom$1'


packages/react/src/components/AnchoredOverlay/index.tsx
```diff
@@ -55,6 +57,33 @@ function getAutoAlignment(
...
+    });
+
+    if (overflow?.top >= 0) {
+      return {
+        reset: {
+          placement: (state.placement.replace(
+            /\w+?(-(start|end))?$/i,
+            'bottom$1'
+          ) || 'bottom') as Placement
+        }
```
Zidious, line 81: nit - for future folks it might be helpful what his regex is achieving

Side note, whats the reason for the cast here?

Zidious (dismissed):
LGTM - discussed on Slack what this PR resolves. As this is very situational in how it triggers, writing a unit test is somewhat difficult, this will be tested downstream but a visual aid on the PR description would be beneficial. 

Just a nit inline nothing blocking.

packages/react/src/components/AnchoredOverlay/index.tsx
scurker: Because `.replace` just returns a string and Placement is a hard value of one of the possible placements and doesn't necessarily satisfy the middleware type. I'll add a comment for the regex part.

Zidious (approved):

PR #2148, opened 2025-10-17
docs: top bar item with action menu screen reader + focus behavior

chornonoh-vova:
# Top bar + action menu correct usage docs

Update the `TopBarItem` + `ActionMenu` usage examples and settings menu in docs
to correctly announce menu and menu items by screen readers and maintain correct
focus behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [6.23.0](https://github.com/dequelabs/cauldron/compare/v6.22.3...v6.23.0) (2025-10-20)
+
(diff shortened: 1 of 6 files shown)


docs/components/Note.tsx
```diff
@@ -10,7 +10,7 @@ interface NoteProps {
 }
 
 export default function Note({ children, title, variant = 'info' }: NoteProps) {
-  const noteTitle = title ?? variant === 'info' ? 'Note' : 'Warning';
+  const noteTitle = title ?? (variant === 'info' ? 'Note' : 'Warning');
```
scurker, line 13: Nit: this appears to be unrelated to the change.

chornonoh-vova: I was trying to use this component with a custom title, and without this change it wasn't working 😢

Issue #1290, opened 2023-11-14
Add Tree Component


scurker:

bobbyomari:
Yesterday I spoke to @anastasialanz about this component and I think for the first iteration we will focus on the primary two use cases: 

- TreeView node (i.e. user can only expand/collapse each node)
- TreeView node with checkbox

The other use cases for trailing actions and checkbox with trailing action are not something that need to be implemented in the immediate. Axe Reports only needs the versions mentioned above. We also agreed that the branches that connect nodes is something we will not implement for the first iteration to reduce scope.

PR #2191, opened 2026-01-12
feat: add TreeView Component

JoshMK:
Closes:

https://github.com/dequelabs/axe-reports/issues/2952
https://github.com/dequelabs/cauldron/issues/1290


packages/react/src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Tree } from 'react-aria-components';
+import TreeViewItem from './TreeViewItem';
+
+export interface TreeViewFileType {
```
anastasialanz, line 5: Does this have to be named "file"? Could it be more generic like `TreeViewNode`?

packages/styles/tree-view.css
```diff
@@ -0,0 +1,74 @@
...
+  --treeview-focus-ring-color: var(--focus-light);
+  --treeview-selected-color: var(--focus-light);
+  --treeview-highlight-background: var(--accent-light);
+}
+
+.cauldron--theme-dark {
+  --treeview-focus-ring-color: var(--focus-dark);
+  --treeview-selected-color: var(--focus-dark);
+  --treeview-highlight-background: var(--accent-dark);
+}
```
anastasialanz, line 12: The variables should be named `--tree-view-*` to be consistent with the rest of the codebase.

packages/react/src/components/TreeView/TreeView.test.tsx
anastasialanz, line 1: In our contributing docs, there's a note about using jest-axe to catch a11y violations. Can we implement that in these tests?

https://github.com/dequelabs/cauldron/blob/3d37eb6a6f3730c3bf856230824aee102eba811d/CONTRIBUTING.md?plain=1#L229-L239

We should include screenshot tests for this component as well.

https://github.com/dequelabs/cauldron/blob/3d37eb6a6f3730c3bf856230824aee102eba811d/CONTRIBUTING.md?plain=1#L205

packages/react/src/components/TreeView/TreeView.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+  expect(getByText('TreeView')).toBeInTheDocument();
+  expect(getByText('Another One')).toBeInTheDocument();
+});
+
+test('selects a tree item on click', () => {
+  const { getByText } = render(
+    <TreeView ariaLabel="Test TreeView" items={items} selectionMode="single" />
+  );
+  const child1 = getByText('TreeView');
+  fireEvent.click(child1);
```
anastasialanz, line 37: We should prefer `userEvent` over `fireEvent` to simulate how a user would interact with this component.

https://testing-library.com/docs/user-event/intro/#differences-from-fireevent

packages/react/src/components/TreeView/TreeViewItem.tsx
anastasialanz, line 1: Can we include tests for things like the `handleOnAction` workaround and checkbox rendering based on `selectionMode`?

packages/react/src/components/TreeView/TreeViewItem.tsx
```diff
@@ -0,0 +1,85 @@
...
+      ref={treeItemRef}
+    >
+      <TreeItemContent>
+        {({ selectionMode, isSelected }: TreeItemContentRenderProps) => (
+          <>
+            <Button slot="chevron">
+              <Icon type="chevron-right" />
+            </Button>
+            {selectionMode !== 'none' ? (
+              <Checkbox
```
anastasialanz, line 60: I'm not able to select an individual checkbox with `space`. As soon as I let go of the space bar, it becomes unchecked again.

https://github.com/user-attachments/assets/cb3aad1f-68fa-4415-910b-9e063b87278c

I'm able to use `space` on the checkbox on https://react-aria.adobe.com/Tree.

mateoviilla1 (changes requested):
## Review

### Critical

**[`TreeViewItem.tsx:60`](https://github.com/dequelabs/cauldron/blob/fc434ff7edbac71c198e66136ac970f5e36debbe/packages/react/src/components/TreeView/TreeViewItem.tsx#L60) — Checkbox Space key bug**

Pressing Space on a focused checkbox doesn't maintain the checked state — it resets immediately on keyup. The combination of `onChangeToggle={false}` on Cauldron's `<Checkbox>` and react-aria's selection state sync causes the visual state to reset on re-render. For an accessibility-first component this is a blocking issue. Consider using react-aria's own `<Checkbox>` from `react-aria-components`, or routing keyboard interaction through `TreeItem`'s selection handler so the state flows through react-aria's unified event model.

---

**[`TreeViewItem.tsx:30–43`](https://github.com/dequelabs/cauldron/blob/fc434ff7edbac71c198e66136ac970f5e36debbe/packages/react/src/components/TreeView/TreeViewItem.tsx#L30-L43) — Direct DOM mutation in `handleOnAction`**

`handleOnAction` calls `setAttribute('aria-selected', ...)` directly on the DOM node to work around react-aria not toggling selection on action. This bypasses React's render cycle, causing the virtual DOM and real DOM to diverge — any subsequent re-render will overwrite the mutation. The correct fix is to manage selection externally via `selectedKeys`/`onSelectionChange` props on `<Tree>`, keeping state in React rather than the DOM.

---

### Important

**[`index.tsx:14`](https://github.com/dequelabs/cauldron/blob/fc434ff7edbac71c198e66136ac970f5e36debbe/packages/react/src/components/TreeView/index.tsx#L14) — `onAction` type mismatch**

`onAction` is typed as `() => void` but `handleOnAction` calls it as `onAction(key)` where `key` is a `string`. TypeScript silently accepts calling a `() => void` with arguments, so this compiles but the public API contract is wrong. Change to `(key: string) => void`.

---

**[`tree-view.css:1–12`](https://github.com/dequelabs/cauldron/blob/fc434ff7edbac71c198e66136ac970f5e36debbe/packages/styles/tree-view.css#L1-L12) — CSS variable naming**

Custom properties use `--treeview-*` (e.g. `--treeview-padding`, `--treeview-focus-ring-color`). The Cauldron codebase convention uses `--tree-view-*` with word separators. Rename all variables accordingly.

---

**[`TreeView.test.tsx`](https://github.com/dequelabs/cauldron/blob/fc434ff7edbac71c198e66136ac970f5e36debbe/packages/react/src/components/TreeView/TreeView.test.tsx#L1) — Test quality**

Three gaps per CONTRIBUTING.md:
1. `fireEvent` should be replaced with `userEvent` from `@testing-library/user-event` to simulate real user interaction (focus/pointer/keyboard sequences).
2. Missing `jest-axe` assertions — at least one `expect(await axe(container)).toHaveNoViolations()` per rendered variant.
3. Missing screenshot tests.

---

**[`index.tsx:5`](https://github.com/dequelabs/cauldron/blob/fc434ff7edbac71c198e66136ac970f5e36debbe/packages/react/src/components/TreeView/index.tsx#L5) — `TreeViewFileType` naming**

The exported type is named `TreeViewFileType`, implying the tree is file-system specific. The component is generic. Rename to `TreeViewItem` or `TreeViewNode`.

---

### Suggestions

**[`TreeViewItem.tsx:73`](https://github.com/dequelabs/cauldron/blob/fc434ff7edbac71c198e66136ac970f5e36debbe/packages/react/src/components/TreeView/TreeViewItem.tsx#L73) — Redundant `items` prop on `<Collection>`**

`<Collection items={children}>` passes `items` but the children are rendered manually as a JSX array. In react-aria, `items` is only consumed when children is a render function — here it's silently ignored. Either remove `items={children}`, or switch to the dynamic collection pattern: `<Collection items={children}>{(child) => <TreeViewItem ... />}</Collection>`.

---

**[`TreeView.mdx:4`](https://github.com/dequelabs/cauldron/blob/fc434ff7edbac71c198e66136ac970f5e36debbe/docs/pages/components/TreeView.mdx#L4) — Incorrect `source` link**

The frontmatter `source` field points to `TreeView.tsx` which doesn't exist — the actual entry point is `index.tsx`. Update to `.../TreeView/index.tsx`.

frankensteinke:
Closing in favor of https://github.com/dequelabs/cauldron/pull/2322

Issue #2067, opened 2025-08-22
TableHeader Visual Variant Enhancement


scurker:
# TableHeader

## Overview

Add a visual variant to the `TableHeader` component that allows it to be styled like a `TableCell` while maintaining its semantic `<th>` element for accessibility purposes.

### Problem Statement

Currently, `TableHeader` components are always styled with a darker background to visually distinguish them as headers. However, there are valid use cases where headers may need to be visually indistinguishable from regular cells while maintaining their semantic meaning for accessibility. This is particularly common in:

- Row headers that shouldn't be visually prominent
- Tables with mixed row/column headers where consistent visual treatment improves readability

### Visual Design

Visually, we want to be able to display table headers that appear like table cells. There should be no visual difference between the two.

### Interface / Props

Introduce new `variant` property that allows for a different visual aesthetic:

```tsx
interface TableHeaderProps {
  variant?: 'header' | 'cell'
}
```

When not defined by the content author, `header` should be the default value.

### Accessibility

Since this is just using a `<th>` under the hood, there shouldn't be any new accessibility considerations.

### Responsiveness

Not applicable, uses same behaviors as `<TableCell>`.

### Variants (optional)

Yes, there would be a `header` and `cell` variant.

### Usage Examples

```tsx
<TableHeader variant="cell">...</TableHeader>
```

### Referencing Implementations

Not applicable.

PR #2196, opened 2026-01-16
feat(react): add `TableHeader` visual cell variant

Zidious:
This patch adds a `variant` prop to the `TableHeader` component allows folks to visually style the header as a cell if required.

Closes: https://github.com/dequelabs/cauldron/issues/2067

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -28,7 +28,7 @@
     "focusable": "^2.3.0",
     "keyname": "^0.1.0",
     "react-id-generator": "^3.0.1",
-    "react-syntax-highlighter": "^15.5.0",
+    "react-syntax-highlighter": "^16.0.1",
     "tslib": "^2.4.0"
   },
   "devDependencies": {
@@ -51,7 +51,7 @@
     "@types/node": "^17.0.42",
     "@types/react": "^18.0.12",
     "@types/react-dom": "^18.0.5",
-    "@types/react-syntax-highlighter": "^15.5.2",
+    "@types/react-syntax-highlighter": "^15.5.13",
     "@types/sinon": "^10",
     "autoprefixer": "^9.7.6",
     "babel-plugin-module-resolver": "^4.0.0",
@@ -79,8 +79,5 @@
   "peerDependencies": {
     "react": ">=16.6 <= 18",
     "react-dom": ">=16.6 <= 18"
-  },
-  "resolutions": {
-    "**/refractor/prismjs": "^1.30.0"
   }
 }
--- a/packages/react/yarn.lock
+++ b/packages/react/yarn.lock
@@ -1152,12 +1152,17 @@
   resolved "https://registry.npmjs.org/@babel/regjsgen/-/regjsgen-0.8.0.tgz"
   integrity sha512-x/rqGMdzj+fWZvCOYForTghzbtqPDZ5gPwaoNGHdgDfF2QA/XZbCBp4Moo5scrkAMPhB7z26XM/AaHuIJdgauA==
 
-"@babel/runtime@^7.12.5", "@babel/runtime@^7.21.0", "@babel/runtime@^7.3.1", "@babel/runtime@^7.8.4":
+"@babel/runtime@^7.12.5", "@babel/runtime@^7.21.0", "@babel/runtime@^7.8.4":
   version "7.26.10"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.26.10.tgz#a07b4d8fa27af131a633d7b3524db803eb4764c2"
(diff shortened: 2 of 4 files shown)


Zidious:
@shawnsharpDQ - Tagging you for design review.

docs/pages/components/Table.mdx
```diff
@@ -334,18 +338,68 @@ function SortableTableExample() {
 </Table>
 ```
 
+### With Row Headings (Cell Variant) and Column Headings
```
Zidious, line 341: This is only new addition are below, above changes are released to formatting/Prettier by the looks of it.

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -57,10 +60,14 @@ const TableHeader = forwardRef<HTMLTableHeaderCellElement, TableHeaderProps>(
...
-        className={classNames('TableHeader', className, {
-          'TableHeader--sort-ascending': sortDirection === 'ascending',
-          'TableHeader--sort-descending': sortDirection === 'descending'
-        })}
+        className={classNames(
+          variant === 'cell' ? 'TableCell' : 'TableHeader',
+          className,
+          {
+            'TableHeader--sort-ascending': sortDirection === 'ascending',
+            'TableHeader--sort-descending': sortDirection === 'descending'
```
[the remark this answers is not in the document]
Zidious: I wasn't sure whether we _only_ wanted these when the variant is `header` as technically it's still marked up as a header.

docs/pages/components/Table.mdx
```diff
@@ -478,62 +536,61 @@ For more advanced usage, the width value does not need to be specified as part o
...
     },
     {
       name: 'align',
       type: ['start', 'center', 'end'],
-      description: 'Only allowed when the table has a grid layout. Overrides any column alignments for this table header.'
+      description:
+        'Only allowed when the table has a grid layout. Overrides any column alignments for this table header.'
+    },
+    {
+      name: 'variant',
```
Bracciata, line 567: While variant for styling matches the rest of the repo, I feel it makes for a slightly confusing experience for developers in this case. Curious others thoughts on it.

Zidious: This is what's specified in the ticket, I'm open to suggestions though

Bracciata: We decided to stick with header and cell.

packages/styles/table.css
```diff
@@ -97,6 +97,10 @@
   outline-offset: unset;
 }
 
+.TableHeader__cell-variant {
```
anastasialanz: To follow BEM syntax:
```suggestion
.TableHeader--cell {
```

https://github.com/dequelabs/cauldron/blob/9830aeddeaecc6bbcd5d1c38742b5d674130d872/docs/pages/styles.mdx?plain=1#L31-L41

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -57,10 +60,16 @@ const TableHeader = forwardRef<HTMLTableHeaderCellElement, TableHeaderProps>(
       <th
         ref={tableHeaderRef}
         aria-sort={sortDirection}
-        className={classNames('TableHeader', className, {
-          'TableHeader--sort-ascending': sortDirection === 'ascending',
-          'TableHeader--sort-descending': sortDirection === 'descending'
-        })}
+        className={classNames(
+          variant === 'cell'
+            ? ['TableCell', 'TableHeader__cell-variant']
```
anastasialanz: I think we'd want to keep the `.TableHeader` class here and then update the stylesheet to include the variant class `TableHeader--cell` anywhere `.TableCell` is declared.

anastasialanz (changes requested):
I have a small change to follow the existing BEM syntax. I was wondering if it would be better to keep `.TableHeader` instead of adding `.TableCell` but it requires modifying the CSS in many places and would mean we'd have to keep cell styles in sync with the modifier and have more duplication.

packages/styles/table.css
Zidious: Applied! I made the naming a bit more explicit to avoid naming confusions here.

packages/styles/table.css
```diff
@@ -97,7 +97,8 @@
   outline-offset: unset;
 }
 
-.TableHeader__cell-variant {
+/* Applies TableCell styles to a <TableHeader> with variant="cell". Only results in visual changes, not semantic ones. */
+.TableHeader--CellVariant {
```
anastasialanz: Can we keep the casing how you had it before `--cell-variant` to follow the casing style for other modifiers?

PR #2206, opened 2026-01-20
fix(react): remove false ellipsis when text fits in Chrome

maksym-shynkarenko:
## Problem                                                                                                                                                                                                                     
Chrome incorrectly displayed ellipsis ("...") for `TextEllipsis` component with `maxLines` prop, even when text fit within the specified number of lines. This was particularly noticeable with short text using `maxLines={2}` or higher.                                                                                                                                                                                                                    

## Root Cause                                                                                                                                                                                                                  
The `.TextEllipsis:after` pseudo-element with `display: block` (added in #1926 as a Safari workaround) caused Chrome to miscalculate overflow when combined with `-webkit-line-clamp` in multiline mode.                       

## Solution                                                                                                                                                                                                                    
Changed the CSS selector from `.TextEllipsis:after` to `.TextEllipsis:not(.TextEllipsis--multiline):after` to apply the Safari workaround only to single-line ellipsis. This prevents the Chrome rendering bug while preserving the original Safari fix.                                                                                                                                                                                                      

## Result                                                                                                                                                                                                                      

  - ✅ Ellipsis now displays only when text genuinely exceeds the line count                                                                                                                                                     
  - ✅ Safari native tooltip workaround preserved for single-line mode                                                                                                                                                           

### Video
<details><summary>Video: fix in Chrome and Safari </summary>

https://github.com/user-attachments/assets/76ad41c0-3577-474b-a27d-b370a36c15f6


</details> 

<details><summary>Video: responsiveness </summary>

https://github.com/user-attachments/assets/a6bda827-13fa-40a3-820b-ae1e8b5c3123


</details> 

Closes: Walnut [#13167](https://github.com/dequelabs/walnut/issues/13167)

### DIFF
--- a/packages/styles/text-ellipsis.css
+++ b/packages/styles/text-ellipsis.css
@@ -8,8 +8,9 @@
 /* Try and prevent Safari from showing a platform
  * tooltip when visible text is truncated
  * see: https://github.com/dequelabs/cauldron/issues/1926
+ * Only apply to single-line ellipsis to avoid Chrome false ellipsis bug
  */
-.TextEllipsis:after {
+.TextEllipsis:not(.TextEllipsis--multiline):after {
   content: '';
   display: block;
 }


packages/react/src/components/TextEllipsis/index.tsx
```diff
@@ -78,11 +71,27 @@ const TextEllipsis = React.forwardRef(
...
           }
 
-          setHasOverflow(
-            typeof maxLines === 'number'
-              ? overflowElement.clientHeight < overflowElement.scrollHeight
-              : overflowElement.clientWidth < overflowElement.scrollWidth
-          );
+          if (typeof maxLines === 'number') {
+            // Apply line-clamp first to measure if text actually overflows
+            overflowElement.style.webkitLineClamp = String(maxLines);
```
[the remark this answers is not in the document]
maksym-shynkarenko: we don't have huge dynamic size changes, so we should not complicate this code

packages/react/src/components/TextEllipsis/index.tsx
```diff
@@ -78,11 +71,27 @@ const TextEllipsis = React.forwardRef(
...
+
+            // Check if there's actual overflow by comparing scroll and client height
+            const hasRealOverflow =
+              overflowElement.scrollHeight > overflowElement.clientHeight;
```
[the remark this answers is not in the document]
maksym-shynkarenko: the same as [above](https://github.com/dequelabs/cauldron/pull/2206#discussion_r2709428015)

Bracciata (dismissed):
This looks good to me!

packages/react/src/components/TextEllipsis/index.tsx
```diff
@@ -63,13 +63,6 @@ const TextEllipsis = React.forwardRef(
       );
     }
 
-    if (typeof maxLines === 'number') {
-      props.style = {
-        WebkitLineClamp: maxLines || 2,
-        ...props.style
-      };
-    }
```
Zidious, line 71: This is correct, the bug here is not actually anything to do with _how_ we're applying the `-webkit-line-clamp`. It's the styling we're applying in the `:after` block in the `text-ellipsis.css` file: 

```css
/* Try and prevent Safari from showing a platform
 * tooltip when visible text is truncated
 * see: https://github.com/dequelabs/cauldron/issues/1926
 */
.TextEllipsis:after {
  content: '';
  display: block;
}
```

Having `display: block;` causes the three trailing `...` to appear regardless of whether it's filled the entire container or not. I've replicated what this component is doing in pure HTML and CSS and you can replicate what the live Cauldron site is doing today which is: https://cauldron.dequelabs.com/components/TextEllipsis#multi-line-ellipsis


JS Fiddle playground: https://jsfiddle.net/t9g6je2v/2/

I think we should investigate that part vs adding this style on the DOM observer.

maksym-shynkarenko:
> This PR references closing [dequelabs/walnut#13260](https://github.com/dequelabs/walnut/issues/13260) but this fix seems unrelated to that, am I incorrect in thinking that?

yeah, sorry, the wrong PR for that ticket, wil lremove it from PR description

packages/react/src/components/TextEllipsis/index.tsx
maksym-shynkarenko: if this issue is not urgent, I will investigate it after my vacation (02/02/2026). For now, I've moved it in draft

maksym-shynkarenko: That work, thanks!

packages/styles/text-ellipsis.css
```diff
@@ -8,8 +8,9 @@
 /* Try and prevent Safari from showing a platform
  * tooltip when visible text is truncated
  * see: https://github.com/dequelabs/cauldron/issues/1926
+ * Only apply to single-line ellipsis to avoid Chrome false ellipsis bug
  */
-.TextEllipsis:after {
+.TextEllipsis:not(.TextEllipsis--multiline):after {
```
Zidious, line 13: I suspect there is no good way for us to write a test for this change?

anastasialanz: @maksym-shynkarenko could you make a screenshot test similar to what Tommy did here https://github.com/dequelabs/cauldron/commit/f92cb92b40f72ecde361f3a9ef30e21a63e21f3c?

https://github.com/dequelabs/cauldron/blob/f92cb92b40f72ecde361f3a9ef30e21a63e21f3c/packages/react/src/components/Button/screenshots.e2e.tsx#L594-L618

maksym-shynkarenko: as we discussed in Slack, we don't need a screenshot for now

PR #2208, opened 2026-01-21
fix: combobox double tab

zlayaAvocado:
Ref: https://github.com/dequelabs/axe-extension/issues/6942

When using `listbox` inside a combobox with a keyboard, it does a double `tab` first focusing on the combobox `input` element and then quietly on the `listbox` with the focus disappearing. The issue is originally from the `axe-extension` but I also was able to replicate it on the cauldron components page with the grouped combobox.

Before:
<details>

https://github.com/user-attachments/assets/087c7e7d-5303-43f9-8fc6-afaa8a1ace56

</details>

After:
<details>

https://github.com/user-attachments/assets/9ebc423c-bcf1-440d-9462-197746340eda

</details>

### DIFF
--- a/packages/react/src/components/Combobox/Combobox.test.tsx
+++ b/packages/react/src/components/Combobox/Combobox.test.tsx
@@ -1663,7 +1663,7 @@ test('should have no axe violations with no matching results', async () => {
   });
 
   const listboxId = combobox.getAttribute('aria-controls');
-  const listbox = document.getElementById(listboxId);
+  const listbox = document.getElementById(listboxId as string);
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Combobox/Combobox.tsx
```diff
@@ -632,7 +632,7 @@ const Combobox = forwardRef<
         onSelectionChange={handleSelectionChange}
         onActiveChange={handleActiveChange}
         ref={listboxRef}
-        tabIndex={undefined}
+        tabIndex={noMatchingOptions ? undefined : -1}
```
anastasialanz, line 635: Can we write a test for this change?

PR #2213, opened 2026-01-22
feat: support fully custom header markup in Dialog/Modal

frankensteinke:
Uses new composable components (DialogHeader, DialogHeading, DialogCloseButton, similar to the existing DialogFooter and DialogContent) to allow a custom markup structure for Dialog headers.

If a `heading` prop is omitted, the default heading markup will not render at all. I threw in some dev-only console warnings (or they could be errors) if the user skips the `heading` _and_ the new composable components.

### DIFF
--- a/packages/react/src/components/Dialog/index.tsx
+++ b/packages/react/src/components/Dialog/index.tsx
@@ -132,7 +132,7 @@ const Dialog = forwardRef<HTMLDivElement, DialogProps>(
     useEffect(() => {
       if (show && !heading && dialogRef.current) {
         const hasHeading = dialogRef.current.querySelector('.Dialog__heading');
-        if (!hasHeading) {
+        if (process.env.NODE_ENV !== 'production' && !hasHeading) {
           throw Error(
             'Dialog: No heading provided. When using a custom header, include a DialogHeading component for accessibility.'
           );


packages/react/src/components/Dialog/index.tsx
```diff
@@ -124,18 +150,14 @@ const Dialog = forwardRef<HTMLDivElement, DialogProps>(
...
-        : 2
-    }` as 'h1';
+    const contextValue: DialogContextValue = {
+      headingId,
+      headingRef,
+      headingLevel,
+      onClose: handleClose,
+      forceAction,
+      closeButtonText
+    };
```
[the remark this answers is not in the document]
frankensteinke: Probably a good idea. Done.

packages/react/src/components/Dialog/index.tsx
```diff
@@ -148,22 +186,21 @@ const Dialog = forwardRef<HTMLDivElement, DialogProps>(
...
+                <DialogHeader>
+                  <DialogHeading>
+                    {typeof heading === 'object' && 'text' in heading
+                      ? heading.text
+                      : heading}
+                  </DialogHeading>
+                  <DialogCloseButton />
+                </DialogHeader>
+              ) : null}
+              {children}
```
anastasialanz, line 196: Would we want someone to use both the `heading` prop and  `children` or only use one?

frankensteinke: We'd want to support both, because `{children}` here also includes the other dialog markup that can be rendered like `<DialogContent />` and `<DialogFooter />`.

anastasialanz: Disregard. I thought the children was for the header!

packages/react/src/components/Dialog/index.tsx
```diff
@@ -113,6 +129,36 @@ const Dialog = forwardRef<HTMLDivElement, DialogProps>(
       initialFocusElement: headingRef
     });
 
+    useEffect(() => {
+      if (show && !heading && dialogRef.current) {
+        const hasHeading = dialogRef.current.querySelector('.Dialog__heading');
+        if (!hasHeading) {
+          throw Error(
```
anastasialanz, line 136: Would we want this to throw when `process.env.NODE_ENV !== 'production'`?

frankensteinke: Sounds good to me! Updated: https://github.com/dequelabs/cauldron/pull/2213/changes/b556f4cf7890f681df9aa39c38b194242ad62d09

Issue #1982, opened 2025-06-17
Remove Button min-width


scurker:

Bracciata:
Should this also remove the following two lines?

```
.Button--condensed {
  min-width: var(--button-height);
  ...
}

.Button--condensed:is(.Button--thin) {
  min-width: var(--button-thin-height);
}
```

PR #2214, opened 2026-01-22
fix: prevent button text wrapping using max-content for button min-width

### LINKED ISSUE #2190: Bug: Cauldron button labels should not wrap
### Steps to reproduce

Buttons contained in a table or in some constrained width starts to cause wrapping of the label of the button.
button with wrapped text

### Expected Behavior

Buttons should not allow for labels to wrap and have a fixed minwidth

Chrome

Bracciata:
Closes #2190 and https://github.com/dequelabs/walnut/issues/13150 #1982

Updates buttons so they do not wrap unless overridden to wrap. Prevents button text from wrapping and no longer makes 6.25rem the minimum width for all buttons.

QA Steps:
Navigate to a page with a button
Shrink the width of the page until it is less than the width of the button
Verify the button doesn't wrap and instead overflows the page.


Bracciata:
> @Bracciata I am seeing that there used to be a max-width of 6.25rem on the button. The change you have implemented is what we would like from design but I am not sure if this will have a cascading effect across our products. Anyway to confirm if this is a breaking change or not?

There is a good chance it is a breaking change within some repo. Thoughts on doing min-width: min(max-content, 6.25rem)?

@shawnsharpDQ

shawnsharpDQ (approved):
2 birds with 1 stone! Love it

gupta-tanvi26:
Wondering if we want to implement this change across across other components too? Like the description list (specifically, I notice that the content wraps around for the description list details)

PR #2228, opened 2026-02-04
fix: remove dialog close button warning for prop-based dialogs

frankensteinke:
Related: https://github.com/dequelabs/axe-extension/issues/7457

`<DialogCloseButton>` will emit a warning whenever a prop-based modal is created with `forceAction` set to true. The warning is still helpful if the user creates a compound header and doesn't include a close button, but this update will get rid of the warning in all other cases.

### DIFF
--- a/packages/react/src/components/Dialog/Dialog.test.tsx
+++ b/packages/react/src/components/Dialog/Dialog.test.tsx
@@ -189,6 +189,22 @@ test('should not render close button when forceAction is true', () => {
   ).not.toBeInTheDocument();
 });
 
+test('should not warn when using built-in heading with forceAction', () => {
+  const consoleWarn = jest
+    .spyOn(console, 'warn')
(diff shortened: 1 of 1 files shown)


packages/react/src/components/Dialog/index.tsx
```diff
@@ -190,7 +190,7 @@ const Dialog = forwardRef<HTMLDivElement, DialogProps>(
                       ? heading.text
                       : heading}
                   </DialogHeading>
-                  <DialogCloseButton />
+                  {!forceAction ? <DialogCloseButton /> : null}
```
chornonoh-vova, line 193: Change looks good.
Can you add a test case that'll verify this scenario?

frankensteinke: @chornonoh-vova done, thanks for pointing it out!

PR #2248, opened 2026-02-11
fix(ActionMenu): correct action menu scrolling on open

Bracciata:
Closes https://github.com/dequelabs/walnut/issues/13307

This fix prevents a bug where when there was a large list of ActionMenus in one container on first open it may scroll to the top of the page.

This bug be reproduced by reverting the changes in this pr and scrolling down to the bottom of a page rendered with this MDX
````jsx example
function ScrollPreservedExample() {
  const projects = [
    { id: '1', name: 'Project One', lastRun: '2 hours ago' },
    { id: '2', name: 'Project Two', lastRun: '1 day ago' },
    { id: '3', name: 'Project Three', lastRun: '3 days ago' },
    { id: '4', name: 'Project Four', lastRun: '1 week ago' },
    { id: '5', name: 'Project Five', lastRun: '2 weeks ago' },
    { id: '6', name: 'Project Six', lastRun: '3 weeks ago' },
    { id: '7', name: 'Project Seven', lastRun: '1 month ago' },
    { id: '8', name: 'Project Eight', lastRun: '2 months ago' },
    { id: '9', name: 'Project Nine', lastRun: '3 months ago' },
    { id: '10', name: 'Project Ten', lastRun: 'Never' },
  ];
  return (
    <div style={{ padding: '0 1rem', minHeight: '200vh' }}>
      <div style={{ marginBottom: '2rem' }}>
        <strong>Scroll down to a row near the bottom, then open Settings.</strong> If the bug is present, the page may jump.
      </div>
      <Panel>
        <Table>
          <TableHead>
            <TableRow>
              <TableHeader scope="col">Project Name</TableHeader>
              <TableHeader scope="col">Last Run</TableHeader>
              <TableHeader scope="col">Actions</TableHeader>
            </TableRow>
          </TableHead>
          <TableBody>
            {projects.map((project) => (
              <TableRow key={project.id}>
                <TableHeader scope="row" variant="cell">
                  {project.name}
                </TableHeader>
                <TableCell>{project.lastRun}</TableCell>
                <TableCell>
                  <div style={{ display: 'flex', gap: '8px', alignItems: 'center' }}>
                    <Link href={`#project-${project.id}`}>Open project</Link>
                    <ActionMenu
                      trigger={
                        <Button
                          variant="secondary"
                          aria-label={`Settings for ${project.name}`}
                        >
                          <Icon type="gears" />
                          Settings
                          <Icon type="chevron-down" />
                        </Button>
                      }
                    >
                      <ActionList>
                        <ActionListLinkItem href={`#configure-${project.id}`}>
                          Configure project
                        </ActionListLinkItem>
                        <ActionListLinkItem href={`#members-${project.id}`}>
                          Members
                        </ActionListLinkItem>
                      </ActionList>
                    </ActionMenu>
                  </div>
                </TableCell>
              </TableRow>
            ))}
          </TableBody>
        </Table>
      </Panel>
    </div>
  );
}`

### CODE REVIEW

File: packages/react/src/components/ActionMenu/ActionMenu.tsx [hunk 1]
```diff
@@ -139,7 +139,15 @@ const ActionMenu = forwardRef<HTMLElement, ActionMenuProps>(
 
     useEffect(() => {
       if (open) {
-        actionMenuListRef.current?.focus();
+        // Use double requestAnimationFrame to ensure layout is complete.
+        // The first RAF schedules work for the next frame, the second ensures
+        // the browser has actually completed the layout pass.
+        // This prevents scroll jumping when opening ActionMenus.
+        requestAnimationFrame(() => {
+          requestAnimationFrame(() => {
```
Bracciata, line 147: It does not work with a singular `requestAnimationFrame`. This is a documented practice in some spots, including the first note on https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -152,6 +152,13 @@ test('should tab to next element when using action list items', async () => {
   const menuTriggerButton = screen.getByRole('button', { name: 'Trigger' });
   const tabButton = screen.getByRole('button', { name: 'Tab to me' });
   await user.click(menuTriggerButton);
+
+  // Wait for ActionList to be focused after double requestAnimationFrame
```
Zidious: Each test is testing specific behaviour - we don't need to assert in every case that it has focus it doesn't provide anymore value, it makes existing unit tests harder to debug ([see docs](https://github.com/dequelabs/results-team/blob/main/policies/unit-testing.md#single-use-case-testing))

packages/react/src/components/ActionMenu/ActionMenu.tsx
Zidious: Are you able to provide the steps you used to repo the issue (or PR description) that this fixes?

Bracciata: Added

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
Bracciata: I simplified it but a wait is still neeeded

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -152,7 +152,16 @@ test('should tab to next element when using action list items', async () => {
   const menuTriggerButton = screen.getByRole('button', { name: 'Trigger' });
   const tabButton = screen.getByRole('button', { name: 'Tab to me' });
   await user.click(menuTriggerButton);
-  await user.keyboard('{Tab}');
+
+  // Wait for menu to be ready for keyboard interaction
```
Zidious: Having to add this to every single unit test seems like an unwanted side effect of adding `requestAnimationFrame` - I'm not very familiar with `requestAnimationFrame` and it's implications to give a alternative.

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -139,7 +139,15 @@ const ActionMenu = forwardRef<HTMLElement, ActionMenuProps>(
 
     useEffect(() => {
       if (open) {
-        actionMenuListRef.current?.focus();
+        // Use double requestAnimationFrame to ensure layout is complete.
```
Zidious, line 142: Is there a way we can add a unit test for this?

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
Bracciata: Swapped to a helper. We looked at alternatives and the only one that we saw that would likely also work was a timeout based approach.

packages/react/src/components/ActionMenu/ActionMenu.tsx
Bracciata: Adding it. Also swapping to a helper pattern in the tests

PR #2256, opened 2026-02-24
fix: href in ActionListLinkItem not working inside TopBar+ActionMenu

chornonoh-vova:
# Fix: Action menu links not navigating when clicked

## Problem

`ActionListLinkItem` elements inside an `ActionMenu` were not navigating to their `href` when clicked. The `onBlur` handler added to `AnchoredOverlay` would fire before the click event could complete, closing the menu and removing the link from the DOM.

## Solution

Updated `handleOverlayBlur` to check `event.relatedTarget` before deciding whether to close the menu:

- Focus moves to the trigger — return early and let `handleTriggerClick` handle closing, preventing a double-close that was causing the menu to re-open and navigate to the first link
- Focus moves inside the overlay — return early so link clicks can complete before the menu closes
- Focus moves anywhere else — close the menu as normal (tabbing out, clicking outside, etc.)

## Testing

Added two new tests to cover the affected scenarios:

- Verifies that clicking an `ActionListLinkItem` navigates to the correct `href`
- Verifies that the menu closes when focus moves outside the overlay (via tab)

### DIFF
--- a/packages/react/src/components/ActionMenu/ActionMenu.test.tsx
+++ b/packages/react/src/components/ActionMenu/ActionMenu.test.tsx
@@ -31,6 +31,17 @@ const defaultProps: React.ComponentProps<typeof ActionMenu> = {
   )
 };
 
+const defaultLinkProps: React.ComponentProps<typeof ActionMenu> = {
+  trigger: <button>Trigger</button>,
+  children: (
+    <ActionList>
+      <ActionListLinkItem href="#target-1">Menu Link 1</ActionListLinkItem>
+      <ActionListLinkItem href="#target-2">Menu Link 2</ActionListLinkItem>
+      <ActionListLinkItem href="#target-3">Menu Link 3</ActionListLinkItem>
+    </ActionList>
+  )
+};
+
 // From the pattern documented in docs/pages/components/TopBar.mdx
 async function waitForMenuFocus() {
   const menu = screen.getByRole('menu');
@@ -684,6 +695,18 @@ test('should navigate to href when an action link item is clicked in TopBar+Acti
   expect(window.location.hash).toBe('#target-1');
 });
 
+test('should navigate to href when an action link item is clicked', async () => {
+  const user = userEvent.setup();
+  render(<ActionMenu {...defaultLinkProps} />);
+
+  await user.click(screen.getByRole('button', { name: 'Trigger' }));
+  expect(await screen.findByRole('menu')).toBeVisible();
+
+  await user.click(screen.getByRole('menuitem', { name: 'Menu Link 1' }));
+
+  expect(window.location.hash).toBe('#target-1');
+});
+
 test('should close menu when focus moves outside in TopBar+ActionMenu pattern', async () => {
   const user = userEvent.setup();
   render(
@@ -700,16 +723,38 @@ test('should close menu when focus moves outside in TopBar+ActionMenu pattern', 
   await user.click(screen.getByRole('menuitem', { name: 'Trigger' }));
   expect(await screen.findByRole('menu')).toBeVisible();
 
-  await user.tab(); // Menu Link 1
-  await user.tab(); // Menu Link 2
-  await user.tab(); // Menu Link 3
-  await user.tab(); // Outside button
+  await waitForMenuFocus();
+  await user.tab();
 
-  expect(screen.getByRole('button', { name: 'Outside' })).toHaveFocus();
+  expect(await screen.findByRole('button', { name: 'Outside' })).toHaveFocus();
   expect(await screen.findByRole('menu', { hidden: true })).not.toBeVisible();
+  expect(screen.getByRole('menuitem', { name: 'Trigger' })).toHaveAttribute(
+    'aria-expanded',
+    'false'
+  );
+});
 
-  // Additionally, make sure that any of the links weren't triggered
-  expect(window.location.hash).toBe('');
+test('should close menu when focus moves outside in action menu with link items', async () => {
+  const user = userEvent.setup();
+  render(
+    <>
+      <ActionMenu {...defaultLinkProps} />
+      <button>Outside</button>
+    </>
+  );
+
+  await user.click(screen.getByRole('button', { name: 'Trigger' }));
+  expect(await screen.findByRole('menu')).toBeVisible();
+
+  await waitForMenuFocus();
+  await user.tab();
+
+  expect(await screen.findByRole('button', { name: 'Outside' })).toHaveFocus();
+  expect(await screen.findByRole('menu', { hidden: true })).not.toBeVisible();
+  expect(screen.getByRole('button', { name: 'Trigger' })).toHaveAttribute(
+    'aria-expanded',
+    'false'
+  );
 });
 
 test('should support className prop', () => {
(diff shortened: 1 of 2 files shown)


packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -118,7 +118,19 @@ const ActionMenu = forwardRef<HTMLElement, ActionMenuProps>(
...
+    const handleOverlayBlur = useCallback((event: React.FocusEvent) => {
+      const relatedTarget = event.relatedTarget;
+
+      // If focus moves to the trigger, let the trigger handle closing
+      if (triggerRef.current?.contains(relatedTarget)) {
+        return;
+      }
+
+      // If focus stays inside the overlay (e.g. clicking a link), don't close prematurely
+      if (actionMenuRef.current?.contains(relatedTarget)) {
```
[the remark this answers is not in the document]
Bracciata: Agree with this and the rest looks good!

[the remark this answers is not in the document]
Bracciata: Also please update the description

maksym-shynkarenko (approved):

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -617,6 +621,52 @@ test('should not trigger action link items when toggling the open state in TopBa
...
+
+  await user.click(screen.getByRole('menuitem', { name: 'Trigger' }));
+  expect(await screen.findByRole('menu')).toBeVisible();
+
+  await user.click(screen.getByRole('menuitem', { name: 'Menu Link 1' }));
+
+  expect(window.location.hash).toBe('#target-1');
+});
+
+test('should close menu when focus moves outside in TopBar+ActionMenu pattern', async () => {
```
anastasialanz, line 710: The blur fix here is really a general `ActionMenu` improvement, but the new tests only exercise the `TopBar` pattern. Worth adding a test for a plain `ActionMenu` with `ActionListLinkItem` too so it's clear this isn't just a `TopBar` thing? You can see it not working [here](https://cauldron.dequelabs.com/components/ActionMenu#links) with the `ActionMenu` by itself.

packages/react/src/components/ActionMenu/ActionMenu.tsx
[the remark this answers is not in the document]
anastasialanz: Copilot's suggestion is incorrect because `handleClickOutside` at lines 114-115 cast to `HTMLElement` and not `Node` so it's already consistent.

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -617,6 +621,52 @@ test('should not trigger action link items when toggling the open state in TopBa
...
+    </>
+  );
+
+  await user.click(screen.getByRole('menuitem', { name: 'Trigger' }));
+  expect(await screen.findByRole('menu')).toBeVisible();
+
+  await user.tab(); // Menu Link 1
+  await user.tab(); // Menu Link 2
+  await user.tab(); // Menu Link 3
+  await user.tab(); // Outside button
```
anastasialanz: If you were to verify this test in the browser, you'd notice that after clicking the trigger, pressing tab does not navigate through each menu item like the comments say. The first tab leaves the menu and goes to the outside button. The up and down arrows are how you navigate the menu links. I think we can keep a single tab here and remove three of them.

https://www.w3.org/WAI/ARIA/apg/patterns/menubar/

> When a menu opens, or when a menubar receives focus, keyboard focus is placed on the first item. Because menubar and menu elements are composite widgets as described in the practice for [Keyboard Navigation Inside Components](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_general_within), Tab and Shift + Tab do not move focus among the items in the menu.

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -617,6 +621,52 @@ test('should not trigger action link items when toggling the open state in TopBa
...
+
+  expect(screen.getByRole('button', { name: 'Outside' })).toHaveFocus();
+  expect(await screen.findByRole('menu', { hidden: true })).not.toBeVisible();
+
+  // Additionally, make sure that any of the links weren't triggered
+  expect(window.location.hash).toBe('');
```
anastasialanz: We don't click on a link in this test so I don't think this is needed.

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -617,6 +621,52 @@ test('should not trigger action link items when toggling the open state in TopBa
...
+  await user.click(screen.getByRole('menuitem', { name: 'Trigger' }));
+  expect(await screen.findByRole('menu')).toBeVisible();
+
+  await user.tab(); // Menu Link 1
+  await user.tab(); // Menu Link 2
+  await user.tab(); // Menu Link 3
+  await user.tab(); // Outside button
+
+  expect(screen.getByRole('button', { name: 'Outside' })).toHaveFocus();
+  expect(await screen.findByRole('menu', { hidden: true })).not.toBeVisible();
```
anastasialanz, line 730: nit: We could also verify that `aria-expanded` is set to `false` which is more specific.

`expect(screen.getByRole('menuitem', { name: 'Trigger' })).toHaveAttribute('aria-expanded', 'false');`

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -118,7 +118,19 @@ const ActionMenu = forwardRef<HTMLElement, ActionMenuProps>(
       }
     }, []);
 
-    const handleOverlayBlur = useCallback(() => {
+    const handleOverlayBlur = useCallback((event: React.FocusEvent) => {
+      const relatedTarget = event.relatedTarget as HTMLElement | null;
```
anastasialanz, line 122: Couldn't we simplify this a bit to be identical to `handleClickOutside`?

```
const handleOverlayBlur = useCallback((event: React.FocusEvent) => {
  if (
    !triggerRef.current?.contains(event.relatedTarget as HTMLElement) &&
    !actionMenuRef.current?.contains(event.relatedTarget as HTMLElement)
  ) {
    setOpen(false);
  }
}, []);
```

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
```diff
@@ -662,6 +677,88 @@ test('should not trigger action link items when toggling the open state in TopBa
...
+      </TopBar>
+      <button>Outside</button>
+    </>
+  );
+
+  await user.click(screen.getByRole('menuitem', { name: 'Trigger' }));
+  expect(await screen.findByRole('menu')).toBeVisible();
+
+  await user.tab();
+  await user.tab();
```
anastasialanz, line 727: We only need the one tab here.

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
anastasialanz: This one still isn't addressed.

packages/react/src/components/ActionMenu/ActionMenu.test.tsx
chornonoh-vova: Unfortunately, it's not working with one tab

anastasialanz: RTL is supposed to mimic user behavior so we should figure out what is wrong.

chornonoh-vova: Fixed

PR #2272, opened 2026-03-12
fix: sync danger & danger-secondary variant disabled state

chornonoh-vova:
# Fix: Sync danger & danger-secondary variant disabled state

## Summary

- Fixes the disabled state styling for `Button--danger` and `Button--danger-secondary` variants to use proper design tokens instead of hardcoded values
- Adds missing disabled state rules for `danger` and `danger-secondary` button variants, aligning them with the existing `error` variant behavior


e2e/screenshots/dark--button-thin-variant-danger-.png
anastasialanz, line 1: The updated disabled state contrast is a little harsh on the eyes.

PR #2288, opened 2026-03-26
feat(AnchoredOverlay): add disable flip property

Bracciata:
Relates to https://github.com/dequelabs/launchpad/issues/1321

Adds a property to disable flipping (trying to keep element on screen by switching position).


docs/pages/components/AnchoredOverlay.mdx
```diff
@@ -133,6 +133,51 @@ function AnchoredOverlayWithOffsetExample() {
 }
 ```
 
+### Disable Flip
```
Zidious: I'm not a huge fan of the naming here, it's referred to as `placement` through other usages, so maybe `disableAutoPlacement`? I won't die on a hill for it, so I'll get Cauldron folks weigh in.

docs/pages/components/AnchoredOverlay.mdx
```diff
@@ -133,6 +133,51 @@ function AnchoredOverlayWithOffsetExample() {
...
+
+When `disableFlip` is set, the overlay will not flip to an alternative placement when space is limited. The overlay will remain at its configured `placement`, though it may still shift along its axis to stay within clipping boundaries.
+
+<Note>
+  Use `disableFlip` with caution. When flipping is disabled, the overlay may render partially or fully outside the visible viewport if there is not enough space at the configured placement. This can create accessibility issues where content is announced by screen readers but not visible to sighted users. This also disables auto placement and top-overflow prevention.
+</Note>
```
Zidious: Will this be a problem? How do we want to handle this if it actually becomes an issue?

Zidious (dismissed):
Had a question inline otherwise LGTM, I'll get other folks review.

docs/pages/components/AnchoredOverlay.mdx
Bracciata: Update to show action plan. This is the only way to meet desired design for a dropdown used. We could name it as `dangerouslyDisableAutoPlacement`

docs/pages/components/AnchoredOverlay.mdx
Bracciata: Updated!

docs/pages/components/AnchoredOverlay.mdx
```diff
@@ -133,6 +133,51 @@ function AnchoredOverlayWithOffsetExample() {
...
+### Disable Auto Placement
+
+When `disableAutoPlacement` is set, the overlay will not automatically reposition itself when space is limited. The overlay will remain at its configured `placement`, though it may still shift along its axis to stay within clipping boundaries.
+
+<Note>
+  Use `disableAutoPlacement` with caution. When automatic placement is disabled, the overlay may render partially or fully outside the visible viewport if there is not enough space at the configured placement. Consumers of this prop should ensure that the configured placement has sufficient space for the overlay content, or implement their own overflow handling to avoid content being hidden from sighted users while still being announced by screen readers.
+</Note>
+
+```jsx example
+function AnchoredOverlayDisableFlipExample() {
```
chornonoh-vova: Should this be renamed?


```suggestion
function AnchoredOverlayDisableAutoPlacementExample() {
```

packages/react/src/components/AnchoredOverlay/index.tsx
```diff
@@ -119,18 +122,20 @@ const AnchoredOverlay = forwardRef(
...
         offsetMiddleware(offset ?? 0),
-        initialPlacement.startsWith('auto')
-          ? autoPlacementMiddleware({
-              alignment: getAutoAlignment(initialPlacement as 'auto')
-            })
-          : flipMiddleware({
-              fallbackAxisSideDirection: 'start'
-            }),
+        disableAutoPlacement
+          ? null
```
chornonoh-vova, line 139: There's a possibility that `disableAutoPlacement` and `placement="auto"` will be in the props at the same time 😅

In this case the position would be forced to `top` (line 122).

Maybe there should be some kind of console warning in this case?

chornonoh-vova (dismissed):

docs/pages/components/AnchoredOverlay.mdx
```diff
@@ -133,6 +133,51 @@ function AnchoredOverlayWithOffsetExample() {
 }
 ```
 
+### Disable Auto Placement
+
+When `disableAutoPlacement` is set, the overlay will not automatically reposition itself when space is limited. The overlay will remain at its configured `placement`, though it may still shift along its axis to stay within clipping boundaries.
+
+<Note>
```
anastasialanz:
```suggestion
<Note variant="warning">
```

Can we make this stand out more visually since this could cause an a11y issue if used improperly?

packages/react/src/components/AnchoredOverlay/index.tsx
```diff
@@ -112,25 +115,37 @@ const AnchoredOverlay = forwardRef(
     const ref = useSharedRef<HTMLElement | null>(refProp);
     const Component = as || 'div';
 
+    if (
+      disableAutoPlacement &&
+      typeof initialPlacement === 'string' &&
+      initialPlacement.startsWith('auto')
+    ) {
+      console.warn(
```
anastasialanz: Can we log this in dev only and in a useEffect? This will warn in production.

docs/pages/components/AnchoredOverlay.mdx
```diff
@@ -133,6 +133,51 @@ function AnchoredOverlayWithOffsetExample() {
 }
 ```
 
+### Disable Auto Placement
```
anastasialanz, line 136: Just curious, what other changes did we try in Launchpad to fix this before making this change here?

chornonoh-vova (approved):

Bracciata:
Closing in favor of a Launchpad change.

PR #2291, opened 2026-03-30
fix(TableHeader): remove sort announcement props as no needed and update aria-sort handling

Bracciata:
## Summary

Fix NVDA screen reader announcing sort state on every table row cell.

**Problem:** 
NVDA was reading "Rule sorted ascending" (or similar) for every cell in a sortable column, not just the column header.

For example:
row 2  RuleRule sorted ascending  column 1  link   and elements...
row 3  RuleRule sorted ascending  column 1  link  All page content...

**Root cause:** 
`TableHeader` rendered an `Offscreen` live region
(`role="status" aria-live="polite"`) inside the sort `<button>`. Because it
was inside the button, screen readers included the announcement text in the
column header's accessible name. NVDA then repeated that name for every cell
in the column.

**Fix:**
Removed the `Offscreen` live region entirely. Sort state is now
communicated solely via the `aria-sort` attribute on `<th>`, which is the
standard ARIA mechanism and is well-supported in modern screen readers (NVDA, JAWS, VoiceOver).

## Changes

- `packages/react/src/components/Table/TableHeader.tsx` — removed `Offscreen`
  live region from inside the sort button; marked `sortAscendingAnnouncement`
  and `sortDescendingAnnouncement` props as `@deprecated` for backwards compatibility

- `packages/react/src/components/Table/index.test.tsx` — updated three tests
  to assert on `aria-sort` attribute instead of the removed live region content

## Before / After (NVDA)

**Before:**
row 2  RuleRule sorted ascending  column 1  link  ...
row 3  RuleRule sorted ascending  column 1  link  ...

**After:**
row 1  column 1  sorted ascending  Rule  button
row 2  First Name  column 1  Frank
row 3  First Name  column 1  Jimmy

Closes: [Walnut #13951](https://github.com/dequelabs/walnut/issues/13951), [Walnut #14061](https://github.com/dequelabs/walnut/issues/14061)


packages/react/src/components/Table/index.test.tsx
```diff
@@ -458,6 +455,50 @@ test('returns 0 axe violations with descending sorting', async () => {
...
+  expect(screen.getByRole('status')).toHaveTextContent('sorted ascending');
+
+  // The live region should be rendered outside the table (in document.body),
+  // not inside the column header, so it is not part of the header's accessible name
+  const columnHeader = screen.getByRole('columnheader');
+  expect(columnHeader).toHaveAttribute('aria-sort', 'ascending');
+
+  const liveRegion = screen.getByRole('status');
+  expect(columnHeader.contains(liveRegion)).toBe(false);
+
```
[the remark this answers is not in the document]
Bracciata: Good call — the DOM containment check was a proxy at best. I'll look into adding a direct `toHaveAccessibleName` assertion here to make the test actually prove what the title claims.

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -56,6 +87,8 @@ const TableHeader = forwardRef<HTMLTableHeaderCellElement, TableHeaderProps>(
           ? sortDescendingAnnouncement
           : '';
 
+    useSortAnnouncement(announcement);
+
```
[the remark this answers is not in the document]
Bracciata: Addressed — the `useSortAnnouncement` hook has been removed entirely. The live region is now a single `createPortal` rendered at the `Table` level via `TableContext`, so non-sortable headers no longer create any DOM. Each `TableHeader` just calls `setSortAnnouncement()` from context when its sort direction changes.

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -21,6 +22,36 @@ interface TableHeaderProps
...
+    liveRegion.setAttribute('role', 'status');
+    liveRegion.setAttribute('aria-live', 'polite');
+    container.appendChild(liveRegion);
+    document.body.appendChild(container);
+    liveRegionRef.current = liveRegion;
+
+    return () => {
+      document.body.removeChild(container);
+    };
+  }, []);
```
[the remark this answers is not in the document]
Bracciata: Agreed — this was the biggest issue. Replaced the imperative DOM approach with `createPortal` + the existing `<Offscreen>` component, matching the pattern already used by `CopyButton`. The live region is now lifted to the `Table` component via `TableContext`, so there's exactly one `role="status"` element per table regardless of column count. React manages the lifecycle automatically.

packages/react/src/components/Table/index.test.tsx
```diff
@@ -204,7 +204,6 @@ test('should render sort button and icons with sortDirection and onSort in Table
   );
 
   expect(screen.getByRole('button')).toBeInTheDocument();
-  expect(screen.getByRole('status').closest('.Icon--sort-triangle'));
   expect(screen.getByRole('status')).toHaveTextContent('');
```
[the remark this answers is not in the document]
Bracciata: Fair point. Worth noting that the original assertions (`screen.getByRole('status').closest('.Icon--sort-triangle')`) were actually no-ops — they returned a value but never asserted on it (no `expect()` wrapper). So they weren't verifying anything before either. I'll add proper icon assertions or update the test names to match what's actually being checked.

packages/react/src/components/Table/index.test.tsx
```diff
@@ -249,7 +248,6 @@ test('should render triangle up Icon and ascending message when sortDirection is
   );
 
   expect(screen.getByRole('status')).toHaveTextContent('up and away');
-  expect(screen.getByRole('status').closest('.Icon--table-sort-ascending'));
 });
```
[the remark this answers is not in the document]
Bracciata: Same situation as above — the removed `.closest()` calls were no-ops (no `expect()` wrapping them). Will add proper icon assertions or update the test names.

packages/react/src/components/Table/index.test.tsx
```diff
@@ -270,7 +268,6 @@ test('should render triangle down Icon and descending message when sortDirection
   );
 
   expect(screen.getByRole('status')).toHaveTextContent('down below');
-  expect(screen.getByRole('status').closest('.Icon--table-sort-descending'));
 });
```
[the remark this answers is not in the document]
Bracciata: Same as above — will address alongside the ascending icon test.

packages/react/src/components/Table/Table.tsx
```diff
@@ -45,6 +47,21 @@ function parseColumnWidth(width?: ColumnWidth): string {
...
+    return null;
+  }
+  return createPortal(
+    <Offscreen>
+      <span role="status" aria-live="polite">
+        {sortAnnouncement}
+      </span>
+    </Offscreen>,
+    document.body
+  ) as React.ReactPortal;
```
Zidious: Casting removes the need for TypeScript, we should avoid casting

packages/react/src/components/Table/TableContext.tsx
```diff
@@ -3,13 +3,15 @@ import React, {
...
-  useRef
+  useCallback
 } from 'react';
 import type { Column } from './Table';
 
 type TableContext = {
   layout: 'table' | 'grid';
   columns: Array<Column>;
+  sortAnnouncement: string;
+  setSortAnnouncement: (announcement: string) => void;
```
Zidious: Why would this be drilled down vs being the responsibility of the component?

Zidious (dismissed):

packages/react/src/components/Table/Table.tsx
Bracciata: It is throwing a linting error without it. Troubleshooting still.

packages/react/src/components/Table/Table.tsx
```diff
@@ -45,6 +47,20 @@ function parseColumnWidth(width?: ColumnWidth): string {
   return width;
 }
 
+function SortAnnouncementPortal(): JSX.Element | null {
+  const { text } = useSortAnnouncementState();
+  if (typeof document === 'undefined') {
```
chornonoh-vova: The `isBrowser` util can be used here instead (from `utils/is-browser`).

```suggestion
  if (!isBrowser()) {
```

The check there is more thorough, and it'll also be consistent with a rest of the codebase.

packages/react/src/components/Table/Table.tsx
```diff
@@ -45,6 +47,20 @@ function parseColumnWidth(width?: ColumnWidth): string {
...
+    return null;
+  }
+  return createPortal(
+    <Offscreen>
+      <span role="status">{text}</span>
+    </Offscreen>,
+    // eslint-disable-next-line ssr-friendly/no-dom-globals-in-react-fc
+    document.body
+  ) as React.ReactPortal;
```
chornonoh-vova: You should be able to avoid this type cast by changing the return type of the `SortAnnouncementPortal` the `React.ReactNode`.

Smth like this should work:

```ts
function SortAnnouncementPortal(): React.ReactNode {
```

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -39,22 +40,38 @@ const TableHeader = forwardRef<HTMLTableHeaderCellElement, TableHeaderProps>(
...
+    useEffect(() => {
+      if (!isSortable) return;
+
+      if (announcement) {
+        announce(ownerToken.current, announcement);
+      }
+      return () => {
+        clear(ownerToken.current);
+      };
+    }, [isSortable, announcement, announce, clear]);
```
chornonoh-vova: It's a minor thing - but it'll currently run clean even if there's no announcement.

I'd suggest something like this:

```ts
useEffect(() => {
  if (!isSortable) return;
  if (announcement) {
    announce(ownerToken.current, announcement);
    return () => {
      clear(ownerToken.current);
    };
  }
}, [isSortable, announcement, announce, clear]);
```

packages/react/src/components/Table/index.test.tsx
```diff
@@ -458,6 +464,266 @@ test('returns 0 axe violations with descending sorting', async () => {
   expect(results).toHaveNoViolations();
 });
 
+test('should not include sort announcement text in the column header accessible name', () => {
```
chornonoh-vova: There's a potentially missing test: for the tables with only non-sortable headers. Maybe it'll be better to make a test case that will make sure that when `TableHeader` is rendered without the `onSort` & `sortDirection` props, the `announce` and `clear` aren't called.

chornonoh-vova (commented):
Overall looks good, left a couple of minor comments for your consideration.

packages/react/src/components/Table/Table.tsx
Bracciata: Added a comment as to why a cast was needed. It also matches the dialog example

chornonoh-vova (dismissed):

packages/react/src/components/Table/Table.tsx
```diff
@@ -45,6 +48,22 @@ function parseColumnWidth(width?: ColumnWidth): string {
   return width;
 }
 
+function SortAnnouncementPortal(): React.ReactPortal | null {
```
denysfedorov: this might be a good case to update types
```ts
    "@types/react": "^18.0.12",
    "@types/react-dom": "^18.0.5",
```

packages/react/src/components/Table/TableContext.tsx
```diff
@@ -3,45 +3,104 @@ import React, {
...
 }
 
-export { TableProvider, useTable };
+function useSortAnnouncementActions(): SortAnnouncementActions {
+  return useContext(SortAnnouncementActionsContext);
+}
+
+function useSortAnnouncementState(): SortAnnouncementState {
+  return useContext(SortAnnouncementStateContext);
+}
```
denysfedorov: Minor suggestion: we could make the contexts nullable instead of using noop defaults. This way if a hook is accidentally used outside a `TableProvider`, it throws immediately instead of silently doing nothing (using noop). Something like this:
```tsx
const SortAnnouncementActionsContext =
  createContext<SortAnnouncementActions | null>(null);

function useSortAnnouncementActions(): SortAnnouncementActions {
  const value = useContext(SortAnnouncementActionsContext);
  if (!value) {
    throw new Error(
      'useSortAnnouncementActions must be used within a SortAnnouncementActionsContext provider'
    );
  }
  return value;
}
```

denysfedorov (commented):
LGTM! Just one small minor suggestion about contexts creation

chornonoh-vova (approved):

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -10,11 +9,15 @@ import useSharedRef from '../../utils/useSharedRef';
...
+interface TableHeaderProps extends Omit<
+  React.ThHTMLAttributes<HTMLTableHeaderCellElement>,
+  'align'
+> {
   sortDirection?: SortDirection;
   onSort?: () => void;
+  /** @deprecated No longer used. Sort state is communicated via aria-sort. */
   sortAscendingAnnouncement?: string;
+  /** @deprecated No longer used. Sort state is communicated via aria-sort. */
   sortDescendingAnnouncement?: string;
```
frankensteinke, line 21: Would we want to remove these from the docs now, or at least mark them as deprecated there as well?

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -46,15 +49,11 @@ const TableHeader = forwardRef<HTMLTableHeaderCellElement, TableHeaderProps>(
...
-      sortDirection === 'ascending'
-        ? sortAscendingAnnouncement
-        : sortDirection === 'descending'
-          ? sortDescendingAnnouncement
-          : '';
+    // Sort state is communicated via the aria-sort attribute on <th>.
+    // A live region (Offscreen) was previously used as a workaround for
+    // limited aria-sort support, but it was removed because screen readers
+    // (e.g. NVDA) included the announcement text in the column header's
+    // accessible name, causing it to be read for every cell in the column.
```
frankensteinke:
```suggestion
```

Good info to know for this PR, but not sure that we need to document it in code. As time goes on, the value of knowing the history of this is less valuable.

frankensteinke (dismissed):

packages/react/src/components/Table/TableHeader.tsx
```diff
@@ -46,15 +49,11 @@ const TableHeader = forwardRef<HTMLTableHeaderCellElement, TableHeaderProps>(
...
```
frankensteinke, line 65: Could you also throw in a deprecation notice via console warning, like https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/IconButton/index.tsx#L84-L93?

Note that this wraps the `useEffect` in a conditional, which seems bad (we'd want conditional checks to happen inside the `useEffect` instead).

packages/react/src/components/Table/TableHeader.tsx
Bracciata: Updated to note in docs as well

packages/react/src/components/Table/TableHeader.tsx
Bracciata: Updated

packages/react/src/components/Table/TableHeader.tsx
Bracciata: Updated to do so!

frankensteinke (approved):
Thanks!

PR #2296, opened 2026-04-02
fix(RadioGroup): prevent radio buttons from becoming unselectable after re-render

chornonoh-vova:
# Summary

The `ref` collection pattern used push-based assignment with a render-time reset:

```ts
inputs.current = [];
```

It ran after `radios.map()` built the JSX.

This created a race condition where the refs array could be empty when a
click handler fired, because the reset ran between React's render and
commit phases.

In Safari specifically, differences in microtask scheduling
widened this window, causing radio clicks via the Icon overlay to be
silently swallowed.

I've replaced the push-based ref collection with index-based assignment:

```ts
inputs.current[index] = input
```

and remove the render-time reset entirely.

Each radio now writes directly to its correct index, so the
array is never cleared and there is no window where refs are missing.

Also added regression tests covering icon overlay clicks after re-renders,
dynamic radio additions, and radio removals.

### DIFF
--- a/packages/styles/base.css
+++ b/packages/styles/base.css
@@ -1,7 +1,6 @@
 :root {
   --workspace-background-color: #f0f2f5;
   --heading-text-color: var(--header-text-color-dark);
-  --heading-font-family: var(--base-font-family);
 }
 
 .cauldron--theme-dark {
@@ -52,7 +51,6 @@ h5,
 h6,
 [role='heading'] {
   color: var(--heading-text-color);
-  font-family: var(--heading-font-family);
 }
 
 ul {
--- a/packages/styles/button.css
+++ b/packages/styles/button.css
@@ -295,21 +295,9 @@ button.Link {
   min-width: var(--button-thin-height);
 }
 
+/* Custom property for button sibling gap. Will be deprecated in https://github.com/dequelabs/cauldron/issues/1944 */
 [class*='Button--'] + [class*='Button--'] {
-  margin-left: var(--button-adjacent-spacing, var(--space-smallest));
-}
-
-/* Gap-based button grouping — use instead of relying on adjacent-sibling margins.
-   Handles flex-wrap correctly and avoids spacing issues in flex containers.
-   See https://github.com/dequelabs/cauldron/issues/1944 */
-.ButtonGroup {
-  display: flex;
-  flex-wrap: wrap;
-  gap: var(--button-group-gap, var(--space-smallest));
-}
-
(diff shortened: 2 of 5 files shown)


packages/react/src/components/RadioGroup/index.tsx
```diff
@@ -83,11 +86,9 @@ const RadioGroup = forwardRef(
               value={radioValue}
               id={id}
               ref={(input) => {
-                if (!input) {
-                  return;
+                if (input) {
+                  inputs.current[index] = input;
```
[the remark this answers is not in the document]
chornonoh-vova: The React should be calling the ref function in case of reordering. Added a test case to verify that it's working.

packages/react/src/components/RadioGroup/index.tsx
```diff
@@ -8,13 +8,15 @@ export interface RadioItem extends React.InputHTMLAttributes<HTMLInputElement> {
...
+export interface RadioGroupProps extends Omit<
+  React.HTMLAttributes<HTMLDivElement>,
+  'onChange'
+> {
   name?: string;
   className?: string;
   radios: RadioItem[];
   defaultValue?: string;
-  value?: any;
+  value?: string;
```
Zidious, line 19: Even though this is optional, this could be classified as breaking change going from `any` to `string` are we OK with that?

markreay: I went back and forth on this a bit and landed here.

Yes, it could break a caller's type assertions, but if anyone was passing value={42} or something, they'd want to know that's a problem. My feel is to allow this.

markreay (approved):

packages/react/src/components/RadioGroup/RadioGroup.test.tsx
```diff
@@ -229,3 +229,177 @@ test('should have no axe violations with radio item and labelDescription', async
...
+
+    // Initially 3 radios
+    expect(screen.getAllByRole('radio')).toHaveLength(3);
+
+    // Add a 4th radio
+    await user.click(screen.getByRole('button', { name: /Add option/ }));
+    const inputs = screen.getAllByRole('radio') as HTMLInputElement[];
+    expect(inputs).toHaveLength(4);
+
+    // Click the newly added radio's icon overlay
```
anastasialanz: Can we update this to click the label text or radio which is the way a user actually selects a radio instead of using implementation details like the class name?

chornonoh-vova: Updated

PR #2301, opened 2026-04-03
feat: add Switch component

ezirlingerDQ:
## Summary
- Add a new accessible `Switch` toggle component with full ARIA `role="switch"` support
- Include controlled/uncontrolled state management, disabled state, error messages, and label descriptions
- Add CSS styles with dark/light theme support via custom properties
- Add comprehensive test suite and MDX documentation with live examples

## Test plan
- [ ] Run `yarn test` and verify all Switch tests pass
- [ ] Run `yarn dev` and verify the Switch docs page renders at `/components/Switch`
- [ ] Test keyboard navigation (Space/Enter to toggle, Tab to focus)
- [ ] Verify screen reader announces switch state correctly
- [ ] Check dark and light theme rendering

### DIFF
--- a/packages/react/src/components/Switch/index.ts
+++ b/packages/react/src/components/Switch/index.ts
@@ -1,0 +1,2 @@
+export { default, Switch } from './Switch';
+export type { SwitchProps } from './Switch';
--- a/packages/styles/switch.css
+++ b/packages/styles/switch.css
@@ -1,0 +1,176 @@
+:root {
+  --switch-track-color: var(--gray-90);
+  --switch-track-color-checked: var(--accent-primary);
+  --switch-track-color-disabled: var(--gray-40);
+  --switch-track-color-checked-disabled: var(
+    --accent-primary-disabled,
+    #78a6d8
+  );
+  --switch-handle-color: var(--gray-90);
+  --switch-handle-color-disabled: var(--gray-40);
+  --switch-focus-color: var(--focus-light);
+  --switch-label-color: var(--gray-90);
+}
+
+.cauldron--theme-dark {
+  --switch-track-color: var(--white);
+  --switch-track-color-checked: var(--accent-light);
+  --switch-track-color-checked-disabled: var(--stroke-dark);
+  --switch-handle-color: var(--white);
+  --switch-handle-color-disabled: var(--stroke-dark);
+  --switch-focus-color: var(--focus-dark);
+  --switch-label-color: var(--white);
+}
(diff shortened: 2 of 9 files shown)


packages/styles/switch.css
```diff
@@ -0,0 +1,176 @@
...
+
+.Switch__track {
+  display: flex;
+  align-items: center;
+  justify-content: flex-start;
+  width: 44px;
+  height: 24px;
+  border: 2px solid var(--switch-track-color);
+  border-radius: 999px;
+  padding: 4px;
```
anastasialanz, line 60: This should be `var(--space-half)` for 4px.

packages/styles/switch.css
```diff
@@ -0,0 +1,176 @@
+:root {
+  --switch-track-color: var(--gray-90);
+  --switch-track-color-checked: var(--accent-primary);
+  --switch-track-color-disabled: var(--gray-40);
+  --switch-track-color-checked-disabled: var(
+    --accent-primary-disabled,
+    #78a6d8
+  );
```
anastasialanz, line 8: `#78a6d8` is the same as `--accent-primary-disabled` which seems unnecessary.

packages/styles/switch.css
```diff
@@ -0,0 +1,176 @@
+:root {
+  --switch-track-color: var(--gray-90);
+  --switch-track-color-checked: var(--accent-primary);
+  --switch-track-color-disabled: var(--gray-40);
```
anastasialanz, line 4: Shouldn't this be `var(--disabled)` instead?

packages/styles/switch.css
```diff
@@ -0,0 +1,176 @@
...
+  --switch-track-color-checked-disabled: var(
+    --accent-primary-disabled,
+    #78a6d8
+  );
+  --switch-handle-color: var(--gray-90);
+  --switch-handle-color-disabled: var(--gray-40);
```
anastasialanz, line 10: Shouldn't this be `var(--disabled)` instead?

packages/react/src/components/Switch/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default, Switch } from './Switch';
```
anastasialanz, line 1: These exports are redundant and should probably follow how we do it in other components like `export { default as ActionMenu } from './ActionMenu';`

```suggestion
export { default as Switch } from './Switch';
```

PR #2303, opened 2026-04-06
fix(styles): replace max-width with width on Link with Icon

denysfedorov:
Closes: https://github.com/dequelabs/walnut/issues/14130

## Summary

Replace `max-width: max-content` with `width: fit-content` on `.Link:where(:has(.Icon))` so consumers can override the link width through normal CSS cascade rules

## Problem

`max-width` acts as a hard ceiling - any consumer setting an explicit `width` on an icon link gets capped to the icon's intrinsic size because `max-width` always wins over `width` per the CSS box model

## Fix

`width: fit-content` prevents the link from stretching (same as before) but can be overridden by downstream CSS:

```css
/* Before: consumer width ignored */
max-width: max-content -> hard cap at ~15px
width: 30px            -> ignored
Result: 15px

/* After: consumer width respected */
width: fit-content     -> from cauldron (specificity 0,1,0)
width: 30px            -> from consumer (later in cascade)
Result: 30px
```


shawnsharpDQ (approved):

Bracciata (approved):
LGTM

chornonoh-vova (approved):

anastasialanz (approved):

PR #2305, opened 2026-04-06
feat: add size option for Tag Buttons

### LINKED ISSUE #1957: Allow Button variant="tag" to support the Tag size prop
The `Tag` component has a `size` prop that we should allow the `Button` component to use.

Bracciata:
Closes #1957

## Summary

- Adds a `size` prop (`'default' | 'small'`) to `Button` (when `variant="tag"`) and `TagButton`
- Exports a shared `TagSize` type from the `Tag` component
- Refactors `ButtonProps` into discriminated union types so `size` is only valid when `variant="tag"` (type error otherwise)
- Adds screenshot tests for the small tag variant for both `Button` and `TagButton`
- Adds documentation and examples for the new `size` prop in both `Button.mdx` and `TagButton.mdx`
- Fixes autoprefixer converting `:user-invalid` to `:invalid` in `select.css`, which caused required selects to show error styling before user interaction
- Fixes `TextEllipsis[hideTooltip][maxLines=2]` test to actually hover and assert no tooltip appears
- Removes unnecessary dark mode screenshots for `TextEllipsis[hideTooltip]` tests (component has no dark mode styling)

## Test plan

- [ ] `Button` with `variant="tag"` and `size="small"` renders with `Tag--small` class
- [ ] `Button` with non-tag variant and `size` prop produces a TypeScript error
- [ ] `TagButton` with `size="small"` passes the size through to the underlying `Button`
- [ ] Screenshot baselines match expected styling for small tag buttons
- [ ] Required `Select` fields show no error border before user interaction
- [ ] Required `Select` fields show error border after user interacts and leaves the field empty
- [ ] `TextEllipsis[hideTooltip]` tests hover and verify no tooltip is shown


chornonoh-vova (dismissed):

packages/react/src/components/Button/index.tsx
```diff
@@ -11,16 +21,18 @@ export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
     | 'danger'
     | 'danger-secondary'
     | 'link'
-    | 'tag'
     | 'badge';
-  thin?: boolean;
+  size?: never;
 }
 
+export type ButtonProps = ButtonTagProps | ButtonNonTagProps;
```
frankensteinke, line 28: Thanks for using a discriminated union here instead of just throwing the new type on the shared props!

packages/react/src/components/Button/screenshots.e2e.tsx
```diff
@@ -566,6 +566,42 @@ test('should have screenshot for Button[variant="tag"]', async ({
...
+        Focus
+      </Button>
+      <Button variant="tag" size="small" disabled>
+        {' '}
+        Disabled
+      </Button>
+    </div>
+  );
+
+  await component.getByText('Hover').hover();
```
frankensteinke:
```suggestion
  await component.getByRole('button', { name: 'Hover' }).hover();
```

Nit: if we want to follow the [recommended RTL query priority](https://testing-library.com/docs/queries/about/#priority), `.getByRole` is probably the right choice here. _I know these are in playwright, but maybe we'd want to defer to that priority list anyway since they mirror the RTL locators?_

packages/react/src/components/Button/index.tsx
```diff
@@ -11,16 +21,18 @@ export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
...
+
 const Button = forwardRef<HTMLButtonElement, ButtonProps>(
   (
     {
       variant = 'primary',
       thin,
+      size = 'default',
```
frankensteinke:
```suggestion
      size,
```

I think you may be able to drop the default value in the destructure. `undefined` seems like it should behave identically, and it will keep the type honest for non-tag variants. This default value will be applied to the non-tag variant as well, as written.

e2e/screenshots/button-variant-tag-size-small-.png
frankensteinke, line 1: Weird that these screenshot file names end with a hyphen, but I don't think it's related to anything in this PR.

frankensteinke (dismissed):
I think we'd probably want to add this new prop to the docs for Button and TagButton. Here's what Claude told me that would involve:

> - [docs/pages/components/Button.mdx](docs/pages/components/Button.mdx) — add a `size` entry to the `ComponentProps` at line ~119 (noting it's only valid when `variant="tag"`), and optionally a new example under the Tag section (~line 81) showing `<Button variant="tag" size="small">Small Tag</Button>`.
> - [docs/pages/components/TagButton.mdx](docs/pages/components/TagButton.mdx) — add a `size` entry to the `ComponentProps` array at line 30, and optionally an example alongside the one at line 15.
>
> Both use the shared `<ComponentProps>` component, so it's just an object append like:
> 
> ```jsx
> {
>   name: 'size',
>   type: ['default', 'small'],
>   defaultValue: 'default',
>   description: '...'
> }
> ```

---

Other than that, I just left some minor pieces of feedback. Thanks for contributing!

e2e/screenshots/button-variant-tag-size-small-.png
Bracciata: Agreed. Maybe we open new issue?

frankensteinke: https://github.com/dequelabs/cauldron/issues/2329 done!

packages/react/src/components/Button/screenshots.e2e.tsx
```diff
@@ -566,6 +566,42 @@ test('should have screenshot for Button[variant="tag"]', async ({
...
+      <Button variant="tag" size="small" disabled>
+        {' '}
+        Disabled
+      </Button>
+    </div>
+  );
+
+  await component.getByRole('button', { name: 'Hover' }).hover();
+  setActive(component.getByText('Active'));
+  await component.getByText('Focus').focus();
```
frankensteinke: Nit: there are some other `.getByText` calls that could be updated to `.getByRole`.

packages/react/src/components/TagButton/screenshots.e2e.tsx
```diff
@@ -0,0 +1,61 @@
...
+        size="small"
+        disabled
+        onClick={noop}
+      />
+    </div>
+  );
+
+  await component.getByText('Hover').hover();
+  setActive(component.getByText('Active'));
+  await component.getByText('Focus').focus();
```
frankensteinke: Nit: more `.getByText` calls that could be updated to `.getByRole`.

frankensteinke (dismissed):
Looks good! Just left some nits about other test `.getByText` calls that could be changed to `.getByRole`.

packages/react/src/components/Button/screenshots.e2e.tsx
Bracciata: Updated

packages/react/src/components/TagButton/screenshots.e2e.tsx
Bracciata: Updated

frankensteinke (approved):
🚀

PR #2310, opened 2026-04-07
fix(Listbox): improve typing for ListBox

### LINKED ISSUE #1890: Bug: Listbox complains about typescript types
### Steps to reproduce

It looks like we don't currently have correct type discrimination setup for `Listbox` as the following typescript error is received if you use a listbox without setting multiselect:

```
Type '{ children: Element; onKeyDown: (event: KeyboardEvent<HTMLUListElement | HTMLAnchorElement>) => void; onActiveChange: (value: ListboxOption) => void; ... 262 more ...; className: string; }' is not assignable to type 'IntrinsicAttributes & PolymorphicComponentProps<SingleSelectListboxProps | MultiSelectListboxProps, ElementType<any, keyof IntrinsicElements>>'.
  Property 'multiselect' is missing in type '{ children: Element; onKeyDown: (event: KeyboardEvent<HTMLUListElement | HTMLAnchorElement>) => void; onActiveChange: (value: ListboxOption) => void; ... 262 more ...; className: string; }' but required in type 'MultiSelectListboxProps'.ts(2322)
Listbox.tsx(40, 3): 'multiselect' is declared here.
```

### Expected Behavior

You should be able to use a Listbox without having to specify multiselect, but when multiselect is set the callback handlers should be discriminative to return the right set of values (single value vs multiple when mulitselect). 

### Version

N/A

_No response_

### Relevant log output

```shell

```

Bracciata:
## Summary

Closes #1890

- Replaces the generic `PolymorphicComponent` cast with a custom `ListboxComponent` overloaded type that properly discriminates between single-select and multi-select variants
- Fixes TypeScript error when using `<Listbox>` without explicitly passing `multiselect` (single-select is now the default)
- Updates `@ts-expect-error` comment in `ActionList` to reflect the actual type conflict (HTML attribute `defaultValue` vs `ListboxValue`)
- Adds compile-time type tests to verify single-select, multi-select, and invalid usage are all typed correctly

## Test plan

- [x] `tsc` passes without errors for single-select `<Listbox value="foo">` (no `multiselect` prop)
- [x] `tsc` catches invalid usage (e.g. `multiselect` with a string value)
- [x] Existing Listbox unit tests still pass


chornonoh-vova (dismissed):

packages/react/src/components/Listbox/index.test.tsx
```diff
@@ -4,6 +4,50 @@ import Listbox from './';
...
+);
+
+// Multi-select should not accept single value
+() => (
+  // @ts-expect-error multi-select value must be an array
+  <Listbox multiselect value="foo">
+    <ListboxOption>A</ListboxOption>
+  </Listbox>
+);
+
```
frankensteinke: I see what you're going for here, but it seems a little unexpected to have these arrow functions in a test file that aren't in `test()` functions. They could be accidentally deleted, and they don't show up in test output.

Could these be real `test()` functions with assertions instead? The TypeScript validation should still work, and you could still use `@ts-expect-error` for the invalid ones ([this is done in the Button test](https://github.com/dequelabs/cauldron/blob/develop/packages/react/src/components/Button/index.test.tsx#L76-L85)). The comments turn into test names. For example,

```js
test('accepts single-select usage without multiselect prop', () => {
  render(
    <Listbox value="foo">
      <ListboxOption value="foo">A</ListboxOption>
    </Listbox>
  );
  expect(screen.getByRole('listbox')).not.toHaveAttribute('aria-multiselectable');
});

```

frankensteinke (dismissed):

frankensteinke (approved):

PR #2311, opened 2026-04-07
fix(select): update default inital styling of required selects

### LINKED ISSUE #1532: A required select field has an initial visual indication of invalid
Cauldron's `Select` component displays a red ring for a required select before the user has made any interactions or form submissions.

We current use `:invalid` to mark the selection as invalid: https://github.com/dequelabs/cauldron/blob/47e0c3edf23721333cb26838433ce9da1309330e/packages/styles/select.css#L66

We should instead use [`:user-invalid`](https://developer.mozilla.org/en-US/docs/Web/CSS/:user-invalid) as it is widely supported and only shows the error after the user has made an input or submitted a form.

Bracciata:
## Summary

Fixes #1532

- Updates CSS selectors in `select.css` from `:invalid` to `:user-invalid` so that required select fields no longer display error styling (red ring) on initial page load
- `:user-invalid` only activates after the user has interacted with the field or submitted the form, matching expected behavior
- Adds screenshot test for `Select[required]` with light and dark theme snapshots

## Test plan

- [x] Verify required select fields do not show red error ring on initial render
- [x] Verify error styling still appears after user interaction with a required field
- [x] Verify explicit `error` prop styling is unaffected
- [x] All 65 screenshot tests pass


chornonoh-vova (dismissed):

e2e/screenshots/dark--select-required--after-interaction.png
anastasialanz, line 1: None of these screenshots show what it looks like for `:user-invalid` to be triggered where the outline would be red.

anastasialanz:
The CSS fix (`:invalid` → `:user-invalid`) is correct and the initial-state screenshots confirm the red ring no longer appears on page load, which resolves #1532.

However, the "after interaction" e2e test (lines 88-91 in `screenshots.e2e.tsx`) does focus → blur to try triggering `:user-invalid`:

```ts
await component.getByLabel('Select').focus();
await component.getByLabel('Select').blur();
```

This is **not sufficient** to trigger `:user-invalid` on a `<select>`. Per the spec, `:user-invalid` requires the user to have "significantly interacted" with the control — for a `<select>`, this means actually **changing the value**, not just focusing/blurring. As a result, the "after interaction" screenshots are visually identical to the initial-state ones — neither shows a red error ring.

To properly trigger `:user-invalid`, the test should change the select value, e.g.:

```ts
const select = component.getByLabel('Select');
await select.selectOption('apple');   // select a valid value
await select.selectOption('');         // change back to invalid
await select.blur();
```

This would produce a screenshot that actually demonstrates the error styling appearing after user interaction, fully proving the fix works in both directions.

anastasialanz (dismissed):
Left a comment about how the screenshot tests don't prove the `:user-invalid` css.

chornonoh-vova (dismissed):

packages/react/src/components/Select/screenshots.e2e.tsx
```diff
@@ -25,6 +25,84 @@ test('should have screenshot for Select', async ({ mount, page }) => {
...
+        <option value="cucumber">Cucumber</option>
+      </Select>
+    </FieldWrap>
+  );
+
+  // Change value then revert to invalid to trigger :user-invalid
+  for (const label of ['Select', 'Hover', 'Focus']) {
+    await component.getByLabel(label).selectOption('apple');
+    await component.getByLabel(label).selectOption('');
+    await component.getByLabel(label).blur();
```
anastasialanz:
```
const select = component.getByLabel(label);
await select.focus();
await select.selectOption('apple');
await select.selectOption('');
await select.blur();
```

Looks like we need to focus first to get the red outline to appear to show these :user-valid styles.

anastasialanz (dismissed):
The screenshots still don't seem right because the red outline is missing for the "after" screenshot. I tried the code out below and was able to generate the error styles. Let me know if it works for you.

anastasialanz (approved):

Issue #2125, opened 2025-09-25
Introduce modal max-height


scurker:
We currently do not apply a max-height constraint to modals, which can push the controls of the modal down below the fold:

<img width="517" height="771" alt="" src="https://github.com/user-attachments/assets/501acf31-6186-49ea-8527-ea279bc38f7f" />

Ideally, we want to keep the controls in view while only scrolling the modal's content.

**Things to consider:**

Does this apply to all viewport sizes?

If we make the modal content scrollable, it can lead to scenarios where the content itself can shrink down to be unusable if the user needs to zoom in and/or has a smaller viewport height. We may need to consider that at a specific breakpoint the modal will need to default to the current behavior to avoid the dialog content shrinking down too small.

## Implementation Details

Dialog inner needs:

```css
max-height: calc(100vh - 130px)
```

And dialog content needs:

```css
overflow-y: auto;
```

PR #2314, opened 2026-04-08
fix: make scrollable modal a separate prop rather than the default

Bracciata:
This addresses a breaking change introduced in #2125.

## Changes

- Moves the scrollable modal behavior from the default to an opt-in `scrollable` boolean prop
- Scopes `max-height` / `overflow-y: auto` CSS rules to `.Modal--scrollable` only
- Non-scrollable modals revert to pre-#2125 behavior (scrim scrolls when content overflows)
- The `scrollable` prop is separate from `variant`, allowing composition with `variant="info"`
- Reduces vertical offset from `6.25rem` to `4rem` for better use of available viewport space
- Sets dialog `z-index` to `25` so modals render above the header
- Adds documentation and an interactive example for the scrollable modal
- Adds unit tests for the `scrollable` class and accessibility
- Adds e2e screenshot tests and a keyboard scrolling test for scrollable modals

## Migration

If you were relying on the scrollable-by-default behavior introduced in #2125, add the `scrollable` prop to your Modal:

```diff
- <Modal show>
+ <Modal show scrollable>
```

The `scrollable` prop can also be combined with `variant="info"`:

```jsx
<Modal show scrollable variant="info">
```


packages/styles/dialog.css
```diff
@@ -12,7 +12,7 @@
   --dialog-footer-border: 1px solid var(--dialog-border-color);
   --dialog-header-height: 2.625rem;
   --dialog-close-button-size: var(--target-size-minimum);
-  --dialog-vertical-offset: 6.25rem;
+  --dialog-vertical-offset: 4rem;
```
Bracciata: This change was recommended by @shawnsharpDQ

shawnsharpDQ (dismissed):

gupta-tanvi26 (dismissed):

chornonoh-vova (dismissed):

chornonoh-vova:
@Bracciata This looks great, but it needs a rebase or merge with develop to fix conflicts

chornonoh-vova (dismissed):

packages/styles/dialog.css
```diff
@@ -12,9 +12,10 @@
...
   --dialog-bottom-spacing: 1.875rem;
   --dialog-content-line-height: inherit;
+  --dialog-z-index: 25;
```
anastasialanz: Why is this 25 now? We also have `z-index` variables in `variables.css` which has `--z-index-dialog: 21;` and `--z-index-dialog-scrim: 20;`.

packages/react/src/components/Dialog/index.tsx
```diff
@@ -231,6 +231,7 @@ const DialogContent = ({
       'text--align-center': align === 'center',
       'text--align-right': align === 'right'
     })}
+    tabIndex={-1}
```
frankensteinke: Should this happen only when scrollable?

packages/styles/dialog.css
```diff
@@ -12,7 +12,7 @@
   --dialog-footer-border: 1px solid var(--dialog-border-color);
   --dialog-header-height: 2.625rem;
   --dialog-close-button-size: var(--target-size-minimum);
-  --dialog-vertical-offset: 6.25rem;
+  --dialog-vertical-offset: 4rem;
   --dialog-bottom-spacing: 1.875rem;
```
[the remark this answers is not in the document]
frankensteinke: Does this change only need to happen to the scrollable version?

packages/styles/dialog.css
Bracciata: I felt the modal should be above the heaing, however, I will revert it to match!

vpats/2026-04-09-cauldron.md
anastasialanz, line 1: Can we drop this vpat from this change?

Bracciata: done!

packages/styles/dialog.css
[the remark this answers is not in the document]
frankensteinke: Looks like this happens everywhere based on the screenshots (and probably should).

I know it's just one line, but would you be able to make this bit into a separate PR? That way, it's easier to track/deploy/revert if needed.

Then we can get this PR merged in. :)

frankensteinke (dismissed):

frankensteinke (approved):
🎉

PR #2315, opened 2026-04-08
fix: revert font-family on heading elements

anastasialanz:
## Summary
- Removes the `--heading-font-family` CSS custom property and `font-family` rule on heading elements introduced in #2287
- Headings will now inherit `font-family` from `body` instead of being explicitly set via a custom property

## Test plan
- [ ] Verify heading elements render with the correct inherited font-family
- [ ] Confirm no visual regressions in light and dark themes


Bracciata (approved):
LGTM

PR #2319, opened 2026-04-09
fix(TableHeader): remove sort announcement props as no needed and update aria-sort handling

maksym-shynkarenko:
## Summary

Fix NVDA screen reader announcing sort state on every table row cell.

**Problem:** 
NVDA was reading "Rule sorted ascending" (or similar) for every cell in a sortable column, not just the column header.

For example:
row 2  RuleRule sorted ascending  column 1  link   and elements...
row 3  RuleRule sorted ascending  column 1  link  All page content...

**Root cause:** 
`TableHeader` rendered an `Offscreen` live region
(`role="status" aria-live="polite"`) inside the sort `<button>`. Because it
was inside the button, screen readers included the announcement text in the
column header's accessible name. NVDA then repeated that name for every cell
in the column.

**Fix:**
Removed the `Offscreen` live region entirely. Sort state is now
communicated solely via the `aria-sort` attribute on `<th>`, which is the
standard ARIA mechanism and is well-supported in modern screen readers (NVDA, JAWS, VoiceOver).

## Changes

- `packages/react/src/components/Table/TableHeader.tsx` — removed `Offscreen`
  live region from inside the sort button; marked `sortAscendingAnnouncement`
  and `sortDescendingAnnouncement` props as `@deprecated` for backwards compatibility

- `packages/react/src/components/Table/index.test.tsx` — updated three tests
  to assert on `aria-sort` attribute instead of the removed live region content

## Before / After (NVDA)

**Before:**
row 2  RuleRule sorted ascending  column 1  link  ...
row 3  RuleRule sorted ascending  column 1  link  ...
<details><summary>Screenshot:</summary>

photo_2026-04-09 13 19 04

</details>

**After:**
row 1  column 1  sorted ascending  Rule  button
row 2  First Name  column 1  Frank
row 3  First Name  column 1  Jimmy
<details><summary>Screenshot:</summary>

photo_2026-04-09 13 19 07

</details>

**DebHub verification:**

photo_2026-04-09 13 19 10

Closes: [Walnut #13951](https://github.com/dequelabs/walnut/issues/13951)

### DIFF
--- a/packages/react/src/components/Table/TableHeader.tsx
+++ b/packages/react/src/components/Table/TableHeader.tsx
@@ -1,7 +1,6 @@
 import type { ColumnAlignment } from './Table';
 import React, { forwardRef } from 'react';
 import classNames from 'classnames';
-import Offscreen from '../Offscreen';
 import Icon from '../Icon';
 import { useTable } from './TableContext';
 import useTableGridStyles from './useTableGridStyles';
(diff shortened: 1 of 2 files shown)


Bracciata:
@anastasialanz as a heads up I am merging this with https://github.com/dequelabs/cauldron/pull/2291 so it is one less PR for same fix

PR #2322, opened 2026-04-09
feat: tree view component

mateoviilla1:
This pull request introduces a new \`TreeView\` component to the React package, providing a fully accessible, customizable tree structure component. It includes comprehensive documentation, tests, and styles, and integrates with the existing library's export system. Additionally, a small enhancement is made to the \`Checkbox\` component to support external state management.

**New TreeView Component**

* Added a new \`TreeView\` component, including its core implementation (\`index.tsx\`) and a supporting \`TreeViewItem\` subcomponent, leveraging \`react-aria-components\` for accessibility and state management.
* Added a complete test suite for \`TreeView\` covering rendering, selection modes, actions, expansion logic, keyboard navigation, axe violations, SSR rendering, and multi-select deselect correctness.
* Added \`TreeView\` documentation with usage examples, prop descriptions, and code samples.
* Exported \`TreeView\` and its type from the package's main entry point.

**Styling and Dependencies**

* Introduced dedicated styles for the tree view, including light/dark theme support, BEM class names, and Cauldron spacing tokens (\`--space-quarter\`, \`--space-half\`), and imported them into the main stylesheet.
* Added \`react-aria-components\` as a new dependency to support accessible tree behavior.

**Checkbox Component Enhancement**

* Added a \`controlled\` prop to \`Checkbox\` to opt out of internal checked-state management, supporting its use as a fully controlled component within \`TreeView\`. When \`controlled\` is set, the component renders the \`checked\` value passed by the parent and reports interactions through \`onChange\` without toggling its own state.


mateoviilla1 (commented):
## Review

### Critical

**[`tree-view.css:14–74`](https://github.com/dequelabs/cauldron/blob/21aa7ce49f593a2f3727b6ff4b5d13c925dc554f/packages/styles/tree-view.css#L14-L74) — CSS nesting will be silently dropped in target browsers**

The stylesheet uses CSS nesting (`&[data-focus-visible]`, nested `.react-aria-TreeItem`, etc.) throughout. Two problems compound each other:

1. The styles package's PostCSS config only runs `postcss-import` and `autoprefixer` — there is no nesting plugin to transform the syntax.
2. The project's browserslist targets **Chrome 55** and **Firefox 54**, which have no native CSS nesting support (landed in Chrome 112 / Firefox 117).

`tree-view.css` is the only file in the package using this syntax — every other stylesheet uses flat selectors. As a result, all the nested rules (focus rings, item indentation, selected-state highlight, chevron animation) will be silently ignored in supported environments.

**Fix**: Flatten all selectors:
```css
.react-aria-Tree[data-focus-visible] { ... }
.react-aria-Tree .react-aria-TreeItem { ... }
.react-aria-Tree .react-aria-TreeItem[data-focus-visible] { ... }
.react-aria-Tree .react-aria-TreeItem[aria-selected='true'] { ... }
```

---

### Important

**[`tree-view.css:51–54`](https://github.com/dequelabs/cauldron/blob/21aa7ce49f593a2f3727b6ff4b5d13c925dc554f/packages/styles/tree-view.css#L51-L54) — `rotate` property not supported in target browsers**

The chevron animation uses the standalone `rotate` CSS property (`rotate: 0deg` / `rotate: 90deg`). This property requires Chrome 104+ / Firefox 72+, but the target is Chrome 55 / Firefox 54. The rest of the codebase uses `transform: rotate()` exclusively. Switch to `transform: rotate(0deg)` / `transform: rotate(90deg)` and `transition: transform 200ms`.

---

**[`index.tsx:11–16`](https://github.com/dequelabs/cauldron/blob/21aa7ce49f593a2f3727b6ff4b5d13c925dc554f/packages/react/src/components/TreeView/index.tsx#L11-L16) — `ariaLabel` prop doesn't support `aria-labelledby`**

`TreeView` accepts `ariaLabel: string`, which only maps to `aria-label`. The ARIA spec requires either `aria-label` or `aria-labelledby`. Other components in the codebase use `Cauldron.LabelProps` to support both — see `Tabs.tsx:21`. Update `TreeViewProps` to extend `Cauldron.LabelProps` and spread those props onto `<Tree>`.

---

**[`TreeView.test.tsx:1`](https://github.com/dequelabs/cauldron/blob/21aa7ce49f593a2f3727b6ff4b5d13c925dc554f/packages/react/src/components/TreeView/TreeView.test.tsx#L1) — Missing jest-axe tests**

For an accessibility-focused component built on react-aria, `jest-axe` coverage is the most important regression guard. CONTRIBUTING.md requires axe tests for new components. Add at least one assertion per rendered variant:

```ts
import { axe } from 'jest-axe';

test('has no axe violations', async () => {
  const { container } = render(<TreeView ariaLabel="Test" items={items} />);
  expect(await axe(container)).toHaveNoViolations();
});
```

---

**[`Checkbox/index.tsx:23`](https://github.com/dequelabs/cauldron/blob/21aa7ce49f593a2f3727b6ff4b5d13c925dc554f/packages/react/src/components/Checkbox/index.tsx#L23) — `onChangeToggle` prop has no test coverage**

The new `onChangeToggle` prop modifies core internal behavior of `Checkbox` but there's no test asserting it. Since `Checkbox.test.tsx` is already comprehensive, add a focused test verifying that when `onChangeToggle={false}`, clicking the checkbox does not change the internal checked state.

---

### Suggestions

**[`TreeView.test.tsx:49`](https://github.com/dequelabs/cauldron/blob/21aa7ce49f593a2f3727b6ff4b5d13c925dc554f/packages/react/src/components/TreeView/TreeView.test.tsx#L49) — Fragile test selector**

`getAllByLabelText('TreeView')[1]` relies on positional ordering and will break silently if the DOM structure changes. Prefer `getByRole('checkbox', { name: 'TreeView' })`.

---

**[`TreeView.test.tsx:2`](https://github.com/dequelabs/cauldron/blob/21aa7ce49f593a2f3727b6ff4b5d13c925dc554f/packages/react/src/components/TreeView/TreeView.test.tsx#L2) — Use `userEvent` instead of `fireEvent`**

`fireEvent` dispatches bare synthetic events that bypass real browser handling. For a component built on react-aria where event sequencing matters, `userEvent` gives more accurate coverage.

packages/react/src/components/Checkbox/index.tsx
```diff
@@ -19,6 +19,7 @@ export interface CheckboxProps extends InputHTMLAttributes<HTMLInputElement> {
   customIcon?: React.ReactNode;
   checkboxRef?: React.ForwardedRef<HTMLInputElement>;
   indeterminate?: boolean;
+  onChangeToggle?: boolean;
 }
```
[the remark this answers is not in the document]
mateoviilla1: This is addressed already on Checkbox.test.tsx:216–224:

packages/react/src/components/TreeView/TreeView.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+test('selects a checkbox when clicked', () => {
+  const { getAllByLabelText } = render(
+    <TreeView
+      ariaLabel="Test TreeView"
+      items={items}
+      selectionMode="multiple"
+    />
+  );
+  const checkbox = getAllByLabelText('TreeView')[1];
```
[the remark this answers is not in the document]
mateoviilla1: This is addressed already

packages/react/src/components/TreeView/TreeView.test.tsx
```diff
@@ -0,0 +1,123 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import TreeView, { TreeViewNode } from '../../../src/components/TreeView';
+
```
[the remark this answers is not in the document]
mateoviilla1: This is already fixed

packages/react/src/components/TreeView/TreeView.test.tsx
```diff
@@ -0,0 +1,171 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { axe } from 'jest-axe';
```
davidg-dq: **Important — Wrong axe import**

This imports `{ axe } from 'jest-axe'` directly. Every other component test in the repo imports the shared axe helper at `../../axe` (or `../../../axe`), which disables the `region` rule — components render outside landmark regions in test isolation, causing false positives without that config.

```ts
import axe from '../../../axe';
```

packages/styles/tree-view.css
```diff
@@ -0,0 +1,75 @@
...
+  --tree-view-focus-ring-color: var(--focus-light);
+  --tree-view-highlight-background: var(--accent-light);
+}
+
+.cauldron--theme-dark {
+  --tree-view-focus-ring-color: var(--focus-dark);
+  --tree-view-highlight-background: var(--accent-dark);
+}
+
+.react-aria-Tree {
```
davidg-dq: **Important — No BEM class names**

This is the only stylesheet in the package targeting react-aria's auto-generated class names (`.react-aria-Tree`, `.react-aria-TreeItem`, `.react-aria-Button`) instead of Cauldron's BEM convention (`.TreeView`, `.TreeView__item`, `.TreeView__chevron`).

This tightly couples styles to react-aria-components' internal class naming — if the library changes them in a major version, styles silently break. It also means consumers can't target standard Cauldron selectors for overrides.

**Suggestion:** Apply BEM class names via `className` props on react-aria components and target those in the stylesheet, consistent with every other component's CSS.

packages/react/src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useState } from 'react';
```
davidg-dq: **Important — Missing screenshot tests**

This component has custom visual styling (focus rings, selection highlights, chevron rotation animations, light/dark theme support) but no `screenshots.e2e.tsx` file. 16 other visual components have screenshot tests. Per project guidelines, visual changes require screenshot tests covering both light and dark themes.

**Suggestion:** Add `screenshots.e2e.tsx` covering default, expanded, single-selection, and multiple-selection states in both themes.

packages/react/src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+type TreeViewProps = Cauldron.LabelProps & {
+  items: TreeViewNode[];
+  onAction?: (key: string) => void;
+  selectionMode?: 'none' | 'single' | 'multiple';
+  defaultExpandedKeys?: string[];
+};
+
+const TreeView = ({
```
davidg-dq: **Important — Missing `forwardRef`, `className`, and rest props**

Every other component in the codebase follows the `forwardRef` pattern, accepts `className` (merged via `classNames()`), and spreads `...other` props to the root element. This allows consumers to pass `data-testid`, custom classes, event handlers, etc.

TreeView is a plain function that only accepts its defined props + label props. This breaks the API contract consumers expect from Cauldron components.

**Suggestion:**
```tsx
const TreeView = forwardRef<HTMLDivElement, TreeViewProps>(
  ({ items, onAction, selectionMode = 'none', defaultExpandedKeys, className, ...other }, ref) => {
    // ...
    return (
      <Tree ref={ref} className={classNames('TreeView', className)} {...other} ...>
```

packages/react/package.json
```diff
@@ -27,6 +27,7 @@
     "classnames": "^2.2.6",
     "focusable": "^2.3.0",
     "keyname": "^0.1.0",
+    "react-aria-components": "^1.13.0",
```
davidg-dq, line 30: **Important — New dependency needs team alignment**

`react-aria-components` adds 1,436 lines to `yarn.lock` — it brings in the full react-aria + react-stately suite. Per project guidelines, new dependencies should be explicitly flagged for team review.

Worth confirming the team is aligned on adopting this. Consider whether importing only the specific packages (`@react-aria/tree`, `@react-stately/tree`) would reduce the footprint, or whether `react-aria-components` is the intended direction for future components.

packages/react/src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+  ...labelProps
+}: TreeViewProps) => {
+  // When onAction is provided, react-aria-components doesn't toggle selected state automatically.
+  // See: https://react-aria.adobe.com/Tree#selection-and-actions
+  // We manage selectedKeys ourselves so selection state stays in React (not the DOM).
+  const [selectedKeys, setSelectedKeys] = useState<Selection>(new Set());
+
+  const handleAction = (key: Key) => {
+    setSelectedKeys((prev) => {
+      const prevSet = prev === 'all' ? new Set<Key>() : new Set(prev);
```
davidg-dq: **Important — Bug: `'all'` selection state silently loses all selections**

`prev === 'all'` is converted to an empty `Set()`. The `Selection` type from react-aria is `'all' | Set<Key>`, and `'all'` can be triggered by Ctrl+A / Cmd+A (react-aria's built-in select-all behavior, forwarded via `onSelectionChange: setSelectedKeys` on line 57).

The bug: User presses Ctrl+A → all items selected → user clicks any item → `handleAction` converts `'all'` to empty set → only clicked item remains → **all other selections lost**.

**Fix:** Expand `'all'` into the full set of item IDs rather than treating it as empty:
```ts
const allIds = new Set<Key>(items.flatMap(flattenIds));
const prevSet = prev === 'all' ? allIds : new Set(prev);
```

packages/react/src/components/Checkbox/index.tsx
```diff
@@ -19,6 +19,7 @@ export interface CheckboxProps extends InputHTMLAttributes<HTMLInputElement> {
   customIcon?: React.ReactNode;
   checkboxRef?: React.ForwardedRef<HTMLInputElement>;
   indeterminate?: boolean;
+  onChangeToggle?: boolean;
```
davidg-dq: **Suggestion — `onChangeToggle` naming**

The `on` prefix reads like an event handler callback, but this is a boolean flag controlling whether the checkbox toggles its internal state. Consider `toggleOnChange` or adding a JSDoc comment to clarify it's a behavior flag, not a callback.

packages/react/src/components/TreeView/TreeView.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+      { id: '5', textValue: 'foo' },
+      { id: '6', textValue: 'bar' }
+    ]
+  }
+];
+
+test('renders tree items', () => {
+  const { getByText } = render(
+    <TreeView aria-label="Test TreeView" items={items} />
+  );
```
davidg-dq, line 30: **Suggestion — Prefer `getByRole` over `getByText`**

Several tests use `getByText('TreeView')` which is lower on the [RTL query priority](https://testing-library.com/docs/queries/about/#priority). Project testing conventions strongly prefer `getByRole` with accessible names:

```ts
getByRole('treeitem', { name: 'TreeView' })
```

This mirrors how assistive technology discovers elements and makes tests more resilient.

packages/styles/tree-view.css
```diff
@@ -0,0 +1,75 @@
+:root {
+  --tree-view-padding: 16px;
```
davidg-dq: **Suggestion — Use `rem` instead of `px`**

The codebase convention prefers `rem` for sizing values. Consider `1rem` here:
```css
--tree-view-padding: 1rem;
```

packages/react/src/components/TreeView/TreeView.test.tsx
```diff
@@ -0,0 +1,171 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { axe } from 'jest-axe';
+import TreeView, { TreeViewNode } from '../../../src/components/TreeView';
+
```
davidg-dq, line 7: **Suggestion — No keyboard navigation tests**

Tree is a composite widget — WAI-ARIA APG requires ArrowUp/Down/Left/Right, Enter, Space, Home, End. For an accessibility-focused design system, keyboard interaction is arguably the most important coverage for this component. All 13 current tests are render checks or click-based.

**Suggestion:** Add tests for at minimum:
- `ArrowDown` moves focus between tree items
- `ArrowRight` expands a collapsed node
- `ArrowLeft` collapses an expanded node
- `Enter` / `Space` activates the focused item

packages/react/src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    });
+    onAction?.(key as string);
+  };
+
+  return (
+    <Tree
+      {...labelProps}
+      selectionMode={selectionMode}
+      defaultExpandedKeys={defaultExpandedKeys}
+      {...(onAction
```
davidg-dq: **Suggestion — Extract conditional props for readability**

The ternary spread here means the component behaves fundamentally differently when `onAction` is provided (controlled selection) vs not (react-aria manages natively). This behavioral fork is easy to miss while scanning JSX.

Extracting it into a named variable above the return makes the JSX more scannable and the fork immediately visible:
```tsx
const actionProps = onAction
  ? { onAction: handleAction, selectedKeys, onSelectionChange: setSelectedKeys }
  : {};

return (
  <Tree {...labelProps} selectionMode={selectionMode} defaultExpandedKeys={defaultExpandedKeys} {...actionProps}>
```

packages/react/src/components/TreeView/TreeViewItem.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { useMemo } from 'react';
+import {
+  Button,
+  Collection,
+  TreeItem,
+  TreeItemContent,
+  TreeItemContentRenderProps
+} from 'react-aria-components';
+import nextId from 'react-id-generator';
+import { TreeViewNode } from '.';
```
davidg-dq: **Suggestion — Circular import**

`TreeViewItem` imports `TreeViewNode` from its own barrel file (`'.'`), which re-exports from `index.tsx`, which imports `TreeViewItem`. This circular dependency works today but is fragile — bundlers and test runners can resolve circular imports inconsistently.

**Suggestion:** Move `TreeViewNode` to a shared `types.ts` within the TreeView directory and import from there in both files.

packages/styles/tree-view.css
```diff
@@ -0,0 +1,75 @@
...
+  outline-offset: -1px;
+}
+
+.react-aria-Tree .react-aria-TreeItem {
+  display: flex;
+  align-items: center;
+  gap: 0.571rem;
+  min-height: 28px;
+  padding: 0.286rem 0.286rem 0.286rem 0.571rem;
+  --padding: var(--tree-view-padding);
```
davidg-dq: **Suggestion — Generic `--padding` variable name**

Every other component uses scoped custom property names (`--button-padding`, `--accordion-trigger-padding`, etc.). The bare `--padding` could collide with parent components and is unclear at a glance.

Since `--tree-view-padding` is already defined in `:root`, just use it directly in the `calc()` on line 50 and remove this intermediate alias.

davidg-dq (dismissed):
## Review Summary

Good progress — the earlier Copilot feedback (CSS nesting, `rotate` property, `ariaLabel` → `LabelProps`, `fireEvent` → `userEvent`, fragile selectors, missing axe tests) has been addressed. This pass focuses on remaining conventions alignment, a selection bug, and readability.

**Important (6):** wrong axe import, no BEM class names in CSS, missing screenshot tests, no `forwardRef`/`className`/rest-props, new dependency needs alignment, selection bug with `'all'` state.

**Suggestions (7):** `onChangeToggle` naming, `getByText` vs `getByRole`, `16px` → `1rem`, keyboard navigation tests, conditional prop spread readability, circular import, generic `--padding` variable name.

Details are inline below.

packages/react/src/components/TreeView/TreeView.test.tsx
mateoviilla1: Fixed: changed to `import axe from '../../axe'` in TreeView.test.tsx, which uses the shared helper that disables the `region` rule.

packages/styles/tree-view.css
mateoviilla1: Fixed: added `className="TreeView"` to `<Tree>`, `className="TreeView__item"` to `<TreeItem>`, and `className="TreeView__chevron"` to the chevron `<Button>`. Updated `tree-view.css` to target these BEM class names instead of the react-aria defaults.

packages/react/src/components/TreeView/index.tsx
mateoviilla1: Fixed: added `screenshots.e2e.tsx` covering default, expanded, single-selection, and multiple-selection states in both light and dark themes.

packages/react/src/components/TreeView/index.tsx
mateoviilla1: Fixed: `TreeView` now uses `forwardRef<HTMLDivElement, TreeViewProps>`, accepts `className` (merged via `classNames('TreeView', className)`), and spreads `...other` onto `<Tree>`.

packages/react/package.json
mateoviilla1: Acknowledged — flagging this here for team visibility. The intent is to adopt `react-aria-components` as the direction for accessible primitives. If the team prefers individual `@react-aria/*` packages or has concerns about the footprint, happy to follow up in a separate discussion.

packages/react/src/components/TreeView/index.tsx
mateoviilla1: Fixed: added a `collectAllKeys` helper that flattens the items tree into a `Set<Key>`. When `prev === 'all'`, `handleAction` now seeds the set from all keys before toggling the clicked one, so Ctrl+A → click correctly deselects one item rather than clearing everything.

packages/react/src/components/Checkbox/index.tsx
mateoviilla1: Added a JSDoc comment clarifying this is a behavior flag, not a callback: `When false, clicks won't update internal checked state — use when a parent controls `checked` externally.`

packages/react/src/components/TreeView/TreeView.test.tsx
mateoviilla1: Fixed throughout the test file — all `getByText` calls targeting tree items are now `getByRole('row', { name: '...' })` (react-aria renders tree items with role=row inside a treegrid).

packages/styles/tree-view.css
mateoviilla1: Fixed: `--tree-view-padding` is now `1rem` instead of `16px`.

packages/react/src/components/TreeView/TreeView.test.tsx
mateoviilla1: Fixed: added 4 keyboard navigation tests covering ArrowDown, ArrowUp, ArrowRight (expand), and ArrowLeft (collapse). All pass via react-aria-components' built-in keyboard handling.

packages/react/src/components/TreeView/index.tsx
mateoviilla1: Fixed: extracted into a named `actionProps` variable above the return statement so the behavioral fork is immediately visible when reading the JSX.

packages/react/src/components/TreeView/TreeViewItem.tsx
mateoviilla1: Fixed: moved `TreeViewNode` to `types.ts`. Both `index.tsx` and `TreeViewItem.tsx` now import from `./types`, breaking the circular dependency.

packages/styles/tree-view.css
mateoviilla1: Fixed: removed the intermediate `--padding` alias. The `calc()` on the chevron now references `--tree-view-padding` directly.

packages/react/src/components/Checkbox/index.tsx
markreay: **Suggestion: Replace `onChangeToggle` with a `controlled` prop**

The goal here makes sense — Checkbox needs a way to let a parent own checked state. But `onChangeToggle` is hard to read and feels specific to TreeView's wiring. I'd suggest a `controlled` prop instead, which could become a standard cauldron pattern.

The implementation change is identical — same spot, same gate:

```
  // Current PR
  if (onChangeToggle) {
    setIsChecked(e.target.checked);
  }
```

```
  // Proposed
  if (!controlled) {
    setIsChecked(e.target.checked);
  }
```

But the calling pattern is much clearer:

```
  // Current PR — reads like disabling a callback
  <Checkbox
    id="item-1"
    label="Select item"
    checked={isSelected}
    onChangeToggle={false}
    onChange={handleSelection}
  />
```

```
  // Proposed — says exactly what's happening
  <Checkbox
    id="item-1"
    label="Select item"
    controlled
    checked={isSelected}
    onChange={handleSelection}
  />
```

controlled is self-documenting: "I own the state, you just render it." `onChangeToggle={false}` requires reading the source to understand.

**Broader pattern: This should become a cauldron convention for any component that manages internal state but needs to support external state management.

Worth documenting as a standard pattern:

Components that manage internal state (Checkbox, Accordion, Disclosure, etc.) support a controlled prop.

> When controlled is set, the component does not manage its own state — it renders what you pass via props and reports interactions through callbacks. Use controlled when state is managed by a parent component (e.g., TreeView managing Checkbox selection).

packages/react/package.json
markreay: I discussed this new component with the cauldron team and they are accepting of it. Good to go.

packages/react/src/components/TreeView/index.tsx
markreay: We need a test to ensure that the scenario that David suggested above cannot happen. If it is there and I missed it, let me know.

packages/react/src/components/TreeView/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+import classNames from 'classnames';
+import { Tree, type Selection, type Key } from 'react-aria-components';
+import { Cauldron } from '../../types';
+import { TreeViewNode } from './types';
+import TreeViewItem from './TreeViewItem';
+
+export type { TreeViewNode } from './types';
+
+type TreeViewProps = Cauldron.LabelProps & {
+  items: TreeViewNode[];
```
markreay, line 11: ### Suggestion: Consider a declarative children API in a future iteration

The items data array works well for the initial use case, but it diverges from how other cauldron components work (ActionMenu, Listbox, Tabs all use JSX children). A declarative API would be more composable and flexible:

```tsx
<TreeView aria-label="Files">
  <TreeViewItem>A</TreeViewItem>
  <TreeViewItem>
    B
    <TreeViewGroup>
      <TreeViewItem>1</TreeViewItem>
    </TreeViewGroup>
  </TreeViewItem>
</TreeView>
```

This also aligns with the RFC's example API. Not blocking — the data-driven approach is fine for now. But I'd recommend planning to add the declarative pattern in a follow-up and eventually deprecating `items` in favor of it, so TreeView stays consistent with the rest of cauldron.

markreay (changes requested):
Very close - I have a suggestion for a slightly different prop for indicating the checkbox will be controlled. Let me know your thoughts on that.

Also there is a test we should add to this.

Additionally, the RFC proposed a declarative pattern for showing the items. Not necessary to do this right now, but there are good reasons to use this for a shared component.

packages/styles/tree-view.css
```diff
@@ -0,0 +1,69 @@
...
+
+.cauldron--theme-dark {
+  --tree-view-focus-ring-color: var(--focus-dark);
+  --tree-view-highlight-background: var(--accent-dark);
+}
+
+.TreeView {
+  display: flex;
+  flex-direction: column;
+  gap: 2px;
```
shawnsharpDQ: should we use our spacing variables here like `--space-quarter`

packages/styles/tree-view.css
```diff
@@ -0,0 +1,69 @@
...
+  overflow: auto;
+  padding: 4px;
```
shawnsharpDQ: same here? `--space-half`

markreay (approved):

packages/react/src/components/Checkbox/index.tsx
mateoviilla1: Done — renamed `onChangeToggle` to `controlled` and flipped the condition to `if (!controlled)`. Updated `TreeViewItem` to use `controlled` as well, and updated the existing test name to match.

packages/react/src/components/TreeView/index.tsx
mateoviilla1: Added — `'deselects only the clicked item when multiple items are selected via onAction'` in `TreeView.test.tsx`. Selects two items via click, deselects one, and asserts the other remains selected, guarding the `prev === 'all'` path in `handleAction`.

packages/react/src/components/TreeView/index.tsx
mateoviilla1: Agreed — the `items` data-driven API works well for the initial use case. A follow-up to design the declarative `<TreeViewItem>` children pattern makes sense once this settles. I'll open a tracking issue for that.

packages/styles/tree-view.css
mateoviilla1: Updated — `gap: 2px` is now `gap: var(--space-quarter)`.

packages/styles/tree-view.css
mateoviilla1: Updated — `padding: 4px` is now `padding: var(--space-half)`.

e2e/screenshots/dark--tree-view-expanded.png
anastasialanz, line 1: Looks like we want to clean up these screenshots since they're very hard to see the text.

PR #2327, opened 2026-04-17
fix(Combobox): add aria attributes on error

chornonoh-vova:
# Summary

- add `aria-invalid` to internal `<input />` when there's an error
- add `aria-describedby` to the internal `<Listbox />`

Together, these 2 attributes form correct representation of the error
state to the screen reader.

When they were missing previously, VoiceOver was not announcing an error
message (even though NVDA did).


Bracciata (approved):
This looks good to me! Claude brought up a concern of consumers overriding props, however, I don't think that is worth us considering.

packages/react/src/components/Combobox/Combobox.test.tsx
```diff
@@ -1909,3 +1909,177 @@ test('should have no axe violations with description and error when expanded', a
...
+      <ComboboxOption>Banana</ComboboxOption>
+      <ComboboxOption>Cantaloupe</ComboboxOption>
+    </Combobox>
+  );
+
+  const listbox = screen.getByRole('listbox');
+  expect(listbox).toHaveAttribute(
+    'aria-describedby',
+    expect.stringContaining('combo-error')
+  );
```
frankensteinke:
```suggestion
  expect(listbox).toHaveAccessibleDescription(
    'This field is required'
  );
```

Nit: this and the other tests assert on the wiring of the IDs, but not the actual description. What do you think about testing on the resolved description instead so that we're testing behavior rather than implementation?

frankensteinke (dismissed):
Approved with nit, flagged inline.

frankensteinke (approved):

PR #2344, opened 2026-04-28
fix: remove unneeded autoprefixer comments

anastasialanz:
When running the docs site locally, these were causing a compile warning to appear. We don't need to turn off auto-prefixing for these.

Screenshot 2026-04-28 at 1 59 21 PM

```
 ⚠  Second Autoprefixer control comment was ignored. Autoprefixer applies control comment to whole block, not to next rules. [autoprefixer]
 ```

https://github.com/postcss/autoprefixer#control-comments

### DIFF
--- a/packages/styles/select.css
+++ b/packages/styles/select.css
@@ -71,7 +71,6 @@
   border-color: var(--field-border-color-focus-hover);
 }
 
-/* autoprefixer: off */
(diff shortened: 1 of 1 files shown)


Bracciata (approved):
LGTM!

frankensteinke (approved):

PR #2350, opened 2026-05-01
fix(Tooltip): cancel hide timeout on unmount

frankensteinke:
## Summary

`Tooltip` schedules a 100 ms `setTimeout` in its `hide` handler that calls `setShowTooltip(false)` when it fires. If the Tooltip unmounts during that 100 ms window — e.g. a parent flips a `{condition && <Tooltip />}` gate while the user is mousing or blurring away from the target — the timeout still fires and runs `setShowTooltip` on an unmounted component, producing the React warning:

> Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak…

The existing `hide` callback returned a cleanup function, but [`useCallback` never invokes the return value of the memoized function](https://react.dev/reference/react/useCallback) — so that cleanup never ran.

This PR:
- Drops the dead `return () => clearTimeout(...)` from the `hide` `useCallback`.
- Adds a `useEffect` whose unmount cleanup clears `hideTimeoutRef.current`, so a pending hide timeout cannot fire after the component is gone.
- Adds a regression test that fails on `develop` and passes with this change. The test asserts the side effect itself (the `cauldron:tooltip:hide` custom event is **not** dispatched after unmount) rather than relying on a console warning, since React 18 no longer emits that warning.

Reproduced in a downstream consumer (axe-extension) where `<LockTooltip>` is conditionally rendered behind auth/license flags that flip frequently.

Results in this error upstream in the extension

## Test plan

- [x] `npx jest src/components/Tooltip/Tooltip.test.tsx` — full Tooltip suite passes (19/19)
- [x] New test fails on `develop` (verified by stashing the `index.tsx` change and re-running) and passes with the fix
- [ ] CI green

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.test.tsx
+++ b/packages/react/src/components/Tooltip/Tooltip.test.tsx
@@ -222,26 +222,30 @@ test('should cancel the pending hide timeout when unmounted during the hide dela
   const hideEvent = spy();
   button.addEventListener('cauldron:tooltip:hide', hideEvent);
 
-  const ShowTooltip = ({ show = true }: { show?: boolean }) =>
-    show ? (
-      <Tooltip target={button} defaultShow={false}>
-        Hello Tooltip
-      </Tooltip>
-    ) : null;
+  try {
+    const ShowTooltip = ({ show = true }: { show?: boolean }) =>
+      show ? (
+        <Tooltip target={button} defaultShow={false}>
+          Hello Tooltip
+        </Tooltip>
+      ) : null;
 
-  const { rerender } = render(<ShowTooltip />);
-  await fireEvent.focusIn(button);
-  expect(await screen.findByRole('tooltip')).toBeInTheDocument();
+    const { rerender } = render(<ShowTooltip />);
+    await fireEvent.focusIn(button);
+    expect(await screen.findByRole('tooltip')).toBeInTheDocument();
 
-  // Schedule the 100ms hide timeout, then unmount before it fires.
-  await fireEvent.focusOut(button);
-  rerender(<ShowTooltip show={false} />);
+    // Schedule the 100ms hide timeout, then unmount before it fires.
+    await fireEvent.focusOut(button);
+    rerender(<ShowTooltip show={false} />);
 
-  // Wait past TIP_HIDE_DELAY so a leaked timeout would have fired.
-  await new Promise((resolve) => setTimeout(resolve, 150));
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -215,6 +215,35 @@ test('should clean up association when tooltip is no longer rendered', async ()
...
+
+  // Schedule the 100ms hide timeout, then unmount before it fires.
+  await fireEvent.focusOut(button);
+  rerender(<ShowTooltip show={false} />);
+
+  // Wait past TIP_HIDE_DELAY so a leaked timeout would have fired.
+  await new Promise((resolve) => setTimeout(resolve, 150));
+
+  expect(hideEvent.called).toBe(false);
+  button.remove();
```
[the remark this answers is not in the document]
frankensteinke: [🤖 AI generated] Good catch — applied in 7bebba8b. Wrapped the body in try/finally and added the `removeEventListener` so a failed assertion can no longer leak the button (or its listener) into later tests.

chornonoh-vova (dismissed):

packages/react/src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -215,6 +215,39 @@ test('should clean up association when tooltip is no longer rendered', async ()
...
+    const { rerender } = render(<ShowTooltip />);
+    await fireEvent.focusIn(button);
+    expect(await screen.findByRole('tooltip')).toBeInTheDocument();
+
+    // Schedule the 100ms hide timeout, then unmount before it fires.
+    await fireEvent.focusOut(button);
+    rerender(<ShowTooltip show={false} />);
+
+    // Wait past TIP_HIDE_DELAY so a leaked timeout would have fired.
+    await new Promise((resolve) => setTimeout(resolve, 150));
```
anastasialanz: Any reason to not use jest fake timers? https://jestjs.io/docs/timer-mocks

frankensteinke: [🤖 AI generated] Good call — switched to jest.useFakeTimers() + jest.advanceTimersByTime(150) in 5401945d. The fake timers are scoped to the focusOut/unmount block so the earlier findByRole still works against real timers.

chornonoh-vova (approved):

anastasialanz (approved):

PR #2382, opened 2026-05-13
fix(Combobox): preserve active descendant after deselecting in multiselect

### LINKED ISSUE #2367: Multiselect Combobox: can't reselect a deselected option without navigating away first
## Problem

In a multiselect Combobox, after selecting and then deselecting an option via keyboard, the option cannot be selected again without first navigating to a different option and back.

## Steps to reproduce

1. Open a multiselect Combobox via keyboard
2. Navigate to an option and select it (Enter/Space)
3. Deselect the same option (Enter/Space again)
4. Try to select it again (Enter/Space)
5. **Actual:** Nothing happens — the option doesn't get selected
6. **Expected:** The option should toggle back to selected
7. **Workaround:** Navigate away to a different option, then navigate back — it can now be selected again

## Verified on

Reproduced on the hosted [Cauldron component library site](https://cauldron.dequelabs.com/components/Combobox#multiselect).

https://github.com/user-attachments/assets/1d5a89ac-2e42-4dca-87f9-5ea8d5aadbb6

## Impact

Keyboard-only users hit a dead end when trying to reselect a just-deselected option. This is a WCAG 2.1.1 (Keyboard) concern — all functionality should be operable via keyboard without requiring extra steps.

Bracciata:
## Summary
- In a multiselect Combobox, deselecting the only selected option via keyboard set `activeDescendant` to `null`, which caused the next Enter keypress on the input to be swallowed by the early return in `handleKeyDown`. Keyboard-only users hit a dead end trying to re-select the just-deselected option (WCAG 2.1.1 concern).
- The `useEffect` that syncs `activeDescendant` to the last-selected value now skips clobbering it in multiselect when the existing active descendant is still a valid matching option. The listbox tracks its own active option via keyboard navigation, so we follow that rather than reset it.

## Test plan
- [x] New regression test `should allow reselecting a deselected option in multiselect via keyboard` — verified to fail on `assertOptionIsActive` after deselect without the fix, passes with it
- [x] Full Combobox + Listbox suites green (141 tests)
- [x] Lint clean
- [x] Manual: open https://cauldron.dequelabs.com/components/Combobox#multiselect locally, navigate via keyboard, select an option, deselect it, re-select it without arrowing away

Fixes #2367

### DIFF
--- a/packages/react/src/components/Combobox/Combobox.test.tsx
+++ b/packages/react/src/components/Combobox/Combobox.test.tsx
@@ -1100,6 +1100,41 @@ test('should handle multiple selections with "enter" keydown event', () => {
   fireEnterKeyPress();
 });
 
+test('should allow reselecting a deselected option in multiselect via keyboard', () => {
+  const onSelectionChange = spy();
+  render(
+    <Combobox label="label" multiselect onSelectionChange={onSelectionChange}>
+      <ComboboxOption>Apple</ComboboxOption>
+      <ComboboxOption>Banana</ComboboxOption>
+      <ComboboxOption>Cantaloupe</ComboboxOption>
+    </Combobox>
+  );
+  const combobox = screen.getByRole('combobox');
+
+  // Note: Combobox forwards events to Listbox via dispatchEvent, but this doesn't
+  // work correctly within jsdom so we fire the events directly on listbox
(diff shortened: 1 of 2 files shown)


rheisler-deque (approved):
Very nice, working as expecting in Amplify

PR #2391, opened 2026-05-14
chore(figma): make Notice body dynamic in Code Connect

frankensteinke:
## Summary
- The Notice Code Connect example was hardcoded to render `Body` as children, so Figma always previewed the same string regardless of the selected variant.
- Wires `children` to the `Text` layer via `figma.textContent('Text')` so the previewed body matches the design.

## Test plan
- [ ] Open the Notice component in Figma and confirm the Code Connect preview shows the actual body text from the selected variant.

### DIFF
--- a/packages/react/src/components/Notice/Notice.figma.tsx
+++ b/packages/react/src/components/Notice/Notice.figma.tsx
@@ -14,11 +14,12 @@ figma.connect(Notice, FIGMA_URL, {
     variant: figma.enum('Size', {
       Small: 'condensed'
     }),
-    title: figma.textContent('Title')
+    title: figma.textContent('Title'),
+    children: figma.textContent('Text')
   },
-  example: ({ type, variant, title }) => (
+  example: ({ type, variant, title, children }) => (
     <Notice type={type} variant={variant} title={title}>
-      Body
+      {children}
     </Notice>
   )
 });


packages/react/src/components/Notice/Notice.figma.tsx
```diff
@@ -14,11 +14,12 @@ figma.connect(Notice, FIGMA_URL, {
     variant: figma.enum('Size', {
       Small: 'condensed'
     }),
-    title: figma.textContent('Title')
+    title: figma.textContent('Title'),
+    children: figma.textContent('Text')
```
rheisler-deque, line 18: I don't know much about Figma Connect, but Claude tells me that this assumes the body layer in the figma component is literally named "Text". Is that true? If so, is that a safe assumption to make?

rheisler-deque (commented):
Looks reasonable, but I asked a question I would need an answer to in order to approve

packages/react/src/components/Notice/Notice.figma.tsx
frankensteinke: That is correct -- it is named "Text".



I updated the connection after adding this mapping and I'm seeing the notice's body text in the code connect markup, which is what we want.



Here's the link to the component in Figma: https://www.figma.com/design/CEFVdiecqDjLSjhorjHUzI/Product-Foundations--Cauldron--Library?node-id=235-3419&m=dev

rheisler-deque: Nice, thanks for connecting the dots for me!

rheisler-deque (approved):

PR #2392, opened 2026-05-14
chore: add Storybook stories for 10 more components

anastasialanz:
## Summary

Adds Storybook stories and Storybook playground links to component docs for 10 more components:

- Accordion
- Address
- Breadcrumb
- Code
- CopyButton
- DescriptionList
- EmptyState
- Icon
- Line
- PageHeader

Each component gets an `index.stories.tsx` covering its main variants/states with `argTypes` configured for Storybook controls, plus a Storybook playground link added to its corresponding `docs/pages/components/*.mdx`.

## Test plan

- [ ] `yarn storybook` — verify each new story renders and controls behave (variants, states, compound subcomponents)
- [ ] Verify the Storybook playground link on each updated docs page opens the correct story


frankensteinke:
This fails coverage, presumably because `.stories.tsx` are factoring into the coverage score. I created https://github.com/dequelabs/cauldron/pull/2393 to fix this; you should be able to update your branch and pass the failing CI after that.

packages/react/src/components/EmptyState/index.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Button variant="secondary">Import</Button>
+      </>
+    )
+  }
+};
+
+export const WithSecondaryActions: Story = {
+  args: {
+    secondaryActions: <>Need help? Contact us!</>
+  }
```
rheisler-deque, line 42: I don't understand, this is just text, not actions. I see the same thing in the [Cauldron Empty State docs page](https://cauldron.dequelabs.com/components/EmptyState#with-secondary-actions).

Should the secondary actions be buttons or some other interactive element? If so, can you update the docs page also?

packages/react/src/components/Accordion/index.stories.tsx
rheisler-deque, line 1: None of these toggle when you click them in the storybook page, except for the Controlled example.

rheisler-deque (changes requested):
Two small issues, push back if they're not a problem for this PR

PR #2396, opened 2026-05-19
fix: typescript useref type errors for react 19

chornonoh-vova:
## What

Fix `useRef` typing across 8 files to be compatible with React 19's stricter type overloads, and add tests for the previously untested `ActionList` component.

## Why

React 19 tightened the `useRef` overloads: calling `useRef<T>()` without an initial value no longer returns a `MutableRefObject<T>`, so existing patterns like `useRef<T>() as MutableRefObject<T>` become TypeScript errors under `@types/react@19`. The fix is to always supply null as the initial value, which produces a correctly typed mutable ref. This is a type-only change — runtime behaviour is unchanged.

## Changes

- `src/utils/useSharedRef.ts` — `useRef<T>()` → `useRef<T | null>(null)`; return type updated to `MutableRefObject<T | null>`; redundant cast removed
- `src/utils/useMnemonics.ts` — `useRef<T>()` → `useRef<T | null>(null)`; updated return type from `RefObject<T>` to `MutableRefObject<T | null>` (only consumer is `ActionList`); restructured outer `useEffect` to extract `listenerTarget` before the early return so TypeScript can narrow away null; removed now-unused `RefObject` import
- `src/utils/useFocusTrap.ts` — collapsed two-line `useRef` + cast into `useRef<HTMLElement | null>(null)`
- `src/components/ActionList/ActionList.tsx` — `useRef<...>()` as `MutableRefObject<...>` → `useRef<...| null>(null)`; removed now-unused `MutableRefObject` import
- `src/components/ClickOutsideListener/index.tsx` — `useRef<HTMLElement>()` → `useRef<HTMLElement | null>(null)`
- `src/components/Dialog/index.tsx` — `useRef<AriaIsolate>()` → `useRef<AriaIsolate | null>(null)`
- `src/components/IconButton/index.tsx` — `useRef() as MutableRefObject<HTMLElement>` → `useRef<HTMLElement>(null)`; removed unused `MutableRefObject` import
- `src/components/TooltipTabstop/index.tsx` — `useRef() as MutableRefObject<HTMLButtonElement>` → `useRef<HTMLButtonElement>(null)`; removed unused `MutableRefObject` import
- `src/components/ActionList/ActionList.test.tsx` (new) — adds 6 tests covering render, role prop, className, ref forwarding, `onAction` callback, and axe accessibility for the main `ActionList` component (previously had no test file)

### DIFF
--- a/packages/react/src/components/Listbox/ListboxOption.tsx
+++ b/packages/react/src/components/Listbox/ListboxOption.tsx
@@ -10,8 +10,9 @@ import type {
 
 export type ListboxValue = Readonly<string | number | undefined>;
 
-interface ListboxOptionProps
-  extends PolymorphicProps<React.HTMLAttributes<HTMLElement>> {
+interface ListboxOptionProps extends PolymorphicProps<
+  React.HTMLAttributes<HTMLElement>
+> {
   value?: ListboxValue;
   disabled?: boolean;
   selected?: boolean;
@@ -58,9 +59,9 @@ const ListboxOption = forwardRef<HTMLElement, ListboxOptionProps>(
       const element = listboxOptionRef.current;
 
       setOptions((options) => {
-        const option = { element, value: optionValue };
         // istanbul ignore next
         if (!element) return options;
+        const option = { element, value: optionValue };
 
         // Elements are frequently appended, so check to see if the newly rendered
         // element follows the last element first before any other checks
@@ -101,7 +102,7 @@ const ListboxOption = forwardRef<HTMLElement, ListboxOptionProps>(
           return;
         }
 
-        onSelect({ element: listboxOptionRef.current, value: optionValue });
+        onSelect({ element: event.target as HTMLElement, value: optionValue });
         onClick?.(event);
       },
       [optionValue, onSelect, onClick, disabled]
--- a/packages/react/src/utils/useSharedRef.ts
+++ b/packages/react/src/utils/useSharedRef.ts
@@ -15,11 +15,13 @@ import setRef from './setRef';
  *   return (<div ref={internalRef}>...</div>)
  * })
  */
-export default function useSharedRef<T>(ref: Ref<T>): MutableRefObject<T> {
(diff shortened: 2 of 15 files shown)


packages/react/src/utils/useSharedRef.ts
rheisler-deque: This comment is out of date now
```suggestion
 * forwarded ref, useSharedRef will return a MutableRefObject<T | null> that will
```

packages/react/src/components/Listbox/ListboxOption.tsx
```diff
@@ -101,7 +102,7 @@ const ListboxOption = forwardRef<HTMLElement, ListboxOptionProps>(
           return;
         }
 
-        onSelect({ element: listboxOptionRef.current, value: optionValue });
+        onSelect({ element: listboxOptionRef.current!, value: optionValue });
```
rheisler-deque: If this ref contained null at this point, clicking the option would do nothing but throw an exception in the console. I was going to suggest you put an early return guard on it like the other refs in this PR, but returning early would also be a bad user experience. It would do nothing without even throwing an error.

I would recommend we use `event.target` here. As far as I can tell, it's the same element that's contained in the ref, but it isn't nullable. If `handleClick` is running, `event.target` is defined, but the ref could still be null because of a bug or regression.

What do you think?

rheisler-deque (changes requested):
Looks great! The types are much better now, We were missing a bunch of potential null exceptions that the new type makes safer.

I requested a couple small changes

packages/react/src/utils/useSharedRef.ts
chornonoh-vova: Updated

packages/react/src/components/Listbox/ListboxOption.tsx
chornonoh-vova: Updated, calling with `event.target` now

rheisler-deque (approved):

PR #2403, opened 2026-05-21
fix(ClickOutsideListener): avoid React 19 element.ref deprecation warning

chornonoh-vova:
## What

Fix the `element.ref` deprecation warning emitted by React 19 in `ClickOutsideListener`, and add missing test coverage for ref-forwarding behavior.

## Why

React 19 moved `ref` from a special element property (`element.ref`) to a regular prop (`element.props.ref`). The previous code accessed `element.ref` directly via destructuring, which triggers a deprecation warning in React 19. A naive fix of switching to `element.props.ref` alone would silently break React 16–18 consumers — in those versions `ref` is always stripped from `element.props` during `createElement`, so consumers like `Dialog`, `Popover`, and `Drawer` would lose their child refs and break focus management.

## Changes

- `utils/getChildRef.ts` (new): extracted the version-aware ref accessor into a testable utility — uses `element.props.ref` on React 19+ and falls back to `element.ref` on React 16–18
- `utils/getChildRef.test.tsx` (new): direct unit tests for both branches using synthetic element objects; covers host elements and function components on both React 18 and React 19 paths
- `ClickOutsideListener/index.tsx`: replaced the inline version-check logic and the incorrect host-element carve-out with a call to `getChildRef`; the `typeof child.type !== 'string'` guard has been removed — `element.props.ref` is safe to access for all element types in React 19
- `ClickOutsideListener/index.test.tsx`: 6 tests added covering previously untested paths (child object ref, child callback ref, forwarded ref, shared ref node, `defaultPrevented` suppression, null children)

> [!NOTE]
> **Canary build support:** An earlier commit attempted feature-detection via `Object.prototype.hasOwnProperty.call(child.props, 'ref')` to handle `0.0.0-experimental-...` canary version strings (which parse to `0` under `parseInt`). That approach was reverted because `child.props.ref` is not part of `@types/react@18`'s element type, causing a TypeScript build error. This edge case is intentionally out of scope: `0.0.0-experimental-...` version strings predate React 19's stable release (Dec 2024). All React 19 RC and canary builds use `19.x.x-...` version strings that `parseInt` resolves correctly.

### DIFF
--- a/packages/react/src/components/ClickOutsideListener/index.tsx
+++ b/packages/react/src/components/ClickOutsideListener/index.tsx
@@ -1,8 +1,7 @@
 import React, { useRef, useEffect } from 'react';
 import setRef from '../../utils/setRef';
 import resolveElement from '../../utils/resolveElement';
-
-const isReact19Plus = parseInt(React.version, 10) >= 19;
+import { getChildRef } from '../../utils/getChildRef';
 
 export interface ClickOutsideListenerProps<
   T extends HTMLElement = HTMLElement
@@ -55,14 +54,8 @@ function ClickOutsideListener(
     childElementRef.current = node;
     // Ref for this component should pass-through to the child node
     setRef(ref, node);
-    // Forward the child's own ref to the cloned node.
-    // React 19 exposes ref via props for function components; host elements (string type)
-    // still use element.ref — accessing props.ref on host elements triggers a React warning.
-    const child = children as React.ReactElement;
-    const childRef =
-      isReact19Plus && typeof child.type !== 'string'
-        ? child.props.ref
-        : (child as any).ref;
+    // Forward the child's own ref. In React 19 ref lives in props; in 16–18 it's element.ref.
+    const childRef = getChildRef(children as React.ReactElement);
     setRef(childRef, node);
   };
 
--- a/packages/react/src/components/ClickOutsideListener/index.tsx
+++ b/packages/react/src/components/ClickOutsideListener/index.tsx
@@ -54,8 +54,12 @@ function ClickOutsideListener(
     // Ref for this component should pass-through to the child node
     setRef(ref, node);
     // If child has its own ref, we want to update
-    // its ref with the newly cloned node
-    const { ref: childRef } = children as any;
+    // its ref with the newly cloned node.
+    // React 19 moved ref into element.props; React 16-18 stores it on the element object.
+    const childRef =
+      parseInt(React.version, 10) >= 19
+        ? (children as React.ReactElement).props.ref
+        : (children as any).ref;
     setRef(childRef, node);
   };
 
--- a/packages/react/src/components/ClickOutsideListener/index.tsx
+++ b/packages/react/src/components/ClickOutsideListener/index.tsx
@@ -55,11 +55,12 @@ function ClickOutsideListener(
     setRef(ref, node);
     // If child has its own ref, we want to update
     // its ref with the newly cloned node.
-    // React 19 moved ref into element.props; React 16-18 stores it on the element object.
-    const childRef =
-      parseInt(React.version, 10) >= 19
-        ? (children as React.ReactElement).props.ref
-        : (children as any).ref;
+    // Feature-detect where ref lives to avoid deprecated element.ref access
+    // on React 19-compatible canary/experimental builds.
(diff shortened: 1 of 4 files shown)


packages/react/src/components/ClickOutsideListener/index.tsx
```diff
@@ -2,6 +2,8 @@ import React, { useRef, useEffect } from 'react';
 import setRef from '../../utils/setRef';
 import resolveElement from '../../utils/resolveElement';
 
+const isReact19Plus = parseInt(React.version, 10) >= 19;
```
rheisler-deque: Heads-up: this branch is dead code in CI. Installed React is 18.3.1, so `isReact19Plus` is always false and none of the six new tests exercise the React 19 path. A future regression in that branch (mistyping `props.ref`, flipping the condition) would pass the suite.

A couple of options:
- Extract a tiny `getChildRef(child, version)` helper and unit-test both branches directly.
- Use `jest.replaceProperty(React, 'version', '19.0.0')` in a test or two to actually hit the React 19 branch, ideally with both a host `<div>` and a forwardRef child.

Not a blocker, but the React 19 behavior this PR exists to fix isn't currently locked down.

packages/react/src/components/ClickOutsideListener/index.tsx
```diff
@@ -53,9 +55,14 @@ function ClickOutsideListener(
...
-    // its ref with the newly cloned node
-    const { ref: childRef } = children as any;
+    // Forward the child's own ref to the cloned node.
+    // React 19 exposes ref via props for function components; host elements (string type)
+    // still use element.ref — accessing props.ref on host elements triggers a React warning.
+    const child = children as React.ReactElement;
+    const childRef =
+      isReact19Plus && typeof child.type !== 'string'
+        ? child.props.ref
+        : (child as any).ref;
```
rheisler-deque: The inline comment at lines 58–60 says: 
> 19 exposes ref via props for function components; host elements (string type) still use element.ref — accessing props.ref on host elements triggers a React warning.

[React 19's docs and upgrade guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#deprecated-element-ref) don't make an exception for host elements. I had claude verify this against react@19.2.6.

I found that accessing `element.props.ref` doesn't warn for host elements either. Only `element.ref` warns, and it warns the same way regardless of element type (host / function / forwardRef).

Do you have evidence to support the comment? If you've seen the warning fire from `props.ref` on a host element in some build I missed, would love to hear about it.

If not, that means host-child consumers still hit `(child as any).ref` here and emit the deprecation warning this PR is trying to suppress. This includes`Dialog`'s `<div role="dialog">` and `Drawer`'s `<div>`, `SideBar`.

I would suggest dropping the `typeof child.type !== 'string'` carve-out. That keeps the React 16–18 path unchanged and finishes the job for host children on React 19.


```suggestion
    const childRef = isReact19Plus
        ? child.props.ref
        : (child as any).ref;
```

rheisler-deque (changes requested):
Thanks for the writeup — the rationale in the PR description made this easy to follow. Two pieces of feedback inline.

One question on commit history: the second commit (fix: detect props.ref on canary builds) replaced parseInt(React.version, 10) >= 19 with Object.prototype.hasOwnProperty.call(child.props, 'ref'), addressing Copilot's note that canary/experimental React builds report React.version === '0.0.0-experimental-...' (parses to 0). The third commit (fix: build error) reverted that to version-detection. Was that intentional — is canary-build support deliberately out of scope?

chornonoh-vova:
> Thanks for the writeup — the rationale in the PR description made this easy to follow. Two pieces of feedback inline.
> 
> One question on commit history: the second commit (fix: detect props.ref on canary builds) replaced parseInt(React.version, 10) >= 19 with Object.prototype.hasOwnProperty.call(child.props, 'ref'), addressing Copilot's note that canary/experimental React builds report React.version === '0.0.0-experimental-...' (parses to 0). The third commit (fix: build error) reverted that to version-detection. Was that intentional — is canary-build support deliberately out of scope?

@rheisler-deque builds started failing after I applied the suggested fix by Copilot, that's why I reverted it.

rheisler-deque (approved):
Looks great, thanks

Issue #2402, opened 2026-05-20
Migrate to pnpm


anastasialanz:
Migrate from `yarn` to `pnpm`.

anastasialanz:
Done in https://github.com/dequelabs/cauldron/pull/2411

PR #2410, opened 2026-05-27
build: convert ad-hoc monorepo to yarn workspaces

frankensteinke:
Relates to: #2402

Converts cauldron from an ad-hoc monorepo to proper yarn workspaces (single root lockfile, `workspaces` field, one-pass CI install) as a prerequisite for the pnpm migration. This is done prior to the pnpm migration, as recommended per the [pnpm-migrate skill](https://github.com/dequelabs/dequelabs-claude-marketplace/blob/main/plugins/dequelabs/skills/pnpm-migrate/SKILL.md).

The screenshot Docker harness had to be rewritten to install the whole workspace after this change, since it used to rely on `packages/react/node_modules` being self-contained. The CI dependencies composite action also got simplified — its three `root`/`packages-react`/`packages-styles` inputs collapsed to a single `install` toggle with all six callers updated.

Hoisting also surfaced two small things that had been masked by the isolated installs:
- `packages/react/jest.config.ts` — the `react-dom/server` mapping switched from a hardcoded path to `require.resolve` so it works regardless of where react-dom installs.
- Added a `typecheck` script inside `packages/react`. yarn v1 resolves bare-binary `yarn tsc` from the workspace root (it picked up root's TS 6.0.2 instead of react's 5.0.4); defined scripts resolve from the package's own bin.

## Verification

- Local: unit tests (1145/1145), typecheck, lint, both package builds, screenshots
- All CI checks on this PR pass
- Built both packages on this branch and on develop and hashed the trees — `lib/` and `dist/` are byte-identical

Besides the screenshot job, nothing else actually broke with this change, but I also cleaned up the `update-browserslist-db` workflow since it still referenced the now-deleted `packages/*/yarn.lock` and would have produced a malformed PR next month.


chornonoh-vova (dismissed):
Looks great!

chornonoh-vova (approved):

rheisler-deque (approved):
Kudos from Claude: "Verification is unusually rigorous. Author ran unit tests (1145/1145), typecheck, lint, both package builds, and screenshots locally, plus all CI checks pass — and on top of that, hashed lib/ and dist/ against develop to prove the published artifacts are byte-identical." I think you really impressed it 😂 

Looks great, tests all pass for me. Gotta love a PR that removes code and complexity!

frankensteinke:
> Kudos from Claude: "Verification is unusually rigorous. Author ran unit tests (1145/1145), typecheck, lint, both package builds, and screenshots locally, plus all CI checks pass — and on top of that, hashed lib/ and dist/ against develop to prove the published artifacts are byte-identical." I think you really impressed it 😂
> 
> Looks great, tests all pass for me. Gotta love a PR that removes code and complexity!

I used Claude to do this work, so this is giving "Obama medaling himself." 🏅😆

PR #2414, opened 2026-05-28
fix: resolve infinite update loop in ActionMenu after React 19 upgrade

chornonoh-vova:
## What
Fix "Maximum update depth exceeded" infinite update loop in `ActionMenu`
introduced by the React 19 upgrade.

## Why
React 19's stricter effect scheduling exposed a latent bug in the
bidirectional `activeOption` sync between `Listbox` and `ActionList`.
`ListboxOption.handleClick` was passing `event.target` (which can be a
child element inside the option) to `onSelect`, so element identity checks
in the sync loop never matched the registered option element, causing
`setActiveOption` to keep firing each render cycle until React hit the
update depth limit.

## Changes
- **`ListboxOption.tsx`**: `handleClick` now passes `listboxOptionRef.current`
(the actual option element) instead of `event.target`, fixing the element
identity mismatch that drove the loop
- **`Listbox.tsx`**: controlled-active-option sync effect compares by
`.element` identity rather than object reference, making the guard
resilient to cases where different wrapper objects represent the same
DOM node
- **`ActionList.tsx`**: extracted `useMnemonics` `onMatch` into a stable
memoized callback; uses a functional state update with an element equality
guard to skip no-op updates when the same element is already active
- **`ActionMenu.test.tsx`**: added regression test covering the scenario of
clicking different items across multiple menu opens


packages/react/src/components/Listbox/ListboxOption.tsx
```diff
@@ -102,7 +102,7 @@ const ListboxOption = forwardRef<HTMLElement, ListboxOptionProps>(
           return;
         }
 
-        onSelect({ element: event.target as HTMLElement, value: optionValue });
+        onSelect({ element: listboxOptionRef.current!, value: optionValue });
```
rheisler-deque, line 105: I remember suggesting this assuming that `event.target` was the same things as the ref. Sorry about that, I didn't realize it could be a child and cause this problem.

rheisler-deque (approved):

rheisler-deque:
@chornonoh-vova The Amplify issue is not related to this code. Ideally, we would merge #2411 first, pull it into this PR, and make sure Amplify works. If you want to merge this one before that, go ahead.

PR #2431, opened 2026-06-08
fix(ci): pass --no-git-checks to pnpm version in canary publishes

frankensteinke:
The canary release publishes two packages in a row. The first one's version bump leaves the git tree dirty, which made pnpm version refuse to run for the second package (it blocks on a dirty tree by default — npm didn't).

To fix this, I added `--no-git-checks` to the pnpm version calls so we don't bail on the dirty tree.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,9 @@ jobs:
           pnpm build:react
           pnpm build:styles
       # The canary jobs mutate package.json's version before publishing, so
-      # pass --no-git-checks; pnpm publish otherwise refuses on a dirty tree.
+      # pass --no-git-checks to both pnpm version and pnpm publish — each
+      # refuses on a dirty tree by default, and the first package's version
+      # bump leaves the tree dirty for the second.
       - name: Publish @deque/cauldron-styles
         working-directory: packages/styles
         run: |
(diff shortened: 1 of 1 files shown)


rheisler-deque (approved):
This solves the issue for canary.

`release.yml` runs `pnpm publish` without `--not-git-checks`, which means it technically could have the same problem if the build steps made the tree dirty. I think we have good evidence that the build steps don't do that - the canary failure logs say `pnpm version` succeeded after `pnpm build:react` and `pnpm build:styles` ran.

So I think we're good, but I'm calling it out in case  you think that could be a problem, or you want to make sure we keep an eye on this issue next time we release.

frankensteinke:
> This solves the issue for canary.
> 
> `release.yml` runs `pnpm publish` without `--not-git-checks`, which means it technically could have the same problem if the build steps made the tree dirty. I think we have good evidence that the build steps don't do that - the canary failure logs say `pnpm version` succeeded after `pnpm build:react` and `pnpm build:styles` ran.
> 
> So I think we're good, but I'm calling it out in case you think that could be a problem, or you want to make sure we keep an eye on this issue next time we release.

@rheisler-deque this should be safe for `release.yml` -- I confirmed that the build steps leave the working tree clean (`buildIconTypes` does generate `types.ts` deterministically, but lib/dist are ignored). Keeping it without `--no-git-checks` on purpose could be a useful guard to catch a genuinely dirty tree.

anastasialanz (approved):

PR #2434, opened 2026-06-09
fix: add accessible dialog semantics to Drawer

### LINKED ISSUE #2428: Drawer: missing role="dialog" and automatic accessible name
The `Drawer` component renders a plain `<div>` with no `role` attribute and no automatic accessible name. Callers who pass `aria-label` or `aria-labelledby` via props are labeling an anonymous div — screen readers ignore that label because there is no role to attach it to. Screen readers do not announce entry into a dialog region when the drawer opens, and virtual-cursor users have no landmark or structural cue that they are in a bounded modal area.

The component already does the right things for containment: `AriaIsolate` hides the rest of the page from AT while open, and `useFocusTrap` constrains keyboard focus. The gap is role announcement and accessible naming.

## Acceptance criteria

- The `Drawer` root `<div>` has `role="dialog"` when rendered.
- The `Drawer` root `<div>` has `aria-modal="true"` when `behavior === 'modal'` (the default).
- The `Drawer` automatically receives an accessible name via `aria-labelledby`, following the same pattern used by `Dialog`:
  - An auto-generated heading ID (via `useId`) is created by `Drawer` and passed through a `DrawerContext`.
  - A `DrawerHeading` sub-component reads the ID from context and applies it as its own `id`, so `aria-labelledby` on the root element always references the visible heading text.
  - A dev-mode warning is raised (matching the existing `Dialog` pattern) if the drawer is open but no `DrawerHeading` is present inside it.
- When the `Drawer` opens, focus moves to the heading rather than the drawer container, matching `Dialog`. Focusing the container draws a focus ring around the entire drawer (which looks bad) and is a weaker announcement; focusing the heading fixes both. The `DrawerHeading` takes `tabIndex={-1}` and a ref carried through `DrawerContext` (like `DialogContext.headingRef`), which `Drawer` passes to `useFocusTrap` as its `initialFocusElement`.

## References

- WAI-ARIA Authoring Practices — [Dialog (Modal) Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)
- `Dialog` component implementation of the same pattern: `packages/react/src/components/Dialog/index.tsx`
- Resolved for `Dialog` in #1259

remove334:
## Summary

- gives Drawer dialog content accessible dialog semantics
- wires the dialog heading context/export so the drawer has an accessible name
- adds focused tests for the dialog/heading behavior

Closes #2428.

## Test plan

- `git diff --check`
- Full test run was not available in this local environment because dependency installation/linking was blocked by network/tooling issues.

### DIFF
--- a/packages/react/src/components/Drawer/DrawerContext.tsx
+++ b/packages/react/src/components/Drawer/DrawerContext.tsx
@@ -1,0 +1,20 @@
+import React, { createContext, useContext } from 'react';
+
+interface DrawerContextValue {
+  headingId: string;
+}
+
+const DrawerContext = createContext<DrawerContextValue | null>(null);
+
+function useDrawerContext(): DrawerContextValue {
+  const context = useContext(DrawerContext);
+  if (!context) {
+    throw new Error(
+      'Drawer compound components must be rendered within a Drawer'
+    );
+  }
+  return context;
+}
+
+export { DrawerContext, useDrawerContext };
+export type { DrawerContextValue };
--- a/packages/react/src/components/Drawer/index.tsx
+++ b/packages/react/src/components/Drawer/index.tsx
@@ -4,10 +4,12 @@ import React, {
   useState,
   useEffect,
   useCallback,
+  useMemo,
   useRef
 } from 'react';
 import { createPortal } from 'react-dom';
 import classnames from 'classnames';
+import { useId } from 'react-id-generator';
 import Scrim from '../Scrim';
 import ClickOutsideListener from '../ClickOutsideListener';
 import useEscapeKey from '../../utils/useEscapeKey';
@@ -16,6 +18,11 @@ import useFocusTrap from '../../utils/useFocusTrap';
 import resolveElement from '../../utils/resolveElement';
 import AriaIsolate from '../../utils/aria-isolate';
 import { isBrowser } from '../../utils/is-browser';
+import {
+  DrawerContext,
+  useDrawerContext,
+  type DrawerContextValue
+} from './DrawerContext';
 
 export interface DrawerProps<
   T extends HTMLElement = HTMLElement
@@ -50,6 +57,7 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
   ) => {
     const drawerRef = useSharedRef(ref);
     const openRef = useRef(!!open);
+    const [headingId] = useId(1, 'drawer-title-');
     const { initialFocus: focusInitial, returnFocus: focusReturn } =
       focusOptions;
     const [isTransitioning, setIsTransitioning] = useState(!!open);
@@ -115,6 +123,24 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
       returnFocusElement: focusReturn
     });
 
+    useEffect(() => {
+      if (open && drawerRef.current) {
+        const hasHeading = drawerRef.current.querySelector('.Drawer__heading');
+        if (process.env.NODE_ENV !== 'production' && !hasHeading) {
+          throw Error(
+            'Drawer: No heading provided. Include a DrawerHeading component for accessibility.'
+          );
+        }
+      }
+    }, [open]);
+
+    const contextValue: DrawerContextValue = useMemo(
+      () => ({
+        headingId
+      }),
+      [headingId]
+    );
+
     const portalElement = resolveElement(portal);
 
     return createPortal(
@@ -127,6 +153,7 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
         >
           <div
             ref={drawerRef}
+            role="dialog"
             className={classnames(className, 'Drawer', {
               'Drawer--open': !!open,
               'Drawer--top': position === 'top',
@@ -135,6 +162,8 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
               'Drawer--right': position === 'right'
             })}
             aria-hidden={!open || undefined}
+            aria-modal={isModal ? true : undefined}
+            aria-labelledby={headingId}
             style={{
               visibility: !open && !isTransitioning ? 'hidden' : undefined,
               ...style
@@ -142,7 +171,9 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
             tabIndex={open ? -1 : undefined}
             {...props}
           >
-            {children}
+            <DrawerContext.Provider value={contextValue}>
+              {children}
+            </DrawerContext.Provider>
           </div>
         </ClickOutsideListener>
         <Scrim show={!!open && !!isModal} />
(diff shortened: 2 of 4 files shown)


packages/react/src/components/Drawer/index.tsx
```diff
@@ -115,6 +123,24 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
+    useEffect(() => {
+      if (open && drawerRef.current) {
+        const hasHeading = drawerRef.current.querySelector('.Drawer__heading');
+        if (process.env.NODE_ENV !== 'production' && !hasHeading) {
+          throw Error(
+            'Drawer: No heading provided. Include a DrawerHeading component for accessibility.'
+          );
+        }
+      }
+    }, [open]);
```
rheisler-deque, line 135: Issue (blocking): This throws whenever there's no `DrawerHeading`. That breaks two cases:
1. an existing drawer with no heading crashes on upgrade, and 
2. a drawer correctly named with `aria-label` gets flagged anyway. The drawer still renders, so this should warn, not crash.

Suggested change — accept any accessible name, and use `console.error`:

```tsx
useEffect(() => {
  if (process.env.NODE_ENV === 'production' || !open || !drawerRef.current) return;
  const hasHeading = !!drawerRef.current.querySelector('.Drawer__heading');
  if (!hasHeading && !ariaLabel && !ariaLabelledby) {
    console.error(
      'Drawer: no accessible name. Add a DrawerHeading, a `heading` prop, or an `aria-label`.'
    );
  }
}, [open, ariaLabel, ariaLabelledby]);
```

`ariaLabel`/`ariaLabelledby` come from the destructure in the next comment. Keep `throw` only for the `useDrawerContext` case below, where nothing can render.

packages/react/src/components/Drawer/index.tsx
```diff
@@ -135,14 +162,18 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
               'Drawer--right': position === 'right'
             })}
             aria-hidden={!open || undefined}
+            aria-modal={isModal ? true : undefined}
+            aria-labelledby={headingId}
```
rheisler-deque, line 166: Issue (blocking): `aria-labelledby` is always set to `headingId`. If there's no `DrawerHeading`, it points at an id that doesn't exist. And if the caller passes `aria-label`, the drawer ends up with both a real label and a dangling reference.

Suggested change — read the label props and set it conditionally:

```tsx
const { 'aria-label': ariaLabel, 'aria-labelledby': ariaLabelledby, ...rest } = props;
const labelledBy = ariaLabelledby ?? (ariaLabel ? undefined : headingId);
// on the div:  aria-label={ariaLabel}  aria-labelledby={labelledBy}  {...rest}
```

Now a caller's `aria-labelledby` wins, an `aria-label` names the dialog cleanly, and the auto id is the fallback.

packages/react/src/components/Drawer/index.tsx
```diff
@@ -156,4 +187,31 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
 
 Drawer.displayName = 'Drawer';
 
+export interface DrawerHeadingProps extends React.HTMLAttributes<HTMLHeadingElement> {
+  children: React.ReactNode;
+  className?: string;
+  level?: number;
+}
```
rheisler-deque, line 194: `level` is typed as `number`, so `level={7}` would render an invalid `<h7>`. Narrowing it to `1 | 2 | 3 | 4 | 5 | 6` would catch that at compile time. `DialogHeading` has the same issue, but while were adding this, let's fix it here.

packages/react/src/components/Drawer/index.tsx
```diff
@@ -156,4 +187,31 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
+  return (
+    <HeadingLevel
+      className={classnames('Drawer__heading', className)}
+      id={headingId}
+      {...other}
+    >
+      {children}
+    </HeadingLevel>
+  );
+};
```
rheisler-deque, line 213: No test asserts the heading's level. If the default or the `h${level}` logic regressed, the suite would still pass. Please add a test that checks the rendered level for both the default and a passed-in `level`.

packages/react/src/components/Drawer/DrawerContext.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+function useDrawerContext(): DrawerContextValue {
+  const context = useContext(DrawerContext);
+  if (!context) {
+    throw new Error(
+      'Drawer compound components must be rendered within a Drawer'
+    );
+  }
+  return context;
+}
```
rheisler-deque, line 17: Please add a test to check that this guard throws if `DrawerHeading` is rendered outside a `Drawer`.

packages/react/src/components/Drawer/index.tsx
```diff
@@ -156,4 +187,31 @@ const Drawer = forwardRef<HTMLDivElement, DrawerProps>(
...
+}
+
+const DrawerHeading = ({
+  children,
+  className,
+  level = 2,
+  ...other
+}: DrawerHeadingProps) => {
+  const { headingId } = useDrawerContext();
+  const HeadingLevel = `h${level}` as 'h1';
```
rheisler-deque, line 203: When level is typed as `1 | 2 | 3 | 4 | 5 | 6`, please also remove `as 'h1';` from here. We shouldn't need the arbitrary typecast anymore because it should infer a type of `h1 | h2 | h3...`

rheisler-deque (changes requested):
Thanks for taking this on! We appreciate you helping us make Cauldron better!

> [!Note]
> I'm leaving the review below for anyone to work on whether or not the original author completes the Contributor License Agreement. We can't merge this PR until they do, but if they don't, we can recreate it. Either way, I think my feedback gets us closer to what we want.

The structure follows `Dialog` as the ticket asked for, but there are a few issues to resolve before merging. Most issues are inline below, but there are a few in this comment as well.

## Documentation
Please update the `docs/pages/components/Drawer.mdx` examples to document the naming options. As written, they'd warn and demonstrate an unnamed drawer.

## PR Name
Since this changes the public API of the component and adds functionality, the PR title should start with `feat:` instead of `fix:`. The main reason is so it will increment the minor version on next release.

## Heading Prop
Please add an optional heading prop like Dialog has, so the name requirement is discoverable from `DrawerProps`, with `DrawerHeading` kept for custom headers. It would take `string | { text: React.ReactNode; level?: number };`

When set, render a `DrawerHeading` from it inside the provider, before children:

```typescript
{heading != null && (
    <DrawerHeading level={typeof heading === 'object' ? heading.level : undefined}>
      {typeof heading === 'object' ? heading.text : heading}
    </DrawerHeading> )}
```
Use the prop or a DrawerHeading child, not both. Passing both would render two heading elements with the same ID.

## Focus issue
My team encountered an issue just last week: When the drawer opens, focus moves to the drawer container. <span id="focus-indicator-description">The focus indicator on the container creates a bad UX - it shows a pink line outside the drawer by a few pixels, and only on one side.</span> See the screenshow below.



Since this PR is about the drawer's accessibility and adding a heading to it, I think it makes sense to do as part of this PR. Could you please move focus to the heading when the drawer opens. `Dialog` does this already. The ticket didn't ask for this originally, but I've updated it to include this detail

Concretely: give `DrawerHeading` a `ref` and `tabIndex={-1}`, carry that ref through `DrawerContext` (like `DialogContext.headingRef`), and pass it to `useFocusTrap` as `initialFocusElement`. That removes the full-drawer focus ring and still announces the title on open.

PR #2444, opened 2026-06-18
ci: enforce Prettier formatting on PRs

### LINKED ISSUE #2347: Enforce Prettier formatting on PRs
## Problem

Prettier is configured (`package.json` `"prettier"` key, Prettier 3) and is run locally via `husky` + `lint-staged` on pre-commit, but it is **not enforced in CI**. The lint workflow only runs `eslint . --quiet` and a typecheck — there is no `prettier --check` step.

As a result, formatting drift slips through whenever the local hook doesn't run (e.g. `--no-verify`, commits via the GitHub web UI, husky not installed, or a contributor's editor reformats on save with a different formatter/Prettier version). This shows up as small, unrelated cosmetic diffs in otherwise-focused PRs — for example the `BaseComboboxProps` reformat in #2327.

A current run of `yarn prettier --check .` reports **132 files** out of style:

| Type | Count |
|------|------:|
| `.mdx` | 58 |
| `.md` | 42 |
| `.yml` / `.yaml` | 13 |
| `.tsx` | 8 |
| `.ts` | 3 |
| `.css` | 3 |
| other (`.mjs`, `.html`, `.babelrc`) | 5 |

Most of the volume is in `docs/` and `vpats/`; only ~11 source files are affected.

## Proposal

Split into two PRs so the mechanical diff is easy to review in isolation:

1. **One-time format sweep PR**: run `yarn prettier --write .` (scoped per the open question below) and commit the result.
2. **CI enforcement PR**: extend `.github/workflows/lint.yml` so PRs fail when files are unformatted.

   ```yaml
   - run: yarn lint
   - run: yarn prettier --check .
   - name: typecheck
     ...
   ```

   Optionally add a `"format:check": "prettier --check ."` script to `package.json` and call that instead, to mirror the `lint` script.

After both land, contributors who skip the local hook will get a clear CI signal instead of merging drift.

## Open question

Should the one-time sweep include `vpats/*.md` and `docs/**/*.mdx`, or should those be added to `.prettierignore` instead? They make up most of the 132 files and may have intentional formatting (e.g. tables in VPATs). Worth a quick look before formatting them wholesale.

frankensteinke:
## Summary

Final PR of the 3-PR stack for #2347. Turns on CI enforcement now that the tree is formatted.

- Add `"format:check": "prettier --check ."` to `package.json`.
- Add a `format check` step to `.github/workflows/lint.yml` (runs after `lint`, before `typecheck`).
- Broaden the existing `fmt` script to `prettier --write .` so the write/check pair covers the same files — the old narrow glob skipped `.mdx`/`.yml`, which is how formatting drifted in the first place.
- Add the generated `packages/react/src/components/Icon/types.ts` to `.prettierignore`. It's already gitignored in a nested `packages/react/.gitignore`, but Prettier (run from the repo root) does not read nested `.gitignore` files, so it must be listed explicitly to avoid a false `format:check` failure when a contributor has built locally.

After this lands, contributors who skip the local husky/lint-staged hook (`--no-verify`, web UI edits, different editor formatter) get a clear CI failure instead of merging drift.

## Stack

1. #2442 — ignore setup ✅ merged
2. #2443 — format sweep — _base: `develop`_
3. **👉 part 3 (this PR)** — CI enforcement — _base: `chore/prettier-format-sweep`_

Closes #2347


.prettierignore
```diff
@@ -2,6 +2,11 @@
 CHANGELOG.md
 pnpm-lock.yaml
 
+# Generated by packages/react/scripts/buildIconTypes.js during the build.
+# It is gitignored in packages/react/.gitignore, but Prettier (run from the
+# repo root) does not read nested .gitignore files, so it is listed here too.
+packages/react/src/components/Icon/types.ts
+
```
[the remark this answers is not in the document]
frankensteinke: `dist` at any depth is gitignored already (https://github.com/dequelabs/cauldron/blob/develop/.gitignore#L1), which Prettier honors.

.github/workflows/lint.yml
```diff
@@ -12,6 +12,8 @@ jobs:
       - uses: actions/checkout@v6
       - uses: ./.github/actions/dependencies
       - run: pnpm lint
+      - name: format check
+        run: pnpm format:check
```
rheisler-deque, line 17: Suggestion: The format check step runs after `pnpm lint`. GitHub Actions stops a job at the first failing step, so a lint failure means format check never runs.

I'd suggest adding `if: always()` so the format check reports on the same run regardless of lint's result:

```suggestion
      - name: format check
        if: always()
        run: pnpm format:check
```

rheisler-deque (dismissed):
Looks great! One suggestion for CI ergonomics.

.github/workflows/lint.yml
frankensteinke: I think this is a good call and a worthwhile change. I went ahead and made it. Thanks!

rheisler-deque (approved):

Issue #2448, opened 2026-06-23
MenuItem/TopBarItem props with defaultProps are typed as required, breaking ComponentProps


frankensteinke:
### Version
`@deque/cauldron-react@6.25.2`

### Description
`MenuItemProps` declares `menuItemRef`, `onClick`, and `onKeyDown` as **required**, even though all three are supplied via `defaultProps`:

```ts
// lib/components/MenuItem/index.d.ts
interface MenuItemProps extends React.HTMLAttributes<HTMLLIElement> {
  children: React.ReactNode;
  menuItemRef: React.Ref<HTMLLIElement>;
  onClick: (e: React.MouseEvent<HTMLLIElement>) => void;
  onKeyDown: (e: React.KeyboardEvent<HTMLLIElement>) => void;
  autoClickLink?: boolean;
}

// MenuItem class
static defaultProps = { menuItemRef, onClick, onKeyDown, autoClickLink };
```

Using `<MenuItem>`/`<TopBarItem>` directly in JSX works, because React's `LibraryManagedAttributes` makes defaulted props optional at the JSX call site. But anyone who derives the prop type with `React.ComponentProps<typeof TopBarItem>` — e.g. to build a wrapper that forwards props — gets the raw interface where those props are still required.

### Reproduction
```ts
import { ComponentProps } from "react";
import { TopBarItem } from "@deque/cauldron-react";

type Forwarded = ComponentProps<typeof TopBarItem>;

// Error: Property "menuItemRef" is missing… (also onClick, onKeyDown)
const props: Forwarded = { children: "x" };
```

A thin wrapper that spreads forwarded props onto `TopBarItem` is forced to wrap everything in `Partial<…>` to compile, which also (incorrectly) makes genuinely-required props like `children` optional.

### Expected
Props that have a `defaultProps` entry should be optional in the interface:

```ts
interface MenuItemProps extends React.HTMLAttributes<HTMLLIElement> {
  children: React.ReactNode;
  menuItemRef?: React.Ref<HTMLLIElement>;
  onClick?: (e: React.MouseEvent<HTMLLIElement>) => void;
  onKeyDown?: (e: React.KeyboardEvent<HTMLLIElement>) => void;
  autoClickLink?: boolean;
}
```

Then `ComponentProps<typeof TopBarItem>` reflects actual usage and wrappers don't need `Partial`.

PR #2460, opened 2026-06-26
fix(ActionMenu): parameterize trigger element type

### LINKED ISSUE #2449: ActionMenu trigger ref/handlers hardcoded to HTMLButtonElement, contradicting documented TopBar/MenuBar nesting
### Version
`@deque/cauldron-react@6.25.2`

### Description
`ActionMenu`'s function-trigger props type the `ref` and event handlers against `HTMLButtonElement`:

```ts
// lib/components/ActionMenu/ActionMenu.d.ts
type ActionMenuTriggerProps =
  Pick<React.HTMLAttributes<HTMLButtonElement>,
    "children" | "onClick" | "onKeyDown" | "aria-expanded" | "aria-haspopup" | "aria-controls">
  & { ref: React.RefObject<HTMLButtonElement> };
```

However, the `renderInTrigger` prop is explicitly documented for the case where the trigger is **not** a button:

> Controls whether the menu should render as a child of the trigger… Intended for use with nested menu patterns, for example when an `ActionMenu` is nested inside a `TopBar`/`MenuBar`.

In that documented pattern the trigger is a `MenuItem`/`TopBarItem`, which renders an `<li>` and takes `menuItemRef: React.Ref<HTMLLIElement>`. So consumers following the documented nesting are forced to cast:

```tsx
<ActionMenu
  renderInTrigger
  trigger={({ ref, children, ...props }) => (
    <>
      <TopBarItem
        // ref is RefObject<HTMLButtonElement>, but it attaches to the <li>
        menuItemRef={ref as unknown as React.RefObject<HTMLLIElement>}
        {...props}  // onClick/onKeyDown typed for button, not li
      >
        …
      </TopBarItem>
      {children}
    </>
  )}
>
```

Both the `ref` cast (via `unknown`, since button/li don't overlap) and a cast on the spread handlers are required.

### Expected
Make the trigger element type a parameter, defaulting to `HTMLButtonElement` so existing usage is unchanged:

```ts
type ActionMenuTriggerProps<E extends HTMLElement = HTMLButtonElement> =
  Pick<React.HTMLAttributes<E>,
    "children" | "onClick" | "onKeyDown" | "aria-expanded" | "aria-haspopup" | "aria-controls">
  & { ref: React.RefObject<E> };
```

Then the documented `TopBar`/`MenuBar` nesting (`<li>` trigger) is expressible without casts, and the common button case is untouched.

### Workaround
Documented `as unknown as RefObject<HTMLLIElement>` cast at the trigger, called out in a comment.

### Related
Surfaces alongside #2448 (defaulted `MenuItem` props typed as required) when wrapping `TopBarItem` in an `ActionMenu` trigger.

frankensteinke:
Closes #2449

`ActionMenu`'s function-trigger props typed the `ref` and event handlers against `HTMLButtonElement`, which contradicted the documented `renderInTrigger` nesting pattern where the trigger is a `MenuItem`/`TopBarItem` (an `<li>`). Consumers following that pattern were forced to cast the ref via `as unknown as RefObject<HTMLLIElement>` and cast the spread handlers.

## Changes

- Made `ActionMenuTriggerProps`, `ActionMenuTriggerFunction`, and `ActionMenuProps` generic over the trigger element type `E extends HTMLElement`, defaulting to `HTMLButtonElement` so existing usage is unchanged.
- The documented `TopBar`/`MenuBar` nesting is now expressible without casts via `ActionMenu<HTMLLIElement>`; the common button case is untouched.
- The component's internal implementation is unchanged — the generic is exposed purely through the public type of the exported component (a cast on the `forwardRef` result), keeping the change additive and low-risk.
- Updated the TopBar-pattern test to use `ActionMenu<HTMLLIElement>` and dropped the `@ts-expect-error`/cast it previously required (now a regression guard).
- Updated the ActionMenu props docs to reflect the parameterized trigger type.

## Verification

- `tsc --noEmit` — clean
- ESLint — clean
- Prettier — clean
- ActionMenu test suite — 46/46 pass

## Related

Surfaces alongside #2448 when wrapping `TopBarItem` in an `ActionMenu` trigger.

### DIFF
--- a/packages/react/src/components/ActionMenu/ActionMenu.tsx
+++ b/packages/react/src/components/ActionMenu/ActionMenu.tsx
@@ -36,6 +36,7 @@ export type ActionMenuTriggerProps<E extends HTMLElement = HTMLButtonElement> =
   Pick<
     React.HTMLAttributes<E>,
     | 'children'
+    | 'id'
     | 'onClick'
     | 'onKeyDown'
     | 'aria-expanded'
--- a/packages/react/src/components/ActionMenu/ActionMenu.tsx
+++ b/packages/react/src/components/ActionMenu/ActionMenu.tsx
@@ -266,11 +266,15 @@ ActionMenuComponent.displayName = 'ActionMenu';
  * This only widens the public types of the `trigger` function; the internal
  * implementation is unaffected. See the ActionMenu docs for a usage example.
  */
-const ActionMenu = ActionMenuComponent as (<
-  E extends HTMLElement = HTMLButtonElement
->(
-  props: ActionMenuProps<E> & React.RefAttributes<HTMLElement>
-) => React.ReactElement) &
-  Pick<typeof ActionMenuComponent, 'displayName'>;
+type ActionMenuType = Omit<
+  React.ForwardRefExoticComponent<ActionMenuProps>,
+  keyof CallableFunction
+> & {
+  <E extends HTMLElement = HTMLButtonElement>(
+    props: ActionMenuProps<E> & React.RefAttributes<HTMLElement>
+  ): React.ReactElement;
+};
+
+const ActionMenu = ActionMenuComponent as ActionMenuType;
 
 export default ActionMenu;
--- a/packages/react/src/components/ActionMenu/ActionMenu.tsx
+++ b/packages/react/src/components/ActionMenu/ActionMenu.tsx
@@ -260,26 +260,11 @@ const ActionMenuComponent = forwardRef<HTMLElement, ActionMenuProps>(
 ActionMenuComponent.displayName = 'ActionMenu';
 
 /**
- * The trigger element defaults to an `HTMLButtonElement`, but the element type
- * is parameterized for nested menu patterns where the trigger is not a button —
- * for example a `MenuItem`/`TopBarItem` (`<li>`) when an `ActionMenu` is nested
- * inside a `TopBar`/`MenuBar`. The internal implementation is unaffected; this
- * only widens the public types of the `trigger` function so consumers can type
- * the trigger's `ref` and event handlers against the actual element without
- * casts. Annotating the trigger param infers the element type, so no JSX type
- * argument is required:
- *
- * ```tsx
- * <ActionMenu
- *   renderInTrigger
- *   trigger={({ ref, children, ...props }: ActionMenuTriggerProps<HTMLLIElement>) => (
- *     <TopBarItem menuItemRef={ref} {...props}>
- *       …
- *       {children}
- *     </TopBarItem>
- *   )}
- * >
- * ```
+ * The trigger element type is parameterized (defaulting to `HTMLButtonElement`)
+ * so it can be widened for nested menu patterns where the trigger is not a
+ * button — e.g. a `MenuItem`/`TopBarItem` (`<li>`) inside a `TopBar`/`MenuBar`.
+ * This only widens the public types of the `trigger` function; the internal
+ * implementation is unaffected. See the ActionMenu docs for a usage example.
  */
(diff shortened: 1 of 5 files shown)


packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -25,26 +25,33 @@ interface ActionMenuListProps {
...
+export type ActionMenuTriggerProps<E extends HTMLElement = HTMLButtonElement> =
+  Pick<
+    React.HTMLAttributes<E>,
+    | 'children'
+    | 'onClick'
+    | 'onKeyDown'
+    | 'aria-expanded'
+    | 'aria-haspopup'
+    | 'aria-controls'
+  > & {
```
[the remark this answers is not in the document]
frankensteinke: Updated!

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -249,6 +257,35 @@ const ActionMenu = forwardRef<HTMLElement, ActionMenuProps>(
...
+ *   renderInTrigger
+ *   trigger={({ ref, children, ...props }: ActionMenuTriggerProps<HTMLLIElement>) => (
+ *     <TopBarItem menuItemRef={ref} {...props}>
+ *       …
+ *       {children}
+ *     </TopBarItem>
+ *   )}
+ * >
+ * ```
+ */
```
anastasialanz, line 268: Do you think we need this entire comment in the component itself or should this be documented in the docs?

frankensteinke: I think you're right. I trimmed this quite a bit and moved a good part of it to the docs.

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -249,6 +257,20 @@ const ActionMenu = forwardRef<HTMLElement, ActionMenuProps>(
...
+ * button — e.g. a `MenuItem`/`TopBarItem` (`<li>`) inside a `TopBar`/`MenuBar`.
+ * This only widens the public types of the `trigger` function; the internal
+ * implementation is unaffected. See the ActionMenu docs for a usage example.
+ */
+const ActionMenu = ActionMenuComponent as (<
+  E extends HTMLElement = HTMLButtonElement
+>(
+  props: ActionMenuProps<E> & React.RefAttributes<HTMLElement>
+) => React.ReactElement) &
+  Pick<typeof ActionMenuComponent, 'displayName'>;
```
anastasialanz: This `Pick<typeof ActionMenuComponent, displayName>` re-derives only `displayName` and drops the rest of the `ForwardRefExoticComponent` surface (`$$typeof`, `defaultProps`, `propTypes`). `Listbox` already solves the same "generic forwardRef component" problem with the idiomatic form — consider converging on it so the library keeps one pattern for generic components:

```ts
type ActionMenuType = Omit<
  React.ForwardRefExoticComponent<ActionMenuProps>,
  keyof CallableFunction
> & {
  <E extends HTMLElement = HTMLButtonElement>(
    props: ActionMenuProps<E> & React.RefAttributes<HTMLElement>
  ): React.ReactElement;
};

const ActionMenu = ActionMenuComponent as ActionMenuType;
```

Keeps the full component surface instead of hand-picking `displayName`, and matches `Listbox`'s `Omit<ForwardRefExoticComponent<…>, keyof CallableFunction> & { <generic call sig> }` idiom. Purely type-level — no runtime change. Non-blocking nit.

frankensteinke: Done!

anastasialanz (approved):

PR #2464, opened 2026-07-02
feat(MetadataList): add MetadataList component

### LINKED ISSUE #1820: Create DefinitionList Component

chornonoh-vova:
## What

Adds the `MetadataList` component — a `dl`-based list for displaying stacked term/value pairs, laid out horizontally (default) or vertically.

## Why

We need a way to display metadata as labeled rows/columns that's visually distinct from the existing `DescriptionList` component.

Closes #1820

## Changes

- Add `MetadataList`, `MetadataListItem`, `MetadataListLabel`, and `MetadataListValue` components (`packages/react/src/components/MetadataList`)
- Add `metadata-list.css` with light/dark theme text colors, following the same token convention as `table.css`/`description-list.css`
- Wire up the new component/styles: export from `packages/react/src/index.ts` and import `metadata-list.css` in `packages/styles/index.css`
- Add unit tests, Storybook stories, and Playwright e2e screenshot tests (horizontal/vertical, light/dark)
- Add MDX documentation page (`docs/pages/components/MetadataList.mdx`)


OlehSihunov (dismissed):

shawnsharpDQ (changes requested):
For vertical list items, the gab between should be 24px

packages/styles/metadata-list.css
```diff
@@ -0,0 +1,61 @@
+:root {
+  --metadata-list-column-gap: var(--space-largest);
+  --metadata-list-row-gap: var(--space-largest);
+  --metadata-list-item-gap: var(--space-half);
+  --metadata-list-padding: var(--space-small) 0;
+  --metadata-list-label-text-color: var(--gray-90);
+  --metadata-list-label-font-size: var(--text-size-small);
+  --metadata-list-label-font-weight: var(--font-weight-medium);
+  --metadata-list-label-line-height: 20px;
```
frankensteinke:
```suggestion
  --metadata-list-label-line-height: 1.5;
```

Could we go with a unitless value here, so that it scales with font size?

docs/pages/components/MetadataList.mdx
```diff
@@ -0,0 +1,95 @@
+---
+title: MetadataList
+description: A component that renders a list of terms and descriptions, stacked and displayed horizontally in a row.
```
frankensteinke: Nit: this only refers to the default orientation (horizontal); would we want to update it to note that vertical orientation is supported as well?

packages/react/src/components/MetadataList/index.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+    expect(screen.getByText(/a/i)).toHaveClass('a');
+
+    render(<MetadataListItem className="b">b</MetadataListItem>);
+    expect(screen.getByText(/b/i)).toHaveClass('b');
+
+    render(<MetadataListLabel className="c">c</MetadataListLabel>);
+    expect(screen.getByText(/c/i)).toHaveClass('c');
+
+    render(<MetadataListValue className="d">d</MetadataListValue>);
+    expect(screen.getByText(/d/i)).toHaveClass('d');
```
frankensteinke: Nit: could be better to limit tests to one `render()` each. There's automatic cleanup that runs as a part of a global `afterEach()` that wouldn't be run within a single test block, so this approach could be a bit fragile.

Two approaches you could take instead: one render per test, or rendering a combined tree that passes class names for all of the compound components.

packages/react/src/components/MetadataList/index.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import classNames from 'classnames';
+
+export type MetadataListProps = {
+  orientation?: 'vertical' | 'horizontal';
+} & React.ComponentPropsWithoutRef<'dl'>;
+
+export function MetadataList({
```
frankensteinke: Could you wrap this and all of the compound components in `forwardRef` so that consumers can use a ref?

frankensteinke (changes requested):
I left some minor feedback inline, I think that they should all be fairly quick updates. Once these and Shawn's feedback is addressed, I think we should be good to go!

frankensteinke (approved):

shawnsharpDQ (approved):

Issue #2348, opened 2026-04-29
chore(react): migrate library build from Rollup to Vite (lib mode)


anastasialanz:
## Summary

Replace the Rollup-based build of `@deque/cauldron-react` with Vite's library mode. `packages/react` is the only package that would change — `packages/styles` (postcss-cli), the docs site (webpack), and Playwright CT (already on Vite) stay as-is.

The migration must preserve the published contract:

- `main: lib/index.js` (CJS), `types: lib/index.d.ts`, `style: lib/cauldron.css`
- Code-split chunks for the icon SVGs (driven by `` import(`./icons/${iconName}.svg`) `` in `packages/react/src/components/Icon/index.tsx`)
- Externals: `dependencies` + `peerDependencies` + `/^react-syntax-highlighter/`
- SVGR svgo config: `removeViewBox: false`, `removeDimensions`, force `width=24`/`height=24`
- `process.env.NODE_ENV` left as a runtime reference (Icon's test-mode guard depends on it)

## Proposed approach

- Add `packages/react/vite.config.ts` using `build.lib` + `@vitejs/plugin-react` + `vite-plugin-svgr` + `vite-plugin-dts`
- Swap `devDependencies`: drop `rollup`, `@rollup/plugin-typescript`, `@rollup/plugin-commonjs`, `@rollup/plugin-dynamic-import-vars`, `@svgr/rollup`; add `vite`, `@vitejs/plugin-react`, `vite-plugin-svgr`, `vite-plugin-dts`
- Update scripts: `build:lib` → `vite build`, `dev` → `vite build --watch` (CSS step unchanged)
- Delete `packages/react/rollup.config.js` after parity is verified
- Out of scope: Jest setup, `.babelrc`, `postcss.config.js`, the docs-site webpack config

## Migration risks

1. Icon dynamic imports — Vite ships `@rollup/plugin-dynamic-import-vars` internally; verify per-icon chunks are still emitted
2. SVGR output parity (both plugins wrap `@svgr/core`; spot-check one icon)
3. `process.env.NODE_ENV` must not be inlined in the lib build
4. `.d.ts` emission via `vite-plugin-dts` instead of `@rollup/plugin-typescript`
5. The `/^react-syntax-highlighter/` external regex must be preserved exactly

## Verification plan

- [ ] `yarn --cwd=packages/react build` produces `lib/` with parity to the previous Rollup output (`diff -r` the trees; bundle size within ~5%)
[...]

PR #2467, opened 2026-07-07
feat(react): support per-component deep imports for tree-shaking

### LINKED ISSUE #2465: chore(react): emit per-component modules + exports map so consumers can tree-shake
## Summary

`@deque/cauldron-react` is published as a single CommonJS barrel (`main: lib/index.js`), with **no `module` field, no `exports` map, and no `sideEffects` flag**. All components — and their transitive dependencies — are required at the top of that one file, so consumers cannot tree-shake. Importing a single component pulls in the entire library.

This got materially worse in 7.x. `react-aria-components` was added (for the new `TreeView`) and `react-syntax-highlighter` has long been there (for `Code`); both are `require`d at the top of the barrel:

```js
// lib/index.js
var SyntaxHighlighter = require('react-syntax-highlighter/dist/cjs/light');
// …language registrations…
var reactAriaComponents = require('react-aria-components');
```

So every consumer bundles TreeView's and Code's dependency trees whether or not they use those components.

## Concrete impact (axe-extension)

Upgrading axe-extension from cauldron 6.25.2 → 7.2.0 pushed `panel.bundle.js` from ~6.4 MiB to **8.22 MiB**, breaking the build's asset-size guard. The extension imports `Modal`, `Button`, `Code`, etc. — but **not** `TreeView`. Stubbing out `react-aria-components` alone (which only `TreeView` uses) dropped the bundle to **5.30 MiB** — a ~2.9 MiB reduction for a component we don't use but are forced to bundle.

## Proposal

Emit **per-component modules** and expose them via a subpath `exports` map, so consumers can import only what they use:

- Build with `output.preserveModules: true` (Rollup) / entry-per-component (Vite `build.lib`) so each component compiles to its own file instead of one barrel.
- Add an `exports` map with per-component subpaths, e.g. `@deque/cauldron-react/Modal`.
- Keep the existing barrel (`.`) export for backward compatibility.

This works **with the current CJS output** — no ESM conversion required — because deep imports let the bundler prune whole components (and their deps) that a consumer never references. Consumers opt in by switching to subpath imports (or via a codemod / `babel-plugin-transform-imports`).

## Coordinate with #2348

The Rollup → Vite migration (#2348) is already touching the build. `preserveModules` / per-entry output is a small addition there rather than a separate effort, so these should land together.

## Out of scope

Making the ergonomic barrel import (`import { Modal } from '@deque/cauldron-react'`) tree-shake automatically — that needs an ESM build + `sideEffects: false`, tracked as a follow-up.

frankensteinke:
Closes #2465

## What

Publishes `@deque/cauldron-react` as **per-component modules** behind a subpath `exports` map, so consumers can import only the components they use instead of bundling the whole library.

- **`rollup.config.js`**: build with `output.preserveModules: true` + `preserveModulesRoot: 'src'`, so each source module compiles to its own file under `lib/` (mirroring `src/`) instead of one bundled barrel. Every `src/components/*/index.{ts,tsx}` is added as an entry point so pure re-export barrels aren't hoisted/elided (which would leave a subpath resolving types but 404ing at runtime). Switched `output.exports` from `auto` → `named` for consistent deep-import ergonomics (default-export components are reached via `.default`, matching the emitted `.d.ts` + esModuleInterop) — this also silences Rollup's mixed-exports warning.
- **`package.json`**: add an `exports` map:
  - `.` → the barrel (`lib/index.js` + types), unchanged for existing consumers.
  - `./*` → `lib/components/*/index.js` (+ types), enabling `@deque/cauldron-react/Modal`.
  - `./internal` and `./internal/*` → `null`, keeping components under `src/components/internal/` off the public deep-import surface (the `./*` wildcard would otherwise expose them).
  - `./cauldron.css`, `./package.json`, and a `./lib/*` passthrough, preserving the documented `@deque/cauldron-react/lib/cauldron.css` path and every other existing deep import.
- **Default exports**: `ActionList`, `ActionMenu`, `Address`, and `DescriptionList` were named-only; they gained a default export so the documented `import X from '@deque/cauldron-react/X'` pattern is uniform across every component.
- **CI guards**: `scripts/verifyExports.js` (`verify:exports`) asserts, post-build, that every component subpath resolves at runtime, exposes a `default`, and ships its `.d.ts`, and that `internal/*` stays private; `type-tests/` (`verify:types`) type-checks deep imports under both `bundler` and `node16` resolution. Both run in the `react` CI job after build.
- **`README.md`**: document the new subpath imports under "Reducing bundle size", the directory-name stability contract, and the `moduleResolution` requirement for TypeScript.

This stays with the **current CJS output** — no ESM conversion. Deep imports let the bundler prune whole components (and their deps) a consumer never references.

## Impact

Importing `Modal` no longer drags in `react-aria-components` (TreeView) or `react-syntax-highlighter` (Code):

```
lib/components/Modal/index.js requires: tslib, react, classnames, ../Dialog
```

`react-aria-components` now only appears in `lib/components/TreeView/index.js`.

## Verification

- `pnpm build` (lib + css) runs clean, no warnings; output is one file per component under `lib/components/*/`, icon chunks co-located under `lib/components/Icon/icons/`.
- `pnpm verify:exports` — all 67 component subpaths resolve at runtime, expose a default, and ship types; the bare `internal` specifier and every internal component are blocked.
- `pnpm verify:types` — deep-import default + named imports and their types resolve under both `moduleResolution: bundler` and `node16`.
- `pnpm typecheck` passes; full react test suite (1155 tests) passes.

## Notes / scope

- Making the **ergonomic barrel** import (`import { Modal } from '@deque/cauldron-react'`) tree-shake automatically is **out of scope** (needs an ESM build + `sideEffects: false`) — tracked as a follow-up per the issue.
- The `./lib/*` back-compat passthrough still resolves `lib/components/internal/...` for anyone who already relied on a full `lib/` path; this preserves pre-`exports` behavior. The `./internal` block closes the new, clean `@deque/cauldron-react/internal/*` surface the `./*` wildcard introduced.
- Coordinates with the Rollup → Vite migration (#2348, still open). Implemented on the current Rollup build; the `preserveModules`/exports approach ports directly to Vite lib mode when that lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


frankensteinke:
Closing after [Cauldron team discussion](https://deque.slack.com/archives/C05G4UJ0M0U/p1783518779308589).

Will wait until we have time to go with the ESM approach here, since consumers will get that for free and it won't tie us to following semver for directory changes.

PR #2475, opened 2026-07-15
feat(TreeView): add opt-in virtualization via height prop

mateoviilla1:
## Summary

Adds opt-in virtualization to `TreeView` via a new `virtualized` prop.

Long lists kept **every row mounted** and re-rendered all of them on each
selection change, so a single checkbox click could freeze for ~1s while
react-aria reconciled the whole tree (surfaced downstream in axe-reports).

With `virtualized` set, the tree becomes a fixed-height scroll region wrapped in
react-aria's `Virtualizer` + `ListLayout`, so only the rows in view are rendered
and selection stays fast regardless of list length. Without it the tree grows to
fit all rows exactly as before — **the non-virtualized path is unchanged**.

```jsx
<TreeView virtualized height="20rem" items={items} selectionMode="multiple" />
```

## Details

- `virtualized` is the mode switch; `height` only sizes the scroll region. They
  were originally the same prop, which meant a height change remounted the tree
  and forced a "0"/"0px" means unset rule that never covered `"400"`, `"100%"`,
  `""` or `"auto"`.
- Selection is controlled whenever `selectionMode !== 'none'`. Toggling
  `virtualized` still remounts the tree, but selection is held in the component
  rather than inside react-aria, so the user's checkboxes survive it.
- `ListLayout` `gap`/`padding` mirror the `--space-quarter` (2px) /
  `--space-half` (4px) tokens. A test asserts the two stay equal rather than a
  comment claiming they do.
- The props type `Omit`s `children`, `role` and `dangerouslySetInnerHTML` —
  react-aria discards all three at runtime, so the type shouldn't advertise them.
- New `.TreeView--virtualized` neutralizes flex/gap/container padding, since
  virtualized rows are absolutely positioned and spacing comes from the layout.
- Composes with the existing `cascadeSelect`/`cascadeDeselect` (test + docs
  example).

## Documented limitations

Virtualized mode removes off-screen rows from the DOM, which costs a few things.
All are listed in the docs; two are worth calling out for review:

- **No horizontal scrolling.** `ListLayout` pins each row to the width of the
  scroll region, so a deeply indented or long-labelled item is clipped rather
  than reachable by scrolling sideways. The non-virtualized tree scrolls
  horizontally. **This conflicts with the TreeView RFC (#1290)**, which requires
  horizontal scrolling — that RFC needs amending to scope it out of virtualized
  mode, or this needs a custom layout. Flagging it rather than deciding it here.
- **`aria-rowindex` is not set on rows.** `useTree` sets the virtualized flag but
  `TreeItem` never forwards it, so `useGridListItem` never reaches its
  `aria-rowindex` branch. `aria-level` / `aria-posinset` / `aria-setsize` are
  correct. I could not find an existing react-spectrum issue covering this; the
  previous link pointed at the bare tracker index and has been removed pending a
  real one.

Virtualized mode also needs a definite width from an ancestor, and browser
find-in-page only matches rendered rows.

## Testing

- `pnpm --filter @deque/cauldron-react test` — 45 TreeView tests pass, incl.
  virtualized render/selection, the style merge, virtualized cascade, selection
  surviving a `virtualized` toggle in both directions, and a `jest-axe` check.
  Existing snapshots unchanged.
- The virtualized tests previously passed with the `<Virtualizer>` deleted,
  because jsdom reports zero dimensions and renders every row. They now assert
  that the virtualizer re-parents rows — a direct `[role="presentation"]` child,
  no direct rows, `aria-rowcount` set — which I verified fails when the wrapper
  is removed.
- Screenshot baselines for the two virtualized cases are committed and pass
  unchanged.
- Docs examples under **TreeView → Virtualized (Long Lists)** and **Virtualized
  with Cascade Selection**.

## Not closed by this PR

This is the library half. `axe-reports#3277` only resolves once the consumer
passes `virtualized` on the subdimensions tree, which is a separate change.

Accessibility review by @VaniChinnam raised three issues against the preview.
Two are pre-existing and unrelated to virtualization: cauldron#2517 (selected
state — closed as expected screen reader behavior, matching the same conclusion
reached in cauldron#1202 and dequelabs/axe-reports#3309) and cauldron#2519 (the
Custom onAction example, fixed separately in #2520). Neither touches code in
this diff.

**cauldron#2518 is fixed in this PR** by bumping `react-aria-components` from
1.13 to 1.20. Arrowing through a virtualized tree scrolled the nearest
scrollable ancestor on every press, marching the tree out of view.
`scrollIntoViewport` walks every scroll parent and scrolls each to reveal the
focused row, which is redundant once a nested scroller has already revealed it —
so it only affected virtualized mode. Measured on the docs page over 30
ArrowDown presses, tree scrolling 724px internally in both cases:

| `react-aria-components` | ancestor scroll |
| --- | --- |
| 1.13.0 | 550px, and still climbing |
| 1.20.0 | none |

1.20.0 rather than 1.21.0 because 1.21.0 and its `react-aria`/`react-stately`
dependencies fall inside the pnpm `minimumReleaseAge` window and the installer
refuses them.

Fallout from the bump is limited to the TreeView SSR snapshots: react-aria
stopped emitting `data-react-aria-hidden` on an internal template element and
its generated ids shifted. Full suite and all 99 screenshots pass.

QA notes: exercise a long tree with `virtualized` set — scrolling windows rows,
and checkbox clicks toggle instantly; confirm a tree without `virtualized` is
unchanged; confirm selections survive turning `virtualized` on and off.

Ref dequelabs/axe-reports#3277


mateoviilla1:
Addressed the review feedback in d14b8cc0:

**Copilot comments**
- **Typed props** — `TreeViewProps` now extends `React.HTMLAttributes<HTMLDivElement>` (matching `ProgressBar`), so `style` / `data-*` / event handlers are typed. Dropped the internal `style` cast and the `as object` cast in the test.
- **`height={0}` docs** — updated the prop description to state that `0` is treated as unset (non-virtualized), matching the implementation.

**Screenshots failure**
The `update-screenshots` label job silently committed nothing because its `git diff --quiet` check ignores *untracked* files (our baselines are brand-new), so the comparison job then failed on the missing snapshot.

While generating the baseline I found the virtualized screenshot rendered the full 200-row tree instead of a clipped scroll region. Root cause was the CT harness template (`playwright/index.html`) applying `min-height`/`max-width: max-content` to `#root > div`, which overrode the fixed `height` and collapsed the absolutely-positioned virtual rows on a top-level mounted component (a plain `<div style="height:240;overflow:auto">` with 200 children reproduced it — not a component bug). Fixed the e2e case by wrapping the tree in a div and giving it an explicit width; it now renders the fixed-height scroll region correctly. The committed baseline shows only the ~8 visible rows of a 200-item list, confirming virtualization windows as intended.

packages/react/src/components/TreeView/screenshots.e2e.tsx
```diff
@@ -86,6 +86,35 @@ test('should have screenshot for TreeView multiple selection', async ({
...
+      <TreeView
+        aria-label="Long list"
+        items={longItems}
+        selectionMode="multiple"
+        height={240}
+        style={{ width: 420 }}
+      />
+    </div>
+  );
+  await expect(component).toHaveScreenshot('tree-view-virtualized');
```
rheisler-deque, line 119: Please add an assertion that only a subset of rows renders to the DOM when the tree is virtualized. We can't test that in the unit tests because the element isn't actually laid out. The e2e case already mounts 200 items, so I think it's reasonable to assert that the number of rendered row elements is far below 200.

Right now, no test verifies that we only render the visible rows. A screenshot only captures the visible region. It looks identical whether 8 rows or 200 rows sit in the DOM. A regression that stopped windowing would pass every test in the suite, including the screenshot baseline.

packages/react/src/components/TreeView/index.tsx
```diff
@@ -36,11 +61,16 @@ const TreeView = forwardRef<HTMLDivElement, TreeViewProps>(
...
+      height,
       className,
       ...other
     },
     ref
   ) => {
+    // A zero height would produce an empty, collapsed scroll region that renders
+    // no rows, so treat `0` (a common "not measured yet" value from layout code)
+    // the same as unset and fall back to the grow-to-fit tree.
+    const isVirtualized = height !== undefined && height !== 0;
```
rheisler-deque: Suggestion: normalize zero before the virtualization check so a string like `"0"` or `"0px"` is treated the same as numeric `0`.

The guard is `height !== undefined && height !== 0`. It catches numeric `0` but not a string. `"0px"` is not strictly equal to the number `0`. The prop accepts `number | string`. Layout code that interpolates a measurement into a template string produces `"0px"` before the element is measured. That is the same not-yet-measured case the numeric guard exists to catch. In that case `isVirtualized` becomes true. The tree renders as a fixed-height scroll region of height `0`. No rows show.

packages/react/src/components/TreeView/TreeView.test.tsx
```diff
@@ -454,3 +454,103 @@ test('cascade is inert in single selection mode', async () => {
...
+test('does not virtualize when height is 0 (treated as unset)', () => {
+  const { getByRole } = render(
+    <TreeView aria-label="Test TreeView" items={items} height={0} />
+  );
+  const tree = getByRole('treegrid');
+  expect(tree).not.toHaveClass('TreeView--virtualized');
+  expect(tree).not.toHaveStyle({ height: '0px' });
+});
+
+test('merges a caller-supplied style with the virtualization height', () => {
```
rheisler-deque, line 485: Suggestion: add a case that passes a conflicting height inside `style` alongside the `height` prop, then assert the tree renders at the prop's height.

The merge exists so the virtualization height wins over a caller's own `style.height`. The current test only passes a non-conflicting `border`. It proves the caller's style survives but never proves the height wins. If someone reordered the spread so a caller's `style.height` overrode the prop, this test would still pass.

packages/react/src/components/TreeView/screenshots.e2e.tsx
```diff
@@ -86,6 +86,35 @@ test('should have screenshot for TreeView multiple selection', async ({
...
+  await setTheme(page, 'dark');
+  await expect(component).toHaveScreenshot('dark--tree-view-virtualized');
+});
```
rheisler-deque, line 122: Suggestion: add a Playwright test that does the following:
1. moves focus into a tall virtualized tree
2. presses PageDown or ArrowDown to a row well past the initial window
3. asserts the row is focused and present in the DOM
4. scrolls to another part of the tree without changing focus
5. asserts the focused row is still present and focused.

Virtualized mode works as expected today, so this is about preventing a regression that would impact keyboard users. We want to make sure they can navigate the list by keyboard, and that their focus is not dropped if their focused element is scrolled out of view.

packages/react/src/components/TreeView/index.tsx
```diff
@@ -13,19 +19,38 @@ import {
...
+    /** When true (multiple selection only), selecting a parent also selects all
+     *  of its (non-disabled) descendants. */
+    cascadeSelect?: boolean;
+    /** When true (multiple selection only), deselecting a parent also deselects
+     *  all of its descendants. */
+    cascadeDeselect?: boolean;
+    defaultExpandedKeys?: string[];
+    /** When set to a non-zero value, the tree becomes a fixed-height scroll
+     *  region and only the rows in view are rendered (virtualized). Use for long
+     *  lists where rendering every row makes selection changes janky. Omit (or
```
rheisler-deque: Suggestion: Rephrase "janky" to describe the actual problem. It's a colloquialism some won't understand, and it's too informal for a public library representing Deque as a company.

Maybe something like, "makes selection changes take a long time while appearing to freeze."

rheisler-deque (commented):
Thank you for submitting this PR to add this feature to Cauldron. I have tested the virtualization functionality in the Amplitude build, and the component is definitely only rendering a small number of rows near the visible set.

This would be the first example of virtualization in Cauldron. We try to make sure Cauldron components are accessible and useful in a general sense, not just to solve a specific problem that one consumer needs to solve. I think adding virtualization to the TreeView is reasonable because it's likely to have an issue with virtualization for any consumer. Virtualization in the TreeView could serve as a model for other components that need it, and later be generalized to wrap any component if needed.

However, I don't think this implementation is ready to be integrated into Cauldron, which will affect all consumers including those outside Deque. I've noted a potential major accessibility issue below, and it's missing features most consumers need when virtualizing a large list of elements. This PR sets the pattern for virtualization we'll use in other elements in Cauldron when more virtualization is needed, so I want to make sure we identify the gaps and make a plan to fill them before this gets merged.

## Accessibility Concern

Elements that are removed from the DOM are also removed from the accessibility tree. That means a user exploring the TreeView with a screen reader may not be able to perceive all the elements without scrolling them into view. Right now, you can scroll elements into view with the Up, Down, PageUp, and PageDown keys, but assistive technologies change key behavior, so we would need to verify that they still work with all major screen readers active, and that mobile screen reader users can navigate the list.

Assistive Technology users often explore a page by looking at all the elements of a certain type. JAWS for example, has a view to show you all the buttons, all the links, all the list items, etc. TalkBack on Android lets you navigate from element to element of a specific type, but only for elements that are already rendered. I believe a user doing either of these things would only perceive the elements of the TreeView that were rendered when they arrived on the page, and I'm not sure they would have a way to actually reach them.

In this implementation, they do have a way to perceive that there are unrendered elements, which is very good. In virtualized mode the rows keep correct `aria-level`, `aria-posinset`, and `aria-setsize`. A child row still reports `posinset=N setsize=200`, so I'm pretty sure it would announce "item 51 of 200" if there were 200 items in the tree.

### Upstream Issue
In virtualized mode the treegrid gains `aria-rowcount` set to the full row total, but the rows never get a matching `aria-rowindex`. When a grid or treegrid sets `aria-rowcount` because the DOM holds only a subset of rows, each rendered row is supposed to carry `aria-rowindex` so assistive technology can place it in the full set. That pairing is missing here.

The cause is in react-aria-components, not this PR. Its `useTree` sets `aria-rowcount` from the virtualized flag, but its `TreeItem` never passes that flag down, so the `aria-rowindex` branch never runs. We cannot fix it from the Cauldron wrapper without patching react-aria. The practical harm is limited, because position still reaches the user through `aria-posinset` and `aria-setsize`, and axe does not flag the missing pair. Suggestion: file this upstream with react-aria and note the inconsistency in the documentation.

## Usability Gaps

- Related to the accessibility issue above, browser-find (pressing CMD+F or Control+F) stops working with this PR. Virtualization removes off-screen rows from the DOM, so browser-find cannot find text in the virtualized rows. TreeView has no search or filter prop, so there is no replacement. A consumer who sets `height` loses find-in-page over the full tree with no warning, making the TreeView less perceivable especially for sighted keyboard users.
- Similarly, there's also no way to reveal or jump to an off-screen node. Consumers often need to scroll to the selected node or to a node a search matched. The public props expose no scroll-to-key or reveal-node method.

## Technical Gaps
- Expansion is uncontrolled, meaning a consumer cannot implement its own solution to searching the tree. TreeView supports only `defaultExpandedKeys` (index.tsx:44), so a consumer cannot expand a matched node's ancestors under program control, which they would need to do to implement a search function. This is pre-existing, but this PR makes the impact worse.
- The whole `items` array must be built up front. There is no lazy or async loading of children. This caps how large a tree the feature really serves. Rendering is much more memory-intensive than holding list items in memory, but I think virtualization as a feature should make it possible to virtualize a list that is too large to hold, not just to large to render. This could be deferred to later because virtualizing rendering solves the more common, more impactful problem.

## Path Forward

I think Virtualization for the TreeView needs a holistic UX design including accessibility consideration because it's a feature, not a bug fix. This design should fill the gaps above as well as any I missed, and explicitly specify the user experience for assistive technology users based on testing with real ATs.

Once we have that, I think your team and the Cauldron maintainers need to decide which parts of the feature will be implemented now and which will be deferred. We should have tickets covering anything that won't be done now, and the docs should mention which parts of this feature are not yet implemented.  

Finally, QA will need to test the user experience with assistive technologies to make sure it matches what is intended holistically, rather than just fixing the bug that led to this PR.

I'm not the gatekeeper of what goes in Cauldron. I'm very happy to discuss my concerns if you disagree with them. I'd like to hear from more people from @dequelabs/cauldron-codeowners about what else this feature would need to be included in Cauldron, and from members of @dequelabs/cauldron-design about what it would take to design this feature comprehensively. Mateo is trying to fix a bug, and I want to give him a realistic idea of what it will take to do that with virtualization.

packages/react/src/components/TreeView/index.tsx
mateoviilla1: Good catch — fixed in f3ac120f. Added an `isZeroHeight` check that normalizes string values too: `height === 0 || (typeof height === 'string' && parseFloat(height) === 0)`, so `"0"`/`"0px"` (and other zero-valued strings) fall back to the non-virtualized grow-to-fit path instead of rendering a zero-height, row-less scroll region. Locked in with a unit test asserting `height="0px"` stays non-virtualized.

packages/react/src/components/TreeView/index.tsx
mateoviilla1: Done in f3ac120f. Replaced "janky" with "makes selection changes take a long time while appearing to freeze" in the JSDoc, the docs prose, and the props table.

packages/react/src/components/TreeView/TreeView.test.tsx
mateoviilla1: Added in f3ac120f. New test passes `style={{ height: 500 }}` alongside `height={200}` and asserts the treegrid renders at `200px`, so a reordered spread that let the caller's `style.height` win would now fail.

packages/react/src/components/TreeView/screenshots.e2e.tsx
mateoviilla1: Added in f3ac120f. The virtualized screenshot test now asserts `getByRole('row').count()` is > 0 and < 50 while 200 items are mounted, so a regression that stopped windowing (all 200 rows in the DOM) fails even though the screenshot looks identical.

packages/react/src/components/TreeView/screenshots.e2e.tsx
mateoviilla1: Added in f3ac120f as a new e2e test: it focuses into the 200-row virtualized tree, ArrowDowns to a row well past the initial ~8-row window, asserts that row is rendered and focused, then scrolls to the end without moving focus and asserts focus is retained.

mateoviilla1:
Thanks @rheisler-deque — this is a really helpful framing, and I agree that virtualization is a feature that deserves a holistic UX + a11y design rather than landing as a bug-fix side effect.

I've addressed the concrete inline suggestions in f3ac120f (string-zero normalization, the "janky" rewording, the windowing DOM-count assertion, the keyboard-focus-retention e2e, and the conflicting-`style.height` test). Separately, I've documented the current gaps you raised directly in the TreeView docs so no consumer enables `height` unaware of them:

- browser find-in-page only matches rendered rows, with no search/filter prop as a replacement yet
- no programmatic scroll-to / reveal-node API yet
- the missing `aria-rowindex` pairing in virtualized mode, noted as a react-aria-components limitation (`useTree` sets the virtualized flag but `TreeItem` never forwards it)

On the bigger questions — the AT experience validated against real screen readers, controlled expansion for search, lazy/async children, and which pieces ship now vs. get deferred with tracking tickets — I don't think I should decide those unilaterally. I'd like to get @dequelabs/cauldron-codeowners and @dequelabs/cauldron-design input on scope before I put more into this PR. I'm happy to drive whatever slice we agree lands now and open tickets for the rest.

Worth noting the original driver was a downstream perf bug (dequelabs/axe-reports#3277); if the holistic design is a larger effort, one option is to keep this PR as the rendering-perf fix behind the documented caveats and track the fuller feature separately. Open to whichever path the team prefers.

frankensteinke:
It looks like there is not a cheaper way to address the performance issue (in Cauldron at least), so virtualization seems like a solid approach.

I think that the per-click freeze we are running into is likely caused by react-aria re-rendering every row on selection change. This may be fixable upstream by having each row subscribe only to its own selection state; could be worth logging an issue in react-aria.

One other thing worth checking during AT testing: because the tab stop pins to the last-focused row, mixed mouse-scroll + keyboard re-entry can snap the view back when the current scroll point differs from the last focused item. This is pretty minor though.

To conclude, I think it's probably worth landing this now. Have you tested on the original subdimensions tree to see if it fixes the sluggishness? I think that some of the improvements that Ryan suggested are probably okay to land as follow-ups, assuming our designers agree with them.

mateoviilla1:
@frankensteinke Yes, I have tested with a subdimensions tree of 400 children, and cascade selection works properly.

mateoviilla1:
## Performance analysis: why the freeze scales with tree size

Sharing the cost model behind the lag, with locally-seeded test trees of increasing size. Separating what's **exact** (complexity + row counts) from what's **estimated** (the ms constants).

### What's actually mounted

The step force-expands the whole tree (`defaultExpandedKeys = allKeys`), and Cauldron's `TreeView` (react-aria `<Tree>`) **with no virtualization keeps every row mounted at once**. So "displayed items" = every node in the catalog:

| Test tree | Parents | Children | **Rows mounted** | ~DOM nodes (~7/row) |
|-----------|---------|----------|------------------|---------------------|
| flat-20   | 20      | 0        | 20               | ~140                |
| flat-50   | 50      | 0        | 50               | ~350                |
| 10×10     | 10      | 100      | 110              | ~770                |
| 100×100   | 100     | 10,000   | **10,100**       | **~70,000**         |

### The cost model

Two regimes, both **linear in N** (rows mounted):

**1. Initial mount/paint** — create + lay out + style-recalc all N rows:
```
T_mount ≈ N · c_mount     (c_mount ≈ 50–100 µs/row: DOM creation + react-aria hook setup + style recalc)
```

**2. Per-click re-render** — the actual pain point. react-aria's Tree **re-renders every mounted row on any `selectedKeys` change**, and React 17 fires 2–3 uncoordinated re-renders per click (pointer / focus / selection events don't batch):
```
T_click ≈ N · c_row · r   (c_row ≈ 15–75 µs/row reconcile; r ≈ 2–3)
```
The O(N) tree build and the cascade Set-update are both dominated by this — cascading a parent adds O(children) keys, then still triggers **one full O(N) re-render**. Nesting itself adds no algorithmic complexity; it just inflates N (a flat 10,100 freezes identically).

Anchoring `c_row` to the ~150 ms/render figure we'd already measured for a "long list":

| Test tree | Rows   | Per-click re-render (× r=2–3)                        | Feel      |
|-----------|--------|-----------------------------------------------------|-----------|
| flat-20   | 20     | ~0.3–1.5 ms                                         | instant   |
| 10×10     | 110    | ~2–8 ms                                             | instant   |
| 100×100   | 10,100 | ~150–760 ms × 2–3 ≈ **0.4–2.3 s** + ~70k-node style recalc | **frozen** |

### Where slowness starts (answering the row-count question)

- It scales linearly, but the constant is brutal because **nothing is culled** — every row re-renders on every click.
- Perceptible jank sets in in the **low-hundreds of rows** (tens of ms/click), and crosses into a clear multi-hundred-ms-to-second freeze in the **low-thousands** and up. The 100×100 tree (~10k rows) is ~92× the 10×10 tree, so each keystroke costs ~92× more.

### What virtualization changes

The `height`-prop fix changes the **complexity class, not the constant**: only ~12–15 rows fit a 480 px viewport, so `N_mounted` becomes viewport-bounded (≈ constant) regardless of catalog size. Per-click cost drops from `O(N)` → `O(viewport)` — roughly **600–800× fewer rows reconciled per render** for the 10k-row tree, turning the ~1 s freeze into <1 ms. The cascade Set-update stays O(descendants), but that's cheap arithmetic, not React reconciliation.

_Numbers from locally-seeded perf trees; complexity/row counts are exact, ms figures are estimates anchored to prior local measurement._

rheisler-deque:
@frankensteinke @mateoviilla1 I met up with the Cauldron design team and @markreay  to discuss this. We agreed we don't want to merge anything into Cauldron that we're not sure is accessible. Mark is taking ownership of the steps forward, which will include gathering information about the severity and impact of the problem, building a prototype including search, and putting that prototype in front of the Jedi Council for feedback on accessibility.

rheisler-deque:
Removing myself as a reviewer from this PR because it isn't waiting on me and I want to clear it out of my to-do queue. Please re-request if and when you're ready.

mateoviilla1:
@rheisler-deque this is ready for another look when you have time. Re-requesting since you stepped off in August — here is what changed.

## The accessibility question that held this up is resolved

@VaniChinnam's pass on the preview raised three issues. All three are now triaged, and **none of them is caused by virtualization**:

- **#2517** (selected state not re-announced on focus return) — **closed as expected behavior**. Vani reached the same conclusion herself on dequelabs/axe-reports#3309, and it matches cauldron#1202 (closed by @scurker in 2023) and jpmorganchase/salt-ds#5405, all citing nvaccess/nvda#12893. I also verified the row semantics are byte-identical virtualized vs not, and `TreeViewItem.tsx` is untouched by this PR.
- **#2518** (focused row scrolls out of the viewport) — **fixed in this PR.** This one was real, and I initially got it wrong: I claimed it was Safari-only and unreproducible, which was incorrect on both counts. Vani's recording put it on Windows + Chrome, and I then reproduced it and root-caused it. Details below.
- **#2519** (Custom onAction example) — **a separate, pre-existing component bug, not touched by this PR.** The example came in with #2427 and is already on `develop`. Tracked separately; the docs half is in #2520.

Her summary also opened with the tree performing well: levels announced correctly, keyboard interaction working as expected.

## #2518: cause and fix

Arrowing through a virtualized tree scrolled its nearest scrollable ancestor on every press, marching the tree out of view. `scrollIntoViewport` in `@react-aria/utils` walks every scroll parent and scrolls each one to reveal the focused row — redundant once a nested scroller has already revealed it, which is why only virtualized mode was affected.

Fixed by bumping `react-aria-components` 1.13 → 1.20. Measured on the docs page over 30 ArrowDown presses, tree scrolling 724px internally in both runs:

| `react-aria-components` | ancestor scroll |
| --- | --- |
| 1.13.0 | 550px, still climbing |
| 1.20.0 | none |

1.20.0 rather than 1.21.0 because 1.21.0 and its `react-aria`/`react-stately` dependencies fall inside the pnpm `minimumReleaseAge` window and the installer refuses them.

## Review feedback from the earlier round is applied

- `virtualized` is now an explicit prop; `height` only sizes the scroll region. Previously `height` did both, so changing it remounted the tree — and that remount **silently discarded the user's selections**. Selection is now controlled whenever `selectionMode !== 'none'`, so it survives.
- The virtualized tests previously passed with the `<Virtualizer>` deleted (jsdom renders every row). They now assert the virtualizer re-parents rows, and I verified they fail when the wrapper is removed.
- Spacing constants are asserted against the `--space-quarter` / `--space-half` tokens rather than a comment claiming they match.
- Props type `Omit`s `children` / `role` / `dangerouslySetInnerHTML`, all discarded at runtime.
- Docs now carry the limitations: no horizontal scrolling (**this conflicts with the TreeView RFC #1290 — flagging rather than deciding it**), needs a definite width, find-in-page only matches rendered rows, and `aria-rowindex` is not set (upstream: `TreeItem` never forwards the virtualized flag).

## State

All checks green — a11y, eslint, packaging, react, screenshots, semantic-pr-title. 45 TreeView unit tests, 99 screenshots, no visual changes.

Two things I'd particularly value your view on: the **RFC #1290 horizontal-scrolling conflict**, and whether the `react-aria-components` bump belongs in this PR or should be split out.

mateoviilla1:
Closing in favour of **#2523**, which contains the same work rebased onto current `develop` as a single commit.

The reason is this PR's preview environment. It was created on 2026-07-15 and **never rebuilt after that**, despite eight subsequent pushes:

```
d14b8cc0  f3ac120f  fd571517  c80c1bd2  1f813571  e4c9d150  7ef29ab1  0b9b1551
```

I confirmed it by content: the preview still lacked the `aria-rowindex` limitation added in `f3ac120f` on 2026-07-21, still used the pre-rename "All Subdimensions" wording, still rendered 500-item examples, and still exhibited the ancestor-scrolling bug that was fixed on 2026-09-06. GitHub Actions ran green on every one of those pushes, so the pushes were fine — the preview simply stopped consuming them.

**This matters for the accessibility review.** @VaniChinnam tested against that preview URL, including the screen recording on #2518. She was looking at a July build the whole time, which is why fixes kept appearing not to work. Her reports were accurate about what was in front of her; the environment was stale. Apologies — I also stated on #2518 that pushing here had rebuilt the preview, and that was wrong.

Nothing is lost from the work itself. #2523 carries all of it:

- `virtualized` as an explicit prop, with `height` only sizing the scroll region
- selection preserved across the remount that toggling it causes
- react-aria-components 1.20, which fixes the ancestor scrolling in #2518
- the documented limitations, including the RFC #1290 horizontal-scrolling conflict
- shortened docs examples, and the cascade example actually virtualized

The review discussion here stays as the record. @rheisler-deque @markreay — sorry for the churn; #2523 is the one to look at, and its preview should reflect the actual code.

Issue #2466, opened 2026-07-07
chore(react): publish a dual CJS/ESM build with sideEffects: false for automatic tree-shaking


frankensteinke:
## Summary

`@deque/cauldron-react` ships **CJS only** today: `rollup.config.js` emits a single
`format: 'cjs'` bundle, and `package.json` exposes only `main`/`types` — no `exports`
map, no `module` field. As a result the ergonomic barrel import
(`import { Modal } from '@deque/cauldron-react'`) can't be tree-shaken: bundlers keep
the whole component graph.

This ticket adds a **dual CJS/ESM build** plus **`"sideEffects": false`** so the barrel
import prunes unused components automatically — no consumer import-style change.

## Payoff

Consumers keep `import { X } from '@deque/cauldron-react'` and automatically bundle only
what they use — no deep imports, no codemod. A consumer importing only `Button` should
pull in neither `react-aria-components` (TreeView) nor `react-syntax-highlighter` (Code).

## Scope notes from an initial code audit

- **The JS bundle imports zero CSS.** All styling lives in `src/index.css` (built
  separately by PostCSS into `lib/cauldron.css`); no `.tsx` module does `import './x.css'`.
  So the correct declaration is `"sideEffects": false`, **not** a `["*.css"]` list.
  `sideEffects: false` does not affect the separate `cauldron.css` artifact — consumers
  still import it explicitly.
- **The only module-scope side effect found** is the `registerLanguage(...)` block in
  `src/components/Code/index.tsx`. It's *local* (only runs when Code is imported) and
  therefore tree-shake-safe. All other `addEventListener`/`window`/`document` usage is
  inside hooks/lifecycle (runtime, not import-time) and irrelevant to `sideEffects`.
- **`Code` hard-imports CJS subpaths** (`react-syntax-highlighter/dist/cjs/light`). This
  is the single most likely thing to break in an ESM consumer graph and must be switched
  to ESM paths before the ESM build is validated.
- Adding an `exports` map is itself a **potentially breaking change**: any consumer
  currently deep-reaching into `lib/*` will hard-error once `exports` is present. The
  barrel, `./cauldron.css`, and `./package.json` must be preserved. Treat as a **minor**
  release with a loud changelog note.
- **Icons stay lazy-loaded and are out of scope.** `Icon` loads SVGs through a dynamic
  `import()`, so they resolve as async chunks rather than eager weight. This ticket does
  not change that. The tree-shaking payoff here is pruning unused *components* and their
[...]

PR #2476, opened 2026-07-15
chore(react): add packaging validation harness + CI gate

frankensteinke:
Adds a `verifyPackaging` script that lints the packaged tarball, checks its types (with `@arethetypeswrong/cli`), and smoke tests it.

This doesn't affect the build output, but will catch any packaging regression in CI. This will be helpful for the move to dual CJS/ESM builds.

Relates to: #2466

### DIFF
--- a/packages/react/scripts/packaging-smoke/smoke.mjs
+++ b/packages/react/scripts/packaging-smoke/smoke.mjs
@@ -1,0 +1,17 @@
+// Smoke test: the packed tarball resolves under native ESM `import`.
+// Runs inside a throwaway consumer where the tarball is installed (not the
+// workspace symlink), so this exercises the real published resolution.
+import assert from 'node:assert';
+
+import cauldron from '@deque/cauldron-react';
+import { Button, Modal } from '@deque/cauldron-react';
+
+assert(Button, 'import: expected `Button` named export');
+assert(Modal, 'import: expected `Modal` named export');
+assert(cauldron.Button, 'import: expected `Button` on the default export');
+assert(
+  typeof Button === 'object' || typeof Button === 'function',
+  'import: `Button` is not a renderable component'
+);
+
+console.log('import "@deque/cauldron-react" OK');
--- a/packages/react/scripts/packaging-smoke/smoke.mjs
+++ b/packages/react/scripts/packaging-smoke/smoke.mjs
@@ -3,12 +3,10 @@
 // workspace symlink), so this exercises the real published resolution.
 import assert from 'node:assert';
 
-import cauldron from '@deque/cauldron-react';
 import { Button, Modal } from '@deque/cauldron-react';
 
 assert(Button, 'import: expected `Button` named export');
 assert(Modal, 'import: expected `Modal` named export');
-assert(cauldron.Button, 'import: expected `Button` on the default export');
 assert(
   typeof Button === 'object' || typeof Button === 'function',
   'import: `Button` is not a renderable component'
--- a/packages/react/scripts/verifyPackaging.js
+++ b/packages/react/scripts/verifyPackaging.js
@@ -94,12 +94,35 @@ try {
       '--no-audit',
       '--no-fund',
       '--no-package-lock',
-      '--no-save'
+      '--no-save',
+      '--ignore-scripts'
     ],
     { cwd: consumerDir }
   );
   run('node', ['smoke.cjs'], { cwd: consumerDir });
   run('node', ['smoke.mjs'], { cwd: consumerDir });
+
+  // `package.json` publishes `style: lib/cauldron.css`, but neither publint
+  // nor attw validates that field — they only cover JS entries and types. A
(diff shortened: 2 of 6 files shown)


packages/react/scripts/verifyPackaging.js
```diff
@@ -0,0 +1,106 @@
...
+    'npm',
+    [
+      'install',
+      tarball,
+      'react@^19',
+      'react-dom@^19',
+      '--no-audit',
+      '--no-fund',
+      '--no-package-lock'
+    ],
```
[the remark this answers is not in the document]
frankensteinke: I'm not sure how much this matters since it's just a throwaway directory, but it's probably worth being explicit. Updated.

packages/react/scripts/packaging-smoke/smoke.mjs
```diff
@@ -0,0 +1,17 @@
...
+// Runs inside a throwaway consumer where the tarball is installed (not the
+// workspace symlink), so this exercises the real published resolution.
+import assert from 'node:assert';
+
+import cauldron from '@deque/cauldron-react';
+import { Button, Modal } from '@deque/cauldron-react';
+
+assert(Button, 'import: expected `Button` named export');
+assert(Modal, 'import: expected `Modal` named export');
+assert(cauldron.Button, 'import: expected `Button` on the default export');
```
rheisler-deque: Suggestion: Drop the assertion that `Button` exists on the default import, or keep it only if a default export becomes a real supported contract.

The barrel in `src/index.ts` has only named exports and no `export default`. Today `import cauldron from '@deque/cauldron-react'` resolves to the whole `module.exports` object only because Node applies CommonJS-to-ESM interop to the CJS build, so it's not really testing what it intends to.

This harness exists to guard the dual CJS/ESM migration. Once a real ESM entry ships with named exports and no default, `import cauldron from ...` becomes `undefined`. This line then throws even when packaging is correct. The gate would then fail on correct packaging, or push someone to add a default export the source never intended. The named-export assertion on line 9 already covers the meaningful ESM case.

packages/react/scripts/verifyPackaging.js
```diff
@@ -0,0 +1,107 @@
...
+  );
+  fs.copyFileSync(
+    path.join(smokeFixtures, 'smoke.mjs'),
+    path.join(consumerDir, 'smoke.mjs')
+  );
+
+  // Install with npm into an isolated dir so resolution is hermetic and does
+  // not touch the pnpm workspace. react/react-dom satisfy the peer range.
+  run(
+    'npm',
```
rheisler-deque, line 88: Suggestion: Add `--ignore-scripts` to the smoke `npm install`.

The smoke install runs with lifecycle scripts enabled. A `postinstall` from a freshly resolved dependency runs during that install. On a push to develop or master, that install runs in a context where the job can request an OIDC token. `--ignore-scripts` stops any dependency install script from running. The smoke test only needs module resolution for `require` and `import`, so it loses nothing.

Here is how the token permission reaches this job:

- `release.yml` sets `id-token: write` at the top (release.yml:9-11). That permission lets a job request an OIDC token.
- `release.yml` calls `tests.yml` as a reusable workflow (release.yml:14-15).
- A reusable workflow inherits the caller's token permissions when it declares none of its own. `tests.yml` declares no `permissions` block, so it inherits `id-token: write`.
- `release.yml` runs on push to develop or master, so the `packaging` job runs with that token on those pushes.

This is a hardening step, not a working exploit. The job sets no `environment`, and npm and cloud OIDC trust normally bind to a specific environment or repo claim. Without that claim, a minted token likely cannot be exchanged for anything. On a normal PR the token does not appear at all. `tests.yml` then runs directly under the `pull_request` trigger, and a fork PR gets a read-only token with no `id-token`.

packages/react/scripts/verifyPackaging.js
```diff
@@ -0,0 +1,107 @@
...
+      'react-dom@^19',
+      '--no-audit',
+      '--no-fund',
+      '--no-package-lock',
+      '--no-save'
+    ],
+    { cwd: consumerDir }
+  );
+  run('node', ['smoke.cjs'], { cwd: consumerDir });
+  run('node', ['smoke.mjs'], { cwd: consumerDir });
```
rheisler-deque, line 103: Suggestion: After the install, assert that `lib/cauldron.css` exists in the installed package and is non-empty.

`package.json` declares `style: lib/cauldron.css` as a published entry (packages/react/package.json:11). publint checks the JS entry fields, and attw checks types. Neither validates a bare `style` field. The smoke scripts load only JS. If the stylesheet fell out of the tarball through a `files` allowlist change, the gate would still pass. A dropped stylesheet is the class of packaging regression this PR exists to catch.

rheisler-deque (changes requested):
Great first phase! I just have a couple things I think should change before merging.

rheisler-deque (approved):

PR #2478, opened 2026-07-16
fix(react): guard ThemeProvider default context against undefined document

frankensteinke:
## Summary

`ThemeProvider`'s default parameter `context = document?.body` throws `ReferenceError: document is not defined` under any SSR runtime (e.g. Next.js App Router prerendering a `"use client"` component). Optional chaining only guards against null/undefined **values** — it does **not** protect against an **undeclared global**, so evaluating the `document` identifier itself throws before `?.` ever runs.

This makes `<ThemeProvider>` (without an explicit `context` prop) unusable in any SSR framework. Passing `context={null}` works around it, confirming the default param is the cause.

### Root cause / history
Introduced on 2024-01-31 in #1328 — a PR that *intended* to make `ThemeProvider` SSR-safe (`document.body` → `document?.body`) but used optional chaining, which never actually protected against the undeclared global. This is a pre-existing bug on `develop`, independent of module format (reproduces on CJS today).

## Fix
```diff
-  context = document?.body,
+  context = typeof document !== 'undefined' ? document.body : undefined,
```
`typeof` is safe on undeclared globals. In the browser, behavior is identical (`document.body`); under SSR, `context` is `undefined`, and the component's existing `context?.` guards (effects + MutationObserver) already bail out on a falsy context.

## Tests
- **New:** `theme.ssr.test.tsx` runs under `@jest-environment node`, where `document` genuinely doesn't exist, and asserts `renderToString(<ThemeProvider …>)` doesn't throw. Verified **red without the fix** (exact `ReferenceError` at `theme.tsx:32`) → **green with it**. A separate node-env file is used rather than deleting `global.document` in the jsdom suite, which destabilizes jsdom's async teardown.
- **setupTests.ts:** guarded the `navigator.clipboard` / `document.execCommand` shims with `typeof … !== 'undefined'` so the shared setup also runs under the node environment.

## Verification
- All context tests pass; the clipboard/`execCommand`-dependent suites (CopyButton, Code) plus TreeView still pass — the setup guard preserves jsdom behavior.
- Lint + prettier clean on touched files.

### DIFF
--- a/packages/react/src/contexts/theme.ssr.test.tsx
+++ b/packages/react/src/contexts/theme.ssr.test.tsx
@@ -18,11 +18,13 @@ const ThemeTester = () => {
 // values, not undeclared globals.
 test('ThemeProvider renders without error when document is undefined (SSR)', () => {
   expect(typeof document).toBe('undefined');
-  expect(() =>
-    renderToString(
+  let html = '';
+  expect(() => {
+    html = renderToString(
       <ThemeProvider initialTheme="dark">
         <ThemeTester />
       </ThemeProvider>
-    )
-  ).not.toThrow();
+    );
+  }).not.toThrow();
+  expect(html).toBe('<div></div>');
 });
--- a/packages/react/src/contexts/theme.ssr.test.tsx
+++ b/packages/react/src/contexts/theme.ssr.test.tsx
@@ -1,0 +1,28 @@
+/**
+ * @jest-environment node
+ */
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+import { ThemeProvider, useThemeContext } from './theme';
+
+const ThemeTester = () => {
+  useThemeContext();
+  return <div />;
+};
+
+// This suite runs under the `node` environment (see the docblock above), where
+// `document` is not a declared global — the same condition as an SSR framework
+// prerendering a client component (e.g. Next.js App Router). A bare
+// `document?.body` default would throw `ReferenceError: document is not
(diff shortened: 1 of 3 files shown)


chornonoh-vova (dismissed):

packages/react/src/contexts/theme.ssr.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+// values, not undeclared globals.
+test('ThemeProvider renders without error when document is undefined (SSR)', () => {
+  expect(typeof document).toBe('undefined');
+  expect(() =>
+    renderToString(
+      <ThemeProvider initialTheme="dark">
+        <ThemeTester />
+      </ThemeProvider>
+    )
+  ).not.toThrow();
```
rheisler-deque: Nit: I think asserting here that something is rendered would technically be a stronger test because it would catch scenarios where there was an error that prevented the UI from rendering, but didn't throw and exception. That's a theoretical problem, I don't know of a way to make that actually happen.

rheisler-deque (dismissed):
This PR looks great, nice job testing to make sure the fix works and to protect against regression!

packages/react/src/contexts/theme.ssr.test.tsx
frankensteinke: I think this is a worthwhile check. I'll keep the `not.toThrow()` assertion, and add another check that the rendered HTML is `<div></div>` (since `<ThemeTester />` renders `<div />`).

rheisler-deque (approved):

PR #2489, opened 2026-07-24
feat(Notice): notice colors redesign

chornonoh-vova:
Redesigns the `Notice` component's `info`, `caution`, and `danger` color variants to use softer, hand-tuned backgrounds paired with bolder, WCAG-compliant icon/border colors, and adds screenshot coverage for the component. Added the `success` variant and a new icon: `report` (used it for the `danger` variant).

image

image

Screenshot 2026-07-27 at 09 41 14

image


jbregar:
This looks great... can we swap the colors over to our new error/warning/success/neutral color set? In this context we'd use:

Danger:
icon and border: #DF5252
background fill: #FDF2F2 

Caution:
icon and border: #A16200
background fill: #FFF8EF

Success (not sure if this exists):
icon and border: #4D8C6E
background fill: #F3FCF7

Info:
icon and border: #547D93
background fill: #F5F7F8

Also, I know this is a bigger ask and perhaps we can talk about it, but it'd be great to add Material Symbols `report` icon to the Icon set and use that for the "Danger" icon so Danger and Caution have unique icons.

packages/react/src/components/Notice/index.test.tsx
frankensteinke: Maybe we could expand this to check icons for all of the different notice types? We could use a structure that you did similar to where you checked axe violations later on in this file, like:

```jsx
test.each([
  ['info', 'info-circle'],
  ['caution', 'caution'],
  ['danger', 'report'],
  ['success', 'check-circle']
] as const)('should render the %s default icon', (type, icon) => { ... });
```

docs/pages/components/Notice.mdx
frankensteinke: There are now 4 icon types used for Notice. Could we update this in the docs to reflect that?

frankensteinke (commented):
Left a few minor notes regarding the notice types, if you're able to take a look!

Also, an additional nit (because Storybook isn't fully adopted in the project yet): there is a story for Notice and it currently has `info`, `caution`, and `danger` types, but no `success` type. Maybe you could add that plus a `Success` story?

frankensteinke (approved):

PR #2490, opened 2026-07-27
fix(react): announce field errors to screen readers

Bracciata:
## Problem

Field errors are rendered into a plain element that is associated with the input only through `aria-describedby`. Errors typically appear **on submit**, when focus does not move — and screen readers only report a changed accessible description when the described element is re-focused. The result is that the message the interaction just produced is never announced.

This affects every component that renders a field error: `TextField`, `Select`, `Checkbox`, `Combobox`, and `FieldGroup`.

Reported downstream as a Serious VPAT finding (WCAG 4.1.3 Status Messages) against the axe Developer Hub — see dequelabs/walnut#14909. The "Edit name" modal there shows `Name is required.` with no announcement, but the defect is in the library, not the consumer.

## Change

Extract the error markup — duplicated near-identically in four of the five components — into an internal `FieldError` component that renders as a live region:

```tsx
<div className="Field__error" role="alert" aria-live="polite" id={errorId}>
```

Both attributes are load-bearing:

- **`role="alert"`** gets the message announced as soon as the node is *inserted*. A bare `aria-live` region inserted with its text already inside is frequently not announced at all, since AT watches pre-existing regions for mutations. The error node does not exist until submit, so this is the case that matters.
- **`aria-live="polite"`** explicitly overrides alert's implicit `assertive`, so a form that reveals several field errors at once queues them rather than interrupting itself.

This mirrors the pattern already used for the Combobox empty state (`Combobox.tsx:120`).

## Notes for review

- **Checkbox keeps its own styling.** It uses `class="Error"` with no caution icon, and `.Field__error` carries additional rules it must not inherit (`styles/forms.css:213`). `FieldError` therefore defaults `className` to `Field__error`, and Checkbox overrides with `className="Error" icon={false}`.
- **Rendered output is otherwise unchanged** — same classes, same icon, same text. The only DOM difference is the two ARIA attributes, so existing screenshots should not move.
- **Non-breaking** by the criteria in CONTRIBUTING.md: no renamed or removed export, no changed prop, no new required prop. `FieldError` is internal and not exported from the public barrel.

### DIFF
--- a/packages/react/src/components/Checkbox/Checkbox.test.tsx
+++ b/packages/react/src/components/Checkbox/Checkbox.test.tsx
@@ -65,6 +65,13 @@ test('should render error checkbox', () => {
   ).toBeInTheDocument();
 });
 
+test('should announce the error via a live region', () => {
+  renderCheckbox({ error: 'you should check this checkbox' });
+  const error = screen.getByRole('alert');
+  expect(error).toHaveTextContent('you should check this checkbox');
+  expect(error).toHaveAttribute('aria-live', 'polite');
+});
+
 test('should toggle checkbox correctly', async () => {
   const user = userEvent.setup();
   const input = renderCheckbox();
--- a/packages/react/src/components/internal/FieldError/index.tsx
+++ b/packages/react/src/components/internal/FieldError/index.tsx
@@ -1,0 +1,49 @@
+import React, { forwardRef, HTMLAttributes, ReactNode } from 'react';
+import Icon from '../../Icon';
+
+export interface FieldErrorProps extends HTMLAttributes<HTMLDivElement> {
+  children: ReactNode;
+  /**
+   * Class applied to the error element. Defaults to the shared field error
+   * style; `Checkbox` styles its error differently.
+   */
+  className?: string;
+  /** Whether to render the caution icon before the message. */
+  icon?: boolean;
+}
+
+/**
+ * The error message for a form field, wired up as a live region.
+ *
+ * Field errors typically appear on submit, at which point focus has not moved.
+ * Screen readers only report a changed accessible description when the element
+ * it describes is re-focused, so associating the message via `aria-describedby`
+ * alone leaves it silent for the very interaction that produced it.
+ *
+ * `role="alert"` gets the message announced as soon as it is inserted, and
+ * `aria-live="polite"` stops a form that reveals several errors at once from
+ * interrupting itself.
+ *
+ * @see https://www.w3.org/WAI/WCAG21/Understanding/status-messages
+ */
+const FieldError = forwardRef<HTMLDivElement, FieldErrorProps>(
+  (
+    { children, className = 'Field__error', icon = true, ...props },
+    ref
+  ): React.JSX.Element => (
+    <div
+      className={className}
+      role="alert"
+      aria-live="polite"
+      ref={ref}
+      {...props}
+    >
+      {icon && <Icon type="caution" />}
+      {children}
+    </div>
+  )
+);
+
+FieldError.displayName = 'FieldError';
+
+export default FieldError;
--- a/packages/react/src/components/Checkbox/index.tsx
+++ b/packages/react/src/components/Checkbox/index.tsx
@@ -9,6 +9,7 @@ import React, {
 import classNames from 'classnames';
 import nextId from 'react-id-generator';
 import Icon, { type IconType } from '../Icon';
+import FieldError from '../internal/FieldError';
 import { addIdRef } from '../../utils/idRefs';
 
 export interface CheckboxProps extends InputHTMLAttributes<HTMLInputElement> {
@@ -164,9 +165,9 @@ const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
           </span>
         )}
         {error && (
-          <div id={errorId} className="Error">
+          <FieldError id={errorId} className="Error" icon={false}>
             {error}
-          </div>
+          </FieldError>
         )}
       </div>
     );
--- a/packages/react/src/components/Combobox/Combobox.test.tsx
+++ b/packages/react/src/components/Combobox/Combobox.test.tsx
@@ -267,6 +267,13 @@ test('should render combobox with error', () => {
     'aria-describedby',
     `other-id ${errorId}`
   );
+
+  // The listbox is closed, so this is the field error rather than the
+  // "no results found" empty state, which is also an alert.
+  const error = screen.getByRole('alert');
+  expect(error).toHaveAttribute('id', errorId);
+  expect(error).toHaveTextContent('You forgot to choose a value.');
+  expect(error).toHaveAttribute('aria-live', 'polite');
 });
 
 test('should render combobox with both description and error', () => {
--- a/packages/react/src/components/Combobox/Combobox.tsx
+++ b/packages/react/src/components/Combobox/Combobox.tsx
@@ -21,7 +21,7 @@ import { addIdRef } from '../../utils/idRefs';
 import TextFieldWrapper from '../internal/TextFieldWrapper';
 import { ListboxValue } from '../Listbox/ListboxOption';
 import ComboboxPill from './ComboboxPill';
-import Icon from '../Icon';
+import FieldError from '../internal/FieldError';
 
 // Event Keys
 const [Enter, Escape, Home, End, Backspace, Delete] = [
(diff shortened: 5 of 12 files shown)


Bracciata:
This should fix https://github.com/dequelabs/walnut/issues/14909

packages/react/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -65,6 +65,13 @@ test('should render error checkbox', () => {
   ).toBeInTheDocument();
 });
 
+test('should announce the error via a live region', () => {
```
rheisler-deque, line 68: Suggestion: Add a Checkbox assertion that the error carries the `Error` class and renders no caution icon.

Checkbox is the only consumer that overrides both `icon={false}` and `className="Error"`, and no Checkbox test checks either one.

packages/react/src/components/internal/FieldError/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * Class applied to the error element. Defaults to the shared field error
+   * style; `Checkbox` styles its error differently.
+   */
+  className?: string;
+  /** Whether to render the caution icon before the message. */
+  icon?: boolean;
+}
+
+/**
```
rheisler-deque, line 15: Nit: Consider trimming the JSDoc to the one fact a reader cannot recover from the code, plus the WCAG link.

Internal components in this directory do not carry JSDoc. The sibling `TextFieldWrapper` has none. Most of this block restates the commit message body, which already records the rationale. The `className` prop comment also points at Checkbox (lines 6-9). That comment will go stale if Checkbox stops overriding the class. The one part a reader cannot recover from the code, that `aria-describedby` alone stays silent, is worth keeping.

packages/react/src/components/internal/FieldError/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    { children, className = 'Field__error', icon = true, ...props },
+    ref
+  ): React.JSX.Element => (
+    <div
+      className={className}
+      role="alert"
+      aria-live="polite"
+      ref={ref}
+      {...props}
+    >
```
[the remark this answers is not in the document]
rheisler-deque: I agree this is a good change for this PR.

I generally disagree with copilot about this for a public component library. One of my frustrations with component libraries is that you can only override props the developer anticipated you would want to override, which usually doesn't include something you need.

For this case, I think this suggestion is OK because it's an internal component. We're just protecting our future selves from making a mistake. If we need to allow overrides later, we can do it more explicitly.

rheisler-deque (commented):
Really nice, well-scoped solution to the problem! I left a couple suggestions.

I have one question just to make sure the PR matches your intention. Here's the scenario: if a user submitted a form with 5 errors in it, all of them would be announced one after the other because of `aria-live="polite"` (I think, I haven't tried it). Then, the user would have to navigate to each field that had an error, and they would hear the same error messages announced again via `aria-describedby`.

Is that what you intend, and is it acceptable behavior to fix the VPAT issue? I usually recommend a couple other patterns over announcing them all via live-region:
1. Move the user's focus to the first field on the page with an error, which then reads that field's error out via `aria-describedby`.
2. Move their focus to an error message at the top of the form describing all the fields that have an error, maybe with a link to focus each one.

Either way, they have the opportunity to discover all the errors in turn. It's less repetitive, and hearing them announced all at once is unfriendly for memory and other cognitive and disabilities, especially on a long form with lots of potential errors.

To be clear, this PR fixes a real issue with a simple change, so I think it's fine to include in Cauldron. I just want to make sure it matches what you expect and talk about whether we should consider a follow-up to improve on the pattern.

packages/react/src/components/internal/FieldError/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @see https://www.w3.org/WAI/WCAG21/Understanding/status-messages
+ */
+const FieldError = forwardRef<HTMLDivElement, FieldErrorProps>(
+  (
+    { children, className = 'Field__error', icon = true, ...props },
+    ref
+  ): React.JSX.Element => (
+    <div
+      className={className}
```
chornonoh-vova, line 35: NIT: maybe it's better to join classnames instead of overriding completely? It's usually the case for other components

frankensteinke (changes requested):
Thanks for checking this out! This is a real gap in Walnut and the `FieldError` extraction itself is nice and clean. My hesitation is only about live regions as the mechanism.

GOV.UK has an example of the type of [Error summary](https://design-system.service.gov.uk/components/error-summary/) component that @rheisler-deque  mentioned. They recommend placing at the top of the form and moving focus to it on submit. Adam Silver also recommends this approach in his legendary Form Design Patterns book. I like this pattern and have used it before.

The other option that Ryan mentioned, moving focus to the first field with the error, is probably the easier solution here, and requires no design changes.

Both of these solutions have an added benefit of scrolling to the part of the page that shows the errors. On a long form, the type of this solution in the PR would let users with screen readers know of errors, but if the errors are off-screen, sighted users would not know about them. Moving focus to the first field with an error works well for everybody.

Live regions added to all errors could be a little noisy, and even duplicative. A consumer that has already moved focus to the first field with an error would hear that error, then the same error (plus all of the other ones) again through the live region.

So I'd rather hold off on this one. I'd recommend trying out a change in Walnut that focuses the first field with an error instead. Let me know what you think!

packages/react/src/components/internal/FieldError/index.tsx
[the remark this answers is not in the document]
Bracciata: Agreed — for an internal accessibility primitive, spreading `...props` before the ARIA attributes is the right call, and @rheisler-deque's framing of it as protecting our future selves matches how I read it. Not applied yet: holding this PR pending a focus-management experiment in Walnut, so leaving the thread open.

packages/react/src/components/Checkbox/Checkbox.test.tsx
Bracciata: Confirmed gap — `Checkbox.test.tsx` only asserts the `alert` role and `aria-live`, and Checkbox is the sole consumer overriding both `className="Error"` and `icon={false}`. Not applied yet: holding this PR pending a focus-management experiment in Walnut.

packages/react/src/components/internal/FieldError/index.tsx
Bracciata: Fair — I checked and the sibling `TextFieldWrapper` carries no JSDoc, so the convention point holds, and the `className` comment would indeed go stale. Not applied yet: holding this PR pending a focus-management experiment in Walnut.

packages/react/src/components/internal/FieldError/index.tsx
Bracciata: These two are mutually exclusive style variants rather than base + modifier: `forms.css:212-217` gives `.Field__error` a different `font-size` plus `align-items`, `margin-bottom` and `line-height` that Checkbox's error deliberately does not get, so joining them would change how Checkbox renders. Holding the PR for now pending a focus-management experiment in Walnut, so nothing changed here yet.

Bracciata:
@frankensteinke @rheisler-deque — holding this one as suggested, and converted it to draft.

To answer Ryan's question directly: announcing all five errors and then re-announcing each one on focus is *not* what I was aiming for, and the duplication you both flagged is real. I'll try the focus-management route in Walnut first — move focus to the first field with an error on submit — which also covers the case a live region does nothing for: a sighted user whose errors are scrolled off-screen.

One thing worth flagging for whenever this comes back around. An error summary and first-error focus both need the form element, the submit handler, and knowledge of sibling fields, so neither can live inside `TextField`/`Select`/`Checkbox` — they are necessarily consumer-side. So the question isn't really live region *vs.* focus management; it's whether Cauldron's field errors should still be live regions once a consumer handles focus properly. If the Walnut change closes dequelabs/walnut#14909 on its own, this PR isn't needed.

Leaving the inline suggestions unresolved so they aren't lost if it does come back.

rheisler-deque:
> To answer Ryan's question directly: announcing all five errors and then re-announcing each one on focus is _not_ what I was aiming for, and the duplication you both flagged is real. I'll try the focus-management route in Walnut first — move focus to the first field with an error on submit — which also covers the case a live region does nothing for: a sighted user whose errors are scrolled off-screen.

@Bracciata FYI The focus-management thing currently partially works in Walnut, but only for required fields, not for validation errors. I'm currently testing a form that I'm working on, and submitting a form without required fields focuses the first empty field in the form. Putting an invalid value in a field and submitting shows the error, but the empty required field lower-down in the form is focused.

I think this is because the field is marked as required and it's in a form, so the browser is automatically handling it. Good for us to know if we address this with focus management.

PR #2504, opened 2026-08-07
ci: Gate PR Preview Deploys from Outside Contributors

### LINKED ISSUE #2486: Gate docs PR preview deploys so outside-contributor code does not deploy without approval
## Summary

An outside contributor can open a pull request against this repo and have their code built and deployed to an Amplify preview automatically, before any maintainer reviews or approves the run.

The preview deploy is triggered by AWS Amplify's own GitHub App, not by a GitHub Actions workflow. GitHub's "require approval to run workflows for fork PRs" setting only gates Actions runs, so it does not cover the Amplify preview path. That is why these deploys happen with no approval step.

The build runs untrusted code. `amplify.yml` runs `pnpm install --frozen-lockfile` and `pnpm build` on whatever the PR contains. Package lifecycle scripts and the docs build both execute contributor-authored code inside Amplify's build environment. Any secrets or IAM permissions attached to that environment are reachable by that code. The output is also served from a Deque-controlled preview domain.

We want previews to deploy automatically for pull requests opened by people inside Deque, and not to deploy contributor code from outside Deque without an explicit, human approval.

## Acceptance criteria

- A pull request opened by a Deque org member deploys a docs preview automatically, with no manual step.
- A pull request opened by an outside contributor does not deploy a preview automatically.
- No deploy credentials are exposed to code from an outside contributor's pull request.
- There is a documented way for a maintainer to approve a preview for a specific outside pull request after reviewing the diff (for example a label a maintainer applies).
- `CONTRIBUTING.md` is updated to describe the new preview behavior. The current text at `CONTRIBUTING.md:213` says previews deploy with each commit to a PR, which will no longer be true for outside contributors.
- The state of Amplify's native pull request previews is settled deliberately. They were turned off as an interim mitigation (see below). If the fix replaces them with a gated Actions workflow, they stay off, and that decision is recorded. If the fix does not provide previews another way, re-enable them in the Amplify Console only after the outside-contributor gate is in place, so the repo is not left with previews on and no gate. [The preview that was disabled is for the develop branch, and you can see it here in the AWS console.](https://us-east-1.console.aws.amazon.com/amplify/apps/d15792l1n26ww3/previews/develop)

## Interim mitigation (already applied)

Amplify pull request previews were disabled to stop the exposure while this ticket is open. In the Amplify Console, under Hosting then Previews, "Pull request previews" was turned off on the `develop` branch. That is the one rule governing all PRs into `develop`. The `develop` branch build is unaffected, so the live docs site still deploys. This is why the AC above asks whoever implements the fix to decide the final state of Amplify previews rather than assume they should go back on.

## Out of scope

- Auditing what secrets or IAM permissions the current Amplify build environment actually holds. That is worth doing separately and would inform how urgent this is.

<details>
<summary>Dev notes (implementation findings so far)</summary>

### Current state

- `amplify.yml` is the only Amplify config in the repo. It is a standard Amplify Console build spec. It builds the docs site (`artifacts.baseDirectory: docs/dist`) via `pnpm install --frozen-lockfile` then `pnpm build`. It is referenced by no GitHub Actions workflow.
- None of the 13 workflows in `.github/workflows/` deploy to Amplify, post a preview URL, use `pull_request_target`, or check `github.event.pull_request.head.repo.fork`.
- The preview flow lives entirely in AWS Amplify's GitHub App integration, configured in the Amplify Console, not in this repo.
- `CONTRIBUTING.md:213` documents the current behavior: previews deploy "with each commit to a PR," and the `aws-amplify` bot posts the link.

### Why the existing GitHub gate does not help

GitHub's fork-PR approval setting only gates GitHub Actions workflow runs. The Amplify GitHub App is a separate integration that reacts to PR webhooks on its own. So that setting never sees the preview deploy.

### Proposed approach

Move the preview deploy off Amplify's native GitHub App trigger and into a gated GitHub Actions workflow.

- Turn off automatic PR previews in the Amplify Console so the native trigger stops firing. This is an AWS console change, not a repo change.
- Add an Actions workflow on the plain `pull_request` trigger. Do not use `pull_request_target`, which would hand secrets to fork code and is the usual source of exploits in this pattern.
- Keep the AWS deploy credentials in a GitHub Environment rather than plain repo secrets.
- Gate the deploy on the PR author's relationship to the repo. The webhook payload exposes `github.event.pull_request.author_association`. Deploy only when it is one of `OWNER`, `MEMBER`, `COLLABORATOR`. Skip otherwise.
- With the `pull_request` trigger, GitHub already withholds secrets from fork PRs by default, so the author-association check is a second layer rather than the only one.
- For the maintainer-approved outside preview, gate on a label a maintainer applies only after reading the diff. This mirrors GitHub's own "approve to run" idea and keeps a human in front of any untrusted deploy.

### Known edges to decide during implementation

- `author_association` describes who opened the PR, not who authored each commit. A member could open a PR whose branch contains outside-authored commits, and it would still deploy. The accountable party is the member who opened it, which is usually the right trust boundary. Stricter handling is possible but should not be the starting point.
- An internal contributor who works from their own fork would be blocked by a fork-only check, which is why the author-association signal is preferred over a fork check.

### Requires access outside this repo

- AWS Amplify Console access to disable the native auto-preview.
- GitHub org and repo admin access to set the Environment, its required reviewers, and the outside-collaborator approval setting.

</details>

rheisler-deque:
## What this does

Until recently, AWS Amplify's own GitHub App built and deployed a docs preview for every PR, including fork PRs, with no approval. That ran untrusted PR code in an environment that holds AWS access. Amplify's native previews were turned off as an interim mitigation. This PR replaces them with a gated GitHub Actions workflow.

A PR from a Deque org member deploys a preview automatically. A PR from an outside contributor pauses before anything builds until a Cauldron maintainer approves it, and every push waits for a fresh approval. No AWS credential ever reaches PR code.

## Files

- `.github/workflows/pr-preview.yml` builds and deploys the preview.
- `.github/workflows/pr-preview-cleanup.yml` deletes the preview branch when the PR closes.
- `CONTRIBUTING.md` describes the new flow for contributors.

## Decision: keep Amplify, gate it with a GitHub Actions workflow

We kept AWS Amplify as the preview host and moved the trigger into a gated GitHub Actions workflow. We did not switch to GitHub Pages.

- Amplify keeps the preview URL and the root-served build. Pages would change the URL and serve from a subpath, which likely needs webpack `publicPath` and router base-path rework.
- Pages would split hosting, because production docs stay on Amplify.
- Pages' one real gain is removing the AWS credential. That gain is smaller than it looks, because Pages still needs the same gated build, the same Environment, and the same maintainer approval.

<details><summary><h2>Implementation details and Decisions for Reviewers</h2></summary>
<p>

### Two gates, and why the build is gated

There are two gates.

- Gate A, in this repo's code: one workflow on `pull_request_target` builds and deploys the preview. The build job carries a GitHub Environment. For an outside PR that environment has required reviewers, so the build pauses before any fork code runs or any artifact uploads. A member PR routes to a no-reviewer environment and runs at once.
- Gate B, a repo setting: "Require approval for all outside collaborators" under Settings, Actions, General. This gates every other fork workflow, mainly `tests.yml`, so a fork's push does not run those or upload their artifacts until a maintainer approves.

The gate sits on the build job, not the deploy job. This is deliberate. If the build ran before approval, an outside author could push commit after commit, each firing a build that uploads a large artifact. That would consume Actions minutes and artifact storage with no approval. Gating the build closes that path. Nothing runs for an outside PR until a maintainer approves.

`pull_request_target` is exempt from the Gate B setting, because it runs trusted base-branch code. So Gate A covers the preview and Gate B covers the rest. Together they mean no fork workflow runs unattended.

### The trust model of the single workflow

The build job runs untrusted fork code (only after review and approval by a member of this repo). The deploy job holds the credential. They are separate jobs, so they never share a runner.

- The build job checks out the fork head and runs `pnpm build` with `permissions: contents: read` and no secret. Even though it runs fork code, there is nothing to steal.
- The deploy job runs no fork code. It downloads the built files as an artifact and pushes them to Amplify. The fork controls the preview's static content. That is the point of a preview. No fork code executes in the job that holds the credential.

`pull_request_target` runs the workflow definition from the base branch, so the workflow code is trusted. `author_association` comes straight from the trusted event, not from the artifact and not from an API call, so a non-member cannot spoof their identity and route their PR to the no-approval environment. Using `pull_request_target` safely depends on one rule: the job that checks out fork code gets minimal permissions and no secret. The design follows that rule.

### Why the gate keys on author_association, not a fork check

The GitHub rule is about write access, not org membership.

- Anyone without write access cannot push a branch to the repo. Their only path is to fork and open the PR from the fork. So every outside contributor PR is a fork PR.
- A Deque org member can still work from a personal fork. Their PR would then be a fork PR even though they are a member.
- A non-member could be granted collaborator write access. Their PR would then be a same-repo PR.

A plain fork check would wrongly block a member who works from a fork. Gating on `author_association` in `OWNER`, `MEMBER`, `COLLABORATOR` follows the trust boundary we want.

### Why not a label

One early idea was to allow members to add a label to a PR they had reviewed and deemed safe, but a label would be exploitable. An outside author could open an innocuous PR, wait for a maintainer to apply the label, then push malicious code that still deploys, because the label persists across pushes.

The Environment approval strategy I used binds to one workflow run and one commit. Each push makes a new run that must be approved again.

### The required PR review is unchanged

Cauldron requires an approving review from a write-access person before any PR can merge. This work does not touch that rule. The preview approval is a separate gate that only controls whether the docs preview deploys. It is not a PR review and does not replace one.

The intended order is that a maintainer reviews the full PR first, including its code and security, and the preview approval is the last step to confirm the change works. So the preview never builds or deploys code a maintainer has not already read.

### One point a reviewer may ask about

The build job checks out the fork head, then runs `./.github/actions/dependencies`. That local action resolves from the checked-out fork tree, so it is the fork's copy of the action, not the base branch's. That is intentional. The build job runs fork code either way, since `pnpm build` is fork code. The job has no secret and only `contents: read`, so running the fork's dependencies action adds no new access. The checkout sets `persist-credentials: false`, so the read-only token is not left in git config.

</p>
</details> 

## Requires setup outside this repo before previews work

> [!IMPORTANT]
>  This PR cannot enable previews on its own. Someone with AWS and GitHub admin access must set up a Github OIDC provider and an IAM role, then in Github set up a repo variable, 2 environments, and turn on a gate.
> 
> The details and order of operations are below. Reviewers please also review that.

Closes: [#2486](https://github.com/dequelabs/cauldron/issues/2486)


rheisler-deque:
To Be reviewed with the code changes in this PR

## Facts you need first

- Region: `us-east-1`.
- Repo: `dequelabs/cauldron`.
- AWS account id: `935235593273`, the account that owns the Amplify apps. Confirm you are in it with `aws sts get-caller-identity --query Account --output text`.

## Part 1: AWS

### Step 1.1: Create the manual-deploy preview app

Create a new Amplify app with no repository. An app created without a repository is in manual-deploy mode, which is what the workflow needs. The workflow deploys with the Amplify manual deploy API, and that API only works on an app not connected to Git. Both existing Cauldron apps are Git-connected, so the previews need their own app.

`aws amplify create-app --name cauldron-pr-previews --platform WEB --region us-east-1 --query app.appId --output text`

`--platform WEB` is a static site. Manual deploy does not support server-side rendered apps, and the docs site is static. The command prints the new app id. Save it as `<PREVIEW_APP_ID>`. You use it in Step 1.4 and in Part 2.

### Step 1.2: Set the site rewrite rules

A new app has no rewrite rules. The docs site needs them, or `/storybook` and every client-side route return 404. These rules match what the existing apps already serve.

Save this as `custom-rules.json`:

```json
[
  { "source": "/storybook", "target": "/storybook/", "status": "301" },
  { "source": "/storybook/", "target": "/storybook/index.html", "status": "200" },
  { "source": "/<*>", "target": "/index.html", "status": "404-200" }
]
```

Apply them:

`aws amplify update-app --app-id <PREVIEW_APP_ID> --region us-east-1 --custom-rules file://custom-rules.json`

The `/<*>` rule must stay last. Amplify reads the rules in order, and it is the catch-all that sends unmatched paths to the single-page-app entry point.

### Step 1.3: Write the trust policy

This account already has the GitHub OIDC provider that GitHub Actions uses, confirmed on 2026-08-18. So there is no provider-creation step.

This says who may assume the role. It allows only the GitHub OIDC provider, only the audience `sts.amazonaws.com`, and only the subject `repo:dequelabs/cauldron:pull_request`. That subject is what a workflow job with no environment gets on a `pull_request_target` event. The deploy and cleanup jobs both match it.

Save this as `trust-policy.json`.

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "arn:aws:iam::935235593273:oidc-provider/token.actions.githubusercontent.com"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "token.actions.githubusercontent.com:aud": "sts.amazonaws.com",
          "token.actions.githubusercontent.com:sub": "repo:dequelabs/cauldron:pull_request"
        }
      }
    }
  ]
}
```

### Step 1.4: Write the permissions policy

This grants only the Amplify actions the workflows call, only on this one app. Save this as `permissions-policy.json`.

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "AmplifyPreviewDeploy",
      "Effect": "Allow",
      "Action": [
        "amplify:GetBranch",
        "amplify:CreateBranch",
        "amplify:DeleteBranch",
        "amplify:CreateDeployment",
        "amplify:StartDeployment"
      ],
      "Resource": [
        "arn:aws:amplify:us-east-1:935235593273:apps/<PREVIEW_APP_ID>",
        "arn:aws:amplify:us-east-1:935235593273:apps/<PREVIEW_APP_ID>/branches/*"
      ]
    }
  ]
}
```

Note. This is the exact set the two workflows call. The deploy job calls `GetBranch`, `CreateBranch`, `CreateDeployment`, and `StartDeployment`. The cleanup job calls `GetBranch` and `DeleteBranch`. The zip upload uses a presigned URL and needs no IAM permission. `amplify:GetJob` is not needed, because the workflow does not poll deployment status. Add it later only if you make the deploy wait on the job.

### Step 1.5: Create the role and attach the policy

Create the role with the trust policy:

`aws iam create-role --role-name cauldron-docs-preview-deploy --assume-role-policy-document file://trust-policy.json --description "GitHub OIDC role for gated Cauldron docs PR previews"`

Attach the permissions policy inline:

`aws iam put-role-policy --role-name cauldron-docs-preview-deploy --policy-name amplify-preview-deploy --policy-document file://permissions-policy.json`

Read back the role ARN. You paste this into GitHub in Part 2.

`aws iam get-role --role-name cauldron-docs-preview-deploy --query Role.Arn --output text`

## Part 2: GitHub

### Step 2.1: Set the repo variables

Set the two Actions variables both workflows read. Use the ARN from Step 1.5 for the first. Use the new app id for the second.

`gh variable set AMPLIFY_PREVIEW_ROLE_ARN --repo dequelabs/cauldron --body "arn:aws:iam::935235593273:role/cauldron-docs-preview-deploy"`

`gh variable set AMPLIFY_APP_ID --repo dequelabs/cauldron --body "d3004bc8l35m2l"`

Console path if you prefer: Settings, then Secrets and variables, then Actions, then the Variables tab, then New repository variable. Create both `AMPLIFY_PREVIEW_ROLE_ARN` and `AMPLIFY_APP_ID`. Both workflows fail if `AMPLIFY_APP_ID` is missing, so set it before the PR merges.

### Step 2.2: Create the member environment

This is the no-reviewer path. Member PRs use it and deploy with no pause.

- Go to Settings, then Environments, then New environment.
- Name it exactly `pr-preview-auto`.
- Add no required reviewers.
- Under Deployment branches and tags, choose Selected branches and tags, and add a rule for `develop`. Under `pull_request_target` the ref is the base branch, which is `develop`.
- Save.

### Step 2.3: Create the gated environment

This is the approval path. Outside PRs use it and pause until a maintainer approves.

- New environment. Name it exactly `pr-preview-gated`.
- Turn on Required reviewers. Add the Cauldron maintainer team or the specific maintainers who should approve previews.
- Set the same Deployment branches and tags rule for `develop`.
- Save.

Note. Required reviewers on an environment are free for public repos. Cauldron is public, so this is available.

Warning. Never add a secret to `pr-preview-auto` or `pr-preview-gated`. The build job sets its environment to one of these and runs fork code inside it. A secret on either environment would be readable by that fork code. These environments carry only reviewers and branch rules, never secrets. The AWS role lives in a repo variable, not here.

### Step 2.4: Turn on Gate B, the fork-approval setting

This gates the other fork workflows, mainly `tests.yml`. The preview workflow does not need it, because `pull_request_target` is exempt, but the rest of the repo does.

- Go to Settings, then Actions, then General.
- Find Fork pull request workflows from outside collaborators.
- Select Require approval for all outside collaborators.
- Save.

## Part 3: Merge the PR

Merge this branch's PR to `develop`. A `pull_request_target` workflow runs the definition from the base branch, so these workflows only run once they are on `develop`.

Do Parts 1 and 2 before you merge. If the workflow goes live before the `pr-preview-gated` environment exists, GitHub auto-creates that environment on first use with no reviewers. An outside PR would then build with no pause during that window. The build holds no secret, so the risk is only CI compute use. The deploy would fail anyway, because the role variable would not exist yet.

## Part 4: Confirm it works

- Open a PR from a branch inside the repo, as a member. The build should run with no pause. The deploy should run. A comment with the preview link should appear. The site should load, including `/storybook`.
- Open a PR from a fork by a non-member, or ask someone outside the org. The build job should pause with a "waiting for approval" state, and no runner should start.
- Approve it. The build should run, the deploy should follow, and the link should comment.
- Confirm the fork PR's `tests.yml` also waited for approval. That confirms Gate B is on.
- Close both PRs. The cleanup workflow should delete the `pr-<number>` Amplify branches with no pause.

## Part 5: Retire the old preview app

Do this only after Part 4 passes and the new previews work. This removes the old `cauldron-preview` app (`d15792l1n26ww3`). Native previews used it. Nothing needs it once previews run on the new app.

First confirm nothing depends on it. Check for a custom domain:

`aws amplify list-domain-associations --app-id d15792l1n26ww3 --region us-east-1`

If that returns any domain, stop. Do not delete the app. A custom domain means something still points at it. Move the domain off it first, or leave the app in place.

If there is no domain, and you have confirmed the app serves nothing you need, delete it:

`aws amplify delete-app --app-id d15792l1n26ww3 --region us-east-1`

Leave `cauldron-dev` (`d1gko6en628vir`) alone. It is a separate app.

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,122 @@
...
+          path: docs/dist
+          retention-days: 7
+
+  deploy:
+    needs: [resolve, build]
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      pull-requests: write
+    steps:
```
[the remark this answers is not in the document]
rheisler-deque: GitHub's fine-grained permissions reference lists this endpoint under both the Issues and Pull requests permissions, so for a PR target `pull-requests: write` should suffice. But field reports of a 403 with `pull-requests: write` alone are common, and the canonical comment actions request `issues: write` too.

I can't prove the PR-only case without running the workflow, so rather than rely on it I've granted `issues: write` on the deploy job in 487ec606. Thanks for the catch.

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,123 @@
...
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
+      - uses: ./.github/actions/dependencies
+      - name: Build docs site
+        run: pnpm build
+      - name: Upload preview site
+        uses: actions/upload-artifact@v7
```
rheisler-deque: Suggestion: Pin these three actions to a full commit SHA, matching `actions/checkout` and `aws-actions/configure-aws-credentials`.

`actions/upload-artifact@v7`, `actions/download-artifact@v7` (line 69), and `actions/github-script@v9` (line 93) are pinned to a floating major tag. A major tag moves when the owner republishes it. This PR closes supply-chain and CI trust gaps, so pinning every third-party action to a SHA is the matching step. All three tags resolve to existing releases today, so this is hardening, not a fix.

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,123 @@
...
+          name: pr-preview-site
+          path: docs/dist
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4.3.1
+        with:
+          role-to-assume: ${{ vars.AMPLIFY_PREVIEW_ROLE_ARN }}
+          aws-region: us-east-1
+      - name: Deploy to Amplify
+        env:
+          APP_ID: d15792l1n26ww3
```
rheisler-deque: Suggestion: Move the Amplify app id `d15792l1n26ww3` into a repo variable like `AMPLIFY_APP_ID`.

It is hardcoded in three places across two files: here, the comment script (line 100), and the cleanup workflow (`pr-preview-cleanup.yml:23`). The role ARN already lives in the variable `AMPLIFY_PREVIEW_ROLE_ARN`. Matching that pattern reduces a three-spot edit to one. This is maintainability, not correctness.

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,123 @@
...
+            *) environment=pr-preview-gated ;;
+          esac
+          echo "pr_number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
+          echo "environment=$environment" >> "$GITHUB_OUTPUT"
+          echo "PR #$PR_NUMBER association $ASSOCIATION routes to $environment"
+
+  build:
+    needs: resolve
+    runs-on: ubuntu-latest
+    environment:
```
rheisler-deque, line 41: Suggestion: Record in the setup runbook that `pr-preview-auto` and `pr-preview-gated` must never hold environment secrets.

The build job sets `environment:` to the routed environment and runs fork code inside it. If either environment is ever given a secret, that fork code can read it. The runbook creates both with only reviewers and branch rules, so nothing leaks today. A one-line warning would stop a future maintainer adding a secret by habit. An alternative is to move the approval gate onto a separate job that does no checkout, so fork code never runs in an environment context.

.github/workflows/pr-preview.yml
rheisler-deque: done

.github/workflows/pr-preview.yml
rheisler-deque: done, and added creating that variable to the runbook

.github/workflows/pr-preview.yml
rheisler-deque: done

chornonoh-vova (dismissed):

Bracciata (dismissed):
Great work!

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,126 @@
...
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
+      - uses: ./.github/actions/dependencies
+      - name: Build docs site
+        run: pnpm build
+      - name: Upload preview site
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
```
anastasialanz:
```suggestion
        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```

This needs the full version number to update correctly with Dependabot.

https://github.com/actions/upload-artifact/releases/tag/v7.0.1

rheisler-deque: Thanks! Fixed

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,126 @@
...
+          BRANCH="pr-${PR_NUMBER}"
+          aws amplify get-branch --app-id "$APP_ID" --branch-name "$BRANCH" >/dev/null 2>&1 || aws amplify create-branch --app-id "$APP_ID" --branch-name "$BRANCH" >/dev/null
+          ( cd docs/dist && zip -r -q "$GITHUB_WORKSPACE/site.zip" . )
+          DEPLOYMENT="$(aws amplify create-deployment --app-id "$APP_ID" --branch-name "$BRANCH")"
+          JOB_ID="$(echo "$DEPLOYMENT" | jq -r '.jobId')"
+          ZIP_URL="$(echo "$DEPLOYMENT" | jq -r '.zipUploadUrl')"
+          curl -fsS -H "Content-Type: application/zip" --upload-file site.zip "$ZIP_URL"
+          aws amplify start-deployment --app-id "$APP_ID" --branch-name "$BRANCH" --job-id "$JOB_ID" >/dev/null
+      - name: Comment preview URL
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
```
anastasialanz:
```suggestion
        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
```

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,126 @@
...
+  deploy:
+    needs: [resolve, build]
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      pull-requests: write
+      issues: write
+    steps:
+      - name: Download preview site
+        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7
```
anastasialanz: We could update to the latest which is v8.0.1

https://github.com/actions/download-artifact/releases/tag/v8.0.1

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,126 @@
...
+        with:
+          name: pr-preview-site
+          path: docs/dist
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4.3.1
```
anastasialanz: This action is currently on v6.2.3

https://github.com/aws-actions/configure-aws-credentials/releases/tag/v6.2.3

.github/workflows/pr-preview.yml
rheisler-deque: Thanks, I didn't realize there were more. All fixed.

.github/workflows/pr-preview.yml
```diff
@@ -0,0 +1,126 @@
...
+      - name: Deploy to Amplify
+        env:
+          APP_ID: ${{ vars.AMPLIFY_APP_ID }}
+          PR_NUMBER: ${{ needs.resolve.outputs.pr_number }}
+        run: |
+          set -euo pipefail
+          BRANCH="pr-${PR_NUMBER}"
+          aws amplify get-branch --app-id "$APP_ID" --branch-name "$BRANCH" >/dev/null 2>&1 || aws amplify create-branch --app-id "$APP_ID" --branch-name "$BRANCH" >/dev/null
+          ( cd docs/dist && zip -r -q "$GITHUB_WORKSPACE/site.zip" . )
+          DEPLOYMENT="$(aws amplify create-deployment --app-id "$APP_ID" --branch-name "$BRANCH")"
```
frankensteinke, line 88: Have you tested this against the Cauldron app yet? `CreateDeployment` is documented as being for manually deployed Amplify apps  (those not connected to a Git repository). Just wanted to make sure it would work here since the app is connected to the repo.

.github/workflows/pr-preview-cleanup.yml
```diff
@@ -0,0 +1,33 @@
+name: PR Preview Cleanup
+
+on:
+  pull_request_target:
+    types: [closed]
+    branches:
+      - develop
+
+permissions:
+  id-token: write
```
frankensteinke, line 10: Do we want to link this to the preview workflow's concurrency group to handle the case of a job running after a PR is closed? The deploy job would create the branch after this cleanup already ran, so nothing would be left to delete it. Narrow use case, but it's a small guard.

```suggestion
concurrency:
  group: PR Preview-${{ github.event.pull_request.number }}
  cancel-in-progress: true

permissions:
  id-token: write
```

frankensteinke (commented):
Thanks for working on this! Looks good overall, just left a couple of minor comments/questions.

.github/workflows/pr-preview-cleanup.yml
rheisler-deque: Good catch, fixed

.github/workflows/pr-preview.yml
rheisler-deque: Very good catch, thank you! All we have to do is create a new Amplify app that isn't connected to Git for the PR previews to deploy to. That's the simplest change that allows the production docs deployments to continue to happen the way they do now.

[I updated my comment with the plan](https://github.com/dequelabs/cauldron/pull/2504#issuecomment-5218980353) to reflect this. Let me know what you think

frankensteinke: Sounds reasonable to me!

frankensteinke (dismissed):

chornonoh-vova (dismissed):

anastasialanz (dismissed):

frankensteinke (approved):

PR #2506, opened 2026-08-12
fix: screen reader announcing submenu items as part of the menu name

chornonoh-vova:
# Summary

Fixes screen readers announcing every submenu item as part of the trigger's accessible name in the `TopBar`/`MenuBar` + `ActionMenu` (`renderInTrigger`) pattern, by pointing both the trigger and the menu at a dedicated `labelId` label element, with a fallback to the previous labelling for triggers that don't render one.

> [!WARNING]
> This can only be tested manually with screen reader, I haven't found a way to test it in unit tests.


packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -176,6 +185,29 @@ const ActionMenuComponent = forwardRef<HTMLElement, ActionMenuProps>(
...
+      if (
+        renderInTrigger &&
+        !labelElement &&
+        process.env.NODE_ENV !== 'production'
+      ) {
+        console.warn(
+          `renderInTrigger should render id="${labelId}" on the element wrapping the trigger's visible label, otherwise the trigger and the menu are named after the entire submenu.`
+        );
+      }
+    }, [renderInTrigger, labelId]);
```
frankensteinke, line 210: Nit: Could you split these, so that we have two single-purpose `useEffect` blocks rather than this all-in-one one?

packages/react/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -176,6 +185,29 @@ const ActionMenuComponent = forwardRef<HTMLElement, ActionMenuProps>(
...
+
+    const labelledBy = hasLabelElement ? labelId : triggerId;
```
frankensteinke: Maybe we could just keep `labelledBy` as `undefined` if a `labelId` isn't provided, that way we aren't explicitly connecting an element that we know is wrong? I think that keeping it undefined would keep the current behavior anyway, so we wouldn't be losing anything.

```suggestion
    let labelledBy: string | undefined = triggerId;
    if (renderInTrigger) {
      labelledBy = hasLabelElement ? labelId : undefined;
    }
```

frankensteinke (approved):

PR #2515, opened 2026-08-28
fix(ActionList): stop mirroring Listbox's active option

### LINKED ISSUE #2512: ActionMenu/Listbox can enter an unbounded render loop under CPU load when `focusStrategy="first"`
## Summary

`Listbox`, `ActionList`, and `ActionMenu` keep the active option in sync through a two-way mirror. Under normal timing this mirror settles after one pass. Under heavy CPU load it can fail to settle and instead loop without end. React then prints thousands of `Maximum update depth exceeded` warnings and the process can run out of memory.

This affects `@deque/cauldron-react`. We saw it in 7.2.0. It is still present in the current release, 7.4.0.

## Where the bug is

The active option is stored in two places at once and each copy updates the other.

- `ActionList` holds its own `activeOption` state and passes it to `Listbox` as a controlled prop. It updates that state from `Listbox` through the `onActiveChange` callback. See `packages/react/src/components/ActionList/ActionList.tsx:32`, `:81`, and `:83`.
- `Listbox` copies the controlled prop into its own internal `activeOption` state. It has two effects that keep the two copies in sync. One effect reports its own active option to the parent through `onActiveChange`. The other effect copies the parent's prop into its own state when the two differ. See `packages/react/src/components/Listbox/Listbox.tsx:155` and `:161`.
- `Listbox`'s focus handler for `focusStrategy === 'first'` calls `setActiveOption(firstOption)` on every focus event. It does not first check whether `firstOption` is already the active option. See `packages/react/src/components/Listbox/Listbox.tsx:300`.
- `ActionMenu` sets `focusStrategy` to `'first'` by default and passes it down on every open. See `packages/react/src/components/ActionMenu/ActionMenu.tsx:84` and `:203`.

## Root cause

The focus handler has no "already active" early return. Each focus event writes the active option again. That write makes `Listbox` report the option to the parent. The parent re-renders and passes the option back down as a prop. The sync effect then copies that prop into `Listbox`'s state again.

When timing is normal, the sync effect sees the same element and stops, so the loop settles after one pass. When the CPU is saturated, the two effects interleave in an order where the two copies never reach a matching state, so the updates repeat without end. React's update-depth guard fires and the process eventually runs out of memory.

We reproduced this three times, each time independently, and captured about 14,600 warning stacks. Every stack ran through the library's own components, not through any of our code. All of the stacks followed the same call path, which points to a single cause.

Limiting the number of test processes that run at once stops the loop. We verified this. It confirms the loop is driven by concurrent load.

## How it was found

We hit this in a test suite in another repo. One test file that mounts `ActionMenu` failed at random. Only that one file failed. The rest of the suite was unaffected. That file passed every time when we ran it alone. It only failed when enough test processes ran at once to saturate the CPU. On a busy machine the failing process printed more than 13,000 `Maximum update depth exceeded` warnings over about three minutes and then died with a V8 heap out-of-memory error.

The failure is a scheduler-timing race, so machines with more cores hit it more often.

## Reproduction

The loop is a timing race, so it does not reproduce in a small or isolated test. It needs many test processes competing for the CPU at once.

- Run a large test suite that mounts `ActionMenu` on a machine with many cores.
- Or run `node --test` with `--test-concurrency=20` over a large `*.test.{ts,tsx}` glob. This reproduces the loop reliably, usually in under two minutes, with the failure at about 135 seconds. `node:test` only spawns as many worker processes as files given, so a small subset will not reproduce it.

## Ruled out

We eliminated these as causes:

- Prop object identity. Memoizing the props passed to `ActionMenu` does not help.
- Trigger ref forwarding.
- The number of test files in the suite. The crash reproduced with a smaller set of files too. The count of files only changes how often the race fires, not whether the bug exists.

There is no consumer-side workaround. The props needed to break the loop from outside are not exposed by `ActionMenu`.

## Suggested fix

Add an "already active" check in `Listbox`'s focus handler so it does not call `setActiveOption` when the target option is already the active one. That removes the repeated write that starts the loop. Apply this check in both the `focusStrategy === 'first'` branch and the `focusStrategy === 'last'` branch at `packages/react/src/components/Listbox/Listbox.tsx:300`. Both branches write on every focus event without checking, so both need the fix. `ActionMenu` triggers the `'first'` branch, but the `'last'` branch has the same bug.

dbjorge:
Closes #2512

## What this changes

The active option was stored in **two places** and synced in **both directions**:

- `Listbox` keeps `activeOption` in local state and reports every change upward via `onActiveChange`.
- `ActionList` stored that reported value in its own state and passed it straight back down as the `activeOption` controlled prop.

`Listbox`'s downward sync effect only re-runs when that prop's *object identity* changes, so it can fire late, carrying a value that is already out of date, and overwrite a newer local one. That stale revert is reported back up, the parent adopts it, and the two copies flip between two items until React's update-depth guard trips.

`Listbox` now owns the active option outright. `ActionList` keeps `activeOption` only as a one-shot mnemonic request — which must be a fresh object on every match, or the request never reaches `Listbox`.

## Why it only showed up under load

I instrumented every write on both sides and captured a real failing run. The healthy and broken orderings differ only in which effect runs first after a commit:

```
HEALTHY: report(180) → parent=180 → click(181) → downSync: no-op → report(181) → parent=181 → no-op
BROKEN:  report(180) → parent=180 → click(181) → report(181) → parent=181 → downSync applies stale 180
```

Under light load React batches the updates and they settle in one pass. Under load they split across separate commits, the downward sync is permanently one step behind, and the ping-pong never converges — ~1,250 alternations per item before the guard fires.

Note that the trigger in the captured trace was an ordinary click (`handleSelect`), **not** the focus handler. That is why the first commit here did not fix it on its own.

## Commits

- `f299065a` — the original focus-handler guard. Still a real improvement (avoids a redundant write and a redundant `onActiveChange` call on every focus) with its own tests, but **not** what resolves this issue. Kept deliberately.
- `466d5956` — the actual fix.
- `fa57a9fe` — drops `activeElement`, which was written on every active-option change and never read by anything. Removing the last mirror write left `handleActiveChange` with nothing to do but that dead write, so the ref, the callback, and the `onActiveChange` prop it was wired to all go together. No behavior change.

## Verification

The previous attempt was merged on reasoning alone and turned out not to fix the bug, so this time the claim is backed by runs against the suite that actually fails.

Method: built `packages/react` from this branch and pointed walnut's `packages/ui-issues` at it via a `file:` override in `pnpm-workspace.yaml`, then ran the failing suite repeatedly:

```
NODE_ENV=test node --require ts-node/register --require ./src/test-setup.ts \
  --test-reporter spec --test --test-concurrency=20 'src/**/*.test.tsx' 'src/**/*.test.ts'
```

| Build | Runs | Runaway loops |
|---|---|---|
| cauldron 7.5.0 (baseline) | 5 | 3 |
| first commit only (`f299065a`) | 14 | 3 |
| **this branch** | **35** | **0** |

A looping run hangs 140–180s and prints `Maximum update depth exceeded` ~13,800 times; a clean run finishes the whole suite in 13–16s. The failure is unmistakable when it happens.

Against the 21% failure rate the previous build showed, 35 clean runs would happen by luck roughly 0.02% of the time. Against the 60% baseline rate it is effectively impossible.

Cauldron's own suite: 1176 tests across 95 suites pass, plus typecheck, lint and format.

One caveat worth stating plainly: **the race itself cannot be reproduced deterministically in a unit test**, because it depends on how React splits commits under scheduling pressure. The new test in `ActionMenu.test.tsx` covers the behavior the fix changes — a mnemonic match must still move the active option after arrow keys have moved away from it — and was written RED first: it fails with the mirror removed but the one-shot request reusing its previous object, and passes with both halves of the fix in place. The 35-run walnut result is the evidence for the race being gone.

Also verified in passing: the 9 `act()` warnings in the walnut runs are unchanged from clean baseline runs, so they are pre-existing and unrelated.

## Follow-up

`Listbox` accepts `activeOption` as a controlled prop while continuing to write its own local copy from 13 internal call sites. Its other controlled prop (`value`) gates internal writes on `!isControlled`, and `Combobox` does the same — `activeOption` is the outlier, so the idiomatic controlled usage is the unsafe one.

I searched the `dequelabs` org: no repo outside cauldron passes `activeOption` to a `Listbox`, so this PR removes the only affected usage. Making `Listbox` properly controlled is a separate change, and per `CONTRIBUTING.md` it needs a deprecation flagged for at least two months first. Tracked separately in #2522 to avoid delaying the bugfix.

### DIFF
--- a/packages/react/src/components/ActionList/ActionList.tsx
+++ b/packages/react/src/components/ActionList/ActionList.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, useCallback, useRef, useState } from 'react';
+import React, { forwardRef, useCallback, useState } from 'react';
 import classnames from 'classnames';
 import { type ListboxOption } from '../Listbox/ListboxContext';
 import Listbox from '../Listbox';
@@ -26,21 +26,12 @@ interface ActionListProps extends Omit<
 
 const ActionList = forwardRef<HTMLUListElement, ActionListProps>(
   ({ selectionType = null, onAction, className, children, ...props }, ref) => {
-    const activeElement = useRef<HTMLLIElement | HTMLAnchorElement | null>(
-      null
-    );
+    // Listbox owns the active option. ActionList must not mirror it back down
+    // as a controlled prop: the two copies then race, and a downward sync
+    // carrying an already-stale value can revert a newer one and oscillate.
+    // `activeOption` below is only a one-shot mnemonic request, never a
+    // mirror. See cauldron#2512.
     const [activeOption, setActiveOption] = useState<ListboxOption>();
-
-    // Deliberately does not mirror the reported option back into
-    // `activeOption`. Listbox owns the active option; feeding its own value
-    // back down as a controlled prop makes the two copies race, and a
-    // downward sync carrying an already-stale value can revert a newer one
-    // and oscillate. See cauldron#2512.
-    const handleActiveChange = useCallback((value: ListboxOption) => {
-      activeElement.current = value?.element as
-        | HTMLLIElement
-        | HTMLAnchorElement;
-    }, []);
 
     const handleAction = useCallback(
       (key: string, event: onActionEvent) => {
@@ -86,7 +77,6 @@ const ActionList = forwardRef<HTMLUListElement, ActionListProps>(
         className={classnames('ActionList', className)}
         activeOption={activeOption}
         {...props}
-        onActiveChange={handleActiveChange}
         navigation="bound"
       >
         <ActionListProvider
--- a/packages/react/src/components/ActionList/ActionList.tsx
+++ b/packages/react/src/components/ActionList/ActionList.tsx
@@ -31,11 +31,15 @@ const ActionList = forwardRef<HTMLUListElement, ActionListProps>(
     );
     const [activeOption, setActiveOption] = useState<ListboxOption>();
 
+    // Deliberately does not mirror the reported option back into
+    // `activeOption`. Listbox owns the active option; feeding its own value
+    // back down as a controlled prop makes the two copies race, and a
+    // downward sync carrying an already-stale value can revert a newer one
+    // and oscillate. See cauldron#2512.
     const handleActiveChange = useCallback((value: ListboxOption) => {
       activeElement.current = value?.element as
         | HTMLLIElement
         | HTMLAnchorElement;
-      setActiveOption(value);
     }, []);
 
     const handleAction = useCallback(
@@ -47,10 +51,12 @@ const ActionList = forwardRef<HTMLUListElement, ActionListProps>(
       [onAction]
     );
 
+    // A new object every match, even for the same element: this is a one-shot
+    // request to Listbox rather than a mirror of its state, so re-using the
+    // previous object would leave the controlled prop unchanged and the
+    // request would never reach Listbox.
     const handleMnemonicMatch = useCallback((element: HTMLElement) => {
-      setActiveOption((prev) =>
-        prev?.element === element ? prev : { element }
-      );
+      setActiveOption({ element });
     }, []);
 
     const containerRef = useMnemonics<HTMLUListElement>({
--- a/packages/react/src/components/ActionList/ActionList.tsx
+++ b/packages/react/src/components/ActionList/ActionList.tsx
@@ -30,7 +30,8 @@ const ActionList = forwardRef<HTMLUListElement, ActionListProps>(
     // as a controlled prop: the two copies then race, and a downward sync
     // carrying an already-stale value can revert a newer one and oscillate.
     // `activeOption` below is only a one-shot mnemonic request, never a
-    // mirror. See cauldron#2512.
+    // mirror. See cauldron#2512; cauldron#2522 tracks making `activeOption`
+    // properly controlled so this constraint lives in the interface instead.
     const [activeOption, setActiveOption] = useState<ListboxOption>();
 
     const handleAction = useCallback(
--- a/packages/react/src/components/Listbox/Listbox.tsx
+++ b/packages/react/src/components/Listbox/Listbox.tsx
@@ -304,7 +304,7 @@ const Listbox = forwardRef<
             ? options.find((option) => !isDisabledOption(option))
             : options[0];
 
-          if (firstOption) {
+          if (firstOption && firstOption.element !== activeOption?.element) {
             setActiveOption(firstOption);
           }
 
@@ -316,7 +316,7 @@ const Listbox = forwardRef<
             ? [...options].reverse().find((option) => !isDisabledOption(option))
             : options[options.length - 1];
 
-          if (lastOption) {
+          if (lastOption && lastOption.element !== activeOption?.element) {
             setActiveOption(lastOption);
           }
 
--- a/packages/react/src/components/ActionMenu/ActionMenu.test.tsx
+++ b/packages/react/src/components/ActionMenu/ActionMenu.test.tsx
@@ -491,6 +491,50 @@ test('should set matching active item on mnemonic key press', async () => {
   });
 });
 
+// Regression test for cauldron#2512. ActionList no longer mirrors Listbox's
+// reported active option, so a mnemonic match must be a fresh request every
+// time -- otherwise re-matching an item that arrow keys have since moved away
+// from is silently dropped.
+test('should set matching active item on mnemonic key press after arrow navigation', async () => {
+  const user = userEvent.setup();
+  render(
+    <ActionMenu {...defaultProps}>
+      <ActionList>
+        <ActionListItem>Apple</ActionListItem>
+        <ActionListItem>Banana</ActionListItem>
+        <ActionListItem>Cherry</ActionListItem>
+      </ActionList>
+    </ActionMenu>
+  );
+
+  screen.getByRole('button', { name: 'Trigger' }).focus();
+  await user.keyboard('{ArrowDown}');
+  await waitForMenuFocus();
+
+  await user.keyboard('c');
+  await waitFor(() => {
+    expect(screen.queryAllByRole('menuitem')[2]).toHaveClass(
+      'ActionListItem--active'
+    );
+  });
+
+  // Move the active option off of Cherry with arrow navigation.
+  await user.keyboard('{ArrowUp}{ArrowUp}');
+  await waitFor(() => {
+    expect(screen.queryAllByRole('menuitem')[0]).toHaveClass(
+      'ActionListItem--active'
+    );
+  });
+
+  // Matching Cherry again must move the active option back to it.
+  await user.keyboard('c');
+  await waitFor(() => {
+    expect(screen.queryAllByRole('menuitem')[2]).toHaveClass(
+      'ActionListItem--active'
+    );
+  });
+});
+
 // Regression test for cauldron#1993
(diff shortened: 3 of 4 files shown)


rheisler-deque:
@dbjorge I ran the failing Walnut suite against a build of this branch, and the runaway loop still happens. Thanks @denysfedorov for helping me reproduce it. I left some suggestions of where else to look that Claude generated from its context after reproing the error, I figured that would be useful for whoever continues working on this.

Here are the details of what I did so anyone can repeat it or point out a mistake in my setup.

<details>
<summary><h2>How I reproduced this locally</h2></summary>

### Reproducing the error on `develop` (baseline)

Walnut's `packages/ui-issues` consumes `@deque/cauldron-react`. The catalog pins `^7.5.0`, which resolves to 7.5.0. From `walnut/packages/ui-issues`, run this command:

```
NODE_ENV=test node --require ts-node/register --require ./src/test-setup.ts --test-reporter spec --test --test-concurrency=20 'src/**/*.test.tsx' 'src/**/*.test.ts'
```

The run fails. `Maximum update depth exceeded` prints tens of thousands of times. The failing test is `src/components/SortByActionMenu/SortByActionMenu.test.tsx`. It hangs for about 90 seconds before the run gives up. `SortByActionMenu` renders cauldron `ActionMenu` and `ActionList`. The log also repeats `An update to Listbox inside a test was not wrapped in act(...)`. So the loop is in the ActionMenu and Listbox path this PR targets. It reproduced on the first run for me.

### Pointing local Walnut at this branch's cauldron build

The idea is to build `@deque/cauldron-react` from this PR, then override Walnut's copy with a local `file:` link.

1. Get this branch's cauldron code. I used a detached worktree at the PR head so my main checkout stayed put.

```
git -C <cauldron> fetch origin pull/2515/head
git -C <cauldron> worktree add --detach <path>/cauldron-pr2515 f299065a
```

2. Install and build the react package in that worktree.

```
pnpm --dir <path>/cauldron-pr2515 install
pnpm --dir <path>/cauldron-pr2515 build:react
```

Note: cauldron pins pnpm 11.4.0 through the `packageManager` field. If your local pnpm differs, corepack refuses to run. Use a matching pnpm, or change the pin in that throwaway worktree.

3. In Walnut, add an override in `pnpm-workspace.yaml` so every `@deque/cauldron-react` resolves to the build. Walnut keeps overrides there, not in `package.json`.

```
overrides:
  '@deque/cauldron-react': 'file:/absolute/path/to/cauldron-pr2515/packages/react'
```

4. Reinstall Walnut.

```
pnpm install
```

5. Verify the loaded copy has the fix. `packages/ui-issues/node_modules/@deque/cauldron-react` should resolve to the `file:` build. Its `lib/index.js` should contain the guard `firstOption && firstOption.element !== ...`.

One note on the build. This PR head is cauldron `develop` plus one commit, this fix. Its parent commit `7edd719e` is the commit `7.5.0` was published from. So this build equals the installed 7.5.0 with only this fix added. Nothing else differs.

</details>

## Outcome

With the fixed build in place, the suite still fails the same way. `SortByActionMenu.test.tsx` still hangs for about 90 seconds. `Maximum update depth exceeded` still prints tens of thousands of times. It reproduced on the first run again.

So the redundant focus write was not the only thing feeding the loop, at least under this suite's load.

## What else could be driving it

- `ActionList` and `ActionMenu` mirror `Listbox`'s active option back down as a controlled prop. Removing one redundant write did not break the cycle. Another write in that sync still runs on each pass. Look for other `setActiveOption` or `onActiveChange` paths that fire when the active element is unchanged but the object identity changes.
- The parent holds a different object for the same element. Any comparison by object identity, rather than by `.element`, keeps re-triggering. This fix added an `.element` comparison in the focus handler. A similar identity-based comparison elsewhere in the sync would keep the loop alive.
- Consider whether one component alone should own the active-option state, so there is no two-way sync to settle.

dbjorge:
@rheisler-deque Thanks for the repro instructions. Was able to reproduce the issue and come up with what I think is a more complete fix.

packages/react/src/components/ActionList/ActionList.tsx
```diff
@@ -26,18 +26,13 @@ interface ActionListProps extends Omit<
 
 const ActionList = forwardRef<HTMLUListElement, ActionListProps>(
   ({ selectionType = null, onAction, className, children, ...props }, ref) => {
-    const activeElement = useRef<HTMLLIElement | HTMLAnchorElement | null>(
-      null
-    );
+    // Listbox owns the active option. ActionList must not mirror it back down
```
dbjorge: Arguably, a better fix here would be for Listbox to treat the `activeOption` as controlled if it's passed as a prop, and to let the `ActionList` own its state exclusively. I didn't go that route for this PR because it would have been a breaking change to `Listbox`.

packages/react/src/components/ActionList/ActionList.tsx
```diff
@@ -80,7 +77,6 @@ const ActionList = forwardRef<HTMLUListElement, ActionListProps>(
         className={classnames('ActionList', className)}
         activeOption={activeOption}
         {...props}
-        onActiveChange={handleActiveChange}
         navigation="bound"
```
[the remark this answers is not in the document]
dbjorge: I don't think this is really accurate; it's true (and by design) that an explicit prop would override the ActionList, but it wouldn't necessarily reintroduce the issue, it just means the caller needs to avoid the issue (the same as ActionList does now).

packages/react/src/components/Listbox/Listbox.tsx
rheisler-deque, line 169: Minor suggestion: Consider adding a development-only warning in `Listbox` that fires when a consumer passes both `activeOption` and `onActiveChange`. This would protect consumers from accidentally re-creating the bug until #2522 takes effect.

```suggestion

    useEffect(() => {
      if (
        isActiveControlled &&
        typeof onActiveChange === 'function' &&
        process.env.NODE_ENV !== 'production'
      ) {
        console.warn(
          'Listbox received both `activeOption` and `onActiveChange`. Feeding the reported active option back into `activeOption` causes an unbounded render loop under load. See cauldron#2512 https://github.com/dequelabs/cauldron/issues/2512. Let Listbox own the active option, or use `activeOption` only as a one-shot request.'
        );
      }
    }, [isActiveControlled, onActiveChange]);
    
```

rheisler-deque (dismissed):
🎉 It works 🎉 Great job, and thanks for both fixing it and capturing #2522! I ran the following tests to confirm the memory issue and the fix:

I ran walnut's `packages/ui-issues` test suite with `--test-concurrency=20`. For the baseline I used the latest cauldron `@deque/cauldron-react` 7.5.0. For the fix I built `@deque/cauldron-react` from this PR branch and pointed walnut at it with a `file:` override in `pnpm-workspace.yaml`, then reinstalled. I confirmed walnut resolved to the local build. I also verified the build contained the focus guard.

- Baseline results: 5 runs all with about 14,000 warnings. Each run took about 100 seconds
- Fix results: 8 runs all with 0 warnings. Each took about 12 seconds.

I left a minor suggestion for making this fix safer before #2522 is done.

packages/react/src/components/Listbox/Listbox.tsx
dbjorge: Added in e683741f. Note that I slightly modified the suggestion to avoid the warning refiring with every render if a consumer ever passed a callback without `useCallback` to `onActiveChange`.

rheisler-deque (approved):

Issue #2447, opened 2026-06-22
react test suite leaks ~5-6 MB/file → OOM in CI under --coverage


frankensteinke:
## Summary

The `@deque/cauldron-react` Jest suite (`jest --maxWorkers=1 --coverage`) leaks memory monotonically across test files and OOMs in CI at Node's default ~4 GB heap ceiling:

```
FATAL ERROR: Ineffective mark-compacts near heap limit
Allocation failed - JavaScript heap out of memory   (exit 134 / SIGABRT)
```

Example failed run: [actions/runs/27957522773 → `react` job](https://github.com/dequelabs/cauldron/actions/runs/27957522773/job/82729766393).

This is **not** a test assertion failure — the worker process runs out of heap part-way through the suite. With `--maxWorkers=1`, every test file runs serially in a single worker, so anything that is retained between files accumulates for the entire run.

## Evidence

Running the full suite locally with `--logHeapUsage` (Jest forces a GC before each measurement, so the numbers below are **retained, non-collectable** memory — not transient garbage) over all 93 suites:

| Point in run | Post-GC heap floor |
|---|---|
| start (Combobox) | 325 MB |
| ~suite 20 | 400 MB |
| ~suite 35 | 478 MB |
| ~suite 53 | 551 MB |
| ~suite 74 | 647 MB |
| end (93 suites) | **857 MB** |

The heap sawtooths — it climbs, GC reclaims 200–300 MB, repeat — but **the floor after each GC rises monotonically and never returns to baseline**. That is ~530 MB retained across the run, roughly **5–6 MB per test file that is never freed**.

### Why it OOMs at 4 GB in CI but passes locally

The suite runs clean locally at `--max-old-space-size=8192`. The 857 MB retained floor is itself well under 4 GB, so the leak alone does not directly explain the crash — the mechanism matters:

- `--logHeapUsage` forces a full GC after every file, which artificially suppresses the peak during measurement.
- Normal CI (no forced GC) lets V8 grow the heap lazily and only collect under pressure, so the heap floats much higher between collections — on top of the ever-growing ~857 MB+ of genuinely un-freeable retained memory.
- Once the retained set plus in-flight transients can no longer be compacted below 4 GB, V8 burns cycles in GC that can't free anything → the `Ineffective mark-compacts` OOM.

In short: the leak sets a rising floor, CI's lazy GC needs headroom above it, and the suite has simply grown until floor + headroom > 4 GB. It will keep crossing that cliff as test files are added.

[...]

PR #2520, opened 2026-09-01
fix(TreeView): make Enter run onAction consistently, and Space select

### LINKED ISSUE #2519: [a11y] Treeview "Custom onAction": additional checkboxes can't be selected, and first selection opens an unannounced dialog
## Problem
In the "Custom onAction" example, once the first checkbox is selected, no further checkboxes can be selected. Activating that first checkbox also opens a pop-up with no advance warning to the user.

## Steps to Reproduce

1. Open the TreeView docs page and go to the "Custom onAction" example.
2. Select the first checkbox — note a pop-up opens.
3. Attempt to select any additional checkbox.
4. 

## Expected Behavior
Any checkbox in the example can be selected or deselected independently. If activating a checkbox opens a dialog, that context change is communicated in advance via instructions or an accessible description.

## Actual Behavior
Only the first checkbox can be selected; subsequent ones do not respond. The first selection opens a pop-up with no prior indication, which is disorienting for keyboard and screen reader users.

mateoviilla1:
## Summary

Fixes the Enter key behaviour @VaniChinnam reported in #2519, implementing the
model she described: **Space selects, Enter runs the action**, each consistently.

## The bug

On a tree with checkboxes, Enter behaved differently from one press to the next
— sometimes running the action, sometimes toggling selection, and after the
first selection doing nothing at all. In her words: *"sometimes it opens the
modal, and sometimes it selects the checkbox. The users are unable to select the
other checkboxes as expected."*

react-aria treats a press as an action only while the selection is empty:

```js
hasPrimaryAction   = allowsActions && (selectionBehavior === 'replace' ? !allowsSelection : !allowsSelection || manager.isEmpty)
hasSecondaryAction = allowsActions && allowsSelection && selectionBehavior === 'replace'
hasAction          = hasPrimaryAction || hasSecondaryAction
```

Checkboxes force `selectionBehavior: 'toggle'`, so `hasSecondaryAction` is
always `false` and `hasPrimaryAction` collapses to `manager.isEmpty`. Selecting
something is the normal case for a checkbox tree, so from the first selection
onward `hasAction` is `false` — `onAction` became unreachable by keyboard, and
Enter stopped toggling selection too, because that path is gated on the same
flag.

Mouse users could still trigger the action on a first click. Same functionality,
available by mouse and not by keyboard — a **WCAG 2.1.1** failure, not just an
inconsistency.

## The fix

Enter is handled on the tree element in the capture phase, so react-aria never
sees it and cannot vary the outcome. Activation and selection are now separate,
and each key does exactly one thing:

- **Space** toggles the item's checkbox, as a standalone checkbox would (this
  already worked, and is unchanged).
- **Enter** runs `onAction` and leaves the selection alone.

Verified across the sequence that used to fail:

| step | action fires | selection |
| --- | --- | --- |
| Enter | yes | unchanged |
| Enter again | yes | unchanged |
| Space | no | toggled on |
| Enter, with a row now selected | **yes** | unchanged |
| Space on a second row | no | both stay selected |

Her second point from #2519 — that a context change should be announced before
it happens — is covered in the docs: the example carries visible instructions
referenced with `aria-describedby` telling users what Space and Enter do.

## Testing

- Four new unit tests, including the previously-silent case (Enter after
  something is selected). Mutation-checked: removing the handler fails two of
  them.
- Full react suite: 1128 tests pass. Docs build passes. eslint and `tsc` clean.

Note the `ActionMenu` suite intermittently reports "suite failed to run" from a
SIGTERM'd jest worker; that is the known leak in #2447 and unrelated to this
change, which touches `TreeView` only.

Closes #2519

### DIFF
--- a/docs/pages/components/TreeView.mdx
+++ b/docs/pages/components/TreeView.mdx
@@ -4,6 +4,7 @@ description: A component that displays a hierarchical tree structure.
 source: https://github.com/dequelabs/cauldron/tree/develop/packages/react/src/components/TreeView/index.tsx
 ---
 
+import { useState } from 'react';
 import { TreeView } from '@deque/cauldron-react';
 
 ```jsx
@@ -179,13 +180,12 @@ function LabelledByTreeView() {
 
 ### Custom onAction
 
-TreeView with a custom action handler for item selection.
+`onAction` runs your own handler when an item is activated, in addition to the
+item's selection change.
 
 ```jsx example
 function ActionTreeView() {
-  const onAction = (key) => {
-    alert(`Selected item: ${key}`);
-  };
+  const [lastAction, setLastAction] = useState(null);
   const items = [
     {
       id: '1',
@@ -199,12 +199,17 @@ function ActionTreeView() {
     }
   ];
   return (
-    <TreeView
-      aria-label="Tree View with Action Handler"
-      selectionMode="single"
-      onAction={onAction}
-      items={items}
-    />
+    <>
+      <TreeView
+        aria-label="Tree View with Action Handler"
+        selectionMode="multiple"
+        onAction={setLastAction}
+        items={items}
+      />
+      <p role="status">
+        {lastAction ? `Last action: item ${lastAction}` : 'No action yet.'}
+      </p>
+    </>
   );
 }
 ```
--- a/docs/pages/components/TreeView.mdx
+++ b/docs/pages/components/TreeView.mdx
@@ -180,8 +180,15 @@ function LabelledByTreeView() {
 
 ### Custom onAction
 
-`onAction` runs your own handler when an item is activated, in addition to the
-item's selection change.
+`onAction` runs your own handler when an item is activated. Activation and
+selection are separate interactions, and each key does one thing only:
+
+- **Space** toggles the item's checkbox, exactly as a standalone checkbox would.
+- **Enter** runs `onAction`, and leaves the selection alone.
+
+Tell users what activation does before they trigger it, particularly if it opens
+a dialog or otherwise moves them somewhere — describe it in nearby instructions
+and point the tree at them with `aria-describedby`.
 
 ```jsx example
 function ActionTreeView() {
@@ -200,8 +207,12 @@ function ActionTreeView() {
   ];
   return (
     <>
+      <p id="action-tree-help">
+        Press Space to select an item, or Enter to run the action for it.
+      </p>
       <TreeView
         aria-label="Tree View with Action Handler"
+        aria-describedby="action-tree-help"
         selectionMode="multiple"
         onAction={setLastAction}
         items={items}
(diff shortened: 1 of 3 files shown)


docs/pages/components/TreeView.mdx
```diff
@@ -179,13 +180,19 @@ function LabelledByTreeView() {
 
 ### Custom onAction
 
-TreeView with a custom action handler for item selection.
+`onAction` runs your own handler when an item is activated. Activation and
+selection are separate interactions, and each key does one thing only:
+
+- **Space** toggles the item's checkbox, exactly as a standalone checkbox would.
+- **Enter** runs `onAction`, and leaves the selection alone.
```
frankensteinke, line 187: The keyboard fix looks right, but it seems like the mouse pointer path still has the behavior that this is fixing for the keyboard.

If I go to http://localhost:8003/components/TreeView and click Documents and then Photos, "last action" is left on the first item.

This is pre-existing, so maybe rather than fixing it here, you could just update the docs to make this clear, and address it as a follow up?

```suggestion
`onAction` runs your own handler when an item is activated. By keyboard,
activation and selection are separate interactions, and each key does one thing
only:

- **Space** toggles the item's checkbox, exactly as a standalone checkbox would.
- **Enter** runs `onAction`, and leaves the selection alone.

A mouse click currently both selects the row and, when nothing is selected yet,
runs `onAction`.
```

I think the closest analogue of how keyboard works would be to make a row click fire the action and the checkbox click select the row. 🤔 Another option would just be to always fire the action on the mouse click.

packages/react/src/components/TreeView/TreeView.test.tsx
```diff
@@ -454,3 +454,78 @@ test('cascade is inert in single selection mode', async () => {
...
+      selectionMode="multiple"
+      onAction={onAction}
+      defaultExpandedKeys={['1']}
+      items={items}
+    />
+  );
+  (getByRole('row', { name: /pizza/ }) as HTMLElement).focus();
+  await userEvent.keyboard('{Enter}');
+  expect(onAction).toHaveBeenCalledWith('2');
+});
```
frankensteinke, line 531: Related to my other comment: it could be worth adding a test for the pointer path now (currently untested). 

```suggestion
});

// Fails today: react-aria only treats a click as an action while the selection
// is empty, so the second click selects without running onAction. Drop the
// `.failing` when the pointer path matches Enter.
test.failing(
  'a row click runs onAction, even once rows are selected',
  async () => {
    const onAction = jest.fn();
    const { getByRole } = render(
      <TreeView
        aria-label="Test TreeView"
        selectionMode="multiple"
        onAction={onAction}
        items={items}
      />
    );
    await userEvent.click(getByRole('row', { name: 'TreeView' }));
    expect(onAction).toHaveBeenCalledTimes(1);

    await userEvent.click(getByRole('row', { name: 'Another One' }));
    expect(onAction).toHaveBeenCalledTimes(2);
  }
);
```

`.failing()` would document the behavior that we want, and you'd just remove that after we make the update.

Issue #448, opened 2021-12-08
Accessibility issue if Layout component has overflow and 0 focusable children


schne324:
We should revisit how this layout container works and potentially move back to the more simple approach of allowing the body to have overflow.

Background on this decision was that we wanted to avoid the browser allowing things to be in focus behind our topbar. By default, the browser will minimally scroll a focused element into "view" but as far as the browser is concerned, being behind the topbar is in "view"

scurker:
@schne324 Do you know if this is still an issue or anything we can do to address it? Is there an example somewhere either in cauldron or one of the webapps?

schne324:
@scurker, sorry for the lack of info. I just updated the title which should help clarify which a11y issue we are referring to here. The gist of it is that sighted keyboard users have no way of scrolling our layout container with a keyboard when there aren't any focusable elements within the layout container. This is a result of our layout container being the scrollable region on the page rather than just the `<body />` element like "normal" web apps (my fault btw--I'm the one who did that 😜 ). The solution for anyone running into this right now would be to set `tabIndex={0}` on the `<Layout />` scrollable element but I think cauldron should consider refactoring our layout to allow the `<body />` to be the scrollable element and let browsers handle reflow and allow keyboard users to use their expected means of scrolling a page with <kbd>↑</kbd> and <kbd>↓</kbd> keydowns

scurker:
Thanks for the background. I'd definitely like to reconsider how we do content layout because it's funky and does cause issues as you mentioned and isn't well documented. Maybe that's something we can get together at some point and discuss in more detail since it might be difficult to avoid breaking changes in cauldron with this change unless we're really creative.

PR #2524, opened 2026-09-08
fix(Dialog): make scrollable content keyboard accessible

denysfedorov:
Scrollable `ModalContent` is given `tabindex="-1"`, which makes it focusable by script only. It is not in the tab order, so a keyboard user cannot reachit. When the content has no focusable children of its own, the region cannot
be scrolled by keyboard at all, and axe reports `scrollable-region-focusable`.

Same defect class as #448, which reports it for `Layout`

## Changes

- `DialogContent` now uses `tabIndex={0}` when scrollable, instead of `-1`
- Rewrites the scrollable Modal keyboard test to press `Tab` and assert the region actually scrolls, rather than calling `.focus()` directly

## Notes

The existing test could not catch this. It asserted `tabindex="-1"` and then called `.focus()`, which succeeds on `-1`, so it passed while the region was unreachable by keyboard

Consumers who worked around this by passing their own `tabIndex` to
`ModalContent` can remove that override

### DIFF
--- a/packages/react/src/components/Dialog/index.tsx
+++ b/packages/react/src/components/Dialog/index.tsx
@@ -238,10 +238,8 @@ const DialogContent = ({
         'text--align-center': align === 'center',
         'text--align-right': align === 'right'
       })}
-      // A scrollable region whose content has no focusable children can only be
-      // scrolled by keyboard if the region itself is tabbable (WCAG 2.1.1). The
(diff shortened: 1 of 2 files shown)


frankensteinke (approved):
Thanks for this fix! Looks like there's a couple of Copilot comments, but I don't think that either of them are worth blocking this from getting in.

Issue #2516, opened 2026-08-28
Update the border radius to match the Panel radius when PanelContent has padding set to false


shawnsharpDQ:
When a table is inside of the PanelContent that has padding set to false, the Table overflows outside of the left and right borders. We need a matching border radius of the Panel when one PanelContent is used as well. The example below uses two `PanelContent` components.

<img width="930" height="442" alt="Image" src="https://github.com/user-attachments/assets/2d9fee16-ff72-4b38-a668-5f3475b002f3" />

```
<Panel padding={false}>
  <PanelContent
    padding={false}
  >
    <Table>
       ...
    </Table>
  </PanelContent>
  <PanelContent>
    <Pagination totalItems={9} itemsPerPage={3} />
  </PanelContent>
</Panel>
```

PR #2531, opened 2026-09-17
fix(Panel): round content that sits on a panel's corners

denysfedorov:
This patch stops a child's square corner from covering the Panel's rounded border.

With `padding={false}`, the first and last children touch the Panel's corners. If a child has its own background, it paints a square corner on top of the Panel's rounded one, so the border looks cut off. #2461 fixed this for `PanelHeader` only. Now it works for any child.

**High level changes**

- Round the top corners of the first child and the bottom corners of the last child
- Add `--panel-inner-border-radius` (panel radius minus border width) and reuse it in the `PanelHeader` rule, instead of repeating the same calc twice
- Publish `--content-start-corner-radius` / `--content-end-corner-radius` for content that draws its own corner, such as a table's cells
- Add screenshot tests: a coloured block flush at the top, flush top and bottom, and a padded panel as a control

**Reviewer notes**

- The two new properties are set only when padding is off. That is how content knows its corner is really exposed. `border-radius: inherit` cannot tell the difference, because every Panel has a radius.
- If a child's background sits inside a transparent wrapper, it is not reached — the wrapper takes the radius and paints nothing. Those cases can use `var(--panel-inner-border-radius)` directly.
- Nothing reads the new properties yet. That comes in the Table PR.

Ref: #2516


packages/react/src/components/Panel/screenshots.e2e.tsx
```diff
@@ -0,0 +1,50 @@
...
+  page
+}) => {
+  const component = await mount(<Panel padding={false}>{banner}</Panel>);
+
+  await expect(component).toHaveScreenshot('panel-flush-content-only');
+  await setTheme(page, 'dark');
+  await expect(component).toHaveScreenshot('dark--panel-flush-content-only');
+});
+
+test('should have screenshot for Panel with flush content', async ({
```
frankensteinke: Nit: maybe we should update the title here to match the screenshot names.

```suggestion
test('should have screenshot for Panel with inset content', async ({
```

frankensteinke (dismissed):

frankensteinke (approved):

PR #2532, opened 2026-09-17
fix(Table): round edge cells when the container publishes a corner

### LINKED ISSUE #2516: Update the border radius to match the Panel radius when PanelContent has padding set to false

denysfedorov:
This patch rounds a table's outer corners when it sits flush inside a rounded container, such as a Panel with `padding={false}`.

A table with `border-collapse: collapse` ignores `border-radius` on the `<table>` itself — the corners you see belong to the edge `th` and `td`. So a flush table always covered the container's rounded corner with a square one. Now the edge cells take the radius the container publishes.

**High level changes**

- Round the outer cells of the first and last rows, using the container's `--content-start-corner-radius` / `--content-end-corner-radius`. The default is `0`, so a table outside such a container does not change
- Make the sort button inherit its cell's radius, so its focus ring follows the corner instead of cutting across it
- Add screenshot tests: flush with a footer, flush top and bottom, as a direct child, an inset control, and a focused sortable header
- Add a "Panel with a Table" example to the Panel docs

**Reviewer notes**

- Needs the Panel PR, which sets those properties. On its own the default is `0` and nothing changes.
- The sort button rule is a no-op outside a rounded container: the cell has no radius, so the button inherits `0`.

Closes: #2516


packages/react/src/components/Table/screenshots.e2e.tsx
```diff
@@ -165,3 +167,98 @@ test('should have screenshot for Table[layout=grid]', async ({
...
+      <PanelContent padding={false}>{flushTable}</PanelContent>
+    </Panel>
+  );
+
+  await component.getByRole('button', { name: 'Name' }).focus();
+
+  await expect(component).toHaveScreenshot('table-flush-sort-focus');
+  await setTheme(page, 'dark');
+  await expect(component).toHaveScreenshot('dark--table-flush-sort-focus');
+});
```
frankensteinke, line 264: May be worth checking the `variant="border"` flush as well:

```suggestion
});

test('should have screenshot for Table[variant=border] flush in a container', async ({
  mount,
  page
}) => {
  const component = await mount(
    <Panel padding={false}>
      <PanelContent padding={false}>
        {React.cloneElement(flushTable, { variant: 'border' })}
      </PanelContent>
    </Panel>
  );

  await expect(component).toHaveScreenshot('table-flush-border');
  await setTheme(page, 'dark');
  await expect(component).toHaveScreenshot('dark--table-flush-border');
});
```

packages/react/src/components/Table/screenshots.e2e.tsx
```diff
@@ -165,3 +167,98 @@ test('should have screenshot for Table[layout=grid]', async ({
...
+        <TableCell>Ada Lovelace</TableCell>
+        <TableCell>Engineer</TableCell>
+      </TableRow>
+      <TableRow>
+        <TableCell>Alan Turing</TableCell>
+        <TableCell>Mathematician</TableCell>
+      </TableRow>
+    </TableBody>
+  </Table>
+);
```
frankensteinke, line 194: Could this be a component with props instead, shaped like so? It would get rid of the need for the funky `cloneElement()` business in my other suggested comment.

```jsx
const FlushTable = ({ variant }: { variant?: 'border' }) => (
  <Table variant={variant}>
    …
  </Table>
);
```

frankensteinke (dismissed):

packages/react/src/components/Table/screenshots.e2e.tsx
denysfedorov: For this use case, I intentionally used `Table` with `TableHeader` to visually test that if we have a cell with a sorting button, it should still have correct (rounded in this case) borders when focused

frankensteinke (approved):

PR #2533, opened 2026-09-18
chore: stop Dependabot from bumping Babel to major versions

rheisler-deque:
## Summary

Adds Dependabot ignore rules so the `update-babel` group no longer opens pull requests for major version updates of `@babel/*` and `babel-*` packages.

Babel 8 raises the minimum Node engine to `^22.18.0 || >=24.11.0` and moves the Babel packages to ESM. The Babel 8 bump also fails CI in this repo, because the current `.babelrc` preset shorthand does not resolve the same way under Babel 8. We do not want that major upgrade to happen automatically, so this stops it while leaving Babel 7 minor and patch updates in place.

This matches the pattern already used in `.github/dependabot.yml` for `@types/node` major versions.

## Change

Two new entries in the npm `ignore:` list:

```yaml
- dependency-name: '@babel/*'
  update-types: ['version-update:semver-major']
- dependency-name: 'babel-*'
  update-types: ['version-update:semver-major']
```

The Babel 7 to 8 upgrade will be tracked as a separate ticket.

No QA Needed


frankensteinke (approved):
Thanks!

